PR #109, opened 2014-06-15
Some testEmptyWallet() asserts -- DON'T MERGE!

schildbach:
This is actually a bug report with a test included.

assertEquals(outputValue, request.tx.getValueSentFromMe(wallet)); fails in line 2207.

This is because the "value sent from me" is double the expected amount, which is because completeTx() was called with an input which was created from the first (failing) completeTx().

If completeTx was paying attention to the pre-existing inputs, it would not have changed anything about the inputs because the correct input is already there.


schildbach:
Closing this for now.

PR #281, opened 2014-11-19
ECKey/DeterministicKey: replace ECPoint with a LazyECPoint wrapper that ...

mikehearn:
...doesn't delays parsing of key bytes into a key structure until it's needed. The process of decoding keys from the wallet previously involved decompressing/recompressing them which was taking ~seconds for hundreds of keys on Dalvik/2012 era Androids. After this patch loading such a wallet takes a few hundred milliseconds, most of which is spent inside RIPEMD160.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/core/ECKey.java
@@ -141,7 +141,7 @@ public class ECKey implements EncryptableItem, Serializable {
     // The two parts of the key. If "priv" is set, "pub" can always be calculated. If "pub" is set but not "priv", we
     // can only verify signatures not make them.
     protected final BigInteger priv;  // A field element.
-    protected final ECPoint pub;
+    protected final LazyECPoint pub;
 
     // Creation time of the key in seconds since the epoch, or zero if the key was deserialized from a version that did
     // not have this field.
@@ -173,11 +173,16 @@ public class ECKey implements EncryptableItem, Serializable {
         ECPrivateKeyParameters privParams = (ECPrivateKeyParameters) keypair.getPrivate();
         ECPublicKeyParameters pubParams = (ECPublicKeyParameters) keypair.getPublic();
         priv = privParams.getD();
-        pub = CURVE.getCurve().decodePoint(pubParams.getQ().getEncoded(true));
+        pub = new LazyECPoint(CURVE.getCurve(), pubParams.getQ().getEncoded(true));
         creationTimeSeconds = Utils.currentTimeSeconds();
     }
 
     protected ECKey(@Nullable BigInteger priv, ECPoint pub) {
+        this.priv = priv;
+        this.pub = new LazyECPoint(checkNotNull(pub));
+    }
+
+    protected ECKey(@Nullable BigInteger priv, LazyECPoint pub) {
         this.priv = priv;
         this.pub = checkNotNull(pub);
     }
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/bitcoinj/core/ECKey.java
```diff
@@ -186,16 +191,24 @@ protected ECKey(@Nullable BigInteger priv, ECPoint pub) {
      * Utility for compressing an elliptic curve point. Returns the same point if it's already compressed.
      * See the ECKey class docs for a discussion of point compression.
      */
-    public static ECPoint compressPoint(ECPoint uncompressed) {
-        return CURVE.getCurve().decodePoint(uncompressed.getEncoded(true));
+    public static ECPoint compressPoint(ECPoint point) {
+        return point.isCompressed() ? point : CURVE.getCurve().decodePoint(point.getEncoded(true));
```
devrandom: I'm not sure if this will further help performance, but instead of encoding/decoding in these functions we could call `Curve.createPoint(point.getX(), point.getY(), true)`, etc. .  After all, compression is just a flag, and doesn't really require any math.

core/src/main/java/org/bitcoinj/core/ECKey.java
```diff
@@ -340,12 +353,12 @@ public ECKey(@Nullable BigInteger privKey, @Nullable byte[] pubKey, boolean comp
...
             if (compressed)
                 point = compressPoint(point);
-            this.pub = point;
+            this.pub = new LazyECPoint(point);
         } else {
             // We expect the pubkey to be in regular encoded form, just as a BigInteger. Therefore the first byte is
             // a special marker byte.
             // TODO: This is probably not a useful API and may be confusing.
-            this.pub = CURVE.getCurve().decodePoint(pubKey);
+            this.pub = new LazyECPoint(CURVE.getCurve().decodePoint(pubKey));
```
devrandom: Why not just `new LazyECPoint(pubKey)` here?

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -413,7 +415,7 @@ public static DeterministicKey deserialize(NetworkParameters params, byte[] seri
         checkArgument(!buffer.hasRemaining(), "Found unexpected data in key");
         if (pub) {
             ECPoint point = ECKey.CURVE.getCurve().decodePoint(data);
-            return new DeterministicKey(path, chainCode, point, null, parent);
+            return new DeterministicKey(path, chainCode, new LazyECPoint(point), null, parent);
```
devrandom: Why not `new LazyECPoint(data)`?

devrandom:
LGTM overall

schildbach:
@mikehearn Did you see @devrandom 's review comments?

mikehearn:
Yes, I did. They just haven't reached the top of my todo list yet. The comments seem correct though. We could probably make those changes.

core/src/main/java/org/bitcoinj/core/ECKey.java
schildbach: All of these methods are deprecated, and also createPoint() takes BigInteger's. Do you have a non-deprecated proposal?

schildbach:
I just started to put devrandoms comments into a PR: https://github.com/bitcoinj/bitcoinj/pull/321

devrandom:
@schildbach thank you

core/src/main/java/org/bitcoinj/core/ECKey.java
devrandom: The deprecation attracts our attention to the fact that ECPoint will not maintain the compression status in the future, and that we should maintain it ourselves.  The existing code manages to avoid the warning through a technicality (i.e. using an encode/decode cycle).  We actually want that warning, to remind us to change the way we handle the compression status.

PR #292, opened 2014-12-06
Add Wallet.isWatching() to determine if the wallet is a watching w...

schildbach:
...allet.

It's useful for the UI to know because it would disable the send button etc.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/core/Wallet.java
@@ -751,6 +751,23 @@ public class Wallet extends BaseTaggableObject implements Serializable, BlockCha
     }
 
     /**
+     * Returns whether this wallet consists entirely of watching keys (unencrypted keys with no private part). Mixed
+     * wallets are forbidden.
+     * 
+     * @throws IllegalStateException
+     *             if there are no keys, or if there is a mix between watching and non-watching keys.
+     */
+    public boolean isWatching() {
+        keychainLock.lock();
+        try {
+            maybeUpgradeToHD();
+            return keychain.isWatching();
+        } finally {
+            keychainLock.unlock();
+        }
+    }
+
+    /**
      * Return true if we are watching this address.
      */
     public boolean isAddressWatched(Address address) {
--- a/core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
+++ b/core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
@@ -551,6 +551,30 @@ public class KeyChainGroup implements KeyBag {
         return keyCrypter != null;
     }
 
+    /**
+     * Returns whether this chain has only watching keys (unencrypted keys with no private part). Mixed chains are
+     * forbidden.
+     * 
+     * @throws IllegalStateException
+     *             if there are no keys, or if there is a mix between watching and non-watching keys.
+     */
+    public boolean isWatching() {
+        Boolean basicChainIsWatching = basic.isWatching();
+        Boolean deterministicChainIsWatching = !chains.isEmpty() ? getActiveKeyChain().isWatching() : null;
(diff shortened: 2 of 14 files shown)


mikehearn:
Could you consider adding a bit of code to throw exceptions when trying to have mixed wallets with both private and public-only keys? If so I'd be happy to merge this.

schildbach:
I'll work on it soon.

schildbach:
Finally I finished this PR. Added some tests as well. These two cases throw an IllegalStateException:
- keys mix watch/non-watch ... we don't want to support that case
- no keys at all ... I'm not sure about this, but I could not find a better solution.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -746,6 +746,22 @@ public DeterministicKey getWatchingKey() {
     }
 
     /**
+     * Returns true if this wallet has watch-only keys only, and false if it has non-watch-only keys only.
```
mikehearn: This is a bit awkwardly phrased. How about:

"Returns whether this wallet consists entirely of watching keys (unencrypted keys with no private part). Mixed wallets are forbidden."

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
```diff
@@ -550,6 +550,28 @@ public boolean isEncrypted() {
...
     }
 
+    /**
+     * Returns true if this chain has watch-only keys only, and false if it has non-watch-only keys only.
+     * 
+     * @throws IllegalStateException
+     *             if there are no keys, or if there is a mix between watch-only and non-watch-only keys.
+     */
+    public boolean isWatching() {
+        Boolean isWatching = null;
```
mikehearn: Why is this boxed?

mikehearn:
It'd be better if the API prevented you from getting the wallet into the bad state to start with, rather than throwing an exception just in isWatching() if it is stuck that way.

E.g. there should be some checks on the codepaths where new keys are added/imported.

schildbach:
I added checks when importing keys and improved the wording of the JavaDoc as you suggested.

I discovered that DeterministicKeyChain uses BasicKeyChain in a way that breaks our "no mixed chains" constraint, so I had to add a mode to BasicKeyChain that disables the checks. Unit tests pass.

mikehearn:
Well not exactly - the keys in the BasicKeyChain inside DKC derive their private keys on demand rather than holding them in RAM. But the private key material is still there, so it's not watching.

schildbach:
I added some JavaDocs. This is now ready for re-review.

mikehearn:
I still don't really understand the need for this isMixed thing. DKC's internal BasicKeyChain is *not* a watching key chain. The private key material is there, it just needs to be derived first (which should happen automatically). Why do you think it's mixed?

schildbach:
It's mixed because it mixes watching keys and non-watching keys (as per the definition of the ECKey.isWatched() method). Just try it: Take out the isMixed check and see many unit tests fail.

mikehearn:
You're defining "watching" wrong then. I suspect it's testing "does this ECKey have a private key field set" but that isn't the right approach, a DeterministicKey can derive its own privkey on demand assuming the wallet isn't encrypted. It's just an optimisation to avoid holding all privkeys in RAM all the time.

schildbach:
You're reviewing the code, ECKey.isWatching() is part of this PR. It's:

return isPubKeyOnly() && !isEncrypted();

So we need to overload isWatching() for DeterministicKey?

mikehearn:
Probably DeterministicKey should override isPubKeyOnly() actually. If findOrDerivePrivateKey returns null then either the parent is encrypted or it's a watch-only hierarchy.

schildbach:
I rebased this one on #345 and guess what: it works (-:

schildbach:
Thanks for merging this.

PR #298, opened 2014-12-19
Extract ECDSASignature.isCanonical()…

schildbach:
…which checks for BIP62 low S value. Adds a test.

And I pushed a second commit on top which restricts access to ECKey.CURVE_PARAMS. Use ECKey.CURVE instead.


mikehearn:
Yes, that sounds good, please do!

schildbach:
After merging and rebasing I split up the remaining commit into two parts:

1) isStandard check for sigs to use canonical encoding.
2) isStandard check for sigs to use low S values.

The remaining commit on this PR only contains 1. Virtually all txns comply to this rule, so I suggest to merge it.

I will present 2 with a later PR. As noted above still ~15% of txns violate this rule.

mikehearn:
That plan sounds good to me.

PR #299, opened 2014-12-19
Creating a useful shaded jar, that can be run with java -jar ...

tchoulihan:
This appeared to be in the video for building a simple wallet, but not in the pom itself.


schildbach:
Merged this a few days ago. Thanks.

PR #302, opened 2014-12-28
NPE when invoking `PeerGroup.broadcastTransaction()` on peer group with no block chain.

mackler:
The modified line here seems to have been assuming that `pinnedTx` was being initialized elsewhere, but it wasn't.  I'm not sure this is the correct fix.  If not, hopefully someone who understands this better can tell me what I'm missing.  Thanks.


schildbach:
Merged, thanks!

PR #305, opened 2015-01-12
LevelDb Full Pruned Block Store

n1bor:
Initial version for review.
Passes unit tests and syncs full chain in about 3 hours


n1bor:
Build is failing due to log being too big. Raised separate issue for the log size.
We were close the the 4Meg limit set by Travis. This pull is the straw that breaks the camels back...

schildbach:
Can you squash your PR into one (or maybe more if justified) commit, eliminating any merge commits?

PR #306, opened 2015-01-13
LevelDb fully pruned block store

n1bor:
and change to test logging to allow
unit tests to work on Travis. Without the log
change the unit tests fail due to hitting 4 Meg limit

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
@@ -1,0 +1,1205 @@
+/*
+ * Copyright 2016 Robin Owens
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bitcoinj.store;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.io.*;
+import java.nio.ByteBuffer;
+
+import org.bitcoinj.core.Address;
+import org.bitcoinj.core.AddressFormatException;
+import org.bitcoinj.core.NetworkParameters;
+import org.bitcoinj.core.ScriptException;
+import org.bitcoinj.core.Sha256Hash;
+import org.bitcoinj.core.StoredBlock;
+import org.bitcoinj.core.StoredUndoableBlock;
+import org.bitcoinj.core.Transaction;
+import org.bitcoinj.core.TransactionOutputChanges;
+import org.bitcoinj.core.UTXO;
+import org.bitcoinj.core.UTXOProviderException;
+import org.bitcoinj.core.VerificationException;
+import org.bitcoinj.script.Script;
+import org.iq80.leveldb.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.fusesource.leveldbjni.JniDBFactory.*;
+
+import com.google.common.base.Stopwatch;
+import com.google.common.collect.Lists;
+
+/**
+ * 
+ * <p>
+ * An implementation of a Fully Pruned Block Store using a leveldb
+ * implementation as the backing data store.
+ * <p>
+ * 
+ * <p>
+ * Includes number of caches to optimise the initial blockchain download.
+ * </p>
+ * 
+ *
+ */
+
+public class LevelDbFullPrunedBlockStore implements FullPrunedBlockStore {
+    private static final Logger log = LoggerFactory
+            .getLogger(LevelDbFullPrunedBlockStore.class);
+
+    NetworkParameters params;
+
+    // LevelDB reference.
+    DB db = null;
+
+    // Standard blockstore properties
+    protected Sha256Hash chainHeadHash;
+    protected StoredBlock chainHeadBlock;
+    protected Sha256Hash verifiedChainHeadHash;
+    protected StoredBlock verifiedChainHeadBlock;
+    protected int fullStoreDepth;
+    // Indicates if we track and report runtime for each method
+    // this is very useful to focus performance tuning on correct areas.
+    protected boolean instrument = false;
+    // instrumentation stats
+    Stopwatch totalStopwatch;
+    protected long hit;
+    protected long miss;
+    Map<String, Stopwatch> methodStartTime;
+    Map<String, Long> methodCalls;
+    Map<String, Long> methodTotalTime;
+    int exitBlock; // Must be multiple of 1000 and causes code to exit at this
+                   // block!
+    // ONLY used for performance benchmarking.
+
+    // LRU Cache for getTransactionOutput
+    protected Map<ByteBuffer, UTXO> utxoCache;
+    // Additional cache to cope with case when transactions are rolled back
+    // e.g. when block fails to verify.
+    protected Map<ByteBuffer, UTXO> utxoUncommittedCache;
+    protected Set<ByteBuffer> utxoUncommittedDeletedCache;
+
+    // Database folder
+    protected String filename;
+
+    // Do we auto commit transactions.
+    protected boolean autoCommit = true;
+
+    // Datastructures to allow us to search for uncommited inserts/deletes.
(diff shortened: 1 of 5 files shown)


mikehearn:
Why do you keep closing and reopening the pull request? Is there some issue with github? You can force push over the branch you are pull requesting to update it.

So - very initial comments without reading the code in much detail:

* Can you give performance numbers vs H2? How much of an improvement is it?
* I see that we're using a Java reimplementation of LevelDB rather than the original. What's the reason for doing that vs using the JNI bindings?
* Watch out for code style. Please always try and match the code style of projects you are contributing to - in this case, if you look at the diff you'll see that indentation is more or less a random mishmash. Probably you are using Eclipse. Configure it to use spaces not tabs and then reformat to ensure it matches. Likewise, spaces around = signs and other things.
* I saw you're adding your own LRU cache and specifying your own Bloom filters. Is that really needed? LevelDB C++ already does its own caching and it shouldn't be needed to specify custom Bloom filters. At least I didn't need to do this when I implemented LevelDB for Bitcoin Core.
* Can you split out the Travis change to a separate pull request please, so it can be reviewed and merged separately?

Thanks!

core/pom.xml
```diff
@@ -470,6 +477,11 @@
             <artifactId>orchid</artifactId>
             <version>1.1</version>
         </dependency>
+        <dependency>
```
schildbach: For each dependency, we need the hash of the artifact in the maven-enforcer-plugin configuration.

n1bor:
In response...
1) H2 sync to block 338000 running now. But think from memory is days not hours!
2) Went with pure java out of ease of integration. Swapping this for JNI should be simple as tricky bit was the transaction/rollback handling. Although JNI might be different!
3) OK - can tidy up.
4) Will rerun benchmark without caches - last test was 5h45 with no cache, 3h45 with bloom and 3h02 with bloom and utxocache (1m) . But fixed many bugs since last test. Bloom was needed as leveldb caching does not help "hasunspentoutputs" as the data is never there (it is check for dup tx hashes). These times were to an SSD - magnetics about 25% slower, on quad core with 12Gig RAM.
5) OK will split travis (did mean too but lost it somehow!) used to clearcase not git!

Below is how many mins it spends in each call with both caches - total time was 182 mins.

image

n1bor:
H2 not good at initial download. Keeps stalling every few hours. But is about 30 times slower than LevelDB once you get to the larger blocks.
H2 Block 180,305 - 186,000 - 70mins
LevelDb Block 180,000 - 186,000- 2mins 30secs. (only takes 20mins from zero to 200,000!).

n1bor:
Below are stats with/without caches:
image

Can not compare to H2 as takes days - is at least 30 times slower... and I have to keep restarting.

Do you want a push with or without the caches? Will then tidy up style, split into 2 commits one with tavis change one with the leveldb change.

Once this merged will look at a getAddressBalance and a getAllUTXOHash just to see fast they are on big data sets.

mikehearn:
OK, I think you proved your point. I'm still curious as to why this happens - but maybe the Java LevelDB reimplementation doesn't do caching or doesn't do it in the same way as the C++ version does.

Anyway it's fine to keep the caches. But please do include some of these data points in comments, so future readers can understand why they are there.

mikehearn:
Hey n1bor, do you plan to revisit this pull req w.r.t the feedback?

n1bor:
Yes - just little tied up.. Did test with native LevelDB and was 30% faster still. Hope to get round to it in week or 2.

schildbach:
@n1bor Are you still interested in this PR?

n1bor:
Seems shame to lose it - let me see if I can tidy up today/tomorrow.

core/pom.xml
```diff
@@ -301,6 +301,13 @@
                 <version>2.12.4</version>
                 <configuration>
                     <argLine>${surefireArgLine}</argLine> <!-- This is required for code coverage to work. -->
+                    <!-- This is added to reduce log output below 4Meg limit of Travis-CI -->
```
schildbach: I think this should go into a separate PR.

core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
```diff
@@ -0,0 +1,1070 @@
+package org.bitcoinj.store;
```
schildbach: All the new files are missing the license header.

core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
```diff
@@ -0,0 +1,1070 @@
...
+ * implementation as the backing data store.<p>
+ * 
+ * <p>Includes number of caches to optimise the initial blockchain download</p> 
+ *  
+ *
+ */
+
+
+public class LevelDbFullPrunedBlockStore implements FullPrunedBlockStore {
+	private static final Logger log = LoggerFactory.getLogger(LevelDbFullPrunedBlockStore.class);
```
schildbach: Leading tab/space mismatches all over the place. We're using 4 spaces per indent level.

core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
```diff
@@ -0,0 +1,1070 @@
...
+	
+	//Standard blockstore properties
+    protected Sha256Hash chainHeadHash;
+    protected StoredBlock chainHeadBlock;
+    protected Sha256Hash verifiedChainHeadHash;
+    protected StoredBlock verifiedChainHeadBlock;
+    protected int fullStoreDepth;
+    //Indicates if we track and report runtime for each method
+    //this is very useful to focuse performance tuning on correct areas.
+    protected boolean instrument=false;
```
schildbach: Space before and after operators like this.

schildbach:
I added a couple of comments, mostly code style. I assume Mike was already happy with what the PR does, so I'll merge as soon as the comments are addressed.

core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
```diff
@@ -0,0 +1,1070 @@
...
+ * <p>Includes number of caches to optimise the initial blockchain download</p> 
+ *  
+ *
+ */
+
+
+public class LevelDbFullPrunedBlockStore implements FullPrunedBlockStore {
+	private static final Logger log = LoggerFactory.getLogger(LevelDbFullPrunedBlockStore.class);
+	
+	NetworkParameters params;
```
schildbach: Should this be `private final`?

core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
```diff
@@ -0,0 +1,1070 @@
...
+
+	//LevelDB reference.
+	DB db=null;
```
schildbach: private?

core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
```diff
@@ -0,0 +1,1070 @@
...
+    protected Sha256Hash chainHeadHash;
+    protected StoredBlock chainHeadBlock;
+    protected Sha256Hash verifiedChainHeadHash;
+    protected StoredBlock verifiedChainHeadBlock;
+    protected int fullStoreDepth;
+    //Indicates if we track and report runtime for each method
+    //this is very useful to focuse performance tuning on correct areas.
+    protected boolean instrument=false;
+    //instrumentation stats
+    long wallTimeStart;
```
schildbach: Perhaps use Guava [Stopwatch](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Stopwatch.html) for this?

core/pom.xml
n1bor: removed - will find out if still needed once I push to travis and will add extra PR if needed for just this.

core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
n1bor: added

core/pom.xml
n1bor: Travis working without this so will skip it! Guess the logging has reduced or travis is allowing more log lines?

core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
n1bor: Fixed

core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
n1bor: Fixed

core/src/main/java/org/bitcoinj/store/LevelDbFullPrunedBlockStore.java
n1bor: Changed.. Could change the LRU cache too - but will do later..

n1bor:
OK - tidy now. Fixed most comments - missed a couple. Will fix in next release. Currently syncs at about 3m/bytes per sec. So taking 4-5hours.

schildbach:
Finally, I merged this. Many thanks for bearing with us for so long!

PR #307, opened 2015-01-13
Travis now lists system info with each build, …

schildbach:
…so don't run lsb_release any more.


mikehearn:
Cool. For Travis stuff you can just go ahead and commit without my review. Thanks.

PR #308, opened 2015-01-16
Make block chain be part of the context, rather than the other way around

mackler:
Right now there's a problem when a `PeerGroup` with no `BlockChain` tries to broadcast a transaction, because the peers don't have access to the same context.  We broadcast to some peers, then wait to hear from others, but in this case those others don't know the transaction they see is being waited for, the confidence is unchanged, and the broadcast procedure times out because [the future that's waiting to be called back never completes](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java#L171).

Rather than having the required `Context` be a member of the optional `BlockChain`, this patch makes the `Context` a member of the `PeerGroup`, and the `BlockChain` an optional member of the `Context`.  Actually, I'm unclear why there needs to be a separate `Context`, since the `PeerGroup` seems to _be_ the context, but I attempted to change as little as necessary for the `PeerGroup` to be able to see that transactions are being broadcast successfully.

I have limited prior experience with the code I altered, so (assuming no other problems) hopefully this will get a careful review from someone competent.  In particular I note:

1. I had to change the access of the `Context` class constructor from `protected` to `public` only because the `TestWithNetworkConnection` class uses it.
2. [Line 109 (né 105) of `TransactionBroadcast.java`](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java#L105) has a `TODO` that sounds as if this might be resolved by this patch, but I'm not sure.  If so, then that line should be changed as well.
3. This runs for me, and the tests compile, but I'm unable to run the tests to completion, so someone else should make sure all the tests will complete.

Incidentally, and slightly off-topic, here's the last thing I see before the tests hang for me; perhaps it gives some indication of what might be wrong.

    11:17:31 789 PeerGroup$7.run: Starting ...
    11:17:31 789 PeerGroup.discoverPeers: Peer discovery took 0msec and returned 0 items
    11:17:31 790 NioClientManager.handleKey: Successfully connected to /127.0.0.1:2001
    11:17:31 790 Peer.connectionOpened: Announcing to /127.0.0.1:2001 as: /bitcoinj:0.13-SNAPSHOT/
    11:17:31 790 Peer.processVersionMessage: Connected to 127.0.0.1: version=70001, subVer='/bitcoinj:0.13-SNAPSHOT/', services=0x1, time=2015-01-15 23:17:31, blocks=1
    11:17:31 790 PeerGroup.handleNewPeer: [127.0.0.1]:2001: New peer
    11:17:31 790 PeerGroup.setDownloadPeer: Setting download peer: [127.0.0.1]:2001
    11:17:32 790 NioClientManager.handleKey: Successfully connected to /127.0.0.1:2002
    11:17:32 790 Peer.connectionOpened: Announcing to /127.0.0.1:2002 as: /bitcoinj:0.13-SNAPSHOT/
    11:17:32 790 Peer.processVersionMessage: Connected to 127.0.0.1: version=70001, subVer='/bitcoinj:0.13-SNAPSHOT/', services=0x1, time=2015-01-15 23:17:31, blocks=1
    11:17:32 790 PeerGroup.handleNewPeer: [127.0.0.1]:2002: New peer
    11:17:32 75 PeerSocketHandler.timeoutOccurred: [127.0.0.1]:18333: Timed out
    11:17:35 75 PeerSocketHandler.timeoutOccurred: [127.0.0.1]:1: Timed out
    11:17:36 75 PeerSocketHandler.timeoutOccurred: [127.0.0.1]:2001: Timed out
    11:17:36 75 PeerSocketHandler.timeoutOccurred: [127.0.0.1]:2002: Timed out
    11:17:36 75 PeerSocketHandler.timeoutOccurred: [127.0.0.1]:2001: Timed out
    11:17:36 75 PeerSocketHandler.timeoutOccurred: [127.0.0.1]:2002: Timed out
    11:17:36 75 PeerSocketHandler.timeoutOccurred: [127.0.0.1]:2001: Timed out
    11:17:36 75 PeerSocketHandler.timeoutOccurred: [127.0.0.1]:2002: Timed out
    11:17:36 75 PeerSocketHandler.timeoutOccurred: [127.0.0.1]:2001: Timed out
    11:17:36 75 PeerSocketHandler.timeoutOccurred: [127.0.0.1]:2002: Timed out

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -86,7 +86,6 @@ public abstract class AbstractBlockChain {
 
     /** Keeps a map of block hashes to StoredBlocks. */
     private final BlockStore blockStore;
-    private final Context context;
 
     /**
      * Tracks the top of the best known chain.<p>
@@ -150,11 +149,6 @@ public abstract class AbstractBlockChain {
         this.params = params;
         this.listeners = new CopyOnWriteArrayList<ListenerRegistration<BlockChainListener>>();
         for (BlockChainListener l : listeners) addListener(l, Threading.SAME_THREAD);
-        context = new Context();
-    }
-
-    public Context getContext() {
-        return context;
     }
 
     /**
--- a/core/src/main/java/org/bitcoinj/core/Context.java
+++ b/core/src/main/java/org/bitcoinj/core/Context.java
@@ -1,15 +1,20 @@
 package org.bitcoinj.core;
 
+import javax.annotation.Nullable;
+
 /**
  * The Context object holds various objects that are scoped to a specific instantiation of bitcoinj for a specific
- * network. You can get an instance of this class through {@link AbstractBlockChain#getContext()}. At the momemnt it
+ * network. You can get an instance of this class through {@link PeerGroup#getContext()}. At the momemnt it
  * only contains a {@link org.bitcoinj.core.TxConfidenceTable} but in future it will likely contain file paths and
  * other global configuration of use.
  */
 public class Context {
     protected TxConfidenceTable confidenceTable;
+    private AbstractBlockChain chain;
 
-    protected Context() {
+    // The only reason this constructor is not protected is so TestWithNetworkConections can use it.
+    public Context(@Nullable AbstractBlockChain chain) {
+        this.chain = chain;
         confidenceTable = new TxConfidenceTable();
     }
 
@@ -21,5 +26,12 @@ public class Context {
      */
     public TxConfidenceTable getConfidenceTable() {
         return confidenceTable;
+    }
+
+    /**
+     * Returns the BlockChain associated with this Context, or null.
+     */
+    public AbstractBlockChain getBlockChain() {
+        return chain;
     }
 }
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -57,7 +57,7 @@ public class Peer extends PeerSocketHandler {
     protected final ReentrantLock lock = Threading.lock("peer");
 
     private final NetworkParameters params;
-    private final AbstractBlockChain blockChain;
+    private final Context context;
 
     // onPeerDisconnected should not be called directly by Peers when a PeerGroup is involved (we don't know the total
     // number of connected peers), thus we use a wrapper that PeerGroup can use to register listeners that wont get
@@ -150,7 +150,7 @@ public class Peer extends PeerSocketHandler {
     @Nullable private SettableFuture<UTXOsMessage> utxosFuture;
 
     /**
-     * <p>Construct a peer that reads/writes from the given block chain.</p>
+     * <p>Construct a peer that reads/writes from the block chain of the given context.</p>
      *
      * <p>Note that this does <b>NOT</b> make a connection to the given remoteAddress, it only creates a handler for a
      * connection. If you want to create a one-off connection, create a Peer and pass it to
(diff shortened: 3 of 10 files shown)


mikehearn:
The idea behind Context is that eventually it will contain data that doesn't obviously fit in any current object, like directories where the library can save temporary files, objects that are shared around lots of places (e.g TxConfidenceTable) and global settings. In other words, sort of like NetworkParameters but not holding only bitcoin network related stuff.

Putting it in BlockChain hasn't worked out, there have been way too many unexpected regressions caused by this. It was only ever meant to be a temporary home anyway. We could move it to PeerGroup for sure. But as I discussed with @devrandom previously, there's probably a better way to phase in this API change - just use a global variable to hold the Context pointer for now and then add it as a c'tor param to the various classes over time, so you can run multiple instances of the library in-process with separate contexts if you really want to.

So that's the rationale as to what's going on.

There is a third possibility that I don't think we discussed before - add all the global state stuff to NetworkParameters even though it's not testnet/mainnet/regtestnet specific, and then do a global rename of NetworkParameters -> BitcoinContext in one fell swoop, sort of like we did for BigInteger->Coin. One issue is that NetworkParameters was always meant to be a global singleton full of constants, and there are a bunch of places in the code that use == rather than .equals() to switch behaviours for this reason. So maybe it's not such a great plan after all.

mackler:
It sounds as if there are two kinds of contexts: one associated with a running application, and another that is specific to a particular network connection.  A temporary file directory would be part of the application context; a transaction confidence table part of a network context.  Right now, the only thing in the Context class is the confidence table, which is specific to a network connection, and I added the block chain, also specific to a network connection.

Why not set aside the Context class for now, put the confidence table and block chain directly in the PeerGroup, and come back to the Context class when you have truly global settings.  In other words, since you cannot have a network connection without a PeerGroup, use that as the context for connection-specific items, eg, the block chain and confidence table.  Leave Context to be a singleton.

As far as using NetworkParameters for global state, what happens if someone wants to have two simultaneous PeerGroups connected to the same network?  Maybe one is listening the other broadcasting or something.  They would need to have separate confidence tables, no?

mikehearn:
We *do* have truly global settings but they tend to get replicated between objects manually at the moment, e.g. "directory to put things in".

We moved the confidence table from PeerGroup to Context as part of making it a global variable. Reason: various things need access to it and the manual wiring of every object to every other object has proven to be a fiddly and error prone process for library users. It's much easier to just have a single object that collects everything together. For example Transaction needs access to the TxConfidenceTable so tx.getConfidence() continues to work even though Transaction no longer holds this data (it can be marked as deprecated, but still). If Context is a global variable/TLS slot then we can keep being backwards compatible. If it's just a thing PeerGroup has then not so much.

mackler:
Is transaction confidence really global?  One factor of confidence is how many peers are seen to have announced a broadcast, correct?  So mightn't two PeerGroups, connected to different peers, have different ideas of a particular transaction's confidence?  For example, suppose the client has a PeerGroup connecting through Tor and does a broadcast but doesn't trust the exit nodes and so has another PeerGroup watching through a bloom filter.  I'm not exactly sure of the details, but my gut is telling me that an application might want more than one TxConfidenceTable at a time.  Am I wrong about that?

Anyway, that's just my attempt to be forward-looking; I definitely have given this far less prospective thinking than you, rather I only introduced this change to address the immediate problem of broadcast futures timing out because they don't see peers echoing the broadcast transaction.  For my case, a single confidence table is fine.  Moving it out of the block chain (which my application lacks) into my PeerGroup made it start working.  If you want me to fix it another way I'll be happy to do so, just to get this particular problem eliminated from the repository.  But as this PR stands, since having the confidence table in the block chain is surely broken, this change seems to make things no worse (for me at least), even if not finally optimal.

devrandom:
The issue of the broader use of Context is a bit difficult as we're discovering.  I've been stuck on it for a while.

For the immediate issue, it seems like you could have confidence building with **either** a BlockChain or a PeerGroup.  For example, you could be reading blocks from a non-network source.  Perhaps the best solution for this PR is to have PeerGroup get the Context from the BlockChain if available, but also able to receive it through a setter (or constructor) if not.  What do you think?

mackler:
> What do you think?

I had not considered it possible to be reading blocks from a non-network source.  Since Bitcoin is a peer-to-peer network I'm unclear how you can be getting blocks other than over a network.  Do you mean reading all-at-once from a file on a disk, or do you mean information is coming in real-time as blocks are mined?  The former seems like a static state where condifence is not changing, and the latter seems like some generalization of a network connection, of which a PeerGroup might be one variety.

My attitude is that we should answer the question "what object is necessary to operate?" and then make that be the Context or make that own the Context.  In other words, make it impossible to operate without a Context, which is the problem right now, since it is possible to operate without a BlockChain.

Now, you say it's possible to operate without a PeerGroup.  So, without understanding that very well, generally I say that if the only commonly-required object class truly is `Context`, then the Context should have as a member, optionally, a `PeerGroup` and also have, optionally, a `BlockChain`, as needed, either directly or indirectly.

I'm sure you have seen that many frameworks have an object called `Context` that gets passed around.  Generally you get it by creating a master object that starts the framework, or the framework gives it to you, such as in Android.  It seems a common pattern for a context to be a parameter to every method invocation, and I cannot help think of the way Scala frameworks use implicit parameters to pass contextual values around without the need to type their names everywhere they are used.  I don't see a problem with passing a context object around, and for purposes of concurrency it seems better than a  global variable since access to its data can by synchronized.

Mike mentions the issue of backwards compatibility.  I am opposed to breaking things that now work, but we're talking about something is is not working as it now is.  Regardless of whether we're fixing something broken or adding a desired new feature, I support the goal of ending up with a more usable library eventually, and if backwards-compatibility is an obstacle, then I support moving through transitional phases that don't break working code but give people warning and time to convert to a new pattern, along with adequate migration documentation.  In other words, I have limited sympathy for the position that "yes it's broken but people are already using it so we cannot fix it."  If that's the case I say then transition through a period of there being two interfaces and phase out the broken one timely.

I'll also note that Java offers stored preferences via its [`Pereferences`](http://docs.oracle.com/javase/7/docs/api/index.html) class.  I'm not sure what the files are that a `Context` is to be storing, but I have used this Java `Preferences` to save and retrieve settings that seem contextual.  My implementation of Java stores this under the user's home directory transparently to me; the only context is the user's account.  If "directory to put things in" is stored as the value of such a preference, then that might eliminate the need for a `Context` object to store it. depending of course on whether there needs to be more than one such directory.

mackler:
I'm trying to understand the state of this issue.  Am I correct that this is being held up because of ongoing changes as discussed [in this thread](https://groups.google.com/forum/#!topic/bitcoinj/KsEWEnSdbN00) and (experimentally) implemented [in pull request #310](https://github.com/bitcoinj/bitcoinj/pull/310)?

Pull request 310 [leaves `Context` as a member field of `AbstractBlockChain`](https://github.com/devrandom/bitcoinj/blob/txconfidence/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java#L89), and so would seem to leave unsolved the problem of `BlockChain`-less transaction-broadcasts being unable to count the number of peers that have echoed the broadcast, which is the purpose of this pull request.  So I am unclear on whether these two pull requests are conflicting, and if so wondering why that one doesn't seem to be addressing the same problem.

In other words, I'm still having this problem and am wondering where I can direct my efforts to move a solution along: either by improving this pull request, or getting involved in that line of development (of which at this time I lack much understanding), or something else.

Thanks.

mikehearn:
It's on my list of things to re-visit and I'm getting back around to it slowly, though I'll be away for a couple of weeks come this weekend. Don't worry: I didn't forget and there's nothing you need to do. It's just bottlenecked on my time unfortunately.

mackler:
> Don't worry: I didn't forget and there's nothing you need to do.

Not worried and thanks for the info.  We wish we had ten of you, but since we don't I count us lucky that your skills include that of prioritizing your limited time.

mikehearn:
I think this patch should no longer be needed on git master?

mackler:
> I think this patch should no longer be needed on git master?

Nor anywhere else.  Thank you!

PR #310, opened 2015-01-17
TXConfidence experiment

devrandom:
This is just an experiment.

I've attempted to pass in the Context to the Transaction constructor in the networking code.  I'm not sure this direction is viable, but wanted to get feedback.


mikehearn:
Hmm, I thought I had reviewed this before. Did I forget? Sorry for the long delay.

This is the sort of direction I've been thinking of, yes. I'll need to take a closer look at how the weak reference and TxConfidenceTable stuff is being done before merging this, and I guess it's gonna be a PITA to get rid of usage of all the deprecated c'tors. 

I need to spend a few days on other things. Then I will return and review this more closely. But in terms of general direction I think it should be viable. What do you think? Do you see it causing big problems?

devrandom:
Actually, you did take a look already, and we exchanged a couple of
messages about it on the mailing list.

I still feel uncertain about the `WeakReference` stuff.  My design
intuition is that weakrefs are for caching stuff that can be
reconstructed.  In this particular use case, we can't recover from
losing track of a transaction - we'll just lose track of the proper
confidence state.  It doesn't feel right that the application will see
different behavior when there is memory pressure.

My current thought is that we should keep track of confidence state when
the application asks for it during the receivePending callback.  We can
add an API to stop tracking a transaction. If there's a worry that the
application will leak memory by forgetting to tell us to stop tracking,
we can have a warning or exception thrown if the number of tracked txs
exceeds a certain bound.  The bound can be configurable (up to
Integer.MAX_VALUE).

mikehearn:
The point of the weak reference is that the app shouldn't see any change in behaviour, once we fix the event listener case. Because if the weak reference gets collected it means the app doesn't have any references to the object, so it can't be affected by that data going away. It'd have no real way to get the data back anyway.

I'm not a big fan of requiring the API user to manually flush things, especially not as it'd apply to every single app.

devrandom:
That's a good point.  Let me take another look today.

mikehearn:
I think the work merged into master should move things forward here

PR #311, opened 2015-01-19
Implement Comparable interface in VersionedChecksummedBytes (and notably the Address subclass)

msgilligan:
By implementing `Comparable` in `VersionedChecksummedBytes` it will make `VersionedChecksummedBytes` and `Address` sortable and usable as keys in sorted Maps.

Unit tests included.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
+++ b/core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
@@ -22,6 +22,7 @@ import java.io.Serializable;
 import java.util.Arrays;
 
 import com.google.common.base.Objects;
+import com.google.common.primitives.UnsignedBytes;
 
 /**
  * <p>In Bitcoin the following format is often used to represent some type of key:</p>
@@ -31,7 +32,7 @@ import com.google.common.base.Objects;
  * <p>and the result is then Base58 encoded. This format is used for addresses, and private keys exported using the
  * dumpprivkey command.</p>
  */
-public class VersionedChecksummedBytes implements Serializable, Cloneable {
+public class VersionedChecksummedBytes implements Serializable, Cloneable, Comparable<VersionedChecksummedBytes> {
     protected final int version;
     protected byte[] bytes;
 
@@ -80,8 +81,6 @@ public class VersionedChecksummedBytes implements Serializable, Cloneable {
     }
 
     /**
-     * Returns the "version" or "header" byte: the first byte of the data. This is used to disambiguate what the
-     * contents apply to, for example, which network the key or address is valid on.
      * {@inheritDoc}
      *
      * This implementation narrows the return type to <code>VersionedChecksummedBytes</code>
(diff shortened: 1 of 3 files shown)


msgilligan:
I don't think build #530 failed because of my changes, it looks like something timed out.

mikehearn:
Could you explain again why you don't just pass the comparator to the maps directly? Why does bitcoinj need to have this notion of comparability in it, itself, given the somewhat arbitrary definition?

msgilligan:
I could pass the comparator to the maps directly, but the code will certainly be much cleaner and readable if `Address` is `Comparable`. 

I can think of many cases where users/apps/tests would want to sort a list of addresses and I don't think it's arbitrary how they would be sorted. In fact, I can only think of one order that anyone would want to use. Can you think of a different sort order that anyone would want to use?

Note: I did verify by checking the algorithm that base-58 and binary have the same sort order.

I can provide more details on the use cases if you think that might be helpful.

mikehearn:
The most obvious way to sort addresses in my mind is by creation/derivation date of their keys. Lexical ordering is something I've never wanted, actually. That's why I'm a bit uncertain about blessing a particular sort order here.

msgilligan:
Information about the creation/derivation date of the keys is not available in an instance of the Address class, so it can't be implemented as a default comparator. Sorting by dates of the keys would be done using the `creationTimeSeconds` of the ECKey object (which according to the comments can be zero/unknown under certain circumstances)

This comparator is technically numeric, not lexical -- the (bitcoin) base58 encoding results in the same ordering as the numeric ordering of the binary values. (See the `ALPHABET` `char[]` array currently on line 41 of `Base58.java`)

I do think that others will find this useful and that there are many applications that will want to sort by hash/address.  Wouldn't an implementation of something like the `addrindex` fork of `bitcoind` require some kind of numeric/lexical comparator?

I had originally written the code using the `String` class, but was trying to refactor to be more strongly-typed and more tightly integrated with bitcoinj, so that's how I discovered that `Address` doesn't implement comparable.

I can "make do" without having this feature upstream and don't want to be overly persistent in advocating for it, but I'm certainly willing to discuss and advocate for it until you make your final decision.

mikehearn:
Alright, fair point that you can't really sort addresses specifically in any other way.

If there are no other objections to this then I'll merge. The only reason I hesitate is I'm not sure if adding this interface would have odd side effects for hashmaps and things ..... but I don't think that can happen, right? They only depend on equals/hashCode?

msgilligan:
Yeah, hashmaps should only depend on `equals()` and `hashCode()`.  When overriding `equals()` and `hashCode()` their behavior needs to be consistent. When implementing `compareTo()` it needs to return `0` where `equals()` would return true and vice-versa. There is no default implementation of `Comparable`.

So, in theory this should have no effect on hashmaps. If it did, I would hope some of the unit tests would detect it. If you have any ideas for specific unit tests that might be useful here, let me know.

mikehearn:
OK, then in that case I'm convinced. Could you squash+rebase this? It's now 7 commits. Then I'll merge it.

msgilligan:
Thanks, Mike. Will do.

msgilligan:
The first rebase didn't work as expected, PLEASE DO NOT MERGE until  the current issue is resolved. I'll post a comment saying "ready" when it is resolved.

Sorry for the confusion,

Sean

msgilligan:
OK, Travis is passing, the commit message is correct -- we're READY TO MERGE! Woo hop!

schildbach:
Merged, thanks!

PR #312, opened 2015-01-23
Allow building a MarriedKeyChain with a watchingKey

oscarguindzberg:
Fix to solve this problem:
During HD / married wallets code refactor KeyChainGroup.addFollowingAccountKeys() was removed and now MarriedKeyChain should be used.
The thing is I find no way to create an HD multisig full watching wallet. MarriedKeyChain builder forces me to supply either a DeterministicSeed, SecureRandom or byte[] entropy (ie to be able to generate at least one side of the private keys of the multisig wallet)

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
@@ -162,6 +162,7 @@ public class DeterministicKeyChain implements EncryptableKeyChain {
         protected long seedCreationTimeSecs;
         protected byte[] entropy;
         protected DeterministicSeed seed;
+        protected DeterministicKey watchingKey;
 
         protected Builder() {
         }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -162,6 +162,7 @@ private int calcDefaultLookaheadThreshold() {
         protected long seedCreationTimeSecs;
         protected byte[] entropy;
         protected DeterministicSeed seed;
+        protected DeterministicKey watchingKey;
```
schildbach: Did you think of persistence (protobuf) for this new field?

devrandom: @schildbach - this is in the Builder inner class, no need to persist.

devrandom:
ACK

mikehearn:
Looks good to me!

mikehearn:
This has been merged

PR #313, opened 2015-01-26
Add fresh-receive-addr and current-receive-addr actions to wallet-tool.

schildbach:


mikehearn:
How is fresh-addr not the same thing as add-key?

schildbach:
From the description it looks like as if add-key is for importing keys.

mikehearn:
Add-key can be used to create new keys as well, no? The help text says:

"Adds a new key to the wallet, either specified or freshly generated."

schildbach:
It's a bit hidden. How do you ask for the current address?

mikehearn:
It's in the wallet dump, but yes, a command to print just that would be a reasonable thing to have for scripting purposes. What is the goal of the changes?

schildbach:
I was looking for a way to send coins into wallet-tool and didn't find any. The wallet dump doesn't print the current address until there is one. I'll reduce the PR to just the current address.

mikehearn:
Oh, I see. I have always just done add-key and then it prints an address. There's no real need to use current addresses with wallet-tool as it's not got much of a UI.

mikehearn:
Anyway this change looks good.

PR #315, opened 2015-01-30
Don't discard hierarchy data when deserializing an HD key.

mackler:
Currently, certain data encoded in an HD key serialized according to [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#Serialization_format), the key's depth and fingerprint of its parent, are used to check consistency if a parent HD is provided while deserializing, but is otherwise discarded.  See [lines 402 et seq.](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java#L402)  If no parent key object is provided then this means information contained in the serialization is lost.

One consequence of this situation is that deserializing an HD key and then reserializing the resulting object will produce a different value than the original serialized key.  This seems wrong.

Another issue is that the lost information may turn out to be critical to using the key.  Specifically, in the context of a system with multiple wallets, one watching, one spending, it may be undesirable for a watching HD key to have a reference to its parent object for security reasons.  However, when it comes time to spend bitcoin value paid to the address of that key or one of its child keys, the identity of its parent key must be known so the signing key can be derived.  The serialization format includes the fingerprint of an HD key's parent, but that fingerprint is currently discarded when deserializing an HD key unless the entire parent object is made available.  The hierarchy depth may be similarly useful in this situation.

Another issue with the status quo is that when an HD key is deserialized without providing a parent, the resulting key is deemed to be either the root of the hierarchy or at a depth of one, even though the serialized data indicate otherwise.  There is no reason to do this.  Just because we don't know where exactly in the hierarchy the key is located, that is not a reason to deem it to be in a location known to be inaccurate.

This patch changes the `DeterministicKey` class so that when an HD key is deserialized, the depth and fingerprint of the parent, if there is one, are not discarded, but rather are saved as private member data, and used when reserializing.  It changes the `getDepth()` method to return the depth as provided by the serialized key, rather than by counting the segments in the path, since the path may not be available if the parent is absent.  It adds a method `getParentFingerprint()` that provides access to the parent's fingerprint (if available) even when the parent object is absent.

Other than changing the access of `getPath()` to `public`, the public interface of the existing methods of this class are unchanged.  The internal representation of the parent HD key is now `Optional<DeterministicKey>` so as to avoid it ever being `null`.  There are two new private constructors that have parameters for the depth and parent fingerprint contained in a serialized HD key.

It should be noted that if an HD key at a depth greater than one is deserialized without a parent, then its path will be shown with fewer segments than its depth value.  Perhaps this should be different, with some indication in the values returned by `getPath()` and `getPathAsString()` that there is missing information about the key's exact ancestry, but this patch stops short of implementing such behavior.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -42,6 +42,8 @@ public class DeterministicKey extends ECKey {
 
     private final DeterministicKey parent;
     private final ImmutableList<ChildNumber> childNumberPath;
+    private final int depth;
+    private final int parentFingerprint; // 0 if this key is root node of key hierarchy
 
     /** 32 bytes */
     private final byte[] chainCode;
@@ -57,6 +59,8 @@ public class DeterministicKey extends ECKey {
         this.parent = parent;
         this.childNumberPath = checkNotNull(childNumberPath);
         this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
+        this.depth = this.childNumberPath.size();
+        this.parentFingerprint = (parent != null) ? parent.getFingerprint() : 0;
     }
 
     public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
@@ -77,6 +81,8 @@ public class DeterministicKey extends ECKey {
         this.parent = parent;
         this.childNumberPath = checkNotNull(childNumberPath);
         this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
+        this.depth = this.childNumberPath.size();
+        this.parentFingerprint = (parent != null) ? parent.getFingerprint() : 0;
     }
 
     /** Constructs a key from its components. This is not normally something you should use. */
@@ -91,6 +97,63 @@ public class DeterministicKey extends ECKey {
         this.keyCrypter = checkNotNull(crypter);
     }
 
+    /**
+     * Return the fingerprint of this key's parent as an int value, or zero if this key is the
+     * root node of the key hierarchy.  Raise an exception if the arguments are inconsistent.
+     * This method exists to avoid code repetition in the constructors.
+     */
+    private int ascertainParentFingerprint(DeterministicKey parentKey, int parentFingerprint)
+    throws IllegalArgumentException {
+        if (parentFingerprint != 0) {
+            if (parent != null)
+                checkArgument(parent.getFingerprint() == parentFingerprint,
+                              "parent fingerprint mismatch",
+                              Integer.toHexString(parent.getFingerprint()), Integer.toHexString(parentFingerprint));
+            return parentFingerprint;
+        } else return 0;
+    }
+
+    /**
+     * Constructs a key from its components, including its public key data and possibly-redundant
+     * information about its parent key.  Invoked when deserializing, but otherwise not something that
+     * you normally should use.
+     */
+    private DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
+                            byte[] chainCode,
+                            LazyECPoint publicAsPoint,
+                            @Nullable DeterministicKey parent,
+                            int depth,
+                            int parentFingerprint) {
+        super(null, compressPoint(checkNotNull(publicAsPoint)));
+        checkArgument(chainCode.length == 32);
+        this.parent = parent;
+        this.childNumberPath = checkNotNull(childNumberPath);
+        this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
+        this.depth = depth;
+        this.parentFingerprint = ascertainParentFingerprint(parent, parentFingerprint);
+    }
+
+    /**
+     * Constructs a key from its components, including its private key data and possibly-redundant
+     * information about its parent key.  Invoked when deserializing, but otherwise not something that
+     * you normally should use.
+     */
+    private DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
+                            byte[] chainCode,
+                            BigInteger priv,
+                            @Nullable DeterministicKey parent,
+                            int depth,
+                            int parentFingerprint) {
+        super(priv, compressPoint(ECKey.CURVE.getG().multiply(priv)));
+        checkArgument(chainCode.length == 32);
+        this.parent = parent;
+        this.childNumberPath = checkNotNull(childNumberPath);
+        this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
+        this.depth = depth;
+        this.parentFingerprint = ascertainParentFingerprint(parent, parentFingerprint);
+    }
+
+    
     /** Clones the key */
     public DeterministicKey(DeterministicKey keyToClone, DeterministicKey newParent) {
         super(keyToClone.priv, keyToClone.pub.get());
@@ -98,6 +161,8 @@ public class DeterministicKey extends ECKey {
         this.childNumberPath = keyToClone.childNumberPath;
         this.chainCode = keyToClone.chainCode;
         this.encryptedPrivateKey = keyToClone.encryptedPrivateKey;
+        this.depth = this.childNumberPath.size();
+        this.parentFingerprint = this.parent.getFingerprint();
     }
 
     /**
@@ -116,13 +181,18 @@ public class DeterministicKey extends ECKey {
         return HDUtils.formatPath(getPath());
     }
 
-    private int getDepth() {
-        return childNumberPath.size();
+    /**
+     * Return this key's depth in the hierarchy, where the root node is at depth zero.
+     * This may be different than the number of segments in the path if this key was
+     * deserialized without access to its parent.
+     */
+    public int getDepth() {
+        return depth;
     }
 
     /** Returns the last element of the path returned by {@link DeterministicKey#getPath()} */
     public ChildNumber getChildNumber() {
-        return getDepth() == 0 ? ChildNumber.ZERO : childNumberPath.get(childNumberPath.size() - 1);
+        return childNumberPath.size() == 0 ? ChildNumber.ZERO : childNumberPath.get(childNumberPath.size() - 1);
     }
 
     /**
@@ -140,14 +210,22 @@ public class DeterministicKey extends ECKey {
     }
 
     /** Returns the first 32 bits of the result of {@link #getIdentifier()}. */
-    public byte[] getFingerprint() {
+    public int getFingerprint() {
         // TODO: why is this different than armory's fingerprint? BIP 32: "The first 32 bits of the identifier are called the fingerprint."
-        return Arrays.copyOfRange(getIdentifier(), 0, 4);
+        return ByteBuffer.wrap(Arrays.copyOfRange(getIdentifier(), 0, 4)).getInt();
     }
 
     @Nullable
     public DeterministicKey getParent() {
         return parent;
+    }
+
+    /**
+     * Return the fingerprint of the key from which this key was derived, if this is a
+     * child key, or else an array of four zero-value bytes.
+     */
+    public int getParentFingerprint() {
+        return parentFingerprint;
     }
 
     /**
@@ -293,7 +371,7 @@ public class DeterministicKey extends ECKey {
                 cursor.pub, new BigInteger(1, parentalPrivateKeyBytes), cursor.parent);
         // Now we have to rederive the keys along the path back to ourselves. That path can be found by just truncating
         // our path with the length of the parents path.
-        ImmutableList<ChildNumber> path = childNumberPath.subList(cursor.getDepth(), childNumberPath.size());
+        ImmutableList<ChildNumber> path = childNumberPath.subList(cursor.getPath().size(), childNumberPath.size());
         for (ChildNumber num : path) {
             downCursor = HDKeyDerivation.deriveChildKey(downCursor, num);
         }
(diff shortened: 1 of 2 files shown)


mikehearn:
We don't use optional anywhere in bitcoinj, just a regular @nullable field is good enough.

I will need to think about this patch. Changing key management code takes more care and effort than other bits of code. Do you really need the xpub serialisation anyway? It never struck me as being very useful. Normally you are embedding an extended key into some other data structure anyway.

mackler:
> We don't use optional anywhere in bitcoinj, just a regular @nullable field is good enough.

Okay; I changed the `Optional`s back to nullables.

> Do you really need the xpub serialisation anyway? It never struck me as being very useful. 

I'm not sure I'm understanding your question.  Do you mean xpub as opposed to xpriv?  Do you mean base58 as opposed to binary?  Or are you talking about BIP32 serialization in general?

I assume you agree there should be _some_ serialization for HD keys available.  If nothing else the BIP32 format is more convenient than separately serializing the public key and chain codes separately since it puts both components into one value.  That's the reason I started using it.

Presuming you're talking about BIP32 format in general, I can offer at least four reasons that address what I understand you're getting at: (1) the value of not losing money, (2) the value of bitcoinj's internal consistency, (3) the value of bitcoinj's BIP32 compliance, and (4) the value of bitcoinj's interoperability with other libraries.  First, yes I have seen that discarding the data contained in the serialized format can mean lost money.  Second, bitcoinj already uses the information in question before discarding it, so we should be consistent and either ignore it completely or else save it.  Third, we already have made changes that serve no purpose other than compliance with the BIP32 format and if that was a good enough reason for past changes then it should apply here as well.  Finally, an added benefit of spec-compliant serialization is interoperability with other libraries.

Starting with the value of the information contained in the serialized form, specifically the parent key's fingerprint,  Recently I've had an incident.  I have a somewhat old HD key, just the public part, on a watching server.  Someone sent money to it, and now I cannot remember how I derived it.  I've been pulling my hair out desperately searching my hard drives for its parent or seed data, and I if only I knew its parent fingerprint maybe--*just maybe*--that would be enough of a clue for me to be able rederive the private key. But that fingerprint was lost when I uploaded it and bitcoinj discarded that information even though it had been contained in the serialized form.  If this patch had been applied when I uploaded that key to my watching server then I would have been in a position to have saved the parent's fingerprint.

Besides the cost of losing those bitcoins, there is intrinsic value in the bitcoinj library behaving consistently, both internally and externally.  Internally, that means consistent treatment of the fingerprint and depth data regardless of whether it comes from the serialized form or a parent `DeterministicKey` object provided by the client.  Externally that means consistency between the serialized values accepted and returned by the methods of the `DeterministicKey` class.

Regarding internal consistency, we all recognize the "H" in HD keys.  BIP32 does so by including the parent fingerprint in the serialized form and bitcoinj does so by accepting a parent `DeterministicKey` object as an argument to its constructor methods.  Moreover, bitcoinj trusts the data in the serialized form enough to throw exceptions if it mismatches a provided parent `DeterministicKey` object, as happens on on lines [404](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java#L404), [406](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java#L406) and [409](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java#L409) of the `deserialize()` method.  If that information is reliable enough to be the basis for  throwing exceptions, and if it provides some of the same information contained in the full parent object, and if the parent object is worth keeping track of as a member variable of `DeterministicKey` class, then by what logic do we simply discard information that we trust is in that parent object merely because the parent object is unavailable?  Either we should ignore information if it doesn't come from the parent object or else if we trust it enough to throw an exception over it then it's worth storing in the state of the deserialized `DeterministicKey` object.  That's what this patch does.

Such internal consistency may be hidden from client users, but the fact that bitcoinj gives back a different value than it is given is inconsistency glaringly exposed.  If a class has, as does `DeterministicKey`, methods named `serialize()` and `deserialize()` yet

    input != serialize(deserialize(input))

that's nothing short of false advertising, is it?  Information deemed not worth saving in the state of a deserialized object has no business affecting the value of the serialized form, does it?  Yet that is the current state of the `DeterministicKey` class's behavior.  If that is a problem, then this patch fixes that problem.

Besides the value of lost money, and the value of bitcoinj's self-consistency, a consistent deference to the BIP32 specification has value.  One does not introduce changes that break existing code without good reason, and [this commit last November](https://github.com/bitcoinj/bitcoinj/commit/41f8f5ccb4aa6b2fa75e771bbdecd8fd38bcf967) broke prior existing client code by introducing the requirement to pass a `NetworkParameters` as an argument to the serialization methods.  The [reason you gave](https://github.com/bitcoinj/bitcoinj/commit/41f8f5ccb4aa6b2fa75e771bbdecd8fd38bcf967#commitcomment-8589184) was compliance with the BIP32 specification, even though we seem to be in agreement there was no good reason for the change besides adherence to the specification.  So are we BIP32 compliant or not?  If not, then I request we undo that previous change.  All it did was to add an additional parameter to the (de)serialization methods and introduce the possibility of errors if the values of those `NetworkParameter` arguments mismatch.  But if we do seek to be BIP32 compliant, then this patch supports that goal.

Besides the value of lost money, internal and external consistency, and BIP32 compliance for its own sake, there is further value in bitcoinj being consistent with other libraries.  For example, the [Trezor Crypto codebase](https://github.com/trezor/trezor-crypto/blob/master/bip32.c) implements BIP32 in C.  The PostgreSQL database system allows user to [define custom data types in C](http://www.postgresql.org/docs/9.4/interactive/xtypes.html).  So by compiling the Trezor Crypto code into a library that PostgreSQL can load, one can have HD keys as a data type in one's database, and the input-output format is the BIP32 serialization.  This would interoperate perfectly with bitcoinj code except that bitcoinj is modifying the serialized form of a given key.  So even though this use case extends beyond bitcoinj, that use is being impaired by bitcoinj's noncompliance with the BIP32 serialization specification.  This patch, by causing bitcoinj to give back HD keys as the same serialized values it accepts would make interoperating with code based on other BIP32-based libraries possible or at least less error-prone.

So, for the sake of lessening the possibility of losing money due to losing track of where in a hierarchy an HD key was derived from, for the sake of bitcoinj's consistency with itself internally, its behavior externally, its consistency with the BIP32 specification and the interoperability with other libraries that such compliance brings, my answer to your original question is "yes."  I do want xpub serialisation.  It is not the case that I always embed an extended key into some other data structure.  I pass extended keys from the machine where I derive them up to my watching server.  I pass extended keys into and out of my database.  I pass extended keys between code that uses the bitcoinj library and code based on other libraries.  The current version of the `DeterministicKey` class works well for these purposes, but it has a problem and this patch addresses that problem.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -42,6 +42,8 @@
 
     private final DeterministicKey parent;
     private final ImmutableList<ChildNumber> childNumberPath;
+    private final int depth;
+    private final byte[] parentFingerprint; // 4 bytes if known, otherwise empty
```
mikehearn: It's probably nicer if this is @Nullable as well and then we don't waste memory on byte[0] allocations that have no purpose.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -91,6 +97,63 @@ public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
         this.keyCrypter = checkNotNull(crypter);
     }
 
+    /**
+     * Return the parent fingerint if known, an empty array if unknown, or raise an exception if
```
mikehearn: typo: fingerint

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -91,6 +97,63 @@ public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
...
+     * the arguments are inconsistent.  This method exists to avoid code repetition in the constructors.
+     */
+    private byte[] ascertainParentFingerprint(DeterministicKey parentKey, byte[] parentFingerprint)
+    throws IllegalArgumentException {
+        checkArgument(parentFingerprint.length == 4 || parentFingerprint.length == 0);
```
mikehearn: As the fingerprint is always four bytes, why not just use an int to store it?

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -91,6 +97,63 @@ public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
...
+        this.childNumberPath = checkNotNull(childNumberPath);
+        this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
+        this.depth = depth;
+        this.parentFingerprint = ascertainParentFingerprint(parent, parentFingerprint);
+    }
+
+    /**
+     * Constructs a key from its components, including its private key data and possibly-redundant
+     * information about its parent key.  Invoked wmhen deserializing, but otherwise not something that
+     * you normally should use.
```
mikehearn: typo: wmhen

mikehearn:
Alright, nice essay, let's do this :) But I'm not keen on all the empty arrays being thrown around here and tested. It's 4 bytes, that's an int, no big deal.

mackler:
Actually I was missing that BIP32 expressly says that the parent fingerprint of a key with no parent is to be 32 zero bits, so the empty array was wrong that way too.  It's an int now.  Thanks for noticing that.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -91,6 +97,64 @@ public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
         this.keyCrypter = checkNotNull(crypter);
     }
 
+    /**
+     * Return the fingerprint of this key's parent as an int value, or zero if this key is the
+     * root node of the key hierarchy.  Raise an exception if the arguments are inconsistent.
+     * This method exists to avoid code repetition in the constructors.
+     */
+    private int ascertainParentFingerprint(DeterministicKey parentKey, byte[] parentFingerprint)
```
mikehearn: Shouldn't parentFingerprint become an int everywhere?

mackler: > Shouldn't parentFingerprint become an int everywhere?

Not unless we're going to change the return value of [getFingerprint()](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java#L143) to `int` also.  The key fingerprints returned by different methods of a single class ought to be represented as a consistent type, especially since one purpose of having the fingerprint of a key's parent is to identify that parent and the probable way to do that is to compare the return values of one key's `getParentFingerprint()` with another's `getFingerprint()`.  It seems wrong that a client in that expectable situation would have to convert the return type of one of those functions just to see if they represent the same fingerprint.  I don't have a strong preference as to which that type should be, but those two methods ought to be consistent with each other, don't you think?

Are you referring as well to the second argument to `ascertainParentFingerprint`?  That byte array already exists before this patch, being set on [line 397](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java#L397).  We can pass that argument as an `int`, but that means either initializing an `int` just to copy its value from another variable that's already initialized, or else deserializing it from the `ByteBuffer` as an `int`.  If you want, I'll do that; I was trying to modify existing code as little as possible.

mikehearn: Oh,hmm, @devrandom do you know of any reason why fingerprints would be represented as a byte[4]? My preference would be for changing it as an int is a lot more efficient.

gary-rowe: If it's any help, the MultiBit Hardware project implements the protobufs from the Trezor device which defines a fingerprint using a simple `int`.

mackler: Int seems better than byte array to me, but I could be missing a reason otherwise.  This file seems to predate github hosting this repository, so the original author is missing.  Personally I'm leaning toward all fingerprints as ints everywhere, including the return values of `getFingerprint()` and `getParentFingerprint()`, but obviously that means changing the existing interface.  If we're going to change it for the better, we should get it over with.

You're the decider, Mike Hearn.  Say how you want it and I'll put it that way.

mikehearn:
Let's do fingerprint's as int everywhere. Thanks!

mackler:
> Let's do fingerprints as int everywhere.

It was my pleasure.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -91,6 +97,63 @@ public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
...
+     * Return the fingerprint of this key's parent as an int value, or zero if this key is the
+     * root node of the key hierarchy.  Raise an exception if the arguments are inconsistent.
+     * This method exists to avoid code repetition in the constructors.
+     */
+    private int ascertainParentFingerprint(DeterministicKey parentKey, int parentFingerprint)
+    throws IllegalArgumentException {
+        if (parentFingerprint != 0) {
+            if (parent != null)
+                checkArgument(parent.getFingerprint() == parentFingerprint,
+                              "parent fingerprint mismatch",
```
mikehearn: Doesn't this need %s %s in it to print the arguments? And if so wouldn't 0x%x 0x%x work too?

mackler: It seems to work for me as-is, and also [the Guava documentation](http://docs.guava-libraries.googlecode.com/git-history/release/javadoc/com/google/common/base/Preconditions.html) says

> In Preconditions error message template strings, only the "%s" specifier is supported, not the full range of Formatter specifiers.

So I'm going to answer "no" on both counts.  Have you observed contrary behavior?

mikehearn: I stand corrected.

mikehearn:
Merged, thanks!

PR #317, opened 2015-02-04
Change PeerGroup.addPeerFilterProvider() to return a ListenableFuture

mackler:
Currently `PeerGroup.addPeerFilterProvider()` returns `void`.  After this patch it returns the `ListenableFuture<BloomFilter>` that returns from its invocation of `recalculateFastCatchupAndFilter()`.

#### Background

I am working with a custom wallet implementation that has a so-called "lookahead" feature to be certain its bloom filter includes a certain number of receive addresses that have not yet been shared with prospective payors.

When the number of spare addresses in the bloom filter is low enough, the lookahead feature invokes `PeerGroup.recalculateFastCatchupAndFilter()`, which results in my `PeerFilterProvider` performing the following procedure:

1. Generate new receive addresses;
2. Stash the new addresses in a temporary variable;
3. Calculate the new bloom filter and give it back to the `PeerGroup`.

The return value of the `recalculateFastCatchupAndFilter()` invocation that triggers this procedure is a `ListenableFuture`, to which I add a listener that, when the future completes, moves the new addresses from the temporary variable described in step 2 into the pool of addresses that are available to be shared with prospective payors.  The rationale for doing this is the same as that for having a lookahead feature: to delay sharing publicly any receive addresses until I am as certain as possible that the peers to which I am connected will match any payments sent to those addresses.

#### The Problem

The foregoing only works when `recalculateFastCatchupAndFilter()` is triggered by the number of spare addresses falling below the lookahead threshold.  But `recalculateFastCatchupAndFilter()` is invoked initially not by that process but rather by the `PeerGroup.addPeerFilterProvider()` by which the bloom filter provider is added to the peer group.  That method discards the future, making it impossible to add a listener that will be triggered when filter calculation is complete and all `Peers` have had their bloom filters set.  This is how I propose to initially populate my pool of available receive addresses.

####  This Patch

This patch simply changes the return value of `addPeerFilterProvider` from `void` to the future returned by its invocation of `recalculateFastCatchupAndFilter`.  Since it currently returns void, it seems unlikely that the change to returning a future will cause a problem for existing code.  Thus far, it is working to solve for me the problem described, and if it is acceptable I request it be merged into the repository.

Alternatively, I am interested to learn of any superior solution to the problem this patch addresses.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerGroup.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerGroup.java
@@ -962,8 +962,14 @@ public class PeerGroup implements TransactionBroadcaster {
      *
      * <p>Note that this should be done before chain download commences because if you add a listener with keys earlier
      * than the current chain head, the relevant parts of the chain won't be redownloaded for you.</p>
+     *
+     * <p>This method invokes {@link PeerGroup#recalculateFastCatchupAndFilter(FilterRecalculateMode)}.
+     * The return value of this method is the <code>ListenableFuture</code> returned by that invocation.</p>
+     *
+     * @return a future that completes once each <code>Peer</code> in this group has had its
+     *         <code>BloomFilter</code> (re)set.
      */
-    public void addPeerFilterProvider(PeerFilterProvider provider) {
+    public ListenableFuture<BloomFilter> addPeerFilterProvider(PeerFilterProvider provider) {
         lock.lock();
         try {
             checkNotNull(provider);
@@ -982,8 +988,9 @@ public class PeerGroup implements TransactionBroadcaster {
             // if a key is added. Of course, by then we may have downloaded the chain already. Ideally adding keys would
             // automatically rewind the block chain and redownload the blocks to find transactions relevant to those keys,
             // all transparently and in the background. But we are a long way from that yet.
-            recalculateFastCatchupAndFilter(FilterRecalculateMode.SEND_IF_CHANGED);
+            ListenableFuture<BloomFilter> future = recalculateFastCatchupAndFilter(FilterRecalculateMode.SEND_IF_CHANGED);
             updateVersionMessageRelayTxesBeforeFilter(getVersionMessage());
+            return future;
         } finally {
             lock.unlock();
         }


mikehearn:
Hmm - bitcoinj already has extensive infrastructure for lookahead/Bloom filter interaction in its HD key chains support. Maybe I missed something but before reviewing this patch, is there some reason that infrastructure can't be reused? It sounds like you're resolving a problem the library already tries to handle.

mackler:
I am somewhat aware of the library's features in this regard--in fact that's how I learned that looking ahead is a good idea.  In my situation I believe bitcoinj's features are mismatched with what I'm doing.

For one thing, I'm under the impression that bitcoinj's HD key chains will only work according to what BIP32 describes as the "[default wallet structure](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#Specification_Wallet_structure)," whereby all receive keys must be derived from child number zero of an "account" key.  For reasons that may bore you I find that layout suboptimal for my purposes (briefly it seems pointless and a privacy risk for a watching server to be able to generate change addresses).  I'm looking at the `KeyChainGroup` and `DeterministicKeyChain` classes; I assume that's where I would find solutions such as you have in mind.

Another issue is that I am not using bitcoinj's `Wallet` class, rather my wallet is built out of a PostgreSQL database, and the HD key derivation is happening inside the database, i.e., not using bitcoinj.  Bitcoinj is doing the networking, including generating the bloom filter.  The master key is in the database where bitcoinj can ignore it, and my `PeerFilterProvider`'s `beginBloomFilterCalculation()` method hits the database to retrieve the next batch of address keys, rather than keeping the master key in memory, using bitcoinj to derive the receive keys and then saving them to the database.

To be honest, I started developing my system before most of bitcoinj's features in this regard were available, and I am not fully aware of everything it does and how it works, so if you think it's possible I could integrate its lookahead behavior with my system, I will investigate further.  I have given it cursory inspection and seems highly tailored to using bitcoinj's `Wallet` class and the BIP32 default wallet structure to the point where getting it to do otherwise would be a significant effort.

mikehearn:
I'd like to see us become less tied to the default BIP32 layout. It's probably not that hard to fix this, it just wasn't a priority.

Anyway, the patch looks OK to me, I'll merge it. I'm not sure how useful it will be for anyone except you, but we can always change things around in future.

mackler:
You can be certain at least one person appreciates this change.  Thank you!

mikehearn:
This was merged, I just forgot to close the pull req.

PR #318, opened 2015-02-06
Modified ALLOWED_TIME_DRIFT...

Bitango:
...to reflect the difference between bitcoinj and bitcoin core.


schildbach:
I think you're breaking rather than repairing something. Bitcoin time values are in seconds, and if you look at the usages of ALLOWED_TIME_DRIFT it's all seconds as well.

PR #319, opened 2015-02-07
Updates PeerGroup's javadoc: PeerGroup is not a guava service 

kmels:
When PeerGroup stopped being a guava service in 27bc229, the javadocs were not updated. This PR updates them. 

Note: even if I don't understand why PeerGroup.stop() should be called, I left the suggestion there.


mikehearn:
Thanks! The stop method just shuts down network sockets and so on. If your app is about to quit anyway, you don't need to call it explicitly. It's useful if you need to shut down and restart the network subsystem though.

PR #21, opened 2014-03-04
Issue #457 - SeedPeers to use sipas DNS dumps

msgilligan:
See issue #457 on Google Code.

* Changes to SeedPeers to read configured IPs of known peers from
  text files in JAR resource
* Fallback option for Android to specify an InputStream
* Unit test update for SeedPeers
* Gradle/Groovy script in contrib/seeds to download/build -seeds.txt for
  production network
* Hand-created -seeds.txt files for test net and unit test configs

PR #320, opened 2015-02-10
Get SeedPeers addresses from NetworkParamters

bitbandi:
Addresss seeds come from network parameters
(testnet addresses are empty, i didnt find that in bitcoin codebase)


schildbach:
Sounds sensible to me.

mikehearn:
I'm not against this but it's overlapping quite a lot with pull #21 - that was never merged primarily because it's still got pending feedback against it. Seeds should be loaded from an external file really which is what that pull req does.

schildbach:
Merged this, because it clearly leads forward and the other PR is in a not so well state and has been like that for quite a while. Thanks for the contribution!

PR #325, opened 2015-02-25
Explicit compression/decompression of ECPoint

devrandom:
* Save an encode/decode round trip
* Expose the fact that ECPoint compression handling is deprecated and we should do our own


mikehearn:
Is there much point in doing this before we actually just store compression state ourselves?

devrandom:
I agree, there's not much functional point.  It's mostly documentary, and can skip it for now.

mikehearn:
Well the change looks correct, at least. I wish we had more unit tests for the compression though. I'd be happier if we had some test vectors that prove this change is a no-op. But I can't quite remember if we do or not :(

mikehearn:
OK, I merged this, as it does appear to be a more direct way to write the same thing.

PR #326, opened 2015-02-25
Make the Context object be a global object

mikehearn:
It is propagated between threads, and created on demand in the PeerGroup constructor. This isn't complete but is a sketch of where we might go. Context propagation is mostly auto-magical at the moment and will become less so with time, making it easier to use multiple instances of the library from the same thread simultaneously. At the moment you would potentially hit issues if using multiple instances from the same thread though and we should fix that before landing (aka don't break devrandom).


mikehearn:
@devrand See what you think of this.

mikehearn:
Oops, I meant to say @devrandom - see what you think

I think this work completes the design doc.

devrandom:
Okay, so this answers my question re the role of the TxConfidenceTable.  It now only tracks previously unseen transactions.  Should it be made to forget Confidence objects once we download a tx, to make space for others?

As to reviewing the PR as a whole.  This is a pretty large change and touches areas that I'm not as familiar with.  To be confident, I would need an interactive walkthrough or more time to review.

mikehearn:
There's no hurry. Any review is appreciated.

> Should it be made to forget Confidence objects once we download a tx, to make space for others?

The table won't go above 1000 elements so eventually the weak reference to the confidence entry will be removed naturally.

core/src/main/java/org/bitcoinj/core/Context.java
```diff
@@ -1,16 +1,119 @@
...
+    /**
+     * Sets the given context as the current thread context. You should use this if you create your own threads that
+     * want to create core BitcoinJ objects. Generally, if a class can accept a Context in its constructor and might
+     * be used (even indirectly) by a thread, you will want to call this first. Your task may be simplified by using
+     * a {@link org.bitcoinj.utils.ContextPropagatingThreadFactory}.
+     *
+     * @throws java.lang.IllegalStateException if this thread already has a context
+     */
+    public static void propagate(Context context) {
+        checkState(slot.get() == null);
```
mackler: Might be helpful to use the form of `checkState()` that includes a message.  Something like

    checkState(..., "Thread %s already has a Context for %s", Thread.currentThread(), slot.get().getParams().getId());

mikehearn:
Merged!

mackler:
Good news.  Thank you!!

PR #327, opened 2015-03-04
Align TransactionOutPoint hashCode with equals contract.

jamespedwards42:
Copied the hashCode from UTXO for consistency.


schildbach:
I think it's not needed. The hash alone should already be unique enough, I'd say.

jamespedwards42:
What if you have two separate outpoints coming from the same transaction?

schildbach:
Hmm thinking about it again you're right.
Who knows, maybe it would improve our problems with handling transactions with *many* outputs?

mikehearn:
Good catch. Our equals/hashCode methods are pretty inconsistent as they were added in an ad-hoc way over time. I checked in a variant of this that uses the IntelliJ default hash code generator, which should result in better mixing.

PR #328, opened 2015-03-09
Implement DeterministicKey.isPubKeyOnly().

schildbach:
The super-implementation in ECKey doesn't take possible rederivation into account.

### DIFF
--- a/core/src/test/java/org/bitcoinj/crypto/ChildKeyDerivationTest.java
+++ b/core/src/test/java/org/bitcoinj/crypto/ChildKeyDerivationTest.java
@@ -171,9 +171,15 @@ public class ChildKeyDerivationTest {
     @Test
     public void pubOnlyDerivation() throws Exception {
         DeterministicKey key1 = HDKeyDerivation.createMasterPrivateKey("satoshi lives!".getBytes());
+        assertFalse(key1.isPubKeyOnly());
         DeterministicKey key2 = HDKeyDerivation.deriveChildKey(key1, ChildNumber.ZERO_HARDENED);
+        assertFalse(key2.isPubKeyOnly());
         DeterministicKey key3 = HDKeyDerivation.deriveChildKey(key2, ChildNumber.ZERO);
-        DeterministicKey pubkey3 = HDKeyDerivation.deriveChildKey(key2.getPubOnly(), ChildNumber.ZERO);
+        assertFalse(key3.isPubKeyOnly());
+        DeterministicKey pubkey2 = key2.getPubOnly();
+        assertTrue(pubkey2.isPubKeyOnly());
+        DeterministicKey pubkey3 = HDKeyDerivation.deriveChildKey(pubkey2, ChildNumber.ZERO);
+        assertTrue(pubkey3.isPubKeyOnly());
         assertEquals(key3.getPubKeyPoint(), pubkey3.getPubKeyPoint());
     }
 
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -277,6 +277,20 @@ public class DeterministicKey extends ECKey {
     }
 
     /**
+     * A deterministic key is considered to be 'public key only' if it hasn't got a private key part and it cannot be
+     * rederived.
+     */
+    @Override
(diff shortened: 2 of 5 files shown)


schildbach:
Note I had to change two usages of .isPubKeyOnly() to !.hasPrivKey() … I hope I didn't break anything due to this.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -309,7 +309,7 @@ protected DeterministicKeyChain(DeterministicSeed seed) {
      * this method to watch an arbitrary fragment of some other tree, this limitation may be removed in future.
      */
     public DeterministicKeyChain(DeterministicKey watchingKey, long creationTimeSeconds) {
-        checkArgument(watchingKey.isPubKeyOnly(), "Private subtrees not currently supported");
```
mikehearn: Shouldn't this still be correct? Why would isPubKeyOnly return false for a watching key?

mikehearn:
Ping?

schildbach:
Yeah, I know )-:

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
schildbach: You're probably right, but then we need to somehow change the failing tests. Because what they currently seem to do is creating a private subtree.

I reverted this one change and rebased on master.

schildbach: The issue is probably the way getPubOnly() works. It just reuses the non-pubOnly parent. Of course a pubOnly key that still has a private key in its parent isn't a pubOnly key… guess we need to rethink this more thoroughly.

schildbach: Maybe we need to forbid getPubOnly on non-root HD keys. To get a pubOnly HD key you'd need to pubOnly the root and re-derive the hierarchy as needed.

core/src/test/java/org/bitcoinj/crypto/ChildKeyDerivationTest.java
```diff
@@ -171,9 +171,15 @@ public void encryptedDerivation() throws Exception {
...
     public void pubOnlyDerivation() throws Exception {
         DeterministicKey key1 = HDKeyDerivation.createMasterPrivateKey("satoshi lives!".getBytes());
+        assertFalse(key1.isPubKeyOnly());
         DeterministicKey key2 = HDKeyDerivation.deriveChildKey(key1, ChildNumber.ZERO_HARDENED);
+        assertFalse(key2.isPubKeyOnly());
         DeterministicKey key3 = HDKeyDerivation.deriveChildKey(key2, ChildNumber.ZERO);
-        DeterministicKey pubkey3 = HDKeyDerivation.deriveChildKey(key2.getPubOnly(), ChildNumber.ZERO);
+        assertFalse(key3.isPubKeyOnly());
+        DeterministicKey pubkey2 = key2.getPubOnly();
+        assertFalse(pubkey2.isPubKeyOnly());
```
mikehearn: Surely these two lines indicate some naming or conceptual confusion? We call a method called getPubOnly and then the return for isPubKeyOnly is false?

schildbach: It looks like you're right. I'm not sure why I wrote this – it's perhaps just a bug and I never got to fix it because I'm stuck with the other test errors. I just switched the asserts, rebased and pushed.

schildbach: But the line you're quoting exhibits the problem I'm stuck with. Yes, the key should be public only. But it isn't because it contains the reference to a private parent. Someone has to make decisions here – I can't do that.

schildbach: Maybe the .getPubOnly() method needs to recursively clone the entire branch of the hierarchy.

mikehearn: Or maybe it should be disconnected from the parent entirely? OK I'll play around a bit.

schildbach:
Thanks for merging this.

PR #329, opened 2015-03-10
Increase wallet autosave delay for wallet-tool.

schildbach:
Since we have some recovery for out-of-sync wallets in place, we can afford to save less often.


mikehearn:
Maybe we should just take that line out and use the default?

schildbach:
Wouldn't that disable autosave? There is no version of autosaveToFile() without time arguments, although there probably should.

schildbach:
I just increased the same delay for WalletAppKit.

mikehearn:
Ok, LGTM

schildbach:
Merged.

PR #330, opened 2015-03-11
Properly ignore PeerGroupTest.peerPriority().

schildbach:


mikehearn:
Ah yes, you're right - for some reason I thought @Ignore just ignored exceptions rather than skipping running the test entirely. But I just checked the docs and you're right, this is the correct way to do it.

schildbach:
Merged.

PR #331, opened 2015-03-11
Support double spend forwarding

oscarguindzberg:


core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1691,24 +1692,46 @@ private boolean checkForDoubleSpendAgainstPending(Transaction tx, boolean takeAc
...
         }
-        if (takeAction && !doubleSpentTxns.isEmpty()) {
-            killTx(tx, doubleSpentTxns);
+        return doubleSpendTxns;
+    }
+
+    /**
+     * Adds to txSet all the txns in txPool spending outputs of txns in txSet, 
+     * and all txns spending the outputs of those txns, recursively.
+     * @param txSet The set of txns whose dependencies should be added. This method updates txSet.
```
mikehearn: For internal methods I don't normally bother with full blown @param and @return type stuff - only for public APIs. Usually just a single line comment is enough to explain how the function behaves assuming it's of a reasonable size.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1691,24 +1692,46 @@ private boolean checkForDoubleSpendAgainstPending(Transaction tx, boolean takeAc
...
+    void addTransactionsDependingOn(Set<Transaction> txSet, Set<Transaction> txPool) {
+        Deque<Transaction> txDeque = new LinkedList<Transaction>();
+        txDeque.addAll(txSet);
+        while(!txDeque.isEmpty()) {
+            Transaction tx = txDeque.poll();
+            for (Transaction anotherTx : txPool) {
+                if (anotherTx.equals(tx)) continue;
+                for (TransactionInput input : anotherTx.getInputs()) {
+                    if (input.getOutpoint().getHash().equals(tx.getHash())) {
+                        if (!txDeque.contains(anotherTx)) {
```
mikehearn: contains() in a LinkedList is quite expensive. Perhaps we want a LinkedHashMap here so we get fast random access? Or maybe it doesn't matter much.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1815,6 +1838,22 @@ private void receive(Transaction tx, StoredBlock block, BlockChain.NewBlockType
                 // this method has been called by BlockChain for all relevant transactions. Otherwise we'd double
                 // count.
                 ignoreNextNewBlock.add(txHash);
+
+                // When a tx is received from the best chain, if other txns that spend this tx are IN_CONFLICT,
```
mikehearn: Maybe this chunk can be factored into a method? IntelliJ can do this sort of refactoring automatically.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1856,6 +1895,65 @@ private void receive(Transaction tx, StoredBlock block, BlockChain.NewBlockType
         saveNow();
     }
 
+    /**
+     * Finds if tx is NOT spending other txns which are in the specified confidence type
```
mikehearn: Again this can just be

// Finds if tx is NOT spending other txns which are in the specified confidence type

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1856,6 +1895,65 @@ private void receive(Transaction tx, StoredBlock block, BlockChain.NewBlockType
...
+     * @param tx
+     * @param confidenceType
+     * @return true if tx is NOT spending other txns which are in the specified confidence type. false otherwise
+     */
+    private boolean isNotSpendingTxnsInConfidenceType(Transaction tx, ConfidenceType confidenceType) {
+        boolean isNotSpendingTxnsInConfidenceType = true;
```
mikehearn: And this can just be boolean result

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1856,6 +1895,65 @@ private void receive(Transaction tx, StoredBlock block, BlockChain.NewBlockType
...
+        for (TransactionInput txInput : tx.getInputs()) {
+            Transaction connectedTx = this.getTransaction(txInput.getOutpoint().getHash());
+            if (connectedTx!=null && connectedTx.getConfidence().getConfidenceType().equals(confidenceType)) {
+                isNotSpendingTxnsInConfidenceType = false;
```
mikehearn: You can just "return false" here, right? Once set it's never reset.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1856,6 +1895,65 @@ private void receive(Transaction tx, StoredBlock block, BlockChain.NewBlockType
...
+    /**
+     * Finds whether txA spends txB 
+     * @param txA
+     * @param txB
+     * @return true if txA spends txB. False otherwise.
+     */
+    boolean spends(Transaction txA, Transaction txB) {
+        boolean spends = false;
+        for (TransactionInput txInput : txA.getInputs()) {
+            spends |= txInput.getOutpoint().getHash().equals(txB.getHash());
```
mikehearn: This can just be "if (txInput.....) return true" right? No need to continue looping after finding that the condition is correct.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1856,6 +1895,65 @@ private void receive(Transaction tx, StoredBlock block, BlockChain.NewBlockType
...
+            if (connectedTx!=null && connectedTx.getConfidence().getConfidenceType().equals(confidenceType)) {
+                isNotSpendingTxnsInConfidenceType = false;
+            }
+        }        
+        return isNotSpendingTxnsInConfidenceType;
+    }
+
+    
+    /**
+     * Creates and returns a new List with the same txns as inputSet but txns are sorted by depencency.
```
mikehearn: This is also known as a topological sort: might as well mention that.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1856,6 +1895,65 @@ private void receive(Transaction tx, StoredBlock block, BlockChain.NewBlockType
...
+     * If tx B spends tx A, then tx A should before tx B on the returned List.
+     * Several invocations to this method with the same inputSet could result in lists with txns in different order, 
+     * as there is no guarantee on the order of the returned txns besides what was already stated.
+     * @param inputSet the txns to sort
+     * @return a List of txns where if tx B spends tx A, then tx A should before tx B.
+     */
+    List<Transaction> sortTxnsByDependency(Set<Transaction> inputSet) {
+        List<Transaction> result = new ArrayList<Transaction>(inputSet);
```
mikehearn: Some JVMs can't do type sharpening e.g. on Android so it's a bit faster to write out ArrayList explicitly here. Not a big deal.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -2146,12 +2250,42 @@ public boolean maybeCommitTx(Transaction tx) throws VerificationException {
...
-            log.info("->pending: {}", tx.getHashAsString());
-            tx.getConfidence().setConfidenceType(ConfidenceType.PENDING);
-            confidenceChanged.put(tx, TransactionConfidence.Listener.ChangeReason.TYPE);
-            addWalletTransaction(Pool.PENDING, tx);
+            
+            Set<Transaction> doubleSpendPendingTxns = findDoubleSpendsAgainst(tx, pending);
+            Set<Transaction> doubleSpendUnspentTxns = findDoubleSpendsAgainst(tx, unspent);
+            Set<Transaction> doubleSpendSpentTxns = findDoubleSpendsAgainst(tx, spent);
+            
+            if (!doubleSpendUnspentTxns.isEmpty() ||
```
mikehearn: How can this case happen? The commit path checks that a tx isn't in the wallet already.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -2146,12 +2250,42 @@ public boolean maybeCommitTx(Transaction tx) throws VerificationException {
...
+                log.info("->dead: {}", tx.getHashAsString());
+                tx.getConfidence().setConfidenceType(ConfidenceType.DEAD);                
+                confidenceChanged.put(tx, TransactionConfidence.Listener.ChangeReason.TYPE);                
+                addWalletTransaction(Pool.DEAD, tx);
+            } else if (!doubleSpendPendingTxns.isEmpty() ||
+                !isNotSpendingTxnsInConfidenceType(tx, ConfidenceType.IN_CONFLICT)) {
+                // tx is a double spend against a pending tx or spends outputs of a tx already IN_CONFLICT
+                // Add tx to the pending pool. Update the confidence type of tx, the txns in conflict with tx and all 
+                // their dependencies to IN_CONFLICT and schedule confidence listener notifications.
+                log.info("->pending: {}", tx.getHashAsString());
```
mikehearn: Maybe the log message here should be more chatty about what's happening?

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -4012,12 +4146,44 @@ public void reorganize(StoredBlock splitPoint, List<StoredBlock> oldBlocks, List
...
+                                doubleSpendTx.getConfidence().setConfidenceType(ConfidenceType.IN_CONFLICT);                    
+                                confidenceChanged.put(doubleSpendTx, TransactionConfidence.Listener.ChangeReason.TYPE);
+                                updateForSpends(doubleSpendTx, false);                                
+                            }
+                        }                    
+                    }
+                }        
             }
 
+            // Note: Is this comment still valid after we implement support for double spend forwarding? I guess it is not.
```
mikehearn: I think it is. The first double spend will be forwarded *if* the user is connected to Bitcoin XT nodes, but the behaviour on re-org is the same: double spent txns are erased from the mempool and will not be resurrected by miners.

core/src/test/java/org/bitcoinj/core/WalletTest.java
```diff
@@ -851,6 +852,418 @@ public void onWalletChanged(Wallet wallet) {
...
+        // txBRoot -> txB2 -+
+        //
+        // txARoot -> txA3
+        //
+        // txA1 is in conflict with txA2 and txA3. txB1 is in conflict with txB2. 
+        
+
+        CoinSelector originalCoinSelector = wallet.getCoinSelector();        
+        try {
+            wallet.setCoinSelector(AllowUnconfirmedCoinSelector.get());
```
mikehearn: There is an alias for this: `wallet.allowSpendingUnconfirmedTransactions()` which is a bit clearer

core/src/test/java/org/bitcoinj/core/WalletTest.java
```diff
@@ -851,6 +852,418 @@ public void onWalletChanged(Wallet wallet) {
...
+
+            Transaction txBRoot = sendMoneyToWallet(valueOf(100, 0), AbstractBlockChain.NewBlockType.BEST_CHAIN);
+            SendRequest b1Req = SendRequest.to(address, valueOf(11, 0));
+            b1Req.tx.addInput(txBRoot.getOutput(0));
+            b1Req.shuffleOutputs = false;
+            wallet.completeTx(b1Req);
+            Transaction txB1 = b1Req.tx;
+            SendRequest b2Req = SendRequest.to(address, valueOf(22, 0));
+            b2Req.tx.addInput(txBRoot.getOutput(0));
+            b2Req.shuffleOutputs = false;
```
mikehearn: Perhaps we should default shuffleOutputs to false for unit test params/testnet - might reduce a lot of the duplication here.

core/src/test/java/org/bitcoinj/core/WalletTest.java
```diff
@@ -851,6 +852,418 @@ public void onWalletChanged(Wallet wallet) {
...
+            unsortedTxns.add(send1c);
+            unsortedTxns.add(send1d);
+            unsortedTxns.add(send1e);
+            unsortedTxns.add(send2a);
+            unsortedTxns.add(send2b);
+            unsortedTxns.add(send2c);
+            List<Transaction> sortedTxns = wallet.sortTxnsByDependency(unsortedTxns);
+            
+            assertEquals(8, sortedTxns.size());
+            assertTrue(sortedTxns.indexOf(send1a) < sortedTxns.indexOf(send1b));
```
mikehearn: Why not just assertEquals(sortedTxns, ImmutableList.of(......));

mikehearn:
This is looking *really* good - thanks for taking this on. The patch looks of very high quality any my comments are mostly cosmetic in nature. Let's get this in to 0.13

oscarguindzberg:
I will continue my work on this next week.

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: done

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: done

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: I suggest to not extract a method here. We don't need this functionality anywhere else so far and Wallet has already an overwhelming amount of methods. Having said that, this is not life or dead to me, so if you insist, I will extract the method.

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: done

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: refactored

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: done

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: done

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: done

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: done

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: I am not following you. I am trying to cover the case of a peer sending us a new tx in conflict with a tx already building (I don't know why a peer would do that, but just in case).

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: done. Please check you are happy with my change. Also updateForSpends() logs about double spends.

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: ok, I will change the code I wrote. Sill todo.

core/src/test/java/org/bitcoinj/core/WalletTest.java
oscarguindzberg: done

core/src/test/java/org/bitcoinj/core/WalletTest.java
oscarguindzberg: Let's do that as part of another pull request.

oscarguindzberg:
Should I squash all commits into one commit before you merge the PR?

mikehearn:
I tried this but I get tons of unit test failures:

java.lang.NullPointerException
	at org.bitcoinj.core.Wallet.addTransactionsDependingOn(Wallet.java:1724)
	at org.bitcoinj.core.Wallet.receive(Wallet.java:1846)
	at org.bitcoinj.core.Wallet.receiveFromBlock(Wallet.java:1760)
	at org.bitcoinj.core.BlockChainTest$1.receiveFromBlock(BlockChainTest.java:78)

Can you rebase on master and see if you observe the same problem?

oscarguindzberg:
yes, sorry I introduced this bug with the latest changes I made. I will
submit the fix soon.

On Sat, Mar 21, 2015 at 12:49 PM, Mike Hearn <notifications@github.com>
wrote:

> I tried this but I get tons of unit test failures:
>
> java.lang.NullPointerException
> at org.bitcoinj.core.Wallet.addTransactionsDependingOn(Wallet.java:1724)
> at org.bitcoinj.core.Wallet.receive(Wallet.java:1846)
> at org.bitcoinj.core.Wallet.receiveFromBlock(Wallet.java:1760)
> at
> org.bitcoinj.core.BlockChainTest$1.receiveFromBlock(BlockChainTest.java:78)
>
> Can you rebase on master and see if you observe the same problem?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitcoinj/bitcoinj/pull/331#issuecomment-84375956>.
>



-- 
Oscar Guindzberg

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: done, pushing soon

core/src/test/java/org/bitcoinj/core/WalletTest.java
oscarguindzberg: Because the exact order of txns in sortedTxns is not guaranteed. Eg send1a,send1b,send1c,send1d,send1e,send2a,send2b,send2c is a valid result but send2a,send2b,send2c,send1a,send1b,send1c,send1d,send1e is also valid.

oscarguindzberg:
Ok, I think fixed all the pending stuff :)

schildbach:
Mike has just stated he wants to another review of this patch, so bear with us.

core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
```diff
@@ -98,6 +98,16 @@
...
 
         /**
+         * If IN_CONFLICT, then it means there is another transaction (or several other transactions) spending one 
+         * (or several) of its inputs but nor this transaction nor the other/s transaction/s are included in the best chain.
+         * The other/s transaction/s should be IN_CONFLICT too.
+         * IN_CONFLICT can be thought as an intermediary state between a) PENDING and BUILDING or b) PENDING and DEAD.
+         * Another common name for this situation is "double spend".
+         */
+        IN_CONFLICT(5),
+
```
schildbach: One empty line should be enough as a separation.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1708,42 +1709,66 @@ public boolean isTransactionRelevant(Transaction tx) throws ScriptException {
...
+        for (Transaction tx : txSet) {
+            txQueue.put(tx.getHash(), tx);            
+        }
+        while(!txQueue.isEmpty()) {
+            Transaction tx = txQueue.remove(txQueue.keySet().iterator().next());
+            for (Transaction anotherTx : txPool) {
+                if (anotherTx.equals(tx)) continue;
+                for (TransactionInput input : anotherTx.getInputs()) {
+                    if (input.getOutpoint().getHash().equals(tx.getHash())) {
+                        if (txQueue.get(anotherTx.getHash())==null) {
```
schildbach: Spaces around "==".

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1902,6 +1946,54 @@ private void receive(Transaction tx, StoredBlock block, BlockChain.NewBlockType
         hardSaveOnNextBlock = true;
     }
 
+    /** Finds if tx is NOT spending other txns which are in the specified confidence type */
+    private boolean isNotSpendingTxnsInConfidenceType(Transaction tx, ConfidenceType confidenceType) {
+        for (TransactionInput txInput : tx.getInputs()) {
+            Transaction connectedTx = this.getTransaction(txInput.getOutpoint().getHash());
+            if (connectedTx!=null && connectedTx.getConfidence().getConfidenceType().equals(confidenceType)) {
```
schildbach: Spaces around "!="

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -2145,7 +2243,7 @@ private void killTx(@Nullable Transaction overridingTx, List<Transaction> killed
             for (TransactionInput deadInput : tx.getInputs()) {
                 Transaction connected = deadInput.getOutpoint().fromTx;
                 if (connected == null) continue;
-                if (connected.getConfidence().getConfidenceType() != ConfidenceType.DEAD) {
+                if (connected.getConfidence().getConfidenceType() != ConfidenceType.DEAD && deadInput.getConnectedOutput().getSpentBy()!=null && deadInput.getConnectedOutput().getSpentBy().equals(deadInput)) {
```
schildbach: Spaces

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -4226,10 +4354,11 @@ public void reorganize(StoredBlock splitPoint, List<StoredBlock> oldBlocks, List
                 // there's another re-org.
                 if (tx.isCoinBase()) continue;
                 log.info("  ->pending {}", tx.getHash());
+                
                 tx.getConfidence().setConfidenceType(ConfidenceType.PENDING);  // Wipe height/depth/work data.
                 confidenceChanged.put(tx, TransactionConfidence.Listener.ChangeReason.TYPE);
                 addWalletTransaction(Pool.PENDING, tx);
-                updateForSpends(tx, false);
+                updateForSpends(tx, false);
```
schildbach: Trailing spaces.

core/src/test/java/org/bitcoinj/core/WalletTest.java
```diff
@@ -864,6 +865,418 @@ public void onWalletChanged(Wallet wallet) {
...
+        // txBRoot (100) -> txB1 (11) -+
+        //
+        // txARoot (10)  -> txA2 (2)  -+
+        //                             |--> txC2 (0.20) -> txD2 (0.02) 
+        // txBRoot (100) -> txB2 (22) -+
+        //
+        // txARoot (10)  -> txA3 (3)
+        //
+        // txA1 is in conflict with txA2 and txA3. txB1 is in conflict with txB2. 
+
```
schildbach: One empty line should be enough.

core/src/test/java/org/bitcoinj/core/WalletTest.java
```diff
@@ -864,6 +865,418 @@ public void onWalletChanged(Wallet wallet) {
...
+        FakeTxBuilder.BlockPair blockPair2 = createFakeBlock(blockStore, blockPair0.storedBlock, doubleSpends.t2);
+        wallet.receiveFromBlock(doubleSpends.t2, blockPair2.storedBlock, AbstractBlockChain.NewBlockType.SIDE_CHAIN, 0);
+        wallet.reorganize(blockPair0.storedBlock, 
+                Lists.newArrayList(blockPair1.storedBlock), 
+                Lists.newArrayList(blockPair2.storedBlock));
+        assertDead(doubleSpends.t1);
+        assertSpent(doubleSpends.t2);
+        assertDead(t1b);
+    }
+
```
schildbach: One empty line.

schildbach:
I just commented some tiny style issues. It's no big deal, I can change those things when merging.

oscarguindzberg:
Thanks @schildbach, I just fixed everything you suggested.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1817,6 +1842,9 @@ private void receive(Transaction tx, StoredBlock block, BlockChain.NewBlockType
             log.info("  <-pending");
 
         if (bestChain) {
+            boolean wasDead = dead.remove(txHash) != null;
```
mikehearn: I don't fully understand this. DEAD means it's been double spent by something in the chain already. So processing a new block on the best chain should never resurrect a transaction. No?

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1842,7 +1870,7 @@ private void receive(Transaction tx, StoredBlock block, BlockChain.NewBlockType
                 // Ignore the case where a tx appears on a side chain at the same time as the best chain (this is
                 // quite normal and expected).
                 Sha256Hash hash = tx.getHash();
-                if (!unspent.containsKey(hash) && !spent.containsKey(hash)) {
+                if (!unspent.containsKey(hash) && !spent.containsKey(hash) && !dead.containsKey(hash)) {
```
mikehearn: Again - is this change right? If a tx appears in a side chain and is considered dead on the main chain, it should not go back into pending.

oscarguindzberg: That is exactly what this new check is doing. It checks a tx is NOT dead before calling commitTx(tx). If tx was dead, it will remain dead.

mikehearn: You're right, sorry.

mikehearn:
Could you rebase this?

core/src/main/java/org/bitcoinj/core/Wallet.java
oscarguindzberg: I disagree. I know miners "in general" remove a tx from their memory pool if a tx in conflict with it is included in a block in the best chain. 
The case I am dealing here is what happens if a miner does not remove that tx in conflict from its memory pool (because she haven't received the new best block yet or she is doing an attack or some other reason) and mines a competitive block including the tx considered dead by the rest of the network and then another block on top of it. In that scenario, the network will accept this 2 new blocks as part of the best chain, the original best block will be considered a fork, the tx that is in the fork will be dead and the tx that was dead will be considered valid.
Am I missing something here?

oscarguindzberg:
Just rebased from master.
@mikehearn Please check my response to your comment on line 
https://github.com/oscarguindzberg/bitcoinj/blob/forward-double-spends2/core/src/main/java/org/bitcoinj/core/Wallet.java#L1845

core/src/test/java/org/bitcoinj/core/WalletTest.java
oscarguindzberg: fixed

core/src/main/java/org/bitcoinj/core/Wallet.java
mikehearn: Resurrecting dead transactions is handled in the reorganize method, right? It basically just undoes the effect of each block on the old chain and then rolls forward on the new chain. So I thought that logic would handle this case.

oscarguindzberg: yes, receive() is called from several methods including reorganize(), so this is part of the reorganize logic.

In a previous version of this PR, the reorganize "undo" logic was:
1) txA is in conflict with txB -> both are in the pending pool, confidence marked for both of them as "in conflict".
2) txA gets into blockX that is part of the best chain -> txA goes to building, txB goes to dead.
3) blockY that competes with blockX is now the new best block. bloc Y does not includes txA nor txB -> txA and txB go back to the pending pool, confidence marked for both of them as "in conflict".
4) If blockY would have included txB or a block on top of blockY includes txB -> txA goes to dead, txB goes to building.

In that code, all the resurrecting code was part of the "undo" logic in reorganize()

Since you told me dead txs are not normally resurrected, I replaced step "3" with this logic:
3) blockY that competes with blockX is now the new best block. blockY does not includes txA nor txB -> txA goes back to the pending pool, txB remains dead.

Now, step "4" logic remains the same, but the "roll forward" logic (ie receive() method) should support the case where we are receiving in a best chain block, a tx that was dead, so that is why this dead handling code here.

mikehearn: Right, thanks for the explanation. It's been too long since I looked at this code and I'm forgetting the details of how it works.

schildbach:
I'm taking this over from Mike. From what I understand, this patch doesn't only support "double spend forwards" (a feature of Bitcoin XT) but also if you receive conflicting pending transactions from different nodes (Core or XT). I am right?

How will estimated balance be calculated in future if I receive two pending tx conflicting each other? Currently, it seems as if all conflicting transactions are counted into estimated balance.

I really like to merge this now, this PR has been waiting long enough. May I ask you for one last rebase?

oscarguindzberg:
I would need some time to review the code since I worked on this a while ago.  When would you need a response from me and the rebase?

schildbach:
There is no hurry at all.

schildbach:
@oscarguindzberg Don't get me wrong. Any changes to balance calculation can come in a later PR. I just wanted to know if this PR touches balance calculation in any way.

oscarguindzberg:
@schildbach Sorry for the delay, just rebased from master.

You are right, if I receive two pending tx conflicting each other, all conflicting transactions are counted into estimated balance.

Another thing I've been thinking in the last few months is there is a major limitation (I won't call it a bug) of this PR: It is almost useless for SPV wallets. They won't be notified of a txs double spending a tx that sends funds to the SPV wallet because the bloom filter just includes the wallet's keys (and a double spend tx probably does not has any reference to the wallet's keys). 
On top of solving that, to make it really useful for SPV wallets, the bloom filter should be updated to track double spend attacks against all the txs in the unconfirmed dependency chain

schildbach:
Removed trailing whitespace and merged. Thanks again for your contribution!

I think the limited usefulness for SPV wallets is ok. At least it helps for some cases of malleability. I had the rare case of wallets receiving two different versions of the same transaction both in pending state (the typical case was one in pending, then another one via a block). And as you say, the bloom filter handling can be improved to cover more cases.

I think a next step would be to do something about how conflicting transactions are counted into the estimated balance. A trivial change would be to simply not count them for now.

schildbach:
Oscar, I just noticed you were editing wallet/Protos.java manually. That file is generated, instead you'd edit wallet.proto and regenerate with `mvn generate-sources -DupdateProtobuf`. But don't worry, I've already pushed a commit on top that fixes the issue.

oscarguindzberg:
ok about protobuf.
Thanks for merging this.

PR #332, opened 2015-03-14
Adjust stall period to 15 seconds and min speed to 5 kB/s. And fix bugs.

schildbach:
The previous 6 seconds caused otherwise quick peers to be disconnected too often.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerGroup.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerGroup.java
@@ -1464,9 +1464,8 @@ public class PeerGroup implements TransactionBroadcaster {
     private class ChainDownloadSpeedCalculator extends AbstractPeerEventListener implements Runnable {
         private int blocksInLastSecond, txnsInLastSecond, origTxnsInLastSecond, stallWarning;
 
-        private static final int STALL_PERIOD_SECONDS = 6;
-        private static final int STALL_MIN_SPEED = 10;
-
+        private static final int STALL_PERIOD_SECONDS = 15;
+        private static final int STALL_MIN_SPEED = 5;
 
         @Override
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/PeerGroup.java
```diff
@@ -1482,9 +1481,9 @@ public synchronized void onBlocksDownloaded(Peer peer, Block block, @Nullable Fi
 
         @Override
         public synchronized void run() {
-            if (blocksInLastSecond > 1) {
```
mikehearn: It's >1 deliberately to avoid triggering this path when a long running connection hears about a newly solved block. I guess there should be a comment explaining this.

schildbach: Why should we ever want to suppress those messages? I was heavily confuzed by not seeing them in some cases.

mikehearn: Because it's meant to be used for tracking download speed. It doesn't tell you much of any use for a single block.

schildbach: Single blocks are also "download". If your goal is excluding blocks received by "inv" rather than "getdata" then the way to go is clearly use *that* info for the conditional, rather than introducing a subtle bug into the logging. As I said, I was missing the message and invested quite some time into debugging.

mikehearn: All blocks are requested in response to an inv even during chain download, so that wouldn't work.

But OK, if you feel strongly about this, let's set it to zero. I don't care much either way. The old code tried not to print download speeds once caught up with the chain head so the new code does too, that's the reason it's done this way.

mikehearn:
LGTM

schildbach:
Merged.

PR #333, opened 2015-03-16
Allow disabling bloom filters on the peer group without using a FullPrunedBlockChain

oscarguindzberg:
On 37coins we handle lots of wallets in the same server, so the overhead of using bloom filters seems to be more than the benefit. The traditional way to avoid bloom filters (Using a FullPrunedBlockChain) is an overkill for us.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerGroup.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerGroup.java
@@ -243,6 +243,9 @@ public class PeerGroup implements TransactionBroadcaster {
     /** The default timeout between when a connection attempt begins and version message exchange completes */
     public static final int DEFAULT_CONNECT_TIMEOUT_MILLIS = 5000;
     private volatile int vConnectTimeoutMillis = DEFAULT_CONNECT_TIMEOUT_MILLIS;
+    
+    /** Whether bloom filter support is enabled when using a non FullPrunedBlockchain*/
+    private volatile boolean vBloomFilteringEnabled = true;
 
     /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/PeerGroup.java
```diff
@@ -243,6 +243,8 @@ public void onPeerDisconnected(Peer peer, int peerCount) {
     /** The default timeout between when a connection attempt begins and version message exchange completes */
     public static final int DEFAULT_CONNECT_TIMEOUT_MILLIS = 5000;
     private volatile int vConnectTimeoutMillis = DEFAULT_CONNECT_TIMEOUT_MILLIS;
+    
+    private boolean bloomFilteringEnabled = true;
```
mikehearn: Make this volatile please, or set it under the peer group lock. Otherwise we're violating the JMM. 

Also you may as well check it directly where the decisions are made and avoid the method call (it's not that big a deal on any reasonable JVM but it can't hurt to access directly).

mikehearn: Oh yes, in bitcoinj volatile fields start with a v, as in vBloomFilteringEnabled. Just a bit of hungarian notation to help us see what's safe and what's not when reading the code.

mikehearn:
Have you measured how much of a performance win this is?

oscarguindzberg:
I just squashed my commits

oscarguindzberg:
I am making a couple of changes and haven't tested performance on this one isolated.

mikehearn:
This has been merged, thanks!

PR #337, opened 2015-03-22
Run core tests in alphabetical order

kmels:
The [default order](http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#runOrder) in maven-surefire-plugin is 'filesystem' which may produce different results across platforms


schildbach:
I can attest this fixes the test failures for me. FWIW, the default of this setting is "filesystem", so Mike is probably right and the outcome of this test depended on the order of files created.

mikehearn:
I'm not sure whether the existence of this setting is genius or madness, but OK, let's go with alphabetical. Perhaps in future we should make it random so once we finally de-flake the tests, Travis can help us find hidden ordering constraints.

PR #338, opened 2015-03-22
Make maven-javadoc-plugin print a lot less clutter.

schildbach:


mikehearn:
Looks good. Feel free to commit small obviously correct things like this directly in future, no need for a pull request.

schildbach:
Merged.

Thanks for your trust. I personally prefer PR because even if the change seems to be obvious, it doesn't hurt to have at least one additional pair of eyes on it. Even the tiniest changes have broken things in subtle ways.

PR #339, opened 2015-03-22
Flexible DKC

devrandom:
* Allows subclassing of DeterministicKeyChain, with factory for deserialization
* Allow overriding of derivation path

To consider:
* Should the derivation path be set in the builder, rather than through subclassing?

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
@@ -112,6 +112,7 @@ public class DeterministicKeyChain implements EncryptableKeyChain {
     // is somewhat arbitrary but can be useful for audits. The first number is the "account number" but we don't use
     // that feature yet. In future we might hand out different accounts for cases where we wish to hand payers
     // a payment request that can generate lots of addresses independently.
+    // The account path may be overridden by subclasses.
     public static final ImmutableList<ChildNumber> ACCOUNT_ZERO_PATH = ImmutableList.of(ChildNumber.ZERO_HARDENED);
     public static final ImmutableList<ChildNumber> EXTERNAL_SUBPATH = ImmutableList.of(ChildNumber.ZERO);
     public static final ImmutableList<ChildNumber> INTERNAL_SUBPATH = ImmutableList.of(ChildNumber.ONE);
@@ -366,6 +367,9 @@ public class DeterministicKeyChain implements EncryptableKeyChain {
         return new DeterministicKeyChain(accountKey, seedCreationTimeSecs);
     }
 
+    /**
+     * For use in {@link KeyChainFactory} during deserialization.
+     */
     protected DeterministicKeyChain(DeterministicSeed seed, @Nullable KeyCrypter crypter) {
         this.seed = seed;
         basicKeyChain = new BasicKeyChain(crypter);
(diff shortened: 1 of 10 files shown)


mikehearn:
The above comments aren't exhaustive, let's think about the approach a bit first.

Is subclassing the right way to go here? It requires that we introduce a factory interface+class. Could we instead just have some kind of plugin object that does transformations on key paths? To what extent can the logic about what hierarchy layout to use encapsulateable? If we can extract it out into a separate class then it seems like it should be easier to use that API as you could just have a Bip44Hierarchy class or whatever, and it can be clearly documented in the javadocs what it's for.

devrandom:
The approach you suggest sounds interesting.  I am would still want subclassing DKC because it's useful for other things.  Here is how we use it:

* Adjusting bloom filter handling (we use p2sh even for non-multisig)
* Vending of keys outside of KCG (we have two active keychains, one for assets and one for bitcoin)
* Derivation path

I look into the encapsulated hierarchy too.

devrandom:
I rebased the stray commit out of existence, which made some of your line comments disappear since they were attached to specific commits...  sorry about that.  I did address all the line comments, including javadocs.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -1227,7 +1253,7 @@ public DeterministicSeed getSeed() {
         ImmutableList.Builder<DeterministicKey> keys = ImmutableList.builder();
         for (ECKey key : getKeys(true)) {
             DeterministicKey dKey = (DeterministicKey) key;
-            if (dKey.getPath().size() > 2) {
+            if (dKey.getPath().size() == getAccountPath().size() + 2) {
```
mikehearn: Hm is this change right? It's gone from everything below a certain path to just that single level.

core/src/test/java/org/bitcoinj/wallet/DeterministicKeyChainTest.java
```diff
@@ -78,6 +79,86 @@ public void getKeys() throws Exception {
     }
 
     @Test
+    public void deriveAccountOne() throws Exception {
+        long secs = 1389353062L;
+        DeterministicKeyChain chain1 = new DeterministicKeyChain(ENTROPY, "", secs) {
```
mikehearn: You could use AccountOneChain here, no?

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
devrandom: I think the old code was a bit too loose.  We know what level the leaves are, so referring to that specific level is better.

core/src/test/java/org/bitcoinj/wallet/DeterministicKeyChainTest.java
devrandom: Agreed, forgot to refactor after I created the class.

mikehearn:
This is merged, thanks! As it touches the key management code quite heavily, we will need some manual testing to go with the unit tests. Have you done any?

devrandom:
Great.  I'll rebase the wallet that motivated this PR and do some manual
tests...

On 2015-06-03 03:52 AM, Mike Hearn wrote:
> This is merged, thanks! As it touches the key management code quite
> heavily, we will need some manual testing to go with the unit tests.
> Have you done any?
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitcoinj/bitcoinj/pull/339#issuecomment-108296614>.
> 

-- 
devrandom / Miron

devrandom:
I did a test with said wallet, and got a match after restoring from seed.  Probably a good next step is testing with Schildbach's.

mikehearn:
OK, Andreas has pushed a version of his wallet that includes this code. I'll close this, thanks.

PR #340, opened 2015-03-23
Add readObject method to initialise listeners object which would be null...

ManfredKarrer:
... in case the confidence object gets created from Serialisation.

I got a null pointer for the listeners object from a serialised transaction. It was a transaction form the own wallet.


mikehearn:
Hm I wonder if we should still support serialization. I'm not really thinking about it any more these days so such bugs are likely to crop up again. Still, this one LGTM.

schildbach:
Merged.

PR #341, opened 2015-03-27
2 - Added unit tests to cover some (not all) red spots in the Cobertura report

haxwell:


mikehearn:
This is looking much better, thanks!

A few comments:

* Variable naming: please try to avoid names like "sut" or "rtn" that don't mean anything (or only mean something to you). The rest of the code uses names like "tx" which is obviously short for transaction, "result", "transaction" etc.
* The newTransaction method puts a blank line between every statement. That just uses up vertical screen space for no real good reason and makes less code fit on the screen. No need for it.
* There are many cases where you don't need easymocks. For instance in the last test you're making a mock Script object. You could as well just use ScriptBuilder to create an OP_RETURN script or something else that doesn't match. Generally I prefer to avoid mocking tools like EasyMock when possible because they don't play nicely with refactoring and it's easy to end up missing bugs because the tests aren't actually testing the interactions between things.

Could you maybe touch up a few of these things and then I'll merge, thanks!

BTW some of the code style conventions are documented here:

https://bitcoinj.github.io/coding-conventions

haxwell:
@mikehearn 

I have changed the variable names as you suggested. I also condensed some code where it made sense, collapsing one-line blocks, removing some single use variable declarations, etc. Finally, the case you mentioned, with the variable mockString, I made use of the ScriptBuilder there instead of creating the one-off mock object. Good call.

I will look into removing the other usages of mocks, as well. I know in some cases they are necessary in order to ensure a specific branch of the code is covered, but at other times they indicate a refactoring would be useful, to remove the need for the mock. I will create another pull request when I'm more clear on this part.

haxwell:
Just FYI, I have the aforementioned "other pull request" ready, but it builds on this one. Once this one is all in and done, I'll submit the next one.

mikehearn:
Cool but why not just squash the two different pull requests together, as they're affecting the same code?

haxwell:
In for a penny, in for a pound, right?

I did think about that this morning, but I guess I didn't want you to think I was just trying to explode on the scene and rewrite everything! Also didn't want you to reject the add'l TransactionTest coverage because the other refactoring code was too over the top.

But I will code more boldly from now on! Starting with Pull Request #344 .

Closing this one.

haxwell:
Bump!

mikehearn:
Merged, thanks!

haxwell:
No, thank you! w00t!

PR #342, opened 2015-03-27
bugfix for incomplete message headers at the end of the buffer

oscarguindzberg:
There is a bug when there is a incomplete message header at the end of the buffer 

This is the description of the bug:

Let's say PeerSocketHandler.receiveBuffer() is called with a Buffer of 65536 bytes which contains:
a) the last part of a Block message (65530 bytes) 
b) the first 6 bytes of the next Block message.

Part "a" is processed using a `largeReadBuffer`.
While processing part "b" https://github.com/bitcoinj/bitcoinj/blob/0a6f901b23524af1ee178174a35d06de3d0e29fd/core/src/main/java/org/bitcoinj/core/PeerSocketHandler.java#L149 throws `BufferUnderflowException` because it can not deserialize the whole message.
Since `i==0`, it attempts to use `largeReadBuffer`
Then https://github.com/bitcoinj/bitcoinj/blob/0a6f901b23524af1ee178174a35d06de3d0e29fd/core/src/main/java/org/bitcoinj/core/PeerSocketHandler.java#L157 also throws BufferUnderflowException because it can not even read the whole header. 
Then `ProtocolException` is thrown https://github.com/bitcoinj/bitcoinj/blob/0a6f901b23524af1ee178174a35d06de3d0e29fd/core/src/main/java/org/bitcoinj/core/PeerSocketHandler.java#L168
That exception will cause bitcoinj to disconnect the Peer.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerSocketHandler.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerSocketHandler.java
@@ -140,6 +140,7 @@ public abstract class PeerSocketHandler extends AbstractTimeoutHandler implement
                         processMessage(serializer.deserializePayload(header, ByteBuffer.wrap(largeReadBuffer)));
                         largeReadBuffer = null;
                         header = null;
+                        firstMessage = false;
                     } else // ...or just returning if we don't have enough bytes yet
                         return buff.position();
                 }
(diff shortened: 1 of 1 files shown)


mikehearn:
This looks good to me!

PR #343, opened 2015-03-31
Make sure connectionClosed is executed just once

oscarguindzberg:
This is a bugfix for https://code.google.com/p/bitcoinj/issues/detail?id=611
I understand this solution is not ideal: it adds another lock object (using the main Peer lock causes circular locks) and we should avoid calling Peer.connectionClosed() several times instead of this "hack".
I just created the PR to open the debate.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -288,17 +288,29 @@ public class Peer extends PeerSocketHandler {
         }
     }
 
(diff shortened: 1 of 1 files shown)


mikehearn:
Yeah that's not a real fix - the real question is why are the connection closed events happening multiple times? The real fix should be at a higher level.

schildbach:
@oscarguindzberg I'm inclined to not merging this, because it's not the real fix.

oscarguindzberg:
ok.

PR #344, opened 2015-04-03
Refactor The TransactionBag Dependency in Transaction.java; Improve TransactionTest coverage

haxwell:
I moved all methods that required a TransactionBag parameter from Transaction to TransactionUtils. I believe this is a more correct placement because when the method is in Transaction, it requires a TransactionBag (ie: a Wallet) to be passed in. So, for TransactionTest, that means a mock. Also, Wallet is the only code that called these methods. So they all support/enhance Wallet's functionality, and not Transaction's. Transaction is not, then, the best place for them.

My original purpose sprung from trying to remove the mocks necessary to test Transaction.isConsistent(). But that mock would always be required, as a TransactionBag of some sort must be passed in to exercise that method's code. I did, however, notice that the implementation of the method did not depend on any private/protected parts of the Transaction object. And further it was only called by Wallet. It could be moved to a neutral location (along with the mock necessary to test it) and Wallet could call it there, keeping all functionality it had, and the mock presently in TransactionTest.isConsistent() would be gone.

There's more to say, but I will let the code talk... I've been in my head on this for days.. do let me know what you think.. I think this is the right move, but it may need some polish.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -17,25 +17,42 @@
 
 package org.bitcoinj.core;
 
+import static com.google.common.base.Preconditions.checkState;
+import static org.bitcoinj.core.Utils.doubleDigest;
+import static org.bitcoinj.core.Utils.reverseBytes;
+import static org.bitcoinj.core.Utils.uint32ToByteStreamLE;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectOutputStream;
+import java.io.OutputStream;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import javax.annotation.Nullable;
+
 import org.bitcoinj.core.TransactionConfidence.ConfidenceType;
 import org.bitcoinj.crypto.TransactionSignature;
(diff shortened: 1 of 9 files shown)


mikehearn:
So, firstly, I agree that these methods shouldn't really be in Transaction. Combining wallet state and Transaction like that was clearly a mistake (from years ago) and we need to divorce it. Part of the wallet changes in the upcoming release are making a start on that.

However. This patch moves public methods. That will break people's code, for no real value. Moving stuff from Transaction to TransactionUtils is just churn and we try to avoid that. We could move the non-public methods of course, but then again, they could also just move to be private/package protected utility methods inside Wallet itself, right? I'm not sure why they need to be put into a new class.

Refactoring libraries is always rather tricky for this sort of reason, and as bitcoinj has grown and acquired users I've done my best to make the upgrade path between releases as smooth as we can whilst still evolving the library. That's one reason why some stuff is still kind of junky/messy.

Given that the *only* purpose of this change is unit tests for relatively small methods, I'm not sure I want to accept it as-is. The cost seems higher than the benefits. If you left the public methods alone, perhaps that'd be better.

haxwell:
You know, I did think about that.. I should have listened to my self.. :)

Regarding why the new class, my thought was that yeah, Wallet is the only thing using those methods now, but the methods are all centered around TransactionBag, so some other implementation of TransactionBag may come around (one that didn't inherit from Wallet. It could happen), and we'd have the same problem.

And I originally deprecated the methods, and then called the new class, rather than deleting them. But then I thought, fuck it. Code Boldly. and I guess I got a bit carried away with Ctrl-D. Note to self!

So, I guess since the refactor is basically (justly) denied, then we're back to #341, yes? All the changes you had requested were made in the most recent iteration of that PR.

PR #345, opened 2015-04-07
IsPubOnly changes

mikehearn:
@schildbach This pull incorporates your change and another one on top that renames some methods and tweaks the call sites to do what they actually expected. Review is appreciated.


schildbach:
Thanks for helping out with this. I included this PR (and mine) in #292.

PR #348, opened 2015-04-29
Makes UTXO use property Script instead of scriptBytes and addressType.

MeLoco1:
There were two fields in UTXO: scriptBytes and addressType. 

The fields substituted with Script script.
All code usages of UTXO getters changed.

FullPrunedBlockChain.getScript() changed to avoid returning null.

Cosmetic changes of UTXO and FullPrunedBlockChain.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/FullPrunedBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/FullPrunedBlockChain.java
@@ -44,10 +44,12 @@ import static com.google.common.base.Preconditions.checkState;
  * serve the full block chain to other clients, but it nevertheless provides the same security guarantees as a regular
  * Satoshi client does.</p>
  */
-public class FullPrunedBlockChain extends AbstractBlockChain {    
+public class FullPrunedBlockChain extends AbstractBlockChain {
     private static final Logger log = LoggerFactory.getLogger(FullPrunedBlockChain.class);
-    
-    /** Keeps a map of block hashes to StoredBlocks. */
+
+    /**
+     * Keeps a map of block hashes to StoredBlocks.
+     */
     protected final FullPrunedBlockStore blockStore;
 
     // Whether or not to execute scriptPubKeys before accepting a transaction (i.e. check signatures).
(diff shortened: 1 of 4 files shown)


MeLoco1:
There is property in Wallet class 
@Nullable volatile private UTXOProvider vUTXOProvider;

But the setter was erroneous: 
    public void setUTXOProvider(@Nullable FullPrunedBlockStore vUTXOProvider)
changed to 
    public void setUTXOProvider(@Nullable UTXOProvider vUTXOProvider)

core/src/main/java/org/bitcoinj/core/FullPrunedBlockChain.java
```diff
@@ -164,18 +176,20 @@ public VerificationException call() throws Exception {
         }
     }
 
-    /** Get the {@link Script} from the script bytes or null if it doesn't parse. */
-    @Nullable
+    /**
+     * Get the {@link Script} from the script bytes or null if it doesn't parse.
```
mikehearn: This JavaDoc is now out of date. Any reason you changed it? An empty Script object is rather unusual for bitcoinj and I don't think we produce them anywhere else.

mikehearn:
Thanks, this looks like a nice cleanup.

I'm pretty sure the answer is "yes", but is this change backwards compatible with existing databases? It's just a simple code change that'd be required, yes?

Re: the change to make getScript never return null. Is this a good idea? Perhaps the exception should just propagate instead?

core/src/main/java/org/bitcoinj/core/FullPrunedBlockChain.java
MeLoco1: When I made my changes and ran the tests without changing the method I had gotten NullPointerException in H2FullPrunedBlockChainTest.AbstractFullPrunedBlockChainTest: 
org.bitcoinj.store.DatabaseFullPrunedBlockStore.addUnspentTransactionOutput(DatabaseFullPrunedBlockStore.java:965)

The error occurs because I started to use 
s.setBytes(5, out.getScript().getProgram());
instead of 
s.setBytes(5, out.getScriptBytes());

There are two methods in this class: getScript() and getScriptAddress() starts from 182 line. The last one is returning blank string if the script is null or script can't be parsed to address. If you analyze both methods using , you will find  they only hides an error and swallows the exceptions. I'd like to refactor it to avoid using the methods. I don't know how to deal with NullPointerException mentioned above. I assume there is something with test data. But the test is to complicated for me to make changes. 

So I made changes to getScript() to return not null. At least now method behaves exactly as getScriptAddress(). 

It's not only place where exceptions are swallowed. Check  AbstractFullPrunedBlockChainTest, line 116 

JavaDoc fixed.

MeLoco1:
Yes, the code is backwards compatible with existing databases.

MeLoco1:
In regards to getScript(), I've added line notes.

mikehearn:
Merged, thanks!

MeLoco1:
There is one unmerged commit.

mikehearn:
Which one? I don't see it.

Also, weirdly I don't see your comment on the github web page ...

MeLoco1:
This one:
In UTXO property scriptBytes changed to class Script.

mikehearn:
It's here: https://github.com/bitcoinj/bitcoinj/commit/bee1873e1a52c2707dcba3964921d45ee61265bf

MeLoco1:
That's wierd. 
I've just checked 0.13-SNAPSHOT and the commit is not included. 

It might be due to failed Travis CI build
https://github.com/bitcoinj/bitcoinj/pull/348#issuecomment-97745321
http://screencast.com/t/yUY3askPiy

mikehearn:
You mean from Maven Central? I only refresh the snapshot builds occasionally, it's done manually not by Travis. I just did it now so see if that fixes it.

MeLoco1:
That's ok now, thank you.

PR #350, opened 2015-05-18
Add hooks for altcoin support, AuxPoW support classes

rnicoll:
Re-opening as changes need discussion. The URI factory is a bit... verbose, as an approach, but does neatly provide space for a modified factory where applicable to an altcoin. Could alternatively parse URIs in NetworkParameters, but cautious of making those classes too long.

Thoughts?

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -829,9 +829,11 @@ public abstract class AbstractBlockChain {
     private static final Date testnetDiffDate = new Date(1329264000000L);
 
     /**
-     * Throws an exception if the blocks difficulty is not correct.
+     * Throws an exception if the block's difficulty is not correct.
+     *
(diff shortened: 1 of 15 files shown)


mikehearn:
Comments:

* They all look good except the AuxPoW stuff - please drop that commit. The rest can go in.
* Please match the JavaDoc style: don't bother with @returns tags because the primary doc line (first sentence) should say this clearly already and the duplication is annoying. Also don't put any text after the /**
* You seem to be adding "this." prefixes a lot where they aren't necessary. They're highlighted red by github ... could you remove them please? It looks like your IDE/tool may be excessively fond of explicitly qualifying names, what are you using?

mikehearn:
Also I am not sure what you mean with URI factory. It might be better for your altcoin library to just provide an AltcoinURI class that is independent of BitcoinURI, if it uses URIs that are significantly different. I don't see any new factories in these commits though.

rnicoll:
That should be all of those changes applied. Apologies, found a better solution than using a factory class, forgot to update comment.

mikehearn:
Merged, thanks!

PR #352, opened 2015-05-19
Initialize context in BlockTest and CoinbaseBlockTest

rnicoll:
Seeing occasional complaints about Context not being constructed, in BlockTest and CoinbaseBlockTest, which this resolves.


mikehearn:
Merged, thanks!

PR #353, opened 2015-05-26
Use ip for seednode

ManfredKarrer:


mikehearn:
@schildbach what do you think?

schildbach:
Hu, where have you got that IP from? It should be 78.46.152.204, no?
I'm ok with adding that IP, it will not change anytime soon.
This PR changes pom dependencies it should not change.

ManfredKarrer:
That is strange I did a ping earlier but get another IP now:

iMac:~ mk$ ping testnet-seed.bitcoin.schildbach.de
PING testnet-seed.bitcoin.schildbach.de (188.165.246.217): 56 data bytes
64 bytes from 188.165.246.217: icmp_seq=0 ttl=54 time=37.716 ms
64 bytes from 188.165.246.217: icmp_seq=1 ttl=54 time=32.001 ms
64 bytes from 188.165.246.217: icmp_seq=2 ttl=54 time=35.262 ms

I agree to better not merge it as it also did not solve the problem.

schildbach:
If you ping it will resolve to random Bitcoin nodes, because that's what it is supposed to do.

mikehearn:
I think this patch is not doing what you think it's doing ;)

PR #354, opened 2015-05-30
Add support for replacing all codes in a MonetaryFormat as a single operation

rnicoll:
Add support for replacing all codes in a MonetaryFormat as a single operation, for example for altcoins where the default codes do not apply.
Replace monetary codes map with an array as a simpler data type given that the data storage needs are small and constant.
.

### DIFF
--- a/core/src/main/java/org/bitcoinj/utils/MonetaryFormat.java
+++ b/core/src/main/java/org/bitcoinj/utils/MonetaryFormat.java
@@ -25,10 +25,9 @@ import static com.google.common.math.LongMath.divide;
 import java.math.RoundingMode;
 import java.text.DecimalFormatSymbols;
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
-import java.util.Map;
 
 import org.bitcoinj.core.Coin;
 import org.bitcoinj.core.Monetary;
@@ -61,6 +60,8 @@ public final class MonetaryFormat {
     /** Currency code for base 1/1000000 Bitcoin. */
     public static final String CODE_UBTC = "µBTC";
 
+    public static final int MAX_DECIMALS = 8;
+
     private final char negativeSign;
     private final char positiveSign;
     private final char zeroDigit;
@@ -69,7 +70,7 @@ public final class MonetaryFormat {
     private final List<Integer> decimalGroups;
     private final int shift;
     private final RoundingMode roundingMode;
-    private final Map<Integer, String> codes;
+    private final String[] codes;
     private final char codeSeparator;
     private final boolean codePrefixed;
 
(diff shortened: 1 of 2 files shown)


schildbach:
Hmm, why don't you just initialize one "base" MonetaryFormat instance and derive from that? Example:

In your Constants.java you define `MonetaryFormat FORMAT = new MonetaryFormat().code(...).code(...).code(...).whateverElseYourAltcoinNeeds()` and then you customize and use the format at any point in the app with `Constants.FORMAT.minDecimals(2).whateverCustomizationYouNeed().format(...)`

Given that you only need the initialization code at one point I wonder if it's worth to introduce an array-based setter.

rnicoll:
It seemed a lot neater to make one call rather than ending up with a bunch of intermediary classes. 

Also avoids encoding knowledge of what the Bitcoin codes are, in another coin's library. Imagine a new 4th decimal place unit is introduced, and my library only replaces 0th, 3rd and 6th (current codes), it would be a bit of a mess. Could have a "removeAllCodes()" method, but doing a full single replacement seemed neater.

core/src/main/java/org/bitcoinj/utils/MonetaryFormat.java
```diff
@@ -233,15 +234,26 @@ public MonetaryFormat noCode() {
...
     /**
+     * Configure all currency codes as a batch.
+     * 
+     * @param codes
+     *            array of currency code representing the code at each shift
+     * from 0-8 decimal places. Use null where no specific code applies.
+     */
+    public MonetaryFormat codes(String[] codes) {
+        return new MonetaryFormat(negativeSign, positiveSign, zeroDigit, decimalMark, minDecimals, decimalGroups,
+                shift, roundingMode, Arrays.copyOfRange(codes, 0, MAX_DECIMALS), codeSeparator, codePrefixed);
```
schildbach: I think if codes.length > MAX_DECIMALS it should warn the caller, rather than silently capping the array.

schildbach:
There is .noCode(), I forgot it in my example. I was just thinking that .codes("BTC", null, null, "mBTC", null, null, "µBTC") ist't a much cleaner API. But having this doesn't hurt of course.

rnicoll:
Ah, missed noCode(), that solves it. I'll drop the new function and this can just be the map/array switch then. I'd like to come back to how MonetaryFormat changes are applied, but I'm thinking a more drastic change involving a factory class, so it's a whole different topic.

rnicoll:
Removed new function, and added a unit test for noCode() as my first version broke it when later trying to set codes.

mikehearn:
I'll let @schildbach handle this one, as he is more qualified anyway, being the GUI dev in the house ;)

schildbach:
@rnicoll Did you see my one code comment? It's just a little thing, but I thought I'd wait what you have to say to the idea before merging.

rnicoll:
@schildbach Should have been clearer, apologies - sidestepped the entire issue by switching to noCodes(), which solves the problem I was having. This is now just the switch to array storage on the backend because it's a little tidier (IMHO)

schildbach:
Merged, thanks!

PR #355, opened 2015-05-30
Add extension hooks for AuxPoW parsing

rnicoll:
This modifies Block to provide extension hooks for blocks with additional/alternative content to conventional Bitcoin blocks, for example altcoins with AuxPoW support. Specifically:

Make headerParsed, transactionParsed, headerBytesValid and transactionBytesValid protected so subclasses in the same package can access them.

Add constructor for use when a block is contained within another object (i.e. AuxPoW header)

Add parseTransactions() method which takes in a payload offset, so block parsers can indicate that transactions do not start at byte 80.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -90,16 +90,16 @@ public class Block extends Message {
     /** Stores the hash of the block. If null, getHash() will recalculate it. */
     private transient Sha256Hash hash;
 
-    private transient boolean headerParsed;
-    private transient boolean transactionsParsed;
+    protected transient boolean headerParsed;
+    protected transient boolean transactionsParsed;
 
-    private transient boolean headerBytesValid;
-    private transient boolean transactionBytesValid;
+    protected transient boolean headerBytesValid;
+    protected transient boolean transactionBytesValid;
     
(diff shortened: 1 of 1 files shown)


rnicoll:
This approach is proving significantly trickier than expected, as block storage is poorly suited to parsing different block types depending on the network parameters (in particular, the existing deserializer is network-agnostic).

Is there any general value in having alternative blocks supported? I'm inclined to make the Block class final, and add a generic "extended header" field to it which the network parameters are responsible for parsing as appropriate, although I'm not thrilled mixing parsing in multiple places. The alternative is to re-introduce the AuxPoW classes, but it would feel that we're fixing this to a very limited set of potential blocks.

Could have the network parameters provide the deserializer class, perhaps. Other concern is that if the block type can vary, it's logical to use generics on StoredBlock and BlockStore to track the content type... which leads to invasive changes throughout the code to propagate those types.

Thoughts?

rnicoll:
Have a solution, which puts the intelligence into BitcoinSerializer and has NetworkParameters handle creating an appropriate serializer. It's a more invasive change though, so will raise as a separate PR later.

mikehearn:
This seems to have moved to a different PR so I'll close this

rnicoll:
Wasn't meant to be part of that one (was trying to keep them separate so it wasn't too big), but must have branched off the wrong branch, and as it's there now, no point worrying about it.

PR #356, opened 2015-05-31
Handle deserialization via NetworkParameters for alternative format support

rnicoll:
Extract common interface from BitcoinSerializer to support alternative serializers, as well as dummy serializer for message classes which do not have their network parameters.
Thread BitcoinSerializer through messages in place of existing parseLazy and parseRetain properties.
Insert dummy serializer into message subclasses when deserialized by Java.
Replace calls to construct Block/Transaction classes from payloads, with calls to MessageSerializer, so alternative formats can be supported elegantly.
Minor documentation changes.


rnicoll:
This also leads on to a Hibernate-esque parsing framework if you wanted to go down that path, where rather than embedding the logic in each Message subclass, more is pushed into the serializer and controlled via annotations & reflection.

mikehearn:
I think we want to avoid reflection as part of the core system for now, for a few different reasons.

Are there any alt coins that don't use Satoshi's binary format? What is motivating this change, specifically?

mikehearn:
Ah, never mind, I see now that this pull is another attempt to do auxpow support. OK, will review this later when I have some time.

rnicoll:
Yup, it's AuxPoW enabling with a bunch of extra hooks while I'm there. I think the approach of tracking parseLazy/parseRetain via the serializer is a neater one, as well, though. In theory also enables weirder formats such as Myriad. Happy to cut this back if I've gone too far adding in hooks.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
```diff
@@ -420,4 +421,21 @@ public int getBip32HeaderPriv() {
      * networks.
      */
     public abstract boolean hasMaxMoney();
+
+    /**
+     * Return the default serializer for this network. This is a shared serializer.
+     * @return 
+     */
+    public final MessageSerializer getDefaultSerializer() {
```
mikehearn: Shouldn't this be synchronized?

mikehearn:
This looks good to me: the changes are all fairly mechanical.

However it's also kind of a big change. Though I think the risk of regressions is fairly low, I'm trying to ship 0.13 at the moment (kinda futile with the block size debate taking up so much time). I'd rather bump this change to immediately after 0.13 is out (ie. 0.14), so there's more time to find regressions.

How does that sound to you? Reasonable?

rnicoll:
Absolutely no hurry on this stuff, just need to know the general theory is sound, and can merge in a few months once I've had a chance to build stuff on the extension library.

So, sounds good.

mikehearn:
We've branched for 0.13 so this can be merged now. @schildbach, does it look OK to you?

schildbach:
It looks correct as far as I can tell. I can't say if the introduced interface is useful to altcoins or not – for this I need to trust Ross.

I'm not so happy about the separation of commits. Two commits seem to have mostly the same description – should they perhaps merged into one? JavaDoc corrections on the other hand could have been put on a different PR – that way we could have merged them separately already weeks ago.

schildbach:
I just saw that there is one inline comment by Mike which is not addressed I think. If you work on the PR, can you rebase on current master? – Thanks.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
rnicoll: In theory it's fine without, as creating two serializers isn't the end of the world, realised should just in case someone makes the serializer mutable.

rnicoll:
Rebased, Javadoc comments moved out, commits squashed, and added synchronization to the serializer construction method. Do let me know if there's any other problems.

rnicoll:
Rebased to bring it up to date with master.

schildbach:
Merged. Many thanks for bearing with us.

mikehearn:
Yes, thanks again.

PR #358, opened 2015-06-01
Fix issued receive keys & child num order

schildbach:
These two commits could be merged separately.


mikehearn:
LGTM

schildbach:
Thanks for the review! I clarified the comment a bit and merged.

PR #359, opened 2015-06-02
Payment channel fixes

fofanov:
Some fixes and tidying up of the payment channels code, with the intention of getting the examples working.

The current design of the StoredPaymentChannels{Client|Server}States requires them to have knowledge of the containing wallet. Currently the containing wallet is only passed to these states via deserialization. There are other cases when this is not possible, such as when a wallet file does not exist.

I have split out the loading of wallet extensions from the main wallet to make it easier to pass the wallet to newly created extensions that need it and changed the WalletAppKit to do this. I've tried to do this without changing too much of the existing design to prevent breaking existing code.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/core/Wallet.java
@@ -4465,6 +4465,7 @@ public class Wallet extends BaseTaggableObject implements Serializable, BlockCha
      */
     public void deserializeExtension(WalletExtension extension, byte[] data) throws Exception {
         lock.lock();
+        keychainLock.lock();
         try {
             // This method exists partly to establish a lock ordering of wallet > extension.
             extension.deserializeWalletExtension(this, data);
@@ -4474,6 +4475,7 @@ public class Wallet extends BaseTaggableObject implements Serializable, BlockCha
             extensions.remove(extension.getWalletExtensionID());
             Throwables.propagate(throwable);
         } finally {
+            keychainLock.unlock();
             lock.unlock();
         }
     }
--- a/core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
+++ b/core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
@@ -369,7 +369,12 @@ public class WalletAppKit extends AbstractIdleService {
             for (WalletExtension e : provideWalletExtensions()) {
                 wallet.addExtension(e);
             }
+
+            // Currently the only way we can be sure that an extension is aware of its containing wallet is by
+            // deserializing the extension (see WalletExtension#deserializeWalletExtension(Wallet, byte[]))
+            // Hence, we first save and then load wallet to ensure any extensions are correctly initialized.
             wallet.saveToFile(vWalletFile);
+            wallet = loadWallet(false);
         }
 
         if (useAutoSave) wallet.autosaveToFile(vWalletFile, 5, TimeUnit.SECONDS, null);
--- a/core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelClientState.java
+++ b/core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelClientState.java
@@ -216,7 +216,6 @@ public class PaymentChannelClientState {
         StoredPaymentChannelClientStates channels = (StoredPaymentChannelClientStates)
                 wallet.getExtensions().get(StoredPaymentChannelClientStates.EXTENSION_ID);
         channels.removeChannel(storedChannel);
-        wallet.addOrUpdateExtension(channels);
         storedChannel = null;
     }
 
@@ -450,7 +449,7 @@ public class PaymentChannelClientState {
         storedChannel.valueToMe = valueToMe;
         StoredPaymentChannelClientStates channels = (StoredPaymentChannelClientStates)
                 wallet.getExtensions().get(StoredPaymentChannelClientStates.EXTENSION_ID);
-        wallet.addOrUpdateExtension(channels);
+        channels.updatedChannel(storedChannel);
     }
 
     /**
@@ -486,7 +485,6 @@ public class PaymentChannelClientState {
         checkState(channels.getChannel(id, multisigContract.getHash()) == null);
         storedChannel = new StoredClientChannel(id, multisigContract, refundTx, myKey, valueToMe, refundFees, true);
         channels.putChannel(storedChannel);
-        wallet.addOrUpdateExtension(channels);
     }
 
     /**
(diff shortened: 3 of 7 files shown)


ollekullberg:
Since I've touched some of that code, I can help out proof-reading your fixes. 

The "Fixing NPE in ExamplePaymentChannelServer due to using unintialized" looks solid. I guess we never ran that example.

ollekullberg:
Re: "Fixing NPE in ExamplePaymentChannelServer" 

Initialization of payment channel extension together with wallet is hard, and I'm not sure I remember everything, but here we go:

At least in the old design, it SHOULD be perfectly OK to provide a StoredPaymentChannelClientStates extension without giving any wallet instance, like this:

  protected List<WalletExtension> provideWalletExtensions() {
                return ImmutableList.<WalletExtension>of(new StoredPaymentChannelClientStates(null));
  }

Another exemple how this works without WalletAppKit is MultiBitHD (Stroem version):         

    WalletExtension[] walletExtensions = new WalletExtension[]{
          new SendFeeDtoWalletExtension(),
          new MatcherResponseWalletExtension(),
          new WalletTypeExtension(),
          new StoredPaymentChannelClientStates(null)
      };

    Wallet wallet = new WalletProtobufSerializer().readWallet(BitcoinNetwork.current().get(), walletExtensions, walletProto);

Like you said, the wallet instance will be given (is mandatory) during the deserializeWalletExtension():

    @Override
    public void deserializeWalletExtension(Wallet containingWallet, byte[] data) throws Exception {
        lock.lock();
        try {
            checkState(this.containingWallet == null || this.containingWallet == containingWallet);
            this.containingWallet = containingWallet;
        ...
        }
    }

So, what to do when we don't have a wallet at all? I think this is the most common scenario:

1. Create a wallet instance, where you have the same impl of provideWalletExtensions() as above.
2. deserializeWalletExtension() will not be called, since there is nothing on disk yet.
3. Save the wallet immediatly, before you try to use the extension.
4. Now, when you re-read the wallet deserializeWalletExtension() will run, an the extension will get it's instance.

fofanov:
Thanks, I probably should have looked at how other wallets do this!

So are you saying that when creating a new wallet with extensions the best practice should be to save and then load before using the wallet? That would simplify my fixes actually. 
I think the reason I didn't consider that was the way in which the WalletAppKit currently returns a fresh wallet without the extensions knowing about the containing wallet (leading to the NPE).

```java
    private Wallet createOrLoadWallet(boolean shouldReplayWallet) throws Exception {
        Wallet wallet;

        maybeMoveOldWalletOutOfTheWay();

        if (vWalletFile.exists()) {
            wallet = loadWallet(shouldReplayWallet);
        } else {
            wallet = createWallet();
            wallet.freshReceiveKey();
            for (WalletExtension e : provideWalletExtensions()) {
                wallet.addExtension(e);
            }
            wallet.saveToFile(vWalletFile);
        }
       return wallet;
```
Doing the save and load would just be a case of reordering this function and I can get rid of the changes I've made to the extension loading. Does that sound better?

fofanov:
@ollekullberg based on your comments I've changed WalletAppKit to save and load any new wallet before use. This makes sense and also simplifies what I was trying to do. Thanks.

ollekullberg:
Looks OK to save and then load the wallet in AppKit, since it's only done when the wallet is created (not often).

About the potential deadlock: if the extension uses the keychain during desirialization, and someone else is accessing the wallet at the same time where they only want the keychainlock, it could be a deadlock. But is it likely that an extension should use the keychainlock during deserialize? I don't think StoredPaymentChannelClientStates does that?

fofanov:
The CycleDetectingLockFactory is picking up the potential deadlock.

When deserializing extensions, StoredPaymentChannelClientStates may call `Wallet#findKeyFromPubKey()`, so currently the lock ordering is StoredPaymentChannelClientStates > KeyChain. 5181cefcf2a4747dfe73d1a9304adc13577e0740 enforces the opposite ordering when printing the wallet and extensions, KeyChain > StoredPaymentChannelClientStates .

The fix is to lock the keychain first before deserializing extensions, so that the ordering becomes KeyChain > StoredPaymentChannelClientStates > KeyChain, which takes advantage of the re-entrant lock.

ollekullberg:
I give up, you win :-D

Now all three commits looks ok in my book.

mikehearn:
Maybe we should just improve the wallet extensions API so we don't need the round-trip hack? The extensions API has never been very good, it was thrown together quickly for this use case and in general isn't as flexible or intuitive as anyone would like.

ollekullberg:
One way would be to build a new extension API, a "WalletExpansion", that could live side by side with WalletExtension. The WalletExtension seems to work well for MultiBitHD when used to store extra info in the wallet, so there's no immediate need to retire it.

ollekullberg:
For payment channels, on the server side, the WalletExtension API is a poor fit mainly because of the storage bottleneck. One would typically not want to serialize and save the entire wallet just because a micropayment has been received, that is too slow. If the new API has a more liberal save method (like "save()"), then a DB could be used to store each single payment, and the number of write-wallet-to-disk operations can be reduced. Note that on the client side the current solution works, so it's tricky to find a generic solution.

mikehearn:
Yeah - perhaps the solution is to not use the server Stored...State class in cases where you need high throughput. 

I was thinking maybe there's a simple new method that can be added to WalletExtension to provide what's needed.  Would we really need a new API entirely? Or can the existing one be adjusted to fit?

ollekullberg:
Thanx Mike. If we write our own version of StoredPaymentChannelServerStates,  we also need to make a change in PaymentChannelServerState.updateChannelInWallet() so it can be configured to not write the wallet to disk. Maybe this can be done without messing everything up?

Maybe the only clean way is to make a WalletExpansion after all? We can subclass it to PaymentChannelServerExpansion, which has some abstract method "save()". Then PaymentChannelServerState.updateChannelInWallet() looks for the PaymentChannelServerExpansion, and if present, it calls paymentChannelServerExpansion.save() , otherwise proceed as normal.

mikehearn:
Ideally PaymentChannelServerState wouldn't assume it's using a Wallet object at all. It'd just ask the StoredPaymentChannelServerStates to save itself and that'd be good enough. I doubt there's any fundamental reason why it works the way it does: just that's what was easiest to code up at the time.

But it's been a while since I looked at this code. If you can find a way to do it more cleanly though that'd be fantastic.

mikehearn:
@ollekullberg Ping - what's the current status of this? Is there a quick, small improvement we can do to avoid the weird round-tripping of the wallet? If so, can we take it and leave a redesign of the extensions API for later?

ollekullberg:
We haven't found out anything quick and pretty (haven't had time to discuss it much here). The last thing we discussed, about server side performance enhancement, should probably be in its own PR anyway.

I suggest we accept @fofanov 's PR. (LGTM)

mikehearn:
Alright. I'll merge this after 0.13 as I don't want to change the new wallet codepath at this late stage.

mikehearn:
Actually, given that 0.13 was delayed already anyway, I changed my mind here. The change is relatively low risk so I've merged these patches. Thanks!

PR #360, opened 2015-06-02
Fixed the TestNet3Params.checkDifficultyTransitions method.

kparmar1:
Fixed the TestNet3Params.checkDifficultyTransitions method.


mikehearn:
Great catch, thanks for debugging this Kalpesh! Do you have a tip address?

PR #361, opened 2015-06-03
Fixed the size of the field 'header' MySQL fullstore

kparmar1:
Fixed the size of the field 'header' in the 'headers' table for the MySQL fullstore; blob changed to mediumblob.

If you want to manually update the MySQL, not delete the tables and re-sync, use the following SQL:-

ALTER TABLE headers MODIFY COLUMN header MEDIUMBLOB;


mikehearn:
Can you rebase on master, please?

Also: the change description doesn't describe the reason for breaking db compatibility. I assume there is  a good one though. Could maybe the code do the auto upgrade itself? Or does it not matter, will the old databases continue to work?

mikehearn:
Ah sorry - just saw your email.

My questions remain for mainnet deployments of this though. Will it break existing databases?

kparmar1:
Actually I’m note now sure that this isn’t some kind of knock on error.  Looking at the spec I would expect the header to be 80 chars ( in fact looking at my data this is what majority of the lengths are). I think this may require further investigation, maybe this is something peculiar to the test network?  At the moment I blindly fixed it, maybe not the best approach.

If this indeed is an error then yes it will effect current DB, although the code will work fine, until this occurrence happens of course.  I noticed this happened on a split chain (on testnet) - this is the point that the length was exceed.

2015-06-03_11:10:55.868 [NioClientManager] INFO  o.b.core.DownloadProgressTracker - Chain download switched to [127.0.0.1]:18333
664 2015-06-03_11:10:57.375 [NioClientManager] INFO  org.bitcoinj.core.AbstractBlockChain - Block is causing a re-organize
665 2015-06-03_11:10:57.376 [NioClientManager] INFO  org.bitcoinj.core.AbstractBlockChain - Re-organize after split at height 446572
2015-06-03_11:10:57.376 [NioClientManager] INFO  org.bitcoinj.core.AbstractBlockChain - Old chain head: 00000000001cb7f2115e7c19f29f7812adb76b8623016cbca44c    5963618ab055
667 2015-06-03_11:10:57.376 [NioClientManager] INFO  org.bitcoinj.core.AbstractBlockChain - New chain head: 00000000000029a6e9a66603c7cd2420ede49c81b53fcb378e15    08aa79bbc398
668 2015-06-03_11:10:57.376 [NioClientManager] INFO  org.bitcoinj.core.AbstractBlockChain - Split at block: 0000000000000a1c6a4f3392bd963ccd11f981c73b13c6110525    283fc99d20ac 

I don’t know the answer to this without further investigation. Maybe you have an idea as to why this header field would be so large ( > 80)

Looking at my test db the following are the header length.  It look like maybe the length really should *NOT* exceed 80.

Testnet max height.
446613

SELECT  length(header), count(*) FROM fullstore_test.headers group by length(header);

length(header), 	count(*)			Header Heights for these counts (where =1)
80				446614			N/A
642				1				446547
4399			1				446567
4654			1				446568
5923			1				446569
82964			1				446573     ****CAUSED ERROR****


If this is indeed an error maybe some kind of upgrade logic could be added to the database stores so that no manual intervention is required…

Maybe hold off on this fix (pull request) unless further investigation, or if anyone can think of why this maybe happening?

K

mikehearn:
I was wondering about that ... I really want to kick maintenance of this code over to someone else entirely though :-)

Bitcoin headers are always 80 bytes, ignoring the trailing transaction varint (but that should not be in the db at all, I think).

What extra junk is appearing in this table? There must be a bug if the records are not always 80 bytes.

kparmar1:
I'll investigate (debug) and see what the heck is going on.  This probably wasn't picked up as the other stores blob fields are not limited in the way MySQL blob fields are. When I implemented (MySQL store) I kept this field small because I made the assumption that headers were always 80 in length.  

I'll contribute when I can, but I'm a little time restricted (and a little skill :-).

kparmar1:
On the positive side I'm using the store again so it's def in my interest to find/fix any bugs :-)
Now I have to get into the shower... the time just vanishes  :-)

ollekullberg:
I don't think blob (or mediumblob) is the correct type for the header. Since we know that the length is 80 bytes, it should be:

BINARY(80)

I've been working with Oracle DBs (partly as a DBA) for 15 years. I know some MySQL too.

mikehearn:
OK, I'll let you guys figure this out as my MySQL knowledge is fairly minimal. BigTable on the other hand .... ;)

ollekullberg:
We can also change:

chainwork mediumblob NOT NULL

to:

chainwork VARBINARY(255) NOT NULL 

(VARBINARY(255) will always be enough for any real world BigInteger) 

Now, we got to the desired point where NO column in this table is a blob, and thus the table is much faster for the MySQL engine to handle.

kparmar1:
Thanks, cool that makes sense and probably why the sync was taking a lot longer that it needed to. I'll make these changes (unless you want to thats cool also). Hopefully there will be no impart on the rest of the stores - implementation is shared - shouldn't be problem.  

If/when I find the actual error that this initial change was supposed to address.

ollekullberg:
I looked at the code and saw that you use the same SQL/Java code in DatabaseFullPrunedBlockStore for the different DBs. This is actually problematic since PostgreSQL does not have the BINARY() data type. I have an idea, and I will do a pulll request to try to change all of the DBs in one go.

kparmar1:
This shouldn’t be a problem… the code to access (read/write) is shared so as long as the fields can be accessed via getBytes, setBytes  it should be fine (and they can). The creation of the tables/databases structures is *not* shared.

kparmar1:
This may not be a quick fix considering I can't replicate it :-( I've started another sync to testnet and hopefully I'll hit this error again ... probably tomorrow.

I must remember to backup my stores in future before I experiment  :-)

Olle you are still doing the datatype changes, if you are there are other fields to consider, txoutchanges, transactions (prob both blobs - future proof as well) and scriptbytes? 

This pull request I'm going to hmm pull as it's no longer relevant :-)

Anyway time for some not so deserved hot dogs :-)

mikehearn:
Enjoy the hotdogs! I always find they go well with mustard.

kparmar1:
Thanks I did - Mustard the the much maligned Ketchup :-)

Olle- In trying to recreate this other bug I did update the fields and you are correct  that there is a definite speed up- like feels almost twice as fast (I haven't scientifically measured this though).

ollekullberg:
Just created a pull request for the changes we talked about. I tested it as much as I could.

PR #362, opened 2015-06-04
MySQL Column types and a name change

ollekullberg:
Change a few column types so that MySQL can perform faster, and hints to the reader the actual size of the binary data. 

Also, the column "index" should be renamed since it is a keyword in some DBs. Hard to predict the effects of using keywords as column names.

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/DatabaseFullPrunedBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/DatabaseFullPrunedBlockStore.java
@@ -74,7 +74,7 @@ import java.util.*;
  * <table>
  *     <tr><th>Field Name</th><th>Type (generic)</th></tr>
  *     <tr><td>hash</td><td>binary</td></tr>
- *     <tr><td>index</td><td>integer</td></tr>
+ *     <tr><td>output_index</td><td>integer</td></tr>
  *     <tr><td>height</td><td>integer</td></tr>
  *     <tr><td>value</td><td>integer</td></tr>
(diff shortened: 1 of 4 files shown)


ollekullberg:
Oups. These 2 commits are actually one commit. Not sure how it ended up this way?

ollekullberg:
Ok, just saw that I forgot to remove 32 on two spots, for example:

private static final String CREATE_OUTPUTS_HASH_INDEX                       = "CREATE INDEX openoutputs_hash_idx ON openoutputs (hash(32)) USING btree

Will do a new PR.

mikehearn:
OK, closing this one then

kparmar1:
Just a note..
It's cool that you changed the column name (index) but this does mean everyone will be forced to recreate their stores (plus potentially change their code). I suppose they should anyway (to get the benefit) but if they didn't want to this would no longer be an option.
For myself I don't really care.

mikehearn:
Yeah I'd rather avoid that if possible. We haven't had any reports of this causing problems so far, even though I agree it'd be better to avoid it. I suspect "index" is a remarkably common column name actually given the generality of the word and concept.

kparmar1:
FYI: I didn't change it when I refactored (deliberately) and it seemed that the only special treatment that was required was for MySQL (wasn't a big deal).

ollekullberg:
Aha, are these tables already running in prod somewhere? It will be painful to type 'index' every time in MySQL, if it's not in prod I would change it now (before it's too late).

PR #363, opened 2015-06-04
Change data types in tables and column namechange.

ollekullberg:
Change data types in tables from blob to binary 

Column namechange: index -> output_index (since index is a keyword in many dbs)

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/MySQLFullPrunedBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/MySQLFullPrunedBlockStore.java
@@ -39,28 +39,28 @@ public class MySQLFullPrunedBlockStore extends DatabaseFullPrunedBlockStore {
     private static final String CREATE_SETTINGS_TABLE = "CREATE TABLE settings (\n" +
             "    name varchar(32) NOT NULL,\n" +
             "    value blob,\n" +
-            "    CONSTRAINT `setting_pk` PRIMARY KEY (name)\n" +
+            "    CONSTRAINT setting_pk PRIMARY KEY (name)  \n" +
             ")\n";
 
     private static final String CREATE_HEADERS_TABLE = "CREATE TABLE headers (\n" +
-            "    hash blob NOT NULL,\n" +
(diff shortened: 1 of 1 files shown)


mikehearn:
Are there really databases that puke on a column named index? If so, why do none of our existing db impls seem to have a problem with it? I'd rather not invalidate everyone's databases in 0.13

ollekullberg:
Ok, didn't know ppl were already using this code. I'll remove the name change then and send a new PR.

mikehearn:
I'm not sure many people are using the MySQL DB except Kalpesh, but the other dbs are being used for sure.

ollekullberg:
Mike, you can close this one.

ollekullberg:
I cleaned up a bit among the backticks. They are needed around "index", so that it won't be a keyword, but I think we should be consistent to only use them there (or on all names, but that seems like a lot of work).

mikehearn:
@kparmar1 You good with this one? It looks OK to me but I'm a non expert reviewer when it comes to MySQL

kparmar1:
LGTM.

kparmar1:
I've just noticed something.....

The fields transactions and txoutchanges *shouldn't* be changed to a blob in undoableblocks (leave as it). This is too small.

kparmar1:
I think this maybe my fault as I previously mentioned that there are other fields (to consider/ look over). I actually meant these were ok, and my fatal error was to use the word blob (I meant that in a generic way - ie not binary).

ollekullberg:
Ah, so you suspect that a txchanges field can be bigger than 64 MB (=2^16 bytes)? In that case we _should_ use mediumblob.

ollekullberg:
Changed back into mediumblob.

mikehearn:
@kparmar1 Looking for re-review of this since the code changed. Can you LGTM it?

kparmar1:
LGTM.

(FYI I'm still trying to figure out why the sync on testnet is failing at times i.e when header > 80 - not releated to this of course)

mikehearn:
OK, merged. Thanks for owning this code.

PR #364, opened 2015-06-09
Wallet size larger than 64MB 

jdojff:
At the moment protobuf deserialization does not allow wallet to be larger than 64MB.
Just created this quick fix and pull request to show the issue. Better solution than increasing limit to 512MB should be implemented.

```
2015-06-07 08:03:02.125 ERROR [MSC service thread 1-12] stderr:71 - Caused by: com.google.protobuf.InvalidProtocolBufferException: Protocol message was too large.  May be malicious.  Use CodedInputStream.setSizeLimit() to increase the size limit.
2015-06-07 08:03:02.125 ERROR [MSC service thread 1-12] stderr:71 - 	at com.google.protobuf.InvalidProtocolBufferException.sizeLimitExceeded(InvalidProtocolBufferException.java:110)
2015-06-07 08:03:02.126 ERROR [MSC service thread 1-12] stderr:71 - 	at com.google.protobuf.CodedInputStream.refillBuffer(CodedInputStream.java:755)
2015-06-07 08:03:02.126 ERROR [MSC service thread 1-12] stderr:71 - 	at com.google.protobuf.CodedInputStream.readRawByte(CodedInputStream.java:769)
2015-06-07 08:03:02.126 ERROR [MSC service thread 1-12] stderr:71 - 	at com.google.protobuf.CodedInputStream.readRawVarint32(CodedInputStream.java:373)
2015-06-07 08:03:02.127 ERROR [MSC service thread 1-12] stderr:71 - 	at com.google.protobuf.CodedInputStream.readBytes(CodedInputStream.java:318)
2015-06-07 08:03:02.127 ERROR [MSC service thread 1-12] stderr:71 - 	at org.bitcoinj.wallet.Protos$TransactionOutput.<init>(Protos.java:5790)
2015-06-07 08:03:02.127 ERROR [MSC service thread 1-12] stderr:71 - 	at org.bitcoinj.wallet.Protos$TransactionOutput.<init>(Protos.java:5735)
2015-06-07 08:03:02.127 ERROR [MSC service thread 1-12] stderr:71 - 	at org.bitcoinj.wallet.Protos$TransactionOutput$1.parsePartialFrom(Protos.java:5833)
2015-06-07 08:03:02.128 ERROR [MSC service thread 1-12] stderr:71 - 	at org.bitcoinj.wallet.Protos$TransactionOutput$1.parsePartialFrom(Protos.java:5828)
[...]

PR #366, opened 2015-06-10
Allowed wallet size increase more than 64MB

jdojff:
Updated pull request based on #364 
To increase allowed wallet size override ```WalletAppKit.getWalletSizeLimit``` method.

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/WalletProtobufSerializer.java
+++ b/core/src/main/java/org/bitcoinj/store/WalletProtobufSerializer.java
@@ -76,6 +76,8 @@ public class WalletProtobufSerializer {
     private static final Logger log = LoggerFactory.getLogger(WalletProtobufSerializer.class);
     /** Current version used for serializing wallets. A version higher than this is considered from the future. */
     public static final int CURRENT_WALLET_VERSION = Protos.Wallet.getDefaultInstance().getVersion();
+    // 512 MB
+    private static final int WALLET_SIZE_LIMIT = 512 * 1024 * 1024;
     // Used for de-serialization
     protected Map<ByteString, Transaction> txMap;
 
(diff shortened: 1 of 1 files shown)


schildbach:
In this version, you made the limit configurable by another `parseToProto` variant. Why don't you increase the max proto size for all cases, like in your last PR? Existing API users will not automatically benefit they way it is done now.

Do you have an example wallet that exceeds the 64 MB? It would be useful to add it to our test cases.

btw. You do not need to submit new PRs if you want to change something. Just amend and/or rebase your branch, then force-push over to GitHub.

jdojff:
I didn't want to introduce magic numbers, so I made configurable method. But there are cases where old ```parseToProto``` is used, however there should be no big problem to use the new one.
Other solution would be to get wallet file size and set the limit automatically.
I have wallet larger than 64MB, but I can't share it. Basically there are ~2k private keys, new key chain with ~5k addresses, and a lot of their transactions.
Ok, thanks for the advice.

mikehearn:
Yeah don't worry about magic numbers, just increase the limit for everyone.

BTW a 64mb wallet is huge. That isn't going to perform well no matter what. After 0.13 I'm planning on slimming down the amount of data in the wallet file by quite a bit, in preparation for shifting to a database/log-structured wallet file. But you're pushing the limits of what bitcoinj was designed to handle right now. Just be aware of that.

jdojff:
Please check it now

core/src/main/java/org/bitcoinj/store/WalletProtobufSerializer.java
```diff
@@ -76,6 +76,10 @@
     private static final Logger log = LoggerFactory.getLogger(WalletProtobufSerializer.class);
     /** Current version used for serializing wallets. A version higher than this is considered from the future. */
     public static final int CURRENT_WALLET_VERSION = Protos.Wallet.getDefaultInstance().getVersion();
+
+    // 512 MB
+    private static final Integer walletSizeLimit = 512 * 1024 * 1024;
```
schildbach: Can you make this an int (native type) and change the name to allcaps "WALLET_SIZE_LIMIT"?

jdojff:
Updated

schildbach:
Merged, thanks!

PR #988, opened 2015-06-12
Use FixedPointCombMultiplier for mul by G

peterdettman:
I noticed a few places where CURVE.getG() is being multiplied by some scalar directly, which won't actually use the fixed-point multiplication for which you have the nice precomputed tables. These changes fix that (ECDSA and other high-level primitives already do this internally of course).

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/core/ECKey.java
@@ -39,6 +39,7 @@ import org.spongycastle.crypto.signers.ECDSASigner;
 import org.spongycastle.crypto.signers.HMacDSAKCalculator;
 import org.spongycastle.math.ec.ECAlgorithms;
 import org.spongycastle.math.ec.ECPoint;
+import org.spongycastle.math.ec.FixedPointCombMultiplier;
 import org.spongycastle.math.ec.FixedPointUtil;
 import org.spongycastle.math.ec.custom.sec.SecP256K1Curve;
 import org.spongycastle.util.encoders.Base64;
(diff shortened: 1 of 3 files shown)


schildbach:
Interesting. Do you have any real world numbers on performance improvement?

core/src/main/java/org/bitcoinj/core/ECKey.java
```diff
@@ -439,10 +440,25 @@ public boolean isWatching() {
...
+     */
+    public static ECPoint publicPointFromPrivate(BigInteger privKey) {
+        /*
+         * TODO: FixedPointCombMultiplier currently doesn't support scalars longer than the group order,
+         * but that could change in future versions.
+         */
+        if (privKey.bitLength() > CURVE.getN().bitLength()) {
+            privKey = privKey.mod(CURVE.getN());
+        }
+        return new FixedPointCombMultiplier().multiply(CURVE.getG(), privKey);
```
mikehearn: How expensive is constructing this FixedPointCombMultiplier object? Would it make sense to stash it in a static field?

peterdettman:
@schildbach I'm not a bitcoin user really, so no "real" data, but in isolated testing on my laptop:

    FixedPointUtil.precompute(CURVE.getG(), 12);    // as used in bitcoinj
    ...
    new FixedPointCombMultiplier().multiply(CURVE.getG(), n);

is slightly over 3 times faster (for 100000 calls to multiply()) than:

    CURVE.getG().multiply(n);

I'd be interested if you could see any performance improvement in loading a wallet or whatever else would be a good test.

core/src/main/java/org/bitcoinj/core/ECKey.java
peterdettman: It's empty constructor, no fields, pure implementation, so I doubt it will make any difference!

mikehearn:
Merged, thanks!

PR #990, opened 2015-06-12
Improve the way compression is handled

peterdettman:
The main change here is the getPointWithCompression method, which swaps the "compression" flag in an ECPoint, as necessary. For going from uncompressed to compressed, this should be much faster, for the reverse direction about the same as before.

It uses the deprecated ECCurve.createPoint, but that's fine since you'll need to change over to "self-managed" compression at some point in any case, as the TODO in ECKey quite rightly points out:

    // TODO: Move this class to tracking compression state itself.
    // The Bouncy Castle guys are deprecating their own tracking of the compression state.


peterdettman:
Rebased to allow merging.

mikehearn:
Merged, thanks!

PR #992, opened 2015-06-12
Improvements in ECKey.extractKeyFromASN1

peterdettman:
- Check there's no extra data following the key
- Confirm the ASN.1 tag number is correct on the publicKey field
- Explicitly check the encoding is one we support
- Rework the sanity check to avoid redundant parsing/multiplies


peterdettman:
Looks like this was merged, should it be closed?

schildbach:
Yes, thanks!

PR #994, opened 2015-06-15
Fix: Block headers received (DB fullstore) that are >80 in length

kparmar1:
Fix: Block headers received (save to fullstore) that are >80 in length i.e includes tx. Tx are removed when saving the header.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -665,14 +665,15 @@ public abstract class AbstractBlockChain {
             // Walk in ascending chronological order.
             for (Iterator<StoredBlock> it = newBlocks.descendingIterator(); it.hasNext();) {
                 cursor = it.next();
-                if (expensiveChecks && cursor.getHeader().getTimeSeconds() <= getMedianTimestampOfRecentBlocks(cursor.getPrev(blockStore), blockStore))
+                Block cursorBlock = cursor.getHeader();
+                if (expensiveChecks && cursorBlock.getTimeSeconds() <= getMedianTimestampOfRecentBlocks(cursor.getPrev(blockStore), blockStore))
                     throw new VerificationException("Block's timestamp is too early during reorg");
                 TransactionOutputChanges txOutChanges;
                 if (cursor != newChainHead || block == null)
                     txOutChanges = connectTransactions(cursor);
                 else
                     txOutChanges = connectTransactions(newChainHead.getHeight(), block);
-                storedNewHead = addToBlockStore(storedNewHead, cursor.getHeader(), txOutChanges);
+                storedNewHead = addToBlockStore(storedNewHead, cursorBlock.cloneAsHeader(), txOutChanges);
             }
         } else {
             // (Finally) write block to block store
--- a/core/src/main/java/org/bitcoinj/store/DatabaseFullPrunedBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/DatabaseFullPrunedBlockStore.java
@@ -619,7 +619,7 @@ public abstract class DatabaseFullPrunedBlockStore implements FullPrunedBlockSto
             s.setBytes(1, hashBytes);
             s.setBytes(2, storedBlock.getChainWork().toByteArray());
             s.setInt(3, storedBlock.getHeight());
-            s.setBytes(4, storedBlock.getHeader().unsafeBitcoinSerialize());
+            s.setBytes(4, storedBlock.getHeader().cloneAsHeader().unsafeBitcoinSerialize());
             s.setBoolean(5, wasUndoable);
             s.executeUpdate();
             s.close();


mikehearn:
Ahhhh, good debugging! 

Perhaps the getHeader() method should always internally ensure it's stripped out the transactions? Or the object should at construct time? Sounds like the method name can be misleading to me. This fix would probably work but might just leave junk lying around for others to trip on.

kparmar1:
Yeah the method name is a a little misleading. The problem with doing this is that it is used in both ways (maybe in other parts of the system too??) Sometimes the tx are required and sometimes not. 

Looking into the other stores (at least). They handle the fact that the header may contain tx. But the fullstore will save the tx from the header too (in a split).
But the basic problem is that it's used in both ways.... (sometime needed sometime not- hopefully this is only for the stores and then it should be an easier fix).

Maybe another method? getHeader(boolean includeTx) ???? Not really too sure here...
I'll could just give it a go and see if it doesn't break everything (hopefully not)....

kparmar1:
Hmm actually you got me/ There is no getHeader() method, maybe that's the problem. Currently the block is passed around with a copyToHeader (so the block is just really the header only)..

Anyway I'll look into it :-)

mikehearn:
There should probably be two methods. I guess the problem is that StoredBlock was originally SPV only. When full verification mode was added, the getHeader method wasn't renamed. Do you have some spare cycles for rationalising the design a bit?

kparmar1:
Ok I guess I could take this on, but I'm not entirely clear as to what I'm trying to do-Yet (I'll have to look at SPV side of things to understand further)...

kparmar1:
Ok so is this correct....
StoredBlock should now contain the block (not header) and the method getHeader() actually to return a header (using copyTo...) Ok. The thing is though IN --> you have the block (plus tx) but OUT (from either store ) you don't.

SPV def not. Fullstore also not (currently, well in a split yes). But now StoredBlock doesn't make sense as it's really just the header... hmmm Somehow consolidate the two? Ok I'm thinking out loud here a little.

Maybe you could elaborate a little further on the second method (and rename). Maybe it will help me to understand a little better. I think I'm missing something....

mikehearn:
No, keep going :) Fresh eyes on the design is always welcome, especially after several years of evolution.

You are correct that StoredBlock comes back out with different data to what goes in. Perhaps it should actually be named StoredHeader and then it should never have any transactions data at all? But then perhaps if someone wanted to build a block store that did actually store every transaction, that'd be a problem.

Perhaps there should be:

StoredHeader > StoredBlock

where StoredBlock inherits from StoredHeader? Then the distinction could be reflected in the API of the FullPrunedBlockStore interface. I realise that internally, the Block class also has the same duality, it can either be a header or a block with transactions too. It's kind of a mess everywhere :)

kparmar1:
Yeah I think that makes sense to me.  That could work fine.... hmmmm The type hierarchy would be fine, as you said, distinction and allow for both the underlying implementations to do their thing.

kparmar1:
FYI: I haven't forgotten about this.  I've made a start but was unable to continue this week but I'll get back to it this weekend.

kparmar1:
Update:-

Ok this is not proving to be particularly easy...  StoredBlock is everywhere and the interdependencies are difficult to refactor within the current structure and as soon as you change one thing it breaks something else.  That's not a problem in itself as I've just been worked through it.

Basically I've got the code to compile and test successful like 2 sec ago :-) ( StoredHeader < StoredBlock as you suggested) but still need to look over it tomorrow, too tired now to continue :-) I'm sure there are many other parts of the lib that will need to be updated too (even if the tests are passing).

I'll try to make further progress... and hopefully have something you can look over soon, even if it's not useable (current state).

kparmar1:
Here is my first attempt at separating the block and header out. I'm not sure if this really improves things or just makes things a little more complicated (internally at least)?

I've got the commit on my fork 
https://github.com/kparmar1/bitcoinj/tree/storedblock_refactor

I didn't want to create a pull request as this is really just experimentation (can if it's easier?) to see if you think this is even worth doing (when you have seen it). 
Anyway when you get time some please take a look (hopefully no glaring errors, can be corrected - sry in advance- just in case). Maybe I'm totally way off base in terms of what you were thinking- if so that's fine and no harm done :-).

mikehearn:
Hmm - interesting commit, thanks for exploring this! I like the idea of parameterising BlockStore in some way. The change is much larger and more invasive than I'd imagined. It may be inevitable though.

For such a large change I think maybe I'd rather do it myself, being guided by your work here. I'd probably split it up into a series of smaller commits, use different names and maybe change more of the API as I go.

So for now I guess I'll take the bugfix patch, close this pull req, and file an issue to remind me to clean this stuff up.

kparmar1:
Yes it def was many more changes than I though it would be. I agree with everything you said 100% (if anything I did helped even a little then I’m happy :-) 
K

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
```diff
@@ -672,7 +672,8 @@ private void handleNewBestChain(StoredBlock storedPrev, StoredBlock newChainHead
                     txOutChanges = connectTransactions(cursor);
                 else
                     txOutChanges = connectTransactions(newChainHead.getHeight(), block);
-                storedNewHead = addToBlockStore(storedNewHead, cursor.getHeader(), txOutChanges);
+                storedNewHead = addToBlockStore(storedNewHead,
+                        cursor.getHeader().transactions == null ? cursor.getHeader() : cursor.getHeader().cloneAsHeader(), txOutChanges);
```
mikehearn: Is the branch ever taken here? We call connectTransactions just above it, so It hink "cursor.getHeader().transactions == null" is always false. This should just be cloneAsHeader()

You could also pull out the repeated cursor.getHeader() calls into a variable here. IntelliJ can do it for you automatically.

core/src/main/java/org/bitcoinj/store/DatabaseFullPrunedBlockStore.java
```diff
@@ -619,7 +619,9 @@ protected void putUpdateStoredBlock(StoredBlock storedBlock, boolean wasUndoable
             s.setBytes(1, hashBytes);
             s.setBytes(2, storedBlock.getChainWork().toByteArray());
             s.setInt(3, storedBlock.getHeight());
-            s.setBytes(4, storedBlock.getHeader().unsafeBitcoinSerialize());
+            s.setBytes(4, storedBlock.getHeader().getTransactions() == null ?
```
mikehearn: Ditto. This is a full store: getTransactions() should never be null, I think? So it needs to always be run through cloneAsHeader()

kparmar1: Bloody hell agreed  (Sry that was a lesson in how to muck up every line you change - however after this weekend it is now so much clearer).
I’ve changed and pushed.
K

> On 23 Jun 2015, at 20:40, Mike Hearn <notifications@github.com <mailto:notifications@github.com>> wrote:
> 
> In core/src/main/java/org/bitcoinj/store/DatabaseFullPrunedBlockStore.java <https://github.com/bitcoinj/bitcoinj/pull/994#discussion_r33082519>:
> 
> > @@ -619,7 +619,9 @@ protected void putUpdateStoredBlock(StoredBlock storedBlock, boolean wasUndoable
> >              s.setBytes(1, hashBytes);
> >              s.setBytes(2, storedBlock.getChainWork().toByteArray());
> >              s.setInt(3, storedBlock.getHeight());
> > -            s.setBytes(4, storedBlock.getHeader().unsafeBitcoinSerialize());
> > +            s.setBytes(4, storedBlock.getHeader().getTransactions() == null ?
> Ditto. This is a full store: getTransactions() should never be null, I think? So it needs to always be run through cloneAsHeader()
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/bitcoinj/bitcoinj/pull/994/files#r33082519>.
> 




Kalpesh Parmar
Tel:  +44 7474575321 (UK)

mikehearn:
Thanks, merged.

PR #999, opened 2015-06-22
Hash-related code cleanup

amichair:
Here are a bunch of refactoring ideas around the Utils/Sha256Hash classes.

Mainly what bothered me were:

- inconsistent use of synchronized MessageDigest singleton vs creating them on the fly - it looks arbitrary which is used when, and it it unclear that a synchronized singleton has better performance than creating instances. If there really is any performance issue here, a ThreadLocal would probably be best. The current code refactoring would make that change easy (encapsulated) as well. This one is more than just refactoring.

- separation between Utils and Sha256Hash methods which seem somewhat arbitrary as well.

- a few other simplifications I came across in this code.

I've split up the refactoring into a few steps so it's easier to follow and modify if necessary.

Let me know which parts you like and which not so much...

Method names are open to improvement as well :-)

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AlertMessage.java
+++ b/core/src/main/java/org/bitcoinj/core/AlertMessage.java
@@ -114,7 +114,7 @@ public class AlertMessage extends Message {
      * doesn't verify, because that would allow arbitrary attackers to spam your users.
      */
     public boolean isSignatureValid() {
-        return ECKey.verify(Utils.doubleDigest(content), signature, params.getAlertSigningKey());
+        return ECKey.verify(Sha256Hash.calcDoubleHashBytes(content), signature, params.getAlertSigningKey());
     }
 
     @Override
--- a/core/src/main/java/org/bitcoinj/core/AlertMessage.java
+++ b/core/src/main/java/org/bitcoinj/core/AlertMessage.java
@@ -114,7 +114,7 @@ public class AlertMessage extends Message {
      * doesn't verify, because that would allow arbitrary attackers to spam your users.
      */
     public boolean isSignatureValid() {
-        return ECKey.verify(Sha256Hash.calcDoubleHashBytes(content), signature, params.getAlertSigningKey());
+        return ECKey.verify(Sha256Hash.hashTwice(content), signature, params.getAlertSigningKey());
     }
 
     @Override
--- a/core/src/main/java/org/bitcoinj/core/Base58.java
+++ b/core/src/main/java/org/bitcoinj/core/Base58.java
@@ -153,7 +153,7 @@ public class Base58 {
         byte[] bytes = copyOfRange(tmp, 0, tmp.length - 4);
         byte[] checksum = copyOfRange(tmp, tmp.length - 4, tmp.length);
         
-        tmp = Utils.doubleDigest(bytes);
+        tmp = Sha256Hash.calcDoubleHashBytes(bytes);
         byte[] hash = copyOfRange(tmp, 0, 4);
         if (!Arrays.equals(checksum, hash)) 
             throw new AddressFormatException("Checksum does not validate");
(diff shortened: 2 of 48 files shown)


amichair:
@schildbach yeah I expected someone to say something about this one :-) It's a matter of taste, I suppose. I prefer less vertical space and less return points/branches and find the flow and logic more readable, but I know for some it's more readable when it's explicit. Whatever y'all say. fwiw, you can still shave off a line by inlining that cast.

amichair:
For naming, another possibility that came up is:

    Sha256Hash wrap(String hexString);
    Sha256Hash wrap(byte[] rawHash);
    Sha256Hash of(byte[] content);
    Sha256Hash of(File content);
    Sha256Hash twiceOf(byte[] content);

etc. for returning Sha256Hash instances (and making the constructor non-public for consistency), and

    byte[] hash(byte[] content);
    byte[] hash(byte[] content, int offset, int length);
    byte[] hashTwice(byte[] content);
    byte[] hashTwice(byte[] content, int offset, int length);

etc. for working directly with bytes.

Another design option is removing the byte[] methods altogether, and always working with Sha256Hash instances, e.g. Sha256.of(content).reverse().getBytes(). This makes for a simpler api and consistent usages, but slightly more verbose calling code and one or two short-lived object creations, if it means anything for performance.

Food for thought.

mikehearn:
I quite like this proposal (with both byte[] and Sha256Hash methods).

amichair:
I updated the code with the proposed API above.

Also added one way of simplifying the reversing code. There are other ways of doing this, e.g. with a single reverse() OOP-chaining-style method, but each has its pros and cons. We can go either way.

mikehearn:
Very nice! Merged! If you see other code cleanups, have time and would like to work on it, please do go ahead!

schildbach:
This whole PR is not nice to review, because it changes things back and forth multiple times. In future, can you try to squash together if you fix things from the same PR?

We write commit messages in present tense and make whole sentences, including a punctuation (dot at the end).

schildbach:
@amichair About that equals/hashCode method, I'm ok if you zip everything in one expression that just contains either && or ||. But as soon as you mix these, errors are almost unevitable. I'd prefer if we at least keep the typical reference equality and type checks on separate lines. For the rest, we started using Guava: https://code.google.com/p/guava-libraries/wiki/CommonObjectUtilitiesExplained

amichair:
@schildbach thanks for the feedback. I'll try to answer the points you raise.

Regarding commit messages, I peeked at the history at the time and saw inconsistent format so I just picked one and went with it. I'll follow the guidelines you mention from now on, and I recommend you add it to the [coding conventions](https://bitcoinj.github.io/coding-conventions) so others will do the same.

As for squashing, I actually thought it would be easier to read when separated rather than squashed, since there are a lot of changes going on, and also wasn't sure all of the steps would be accepted so it would be easier to rewrite this way (I separated them into logical atomic steps - at least logical to me :-) ). The last two commits that re-renamed stuff came later, after a discussion with @mikehearn culminated with the suggested API which I summed up in the comment above, and after his approval I committed it. Then added another independent but related simplification step. I think squashing everything would have made it harder to re-review everything, but if you guys prefer squashed PRs, of course I'll follow your lead.

About the equals expressions, I brought up your remark in the irc channel to see if I should undo the change, and @mikehearn was there and said he prefers the shorter one-line expression and it is more in line with the coding conventions page which stresses reducing vertical space. I suggested to bring it up for a vote, and a funny discussion ensued :-)

Finally, there's https://github.com/bitcoinj/bitcoinj/pull/1003. I split that one up too into separate commits, since one is about logic-changing bugs/fixes, and the second is more about refactoring/simplification/consistency. I'm guessing you'll love half of it (using Objects.hashCode alot) and hate the other half (using the short logic expressions for equals).

I suggest you and mikehearn (and any other devs in the house?) discuss it to make sure you agree on some standard, and I'd be happy to change things according to whatever you decide - I don't want to step on anyone's toes. But either way, I think it's important to make it all more consistent, since it's easier for bugs to hide in inconsistent code (as the first commit there might suggest).

We can discuss on irc if more appropriate...

PR #1000, opened 2015-06-23
remove unused imports and unused local variables

p-:
Hi
I removed some unused imports and unused local variables. (I did not modify the protos or variables used in tests.) I think it's important to get rid of unused code as it may shadow other problems.


mikehearn:
Thanks!

PR #1001, opened 2015-06-24
reduced unnecessary UnsupportedEncodingException handling

amichair:
... and little cleanups.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Base58.java
+++ b/core/src/main/java/org/bitcoinj/core/Base58.java
@@ -16,7 +16,6 @@
 
 package org.bitcoinj.core;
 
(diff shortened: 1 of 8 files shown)


mikehearn:
This doesn't actually compile. Unfortunately the orchid module cannot depend on core. So you may wish to introduce the same method into the Orchid Utils equivalent (in com.subgraph.orchid.misc)

amichair:
Hmmm... it looks like running 'mvn clean package' on root folder doesn't build orchid (it's not one of the aggregate modules), which is why I got no errors. I undid the changes in the orchid module (if it's some sort of copy or close fork to an upstream project, probably best not to change it much here, no?)

mikehearn:
Thanks!

PR #1002, opened 2015-06-26
Fix consistency issues with equals/hashCode in LazyECPoint

peterdettman:
The existing implementation could give inconsistent results, potentially returning differing hashcodes for, or inequality between, points with the same logical value. I wasn't able to see any problems actually happening during unit testing, but clearly there's room for serious (and subtle) problems.

This PR changes equals and hashCode to both be based on a "canonical encoding", which is chosen as the compressed encoding, mainly because it can still be generated cheaply if the original encoding is uncompressed (the reverse is not true). Also that encoding is an external standard and not BC-specific.

I've made no attempt at cacheing the canonical encoding or the hashcode, but that's obviously a possibility if performance becomes an issue.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
+++ b/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
@@ -168,20 +168,16 @@ public class LazyECPoint {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
-        LazyECPoint point1 = (LazyECPoint) o;
-        if (bits != null && point1.bits != null)
-            return Arrays.equals(bits, point1.bits);
-        else
-            return get().equals(point1.get());
+        return this == o || o != null && getClass() == o.getClass()
+            && Arrays.equals(getCanonicalEncoding(), ((LazyECPoint)o).getCanonicalEncoding());
     }
 
     @Override
     public int hashCode() {
-        if (bits != null)
-            return Arrays.hashCode(bits);
-        else
-            return get().hashCode();
+        return Arrays.hashCode(getCanonicalEncoding());
+    }
+
+    private byte[] getCanonicalEncoding() {
+        return getEncoded(true);
     }
 }


peterdettman:
Re-formatted to address line comments from @amichair, and rebased.

mikehearn:
Yes this stuff is horribly subtle. Point compression makes everything more complicated, I hate it. I wish we'd been using compressed points right from the start. At some point we might want to force a key rotation on any wallet that has uncompressed keys/addresses in it and forbid the import of them just so we can clear out the support for it.

I don't think we will ever attempt to mix up compressed and uncompressed points (if we do, that is of course a different bug) .... despite their logical equality the way Bitcoin turns points into addresses means that effectively a compressed/uncompressed point are different even if they share coordinates.

Still, you're right that this seems like an improvement, even if it shouldn't impact any existing users. So it LGTM. @schildbach may wish to see if it causes any performance regression on Android though.

peterdettman:
Someone may also want to just check ECKey.PUBKEY_COMPARATOR; it could be suffering from a similar problem with it's equality and/or ordering, but I don't understand what it's being used for well enough to say.

schildbach:
@troggy Can you comment on Peter's last comment?

troggy:
Haven't worked with bitcoinj for awhile, so I might not be correct here.
ECKey.PUBKEY_COMPARATOR is used to order pubkeys in redeem scripts for P2SH outputs. Bitcoinj itself doesn't use this ordering while spending (instead we are looking for a key position in the script and use this position to place signature at the proper place), so this PR should not affect P2SH spending in bitcoinj. But it could affect other clients willing to spend P2SH UTXOs with scripts created by bitcoinj, if these clients rely on keys order (why should they?).
Frankly, I don't remember why exactly this sorting is needed. @devrandom Dou you remember the reason we have public keys sorted in P2SH redeem scripts?

devrandom:
@troggy the reason for sorting is so that the P2SH address is consistent, regardless of who constructs the redeem script.   It makes interoperability a bit easier between different implementations of P2SH apps. This is standardized here: https://github.com/bitcoin/bips/pull/146

schildbach:
I didn't see any noticable impact on Android performance. I'd say let's merge this for now, we can undo it later if it shows up in profilers. Agree?

mikehearn:
OK, sounds good

schildbach:
Merged, thanks!

PR #1003, opened 2015-06-28
Equals-hashCode-copmareTo fixes

amichair:
The law of bug clustering strikes again! Following the bug reported in https://github.com/bitcoinj/bitcoinj/pull/1002, I had a look at other equals/hashCode/compareTo implementations, and indeed found several issues:

- DumpedPrivateKey incorrectly uses byte array reference rather than its content in hashCode
- GetBlockMessage/GetHeadersMessage equals/hashCode ignore list ordering - it's undocumented, but I assume that explains the current implementation. I added a comment, but perhaps the correct fix is to use Sets instead of Lists, if that's how they are to be treated?
- TransactionInput hashCode/equals include scriptSig which is a WeakReference (and scriptBytes already contains the same data anyway)
- TransactionOutPoint hashCode does not call getIndex/getHash accessors, so they are not parsed properly
- TransactionOutput null checks on parent cause equals to not be reflexive and hashCode to break the equals contract (what is the desired behavior here regarding parent?)
- Sha256Hash,Utils.Pair,Wallet.TxOffsetPair,ChildNumber,ExponentialBackoff implement compareTo that is not consistent with equals. I just added a comment documenting this for now, since some of these are privately used in only one or two methods and I assume they work correctly, but if necessary I can add equals/hashCode so that the contracts are respected.

I fixed the ones that were obvious, but for the others I don't know what the intent or use-case was for the original code or what the desired behavior should be. If you let me know, I'll update the code.

I also added a second commit which is supposed to contain only code cleanup rather than fixing anything. Behavior might be slightly changed (e.g. actual hashCode values) but not in a meaningful way.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -809,8 +809,7 @@ public class Block extends Message {
     public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
-        Block other = (Block) o;
-        return getHash().equals(other.getHash());
+        return getHash().equals(((Block)o).getHash());
     }
 
     @Override
--- a/core/src/main/java/org/bitcoinj/core/BloomFilter.java
+++ b/core/src/main/java/org/bitcoinj/core/BloomFilter.java
@@ -357,9 +357,7 @@ public class BloomFilter extends Message {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         BloomFilter other = (BloomFilter) o;
-        return hashFuncs == other.hashFuncs &&
-               nTweak == other.nTweak &&
-               Arrays.equals(data, other.data);
+        return hashFuncs == other.hashFuncs && nTweak == other.nTweak && Arrays.equals(data, other.data);
     }
 
     @Override
--- a/core/src/main/java/org/bitcoinj/core/Coin.java
+++ b/core/src/main/java/org/bitcoinj/core/Coin.java
@@ -18,6 +18,7 @@ package org.bitcoinj.core;
 
 import org.bitcoinj.utils.MonetaryFormat;
 import com.google.common.math.LongMath;
+import com.google.common.primitives.Longs;
 
 import java.io.Serializable;
 import java.math.BigDecimal;
@@ -256,14 +257,9 @@ public final class Coin implements Monetary, Comparable<Coin>, Serializable {
 
     @Override
     public boolean equals(final Object o) {
-        if (o == this)
-            return true;
-        if (o == null || o.getClass() != getClass())
-            return false;
-        final Coin other = (Coin) o;
-        if (this.value != other.value)
-            return false;
-        return true;
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        return this.value == ((Coin)o).value;
     }
 
     @Override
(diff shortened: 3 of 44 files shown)


amichair:
Rebased and updated equals implementations to all match the format we agreed upon (3-line).

schildbach:
Awesome, thanks for working on these cleanups.

amichair:
Regarding TransactionOutput, I wasn't sure if this was just a simple bug or there was some other intention here. So it's correct to use regular equality here? i.e. two TransactionOutputs are equal if they either have an equal parent or their parents are both null?

And regarding the hashCode bug, I suppose we can just remove index() from the mix. Sound ok?

amichair:
I updated TransactionInput/TransactionOutput with the suggested fixes. Still waiting for confirmation that this is the intended behavior, since the previous one (in both classes) looks like it may have had reasoning behind it, even if buggy.

mikehearn:
TransactionOutput/Input equality/hashcode is a really tricky area that has changed a few times in the past, unfortunately. Changing it has a high risk of breaking things. I'd want to test it out locally for a while first.

The issue is that there are valid use cases for both ignoring the parent and taking it into account, depending on what you're trying to do. The current approach is a kludge whereby if parent==null it's of course ignored, but this can lead to subtle bugs where two outputs/inputs that you expect to compare equal actually don't. The fix is usually to stick a call to duplicateDetached() onto one of them, but I'd much rather find a type safe solution: possibly by breaking TransactionOutput into two classes, one that contains the Script+Value and one that contains the first object+parent pointer.

amichair:
I'm glad I wasn't paranoid there for nothing - it seemed fishy!

Just to emphasize the current contract breaks: the parent == null check in equals breaks reflexivity (x.equals(y) != y.equals(x) if x has a parent set and y doesn't), and the index in hashCode can give different hashcodes to equal instances. These are very basic assumptions explicit in the Object contract of these methods, and relied upon in just about all collections). If anything worked until now, it was probably by chance :-)

So what shall we do with this issue while you research it? Unfortunately, I don't have enough understanding of the history and usages of all this parent stuff to evaluate potential solutions. Are the first two commits good? Shall I remove the third? Anything else I can do? Maybe open a separate issue just for the parent stuff?

schildbach:
Ok let's exclude TransactionInput and TransactionOutput from this PR for now – maybe put those changes on a separate PR?

@amichair The remainder of this PR I think can be squashed into one commit, and could you rebase that on current master? It will perhaps clash with Peter's changes to LazyECKey. I'd like to keep his logic but feel free to adapt code style to the other equals/hashCode methods if you like.

amichair:
OK, I removed the third commit (TransactionOutput/TransactionInput), rebased, and formatted LazyECKey.

Are you sure you want to squash? One commit actually changes logic and fixes bugs, whereas the other one has many changes but is not supposed to change behavior - I think it's easier to look at them separately so the important changes aren't lost in the noise. But I'll squash them if you still prefer it :-)

schildbach:
Well I'd love if those two commits were truely separate but if I look at the "Fix" commit I see plenty of style changes (which would belong in the second commit, no?)

Your latest version still changes the semantics of TransactionInput.equals/hashCode.

amichair:
You're right, I was too lenient just then. How about now?

While the warning comments are not behavioral changes, to me their absence is a bug (and a real bug is just waiting to happen), so I left them along with the fixes rather than moving them into styling. Concur?

The TransactionInput/Output changes that were made were the easy ones (removing scriptSig which is redundant and inconsistent, and removing index which broke the equals/hashCode contract and is not needed). The hard part is what to do with the parent... that part is unchanged for now, and I'll open a separate issue for it.

schildbach:
Looks good now, and I merged. Many thanks for responding to my nitpicking for so long!

amichair:
My pleasure :-)

Opened https://github.com/bitcoinj/bitcoinj/issues/1025 for the TransactionOutput issue.

PR #1004, opened 2015-06-28
Little cleanups

amichair:
Misc little refactorings, style/formatting issues, simplification, consistency, etc.

Let me know if there's anything in there you don't like. Mostly I tried to keep the code consistent with itself and best practices/guidelines, but there is some room for personal taste, I suppose.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/UnsafeByteArrayOutputStream.java
+++ b/core/src/main/java/org/bitcoinj/core/UnsafeByteArrayOutputStream.java
@@ -63,7 +63,7 @@ public class UnsafeByteArrayOutputStream extends ByteArrayOutputStream {
      * @param len the number of bytes to write.
      */
     @Override
-    public void write(byte b[], int off, int len) {
+    public void write(byte[] b, int off, int len) {
         if ((off < 0) || (off > b.length) || (len < 0) ||
                 ((off + len) > b.length) || ((off + len) < 0)) {
             throw new IndexOutOfBoundsException();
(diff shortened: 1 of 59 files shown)


core/src/main/java/org/bitcoinj/core/UnsafeByteArrayOutputStream.java
```diff
@@ -63,7 +63,7 @@ public void write(int b) {
      * @param len the number of bytes to write.
      */
     @Override
-    public void write(byte b[], int off, int len) {
+    public void write(byte[] b, int off, int len) {
```
mikehearn: Heh, I didn't even know that was valid Java syntax.

mikehearn:
Wow, thanks! That's a huge cleanup! Looks great, I'll merge this now.

Probably we should augment the style guide on the website to specify things like ordering of modifiers explicitly.

amichair:
Or create a checkstyle configuration or somesuch and not leave it up to chance ;-)

btw it's worthwhile running an IntelliJ inspections analysis on the whole project once in a while and sifting through the results to apply the ones you like. That's what I partially did here.

mikehearn:
Yes, I have the inspector running in the background. Doing a whole program analysis tends to throw up a lot of small issues and then I always figure that other things need my time more (like code review, doh). So I clean up as I go and try to avoid introducing new issues. But sometimes contributed code can generate more.

amichair:
The full analysis seems to come up with lots of stuff that the online lint one doesn't. It's true that there's a lot of false positives (or just uninteresting ones) and it can take a couple hours to sort through them (it would have been far worse without ctrl-alt-down, alt-enter, etc.) - that's why I do it maybe once a year, to clean up whatever got messed up since the last time. Or on new projects that seem like they can use it ;-)

PR #1005, opened 2015-06-29
fix docblock for clearOutputs()

jlopp:
Looks like the docblock for clearOutputs was copypasta from clearInputs


schildbach:
Thanks, merged.

PR #1006, opened 2015-06-30
Support testnet and regtest in BuildCheckpoints tool

devrandom:


tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
```diff
@@ -59,22 +83,26 @@ public static void main(String[] args) throws Exception {
...
         peerGroup.addAddress(peerAddress);
         long now = new Date().getTime() / 1000;
         peerGroup.setFastCatchupTimeSecs(now);
 
-        final long oneMonthAgo = now - (86400 * 30);
+        // Only go back 1 day for testnet, so we can quickly get past block storms
+        final int daysAgo = params.getId().equals(NetworkParameters.ID_MAINNET) ? 30 : 1;
+
+        final long timeAgo = now - (86400 * daysAgo);
+        System.out.println("Checkpointing up to " + new Date(now * 1000));
```
schildbach: Can you use Utils.dateTimeFormat()?

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
```diff
@@ -46,12 +51,31 @@
...
 public class BuildCheckpoints {
-
-    private static final NetworkParameters PARAMS = MainNetParams.get();
-    private static final File PLAIN_CHECKPOINTS_FILE = new File("checkpoints");
-    private static final File TEXTUAL_CHECKPOINTS_FILE = new File("checkpoints.txt");
+    private static NetworkParameters params;
 
     public static void main(String[] args) throws Exception {
+        OptionParser parser = new OptionParser();
+        OptionSpec<NetworkEnum> netFlag = parser.accepts("net").withRequiredArg().ofType(NetworkEnum.class).defaultsTo(NetworkEnum.MAIN);
```
schildbach: Does this mean the "net" arg will be required? I'd prefer a default.

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
```diff
@@ -59,22 +83,26 @@ public static void main(String[] args) throws Exception {
...
+        final PeerGroup peerGroup = new PeerGroup(params, chain);
         final InetAddress peerAddress = InetAddress.getLocalHost();
         System.out.println("Connecting to " + peerAddress + "...");
         peerGroup.addAddress(peerAddress);
         long now = new Date().getTime() / 1000;
         peerGroup.setFastCatchupTimeSecs(now);
 
-        final long oneMonthAgo = now - (86400 * 30);
+        // Only go back 1 day for testnet, so we can quickly get past block storms
+        final int daysAgo = params.getId().equals(NetworkParameters.ID_MAINNET) ? 30 : 1;
```
schildbach: I'm not so sure if this is a good idea. The idea is checkpoints will never change. On testnet, larger forks have been seen than on mainnet.

schildbach:
@devrandom Did you see my inline comments?

devrandom:
@schildbach - not sure how I missed your comments, sorry about that.  Looking now.

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
devrandom: There is a default - see last clause of statement.  That means that the user can either omit the --net option entirely, but if they do supply it they have to provide an argument.

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
devrandom: Changing this to a flag, default to 30 days.

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
devrandom: Just to clarify - the use of withOptionalArg elsewhere in the code was incorrect.  It doesn't make sense to specify this option without its argument.

devrandom:
I think this is ready for review.

mikehearn:
@schildbach happy to let you handle this one

schildbach:
Merged, thanks!

I think now that BuildCheckpoints has graduated into a command line tool, we could use a set of wrapper scripts, similar to the ones for wallet-tool.

PR #1007, opened 2015-06-30
Fix peerTimeoutTest intermittent failures.

amichair:
The build fails occasionally with GitHub/Travis, or when I run 'mvn test -pl core -Dtest=PeerGroupTest' explicitly, although usually not when I do a full build. Can't be sure it fixes it (since failure is inconsistent), but I think it should help. Suggestions?

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/PeerGroupTest.java
+++ b/core/src/test/java/org/bitcoinj/core/PeerGroupTest.java
@@ -483,8 +483,9 @@ public class PeerGroupTest extends TestWithPeerGroup {
 
     @Test
     public void peerTimeoutTest() throws Exception {
+        final int timeout = 100;
         peerGroup.start();
-        peerGroup.setConnectTimeoutMillis(100);
+        peerGroup.setConnectTimeoutMillis(timeout);
 
         final SettableFuture<Void> peerConnectedFuture = SettableFuture.create();
         final SettableFuture<Void> peerDisconnectedFuture = SettableFuture.create();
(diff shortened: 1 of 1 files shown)


mikehearn:
LGTM, thanks. Thread.sleep needs to be expunged.

schildbach:
Unfortunately the test still seems to be unstable:
https://api.travis-ci.org/jobs/69009745/log.txt?deansi=true

amichair:
I'll explain:

The old implementation was failing often since it was sleeping 110ms, when configured with a 100ms timeout, and then checking if the disconnect even had been fired, which it often wasn't (I could reproduce it often here, and saw it a bunch on Travis).

The new implementation gives it more max time (300ms), but doesn't wait any longer than it should - when the disconnect event is triggered, it continues right away. So it's more lenient on heavy loaded servers with long delays, and more efficient when things do work out ok.

The old one also did an arbitrary check 50ms into the test to see that it was not disconnected yet. The new one replaced it with a more exact one (so I thought) that checks that the elapsed time until the event was triggered is no less than the original 100ms timeout. This is where it failed now.

With the current code, the only scenario I could think of that would cause this is if there was a significant delay between when the connection was opened and the start timestamp was taken. I can swap the two statements, but that would replace the false positive with a false negative, i.e. if it got disconnected before the timeout period, we may miss it (not unlike the old test). I guess this would be the better option, unless someone has additional insights on the matter, or I just missed something...

amichair:
Can someone reopen this PR so I can push another commit?

mikehearn:
It says "@amichairamichair deleted the amichair:peer-timeout-test branch 7 hours ago"

So I think you have to open a new one.

amichair:
I've restored the branch to its former state (so I see the 'delete branch' button again)... but I guess that doesn't cut it. I've opened a new one - https://github.com/bitcoinj/bitcoinj/pull/1008.

PR #1014, opened 2015-07-04
remove useless math like >> 0

C-Otto:


schildbach:
Looks all correct to me.

I guess these numbers are for formatting, but frankly I don't mind. If no one objects, I'll merge this PR.

C-Otto:
My intention is to remove all warnings (as shown in IntelliJ IDEA) so that the remaining warnings point to real issues. To rephrase, with code like this it may easily happen to overlook a more important warning.

schildbach:
Fair enough. I merged this. Thanks!

PR #1016, opened 2015-07-04
convert to local variable

C-Otto:


C-Otto:
I modified this PR by not inlining a constant, as suggested by @schildbach. The constant is now properly introduced in PR #1018.

schildbach:
Merged, thanks!

PR #1017, opened 2015-07-04
remove useless casts

C-Otto:


schildbach:
I merged the first commit – thanks!
The second will be fixed in PR #1003 already, so I won't merge it.

PR #1019, opened 2015-07-04
disable warnings for generated classes

C-Otto:


core/src/main/java/org/bitcoin/crawler/PeerSeedProtos.java
```diff
@@ -3,6 +3,7 @@
 
 package org.bitcoin.crawler;
 
+@SuppressWarnings("ALL")
```
schildbach: This requires more explanation. Suppressing warnings is generally a solution.

C-Otto: You mean "no solution"? My explanation is in the commit's title: these classes are generated, it does not make any sense to deal with warnings generated for these.

schildbach: Ah ok I see. But if they are generated, your annotations will be gone soon? I think we'd need to find a way so that the source file contains the annotation, or an instruction to generate the annotation.

C-Otto: Yes, I'm afraid. But if the changes are committed manually, one might just leave the annotation in the file? I'm with if this is just not merged at all, though.

schildbach:
@mikehearn What do you think? Does it make sense to "fix" generated source files like this?

mikehearn:
It can't hurt. I doubt I'll regenerate that file very often.

schildbach:
Well well, so I merged.

PR #1008, opened 2015-06-30
Fix another peerTimeoutTest failure mode.

amichair:
Continuation of https://github.com/bitcoinj/bitcoinj/pull/1007.

PR #1020, opened 2015-07-05
Base58 cleanup

amichair:
Simplified and clarified code, slight optimizations, better documentation, corrected exception type.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AddressFormatException.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressFormatException.java
@@ -17,12 +17,16 @@
 package org.bitcoinj.core;
 
 @SuppressWarnings("serial")
-public class AddressFormatException extends Exception {
+public class AddressFormatException extends IllegalArgumentException {
     public AddressFormatException() {
         super();
     }
 
     public AddressFormatException(String message) {
         super(message);
+    }
+
+    public AddressFormatException(Throwable cause) {
+        super(cause);
     }
 }
--- a/core/src/main/java/org/bitcoinj/core/Base58.java
+++ b/core/src/main/java/org/bitcoinj/core/Base58.java
@@ -95,9 +95,9 @@ public class Base58 {
      *
      * @param input the base58-encoded string to decode
      * @return the decoded data bytes
-     * @throws AddressFormatException if the given string is not a valid base58 string
+     * @throws IllegalArgumentException if the given string is not a valid base58 string
      */
-    public static byte[] decode(String input) throws AddressFormatException {
+    public static byte[] decode(String input) throws IllegalArgumentException {
         if (input.length() == 0) {
             return new byte[0];
         }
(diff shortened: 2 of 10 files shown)


amichair:
Build is failing due to https://github.com/bitcoinj/bitcoinj/pull/1008... hopefully that fix will be the end of these PR build failures.

amichair:
If you're uncomfortable with changing the API (declared exception), feel free to take only the first commit and close this issue. If there are any comments about the first commit, let me know and I'll be happy to change it.

mikehearn:
Hmm, not sure about the exception change: yes it's quite specific but as it's checked it forces the caller to consider the case where the address is invalid, which is quite a common problem as they're often typed in. @schildbach Andreas, what do you think?

schildbach:
In terms of API compatibility it would better to at least keep the AddressFormatException, but change it's base class to IllegalArgumentException.

Generally we want to migrate away from manually transmitting addresses (via clipboard is probably the most prominent way) to protocols like BIP70, and for use in protocols I think the exception doesn't need to be checked any more.

mikehearn:
That sounds like a reasonable solution.

amichair:
I changed AddressFormatException to extend IllegalArgumentException, but I'm not entirely sure if this is what you meant. And what about places that should need a checked AFE exception? Or some other *FormatExceptions where neither AFE or IAE are the best choice (e.g. BIP38PrivateKey constructor)? I'm thinking maybe to revert the whole exception thing and let it be for now.

schildbach:
Yes, our change would also affect the other subclasses of DumpedPrivateKey. And yes, the base class change is what I meant.

I think AddressFormatException should still be declared to be thrown, because they are.

I merged the base58 cleanup commit already.

amichair:
After a discussion on #bitcoinj, two possible goals emerged for changing the exceptions:

- change AddressFormatException to extend IllegalArgumentException so that it becomes an unchecked exception, in order to reduce boilerplate code.

- rename AddressFormatException, which is used in places unrelated to addresses (base58 utility class being one of them), to something more generic like DataFormatException, in order to clean up the API. Possibly use more specific subclasses where appropriate (e.g. AddressFormatException in Address extends DFE).

These should be considered while looking at all the current usages of AFE. They can be done together or independently. Or neither. If going for a rename, it can be done before a major release where API changes are expected.

schildbach:
@mikehearn What do you think? You already agreed to the hierarchy change but what do you think about the renaming of the exception to a more appropriate name?

About the time if the change, I think now would be the right time: we just started the 0.14 cycle.

mikehearn:
I leave the discussion over the exception API in your capable hands - I don't have any strong opinions on this.

mikehearn:
Any motion here? I didn't want to stall things.

schildbach:
I've now pushed a simpler commit than this: 200f2368c65d5fd1dae8f6f7ce6307c2e09c16c4
It just changes AddressFormatException to unchecked.

This commit needs a rebase if we still want to clean up the "overly-specific" usage of AddressFormatException. Or maybe we just rename the exception and that's it?

schildbach:
For now, I'm closing this. If there is anything left to merge, please feel free to rebase and submit as a new PR.

PR #1021, opened 2015-07-05
Cleanup HDUtilsTest

C-Otto:
Introduce indirection with the benefit of readability.
Use less (unsafe) type casts.
Use speaking names for variables and methods.
Use Hamcrest's assertThat instead of assertEquals (with possibly wrong argument order).


schildbach:
I think the two commits can be squashed into one, since the adding of the two constructors doesn't make sense without using them.

PR #1022, opened 2015-07-05
Cleanup MAX_SATOSHIS pseudo-constant

C-Otto:
This commit inlines the pseudo constant to the only method using it.

Due to cyclic dependencies (NetworkParameters vs. Coin) MAX_SATOSHIS
cannot be made a true constant.


schildbach:
Merged, thanks!

PR #1023, opened 2015-07-05
Refactor Wallet: Move broadcasting code to own class

C-Otto:
The class Wallet is way too big. This is my first attempt to find something worth moving out of it. I'm not too happy about the package for the new class, feel free to suggest something else.

Note that this uses its own logger (which could be changed easily).

In the long run I'd like to move all code related purely to wallet maintenance out.

Do you know if broadcasting transactions like this is done elsewhere, too?

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/core/Wallet.java
@@ -4811,8 +4811,8 @@ public class Wallet extends BaseTaggableObject implements Serializable, BlockCha
      * A wallet app should call this from time to time in order to let the wallet craft and send transactions needed
      * to re-organise coins internally. A good time to call this would be after receiving coins for an unencrypted
      * wallet, or after sending money for an encrypted wallet. If you have an encrypted wallet and just want to know
-     * if some maintenance needs doing, call this method with andSend set to false and look at the returned list of
-     * transactions. Maintenance might also include internal changes that involve some processing or work but
+     * if some maintenance needs doing, call this method with {@code signAndSend} set to false and look at the returned
+     * list of transactions. Maintenance might also include internal changes that involve some processing or work but
      * which don't require making transactions - these will happen automatically unless the password is required
      * in which case an exception will be thrown.
      *
--- a/core/src/main/java/org/bitcoinj/core/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/core/Wallet.java
@@ -37,6 +37,7 @@ import org.bitcoinj.utils.*;
 import org.bitcoinj.wallet.*;
 import org.bitcoinj.wallet.Protos.Wallet.*;
 import org.bitcoinj.wallet.WalletTransaction.*;
+import org.bitcoinj.wallet.maintenance.BroadcastTransactionsAndLog;
 import org.slf4j.*;
 import org.spongycastle.crypto.params.*;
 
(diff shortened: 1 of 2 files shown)


mikehearn:
Broadcasting logic was moved into the TransactionBroadcast class some time ago. Your change actually makes the code bigger overall, not smaller, which isn't really a win. Moving code from one file to another whilst making it more complex and with more LOC overall is something I'd rather avoid.

I don't think a pure "broadcast and log" pattern is used anywhere else. I might be wrong though. If it were common the argument for this change would be stronger.

I agree that the Wallet class is too big, but as you have noticed, it's harder to shrink than it appears. The next big win for shrinking and simplifying Wallet is to stop using the Transaction class for tracking spentness, etc, and then eliminate the pools stuff. Since the myUnspents field was introduced during the 0.13 cycle, some of the older code is now redundant. 

However that is a very deep and complex change.

C-Otto:
@mikehearn 
I agree that I created more LOC, but nevertheless managed to make the Wallet class smaller - my main goal.

I don't agree that I introduced complexity. In fact, I identified a "single responsibility" (cf. SOLID) that was scattered around the Wallet class, and moved into its own class. This should be enough reason for it to have its own class, even though it is only used once. And, at least by my standards, the new class is way more readable and testable, which in itself should merit a merge. Besides, if I removed the setters and comments (both giving extra value, I think), the size difference would not really be noticable.

To ask slightly provocative but important questions:
Do we agree on the fact that Wallet is way too big?
What else should be done to shrink Wallet?
Are deep and complex steps the only way out?

I'll have a look at TransactionBroadcast soon.

PS: As said, I'd like to move all of Wallet maintenance out of Wallet - similar to how I did it in this PR. What do you think about that?

C-Otto:
Oh, an addition: If in the not-so-distant-future we need to broadcast-and-log in another place, would the current code be more suitable for that compared to the code of this PR? I think we'd either do a similar change then, or revert to repeating the code (maybe even copy&paste).

C-Otto:
Ah, I see that TransactionBroadcast ist the code already used. I guess the name BroadcastTransactions is not the best, as it merely delegates and logs, but other than that I'd like to refactor this "logic" out of Wallet.

mikehearn:
What I meant was, you removed 21 lines of code and added 93, all of which is still logically a part of the wallet. So you're making the wallet code larger, not smaller, even though a chunk may now be in a separate file. There's little need to unit test such straightforward code that only prints something to the logs and has no other purpose.

> Do we agree on the fact that Wallet is way too big?

Sure, I already said this above.

> What else should be done to shrink Wallet?

Good question. It's not easy because the wallet class has so many features. Most of the biggest wins are   complex redesigns rather than simple code movement.

For instance, it'd make sense to move tracking of outputs state, handling of reorgs etc into a separate class. But that code is quite deeply intertwined with the core of how the wallet works, so such a refactoring is quite advanced. I'd probably want to do it myself.

A smaller change may be to move the Bloom filtering logic into a separate class. Some of it has already been moved, but the PeerFilterProvider methods could be implemented by a private class which is then exposed via a single method rather than the current approach where the methods are implemented directly by the class. Moving this would also have the advantage of removing methods from the public Wallet API that are essentially internal and aren't meant to be used directly.

> PS: As said, I'd like to move all of Wallet maintenance out of Wallet - similar to how I did it in this PR. 

If the total LOC isn't wildly different and the code does get simpler as a result then sure, but I'd have to see what you come up with. One reason the wallet class is so huge is that moving stuff into other files just for the sake of it doesn't always make things easier or better.

C-Otto:
We seem to have a very different understanding of how code should look like, it seems. @schildbach maybe you want to share your thoughts on this?

schildbach:
I agree with Mike on this one.

PR #1024, opened 2015-07-06
DefaultRiskAnalysis update

stonecoldpat:
An update to some of the IsStandard() rules to take into account the
following:
-A transaction’s total size cannot be greater than 100kb.
-The size of each individual transaction input cannot be greater than
1650 bytes.
-A multi-sig output cannot have more than three public keys or require
more than three signatures (or less than 1).
-Checks if more than one OP_RETURN is included in a transaction.
-Checks if a transaction output only contains push data.

More checks can be done, for example:
-Check for Coinbase inputs.
-Check P2SH inputs (only check multis outputs at the moment)

Let me know if there are any mistakes.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -170,7 +170,7 @@
                                         <urn>com.fasterxml.jackson.core:jackson-core:2.5.1:jar:null:test:e2a00ad1d7e540ec395e9296a34da484c8888d4d</urn>
                                         <urn>com.fasterxml.jackson.core:jackson-databind:2.5.2:jar:null:test:2b4dd13fbe6f8c6b146d4c3b7fd70862f136802d</urn>
                                         <urn>com.google.code.findbugs:jsr305:2.0.1:jar:null:compile:516c03b21d50a644d538de0f0369c620989cd8f0</urn>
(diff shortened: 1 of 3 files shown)


stonecoldpat:
Awesome, I've fixed the spaces you mentioned earlier @schildbach. They seem to be in this update as well.

schildbach:
I had the WatchMempool tool running on your PR, and here are the results:

```
Runtime: 45 minutes
  TOTAL                                    17810  (100% of total)
  NON_STANDARD-DUST                           80  (0% of total)
  NON_STANDARD                                80  (0% of total)
  OK                                       17730  (99% of total)
```

I think we can safely say at least there are no false positives introduced by this patch.

I noticed there is a bit of space/tabs mixup in the indentation. We only use spaces, 4 at a time.

stonecoldpat:
Hopefully that should be all the tabs/spaces fixed... let me know if I rebased it correctly or not

schildbach:
I see you have added a commit called squash (but it isn't actually squashed with the commit before). Also, there is a bunch of commits from master. These do not belong on your branch. Can you just rebase your branch on master and squash the "squash" commit into the the first commit?

Regarding the content of your PR. I did a bit of testing and it looked good, see my previous comment. What's left is I think someone should review the logic and compare it with the specification. Don't worry, I don't expect any serious issue here. I'm currently busy with some things and can do the review later this week. If anyone else want to beat me feel free.

PR #1027, opened 2015-07-08
TestFeeLevel: Create transactions with two outputs (and thus most of …

schildbach:
…the time two inputs). This is more close to standard wallet transactions.


schildbach:
Merged.

mikehearn:
LGTM

PR #1031, opened 2015-07-14
Context and save

devrandom:
Fixes the issue at https://github.com/bitcoinj/bitcoinj/commit/50e7e14763781db9069b2d5cf93d8df4f03c31a7#diff-b865a4a76bc217a5b92752e6c820e0efR91

Bonus - strict mode for Context propagation.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Context.java
+++ b/core/src/main/java/org/bitcoinj/core/Context.java
@@ -86,6 +86,7 @@ public class Context {
             log.error("This error has been corrected for, but doing this makes your app less robust.");
             log.error("You should use Context.propagate() or a ContextPropagatingThreadFactory.");
             log.error("Please refer to the user guide for more information about this.");
+            log.error("Thread name is {}.", Thread.currentThread().getName());
             // TODO: Actually write the user guide section about this.
             // TODO: If the above TODO makes it past the 0.13 release, kick Mike and tell him he sucks.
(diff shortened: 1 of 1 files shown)


mikehearn:
Ouch, that's a bad bug. I did some changes similar to yours, but without losing the exception handler and min thread priority. If you rebase the first commit should be no longer needed, and I'll take the strict mode - though the TLS magic stuff is only intended to be temporary during the transition to using Context everywhere. Then we should remove it.

devrandom:
Thank you, rebased.

devrandom:
I experimented with turning this on.  I think the code doesn't yet pass the context in all cases, and perhaps you already knew this.

For example, when you try to commit a new transaction, it gets logged to stdout, which does a naked getConfidence() call without a context.

mikehearn:
Yeah, the transition isn't complete. We need to move most c'tors that take NetworkParameters to taking Context instead. Then things like toString can pass Context into getConfidence() and we can remove or deprecated the no-args getConfidence method.

mikehearn:
Thanks, this new version LGTM

schildbach:
I squashed the commits and merged.

@devrandom Your commit comments could use more explanation. "Show thread name" isn't really useful if you read it without previous knowledge. A good way to get into the right mood is try to formulate complete sentences, along with proper punctuation.

PR #1034, opened 2015-07-14
Refactor listeners into their own package

rnicoll:
The org.bitcoinj.core package contains a huge number of classes, making navigating it difficult. This PR moves the listeners into their own package to try thinning it out a bit.


schildbach:
What's the gain? You're spreading things that belong together over multiple packages. Note we don't have an "exception" package for the same reason.

rnicoll:
The idea is to improve discoverability, at the moment it can difficult to find the right class to do something, and this isn't aided by org.bitcoinj.core having a huge number of classes to sort through. Particularly tricky cases include "NetworkParams" being in core rather than params and Wallet being in core instead of wallet.

Moving exceptions into their own package would in fact have been the next suggestion!

Aware this may simply be where we disagree on architecture, but thought it was worth making the suggestion, and a patch is often clearer than just opening an issue.

schildbach:
I agree NetworkParameters and Wallet are in the wrong package. That's for historical/compatibility reasons.

Perhaps we should collect a bunch of API renames. Those things are best done just before branching off the release branch, so in roughly half a year.

mikehearn:
Thanks for taking on more refactoring work!

I agree with Andreas that simply moving things around just breaks code for no significant gain (if you navigate using your IDE features like I do the size of a package makes no difference). We try and avoid  breaking user code unless they get something back in return.

But! This argument goes away if the change is rolled in with a more substantial refactoring of how events work to fix warts and rough edges.

And I've wanted such a refactoring for a while. So if you're up for it, that'd be a fantastic cleanup and contribution to the next release ... albeit a little bit more work. I've posted details of what I'm thinking about to the list:

https://groups.google.com/forum/#!topic/bitcoinj/AE7_XZtvLNs

rnicoll:
Yup, happy to tackle that - it will probably be next week before I get to it, but consider it on the to-do list!

rnicoll:
Pushing work in progress for reference/backup, but not yet finished.

rnicoll:
@mikehearn Could you take a look, see if this is going in the right direction, and that the result is as expected? Splitting the wallet events down mostly ended with a lot of excessive individual classes, so in the end split it into two, but can subdivide further if you really want.

mikehearn:
That's looking good! Yes, I worry about the number of interfaces: Java really screwed up its design of SAM interfaces. There is all this BiConsumer<A,B> nonsense. Kotlin has Function1<A>, Function2<A, B> etc which is much more sensible. Some of the interfaces can be collapsed using generics and introducing a similar thing to Function1, Function2 etc.

We can always split the 2-method wallet event listener later, if need be. But I think this is definitely a step in the right direction for improving things in lambda-supporting languages.

rnicoll:
Rebasing this appears to cause some unit tests to fail. Trying to trace down what's happened, but no ETA

rnicoll:
Rebased and tests fixed. Travis claims failure, but from the raw log ( https://s3.amazonaws.com/archive.travis-ci.org/jobs/72576671/log.txt ) looks to be after the tests run. If I've broken something and missed it, please do let me know, though!

schildbach:
Generally I like this refactoring, and I'm happy to merge if Mike is.

The only thing I don't like is putting all listeners into a separate package. I think they always belong with the class you're listening.

schildbach:
General note: I try to use packages to separate use cases, rather than technology.

mikehearn:
Yes, this looks OK to me as long as (obviously) it doesn't cause any regressions.

rnicoll:
Rebased on current master

I do see what you mean about by use case rather than technology, but I'm not sure there's much we can do to trim core down by use case unfortunately, and the huge number of classes makes it hard to navigate/learn IMHO.

schildbach:
Breaking up classes that belong together (by usecase) make it even harder to navigate imho.

Anyway, currently the examples are broken.

rnicoll:
Sorry about that, keep compiling just core and missing the rest. Fixed.

schildbach:
Is there a reason you changed the parameter order of ```public void addEventListener(Executor executor, WalletEventListener listener)```? For compatibility reasons it would be better to keep it.

mikehearn:
I asked for that. Some languages work better when callbacks are the last
argument.

schildbach:
Ah ok. So I'll merge.

schildbach:
Thanks for this big chunk of refactoring!

mikehearn:
I think this needs more work (we can do it, I guess). The loss of backwards compatibility is pretty severe, most of my own software has broken, and because so many interfaces aren't SAM, the benefits don't normally appear anyway.

Currently the BlockChain listeners could use the old method back, AbstractBlockChainListener too.

schildbach:
Functionally it seems ok though. I've been running it with Bitcoin Wallet from time to time.

mikehearn:
Right, I'm sure the code itself works fine. It just could do with more backwards compat. I'll add a few commits.

rnicoll:
Happy to look at this, but likely not before Monday, so probably does make more sense for you to reintroduce parts if you're okay to do so?

mikehearn:
Already done, don't worry about it.

schildbach:
I think this PR introduced a regression in that it is not possible to register a listener to the keyChainGroup. Nor does the old deprecated `Wallet.addEventListener()` do.

schildbach:
@rnicoll Saw your question on IRC. If you look at the release-0.13 branch, `Wallet.addEventListener()` calls through to `keychain.addEventListener()`.

rnicoll:
@schildbach found it... addEventListener() calls addChangeEventListener(), which calls keyChainGroup.addEventListener(). That was renamed from keychain in https://github.com/bitcoinj/bitcoinj/commit/fe7b7032371d8b6c1c8b870d0e4aa9508aa9bbd4

Do let me know if I've missed anything, please.

schildbach:
@rnicoll Ah you're right. Sorry about that.

PR #1038, opened 2015-07-17
Allow unmarried chains to have P2SH addresses

devrandom:
Allow subclasses of DKC to vend P2SH addresses even if they are unmarried.


mikehearn:
What is the use case for this? Why would you ever use P2SH for a single key output?

mikehearn:
BTW the change itself looks OK, even if I'm still unsure p2sh without multisig is really that useful :)

devrandom:
OK.  I was wondering whether to add a non-final `vendsP2SH()` to both DKC and KCG, the former delegating to isMarried by default.  Also, would change the `checkState` statements to call the new functions.   That would let the existing behavior continue, but would let the API user override for non-multisig P2SH.

Or could just go with this existing change.

mikehearn:
As this change just relaxes a few checks and is so small/simple, let's just go with it. LGTM.

schildbach:
Merged.

PR #1039, opened 2015-07-18
Provide a consistent fromBase58/toBase58 API for the VersionChecksumm…

schildbach:
…edBytes hierarchy.

Also makes DumpedPrivateKey use WrongNetworkException when appropriate. Adds missing testcases and corrects/adds a bit of JavaDoc.


mikehearn:
LGTM

Should the old c'tor be deprecated, as the factory method is clearer?

schildbach:
Thanks! I just deprecated the old constructors and migrated bitcoinj itself to use the new API.

PR #1040, opened 2015-07-19
Connect bitcoinj to hidden services

cpacia:
Pull to allow bitcoinj to connect to hidden services. 

You can test with:
```java
InetSocketAddress onion = InetSocketAddress.createUnresolved("2p7lrn6hrjgb3j6l.onion", MainNetParams.get().getPort());
PeerAddress addr = new PeerAddress(onion);
peergroup.addAddress(addr);
peergroup.setMaxConnections(1);
```

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerAddress.java
@@ -40,6 +40,7 @@ public class PeerAddress extends ChildMessage {
     static final int MESSAGE_SIZE = 30;
 
     private InetAddress addr;
+    private String hostname; // Used for .onion addresses
     private int port;
     private BigInteger services;
     private long time;
@@ -97,8 +98,20 @@ public class PeerAddress extends ChildMessage {
         this(addr, MainNetParams.get().getPort());
     }
 
+    /**
+     * Constructs a peer address from an {@link InetSocketAddress}. An InetSocketAddress can take in as parameters an
+     * InetAddress or a String hostname. If you want to connect to a .onion, set the hostname to the .onion address.
+     */
     public PeerAddress(InetSocketAddress addr) {
-        this(addr.getAddress(), addr.getPort());
+        if (addr.getHostName() == null || !addr.getHostName().toLowerCase().endsWith(".onion")) {
+            this.addr = checkNotNull(addr.getAddress());
+        } else {
+            this.hostname = addr.getHostName();
+        }
+        this.port = addr.getPort();
+        this.protocolVersion = NetworkParameters.PROTOCOL_VERSION;
+        this.services = BigInteger.ZERO;
+        length = protocolVersion > 31402 ? MESSAGE_SIZE : MESSAGE_SIZE - 4;
     }
 
     public static PeerAddress localhost(NetworkParameters params) {
@@ -163,6 +176,11 @@ public class PeerAddress extends ChildMessage {
         return length;
     }
 
+    public String getHostname() {
+        maybeParse();
+        return hostname;
+    }
+
     public InetAddress getAddr() {
         maybeParse();
         return addr;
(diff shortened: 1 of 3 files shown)


schildbach:
If you make changes to orchid, you need to advance the version of our orchid artifact.

Do you plan more changes to orchid? In this case we should switch it to -SNAPSHOT with a first commit and release it with the next bitcoinj release.

If this is expected to be the only change to orchid, then as a shortcut we can just advance the release number in the same commit.

cpacia:
I don't foresee making any other changes to orchid. Although I just noticed I've got some kind of bug with the PeerAddress equals method that I have to fix. I'll have to make the changes in the morning.

mikehearn:
Could you adjust the example in examples/src/main/javascript/tor.js too, as it no longer needs some hack after this patch?

examples/src/main/javascript/tor.js
```diff
@@ -6,19 +6,15 @@ var params = bcj.params.MainNetParams.get();
 var context = new bcj.core.Context(params);
 bcj.utils.BriefLogFormatter.init();
 
-var InetAddress = Java.type("java.net.InetAddress");
+var PeerAddress = Java.extend(Java.type("org.bitcoinj.core.PeerAddress")
```
mikehearn: Java.extend is for when you want to subclass it. In this case it's no longer needed, so it can be just a plain call to Java.type

mikehearn:
Thanks Chris! If you just delete the now unneeded Java.extend call from the js then this LGTM

AUTHORS
```diff
@@ -10,6 +10,7 @@ Andreas Schildbach
 Bennett Hoffman
 Carlos Lopez-Camey
 Carsten Otto
+Chris Pacia
```
schildbach: Note that on the next release, this will be auto-generated from your author info in the commit. Feel free to keep this chunk though.

core/src/main/java/org/bitcoinj/core/PeerAddress.java
```diff
@@ -40,6 +40,7 @@
     static final int MESSAGE_SIZE = 30;
 
     private InetAddress addr;
+    private String hostname; //Used for .onion addresses
```
schildbach: Can you add a space after the // ?

core/src/main/java/org/bitcoinj/core/PeerAddress.java
```diff
@@ -216,6 +234,9 @@ public void setTime(long time) {
 
     @Override
     public String toString() {
+        if (hostname != null){
```
schildbach: missing space, or just remove the braces

core/src/main/java/org/bitcoinj/core/PeerAddress.java
```diff
@@ -232,8 +253,13 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hashCode(addr, port, time, services);
     }
-    
+
     public InetSocketAddress toSocketAddress() {
-        return new InetSocketAddress(addr, port);
+        // Reconstruct the InetSocketAddress properly
+        if (hostname != null){
```
schildbach: missing space

cpacia:
Mike, Andreas. I'll clean it up first thing in the morning. Thanks!

cpacia:
OK I amended that last commit to fix the PeerAddress extension in the javascript example, gave it a better commit message, and fixed the spacing in three places.

mikehearn:
Nice job!

cpacia:
Thanks. Give me one sec though, missed a ) on line 17.

cpacia:
Should be good now. Just moved those spacing changes into the original commit so they weren't part of the javascript commit.

schildbach:
Excellent! Merged.

PR #1042, opened 2015-07-19
Don't derive private keys just to check if they exist

devrandom:
This speeds up `calculateAllSpendCandidates` and `getBalance` by a couple of orders of magnitude, which is especially noticeable on Android.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -320,7 +320,7 @@ public class DeterministicKey extends ECKey {
     /** {@inheritDoc} */
     @Override
     public boolean hasPrivKey() {
-        return findOrDerivePrivateKey() != null;
+        return findParentWithPrivKey() != null;
     }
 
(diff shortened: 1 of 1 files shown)


schildbach:
Nice find! But rather than duplicating the code, why not extract a "private DeterministicKey findPrivateKey()" and use that?

devrandom:
I refactored along the lines you suggested, but since it looks for a parent only, I called it `findParentWithPrivKey`.

schildbach:
LGTM, I'll merge if no one objects.
I think this also qualifies for backporting to 0.13.

kparmar1:
Yes backporting would def be a good thing.

kparmar1:
Versioning and update behaviour could be coded but the potential issues 
and increased complexity is probably not worth it given that these data 
structures shouldn't really change (when stable).

devrandom:
Not sure I understand, this code doesn't affect any data structures.

schildbach:
Merged, thanks.

kparmar1:
Apologies I seem to have replied to something completely different.
K

PR #1043, opened 2015-07-20
MySQL data type fixed. Changed binary to varbinary.

kparmar1:
I received an email from Daniel Yin for a bug in the MySQL store. This was a change that happened recently if you remember (not sure why the unit test didn't flag this then- maybe it wasn't run I guess).

The problem is to do with padding of the values that we really shouldn't need to worry about so hence this correction- it's now equivalent to the blob that was used previously. 

(I've also informed Daniel that he shouldn't email me directly. I only realised when I sent the reply. Create an issue or post on the forum for obvious reasons - the funniest being me and a big red bus :-)


kparmar1:
Tests passing locally and everything is up to date.

schildbach:
Merged, thanks.

PR #1046, opened 2015-07-22
Remove lots of Java serialization.

schildbach:
Remove Java serialization from

* NetworkParameters,
* the entire Message hierarchy,
* Block,
* StoredBlock,
* StoredUndoableBlock,
* TransactionConfidence,
* UTXO,
* DeterministicHierarchy,
* MemoryFullPrunedBlockStore and
* Wallet

It was largely untested, in many cases not working any more and we don't see a reason for supporting it.


schildbach:
I pushed another commit on top which removes the transient modifiers for members of above classes. Transient seems to affect only Java serialization.

PR #1047, opened 2015-07-22
Prevent null pointer exception if DNS seeds parameter is null.

jlopp:
I'm working on upgrading BitcoinJ from 0.12 to 0.13 and some of my tests were failing due to a null pointer exception that gets thrown if you instantiate a DnsDiscovery object with the UnitTest NetworkParams.

### DIFF
--- a/core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
+++ b/core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
@@ -56,9 +56,10 @@ public class DnsDiscovery extends MultiplexingDiscovery {
     }
 
     private static List<PeerDiscovery> buildDiscoveries(NetworkParameters params, String[] seeds) {
-        List<PeerDiscovery> discoveries = new ArrayList<PeerDiscovery>(seeds.length);
-        for (String seed : seeds)
-            discoveries.add(new DnsSeedDiscovery(params, seed));
+        List<PeerDiscovery> discoveries = new ArrayList<PeerDiscovery>();
+        if (seeds != null)
+            for (String seed : seeds)
+                discoveries.add(new DnsSeedDiscovery(params, seed));
         return discoveries;
     }
 


core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
```diff
@@ -56,7 +56,7 @@ public DnsDiscovery(String[] dnsSeeds, NetworkParameters params) {
     }
 
     private static List<PeerDiscovery> buildDiscoveries(NetworkParameters params, String[] seeds) {
-        List<PeerDiscovery> discoveries = new ArrayList<PeerDiscovery>(seeds.length);
+        List<PeerDiscovery> discoveries = new ArrayList<PeerDiscovery>(seeds != null ? seeds.length : 0);
```
schildbach: This will just move the NPE one line lower.

Why not remove the entire parameter in the ArrayList constructor, it isn't much useful?

schildbach:
Thanks! Can you make your commit message a real sentence by starting with a capital letter and ending with a punctuation mark?

core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
jlopp: Ah you're right; I was thinking that enhanced for loops would skip null variables. I'm assuming that the sending of the size parameter is for performance & security reasons, so I didn't want to remove it.

schildbach: This part of the code is not performance critical.
Security: how do you mean?

jlopp: For some reason I thought that it was both an initial and permanent capacity parameter, but I see it's only initial capacity so it wouldn't prevent other code from adding more items after all.

jlopp:
Alright, this approach seems to be cleaner. I could also see another approach of just changing the predefined seeds to be empty arrays rather than null, but that doesn't seem as robust.

core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
```diff
@@ -56,6 +56,9 @@ public DnsDiscovery(String[] dnsSeeds, NetworkParameters params) {
     }
 
     private static List<PeerDiscovery> buildDiscoveries(NetworkParameters params, String[] seeds) {
+        if (seeds == null)
```
schildbach: I propose the following code:

```
List<PeerDiscovery> discoveries = new ArrayList<PeerDiscovery>();
if (seeds != null)
    for (String seed : seeds)
        discoveries.add(new DnsSeedDiscovery(params, seed));
```

jlopp:
Good to go, though let me know if there's a documented procedure to squash all of those revisions into a single commit. I tried to do it on my own but it looks like it didn't work.

schildbach:
I can do it for you.

The recommended procedure is to interactively rebase (git rebase -i) your branch on master then use the "fix" keyword for those commits that shall be squashed into the previous one (or use "squash" if you want to edit the resulting comment). Then force-push (git push -f) your branch onto your remote.

schildbach:
Ah, and if that fails that's because you've got merge commits on your branch.

jlopp:
Should be all cleaned up now :-)

schildbach:
Thanks! Looks perfect now. I just merged.

PR #1048, opened 2015-07-22
Don't call Wallet.saveNow more than necessary

devrandom:
* no need to saveNow on key derivation, since the keys can always be re-derived
* no need to saveNow on tx receipt
* default to saveNow on tx commit, but let the API user saveLater instead

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/core/Wallet.java
@@ -428,9 +428,7 @@ public class Wallet extends BaseTaggableObject implements Serializable, BlockCha
         } finally {
             keychainLock.unlock();
         }
-        // Do we really need an immediate hard save? Arguably all this is doing is saving the 'current' key
-        // and that's not quite so important, so we could coalesce for more performance.
-        saveNow();
+        saveLater();
         return keys;
     }
 
@@ -453,7 +451,7 @@ public class Wallet extends BaseTaggableObject implements Serializable, BlockCha
         } finally {
             keychainLock.unlock();
         }
-        saveNow();
+        saveLater();
         return key;
     }
 
@@ -1617,7 +1615,7 @@ public class Wallet extends BaseTaggableObject implements Serializable, BlockCha
             // If this tx spends any of our unspent outputs, mark them as spent now, then add to the pending pool. This
             // ensures that if some other client that has our keys broadcasts a spend we stay in sync. Also updates the
             // timestamp on the transaction and registers/runs event listeners.
-            commitTx(tx);
+            commitTx(tx, false);
         } finally {
             lock.unlock();
         }
@@ -1838,7 +1836,7 @@ public class Wallet extends BaseTaggableObject implements Serializable, BlockCha
                 if (!unspent.containsKey(hash) && !spent.containsKey(hash)) {
                     // Otherwise put it (possibly back) into pending.
                     // Committing it updates the spent flags and inserts into the pool as well.
-                    commitTx(tx);
+                    commitTx(tx, false);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -1617,7 +1615,7 @@ public void receivePending(Transaction tx, @Nullable List<Transaction> dependenc
             // If this tx spends any of our unspent outputs, mark them as spent now, then add to the pending pool. This
             // ensures that if some other client that has our keys broadcasts a spend we stay in sync. Also updates the
             // timestamp on the transaction and registers/runs event listeners.
-            commitTx(tx);
+            commitTx(tx, false);
```
schildbach: We should continue saving on tx receive. There is no guarantee we'll ever receive the tx again.

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -428,9 +428,7 @@ public DeterministicKey freshKey(KeyChain.KeyPurpose purpose) {
         } finally {
             keychainLock.unlock();
         }
-        // Do we really need an immediate hard save? Arguably all this is doing is saving the 'current' key
-        // and that's not quite so important, so we could coalesce for more performance.
-        saveNow();
+        saveLater();
```
schildbach: Agree on these two.

core/src/main/java/org/bitcoinj/core/Wallet.java
devrandom: If we die before the autosave, when we start up again, we will start scanning from a block that is at or before where the tx appeared, because the wallet save file contains the last block we processed.  There is no way we can miss a confirmed tx by delaying the save.

schildbach: I was thinking about unconfirmed tx.

devrandom: There's still no way to miss it:

* When you come up again, if it's unconfirmed, it will normally be sent to you from the mempool of the peers you connect to
* If it gets confirmed in the mean time, you will scan it from the blockchain in the normal way

The only way to miss it completely, is if it didn't propagate in the network and never confirms, which is a bad situation anyway.

schildbach: Each mempool is different. You'll get connected to different peers. The transaction can have arrived through other channels, e.g. Bluetooth. It's important the transaction gets saved to the wallet, such that it can be rebroadcasted to connecting peers.

devrandom: In our use case, there is a high transaction rate, so saving the wallet and forcing a sync every time is not performing well.  Would it be better to have this be configurable, with a default of saving now?

I can definitely see that you'd want to persist in the Bluetooth case.  In the other cases, your effect on the propagation of the tx is likely to be minimal, since you are just an SPV client with a small number of connections.

schildbach: Configurable is a compromise, if you really need the ability to mass-receive transactions.

mikehearn: I guess it being configurable is OK, but really this autosave crap needs to die in favour of pluggable backends with e.g. a database underneath it.

devrandom: I can work on that instead or in addition, since that's exactly what I'm up to right now. ;)

schildbach:
Mike has worked on part of this issue via commit 5f3d3a7f846d33c4001938fab778afdb26fb6ac7.

schildbach:
@devrandom Could you rebase this on master, because some of your proposals have been addressed on master already. I'm not sure if there is anything left.

schildbach:
For now, I'm closing this. If there is anything left to merge, please feel free to rebase and submit as a new PR.

Issue #1055, opened 2015-08-01
Cases are inconsistent for table and column names in the SQL code


traber:
The SQL statements declared in the class DatabaseFullPrunedBlockStore and MySQLFullPrunedBlockStore are inconsistent. Sometimes the names are kept in camel case (e.g. "openOutputs" or "scriptBytes") and sometimes they are not (e.g. "openoutputs" or "scriptbytes"). This might work on a Windows system but not on a UNIX system. In consequence the current implementation is resulting in MySQLSyntaxErrorExceptions being thrown while creating or using a MySQLFullPrunedBlockStore with a database running on a UNIX system.

kparmar1:
Hmmm yes you are correct...inconsistencies.  I think it would be better 
to change all to lower case?
Any objections?

This would keep it all consistent with Postgres store (I just checked 
and it uses lowercase) (I haven't ever used cases for column names 
before, usually a _ for rational DB's anyway).

I didn't realise there were inconsistent names between Postgres and H2 
(case wise) when I factored out the code (I must have used H2 as the 
guild for MySQL and not Postgres )- My Bad - I should have noticed that 
(OSx obviously more like windows although I would have expected the 
opposite).

I can make the fixes for this.

K

kparmar1:
Hi Thomas,

I was wondering If you can test out the fix as I don't  have another OS up and running and configured at the moment (I will do but not just at this moment)?
 
https://github.com/bitcoinj/bitcoinj/pull/1056

Thanks.

mikehearn:
OS X has a case insensitive file system.

traber:
The fixed code works for me but presumably I did not trigger the execution of all SQL statements during my tests. This is why I also manually checked the fixed source code. In DatabaseFullPrunedBlockStore I found one left over of a camel case column name in the SQL statement provided by the INSERT_OPENOUTPUTS_SQL constant ("toAddress"). This is no problem since column names are not case sensitive in MySQL on any platform.

kparmar1:
Yeah I looked it up and was surprised. I assumed it would be closer to Linux. Anyway I'm moving over to Ubuntu now and surprisingly I actually like the experience (after I got it all working- a little bit of a pain involved).

PR #1056, opened 2015-08-01
Database (stores) column names/objects lower case fixes.

kparmar1:
Database (stores MySQL/Postgres/H2) column names/objects changed to lower case.
The issue raised was #1055

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/DatabaseFullPrunedBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/DatabaseFullPrunedBlockStore.java
@@ -103,27 +103,27 @@ public abstract class DatabaseFullPrunedBlockStore implements FullPrunedBlockSto
     private static final String INSERT_SETTINGS_SQL                             = "INSERT INTO settings(name, value) VALUES(?, ?)";
     private static final String UPDATE_SETTINGS_SQL                             = "UPDATE settings SET value = ? WHERE name = ?";
 
-    private static final String SELECT_HEADERS_SQL                              = "SELECT chainWork, height, header, wasUndoable FROM headers WHERE hash = ?";
-    private static final String INSERT_HEADERS_SQL                              = "INSERT INTO headers(hash, chainWork, height, header, wasUndoable) VALUES(?, ?, ?, ?, ?)";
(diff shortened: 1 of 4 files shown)


traber:
It seems to work for me but presumably I did not trigger the execution of all SQL statements during my tests. This is why I also manually checked the source. In DatabaseFullPrunedBlockStore I found one left over of a camel case column name in the SQL statement provided by the INSERT_OPENOUTPUTS_SQL constant ("toAddress").

kparmar1:
Hi, Thanks. I'll fix the one I missed.
K.

kparmar1:
I've made the fix post Mission Impossible (Not as good as the reviews suggested but well worth a watch :-)

schildbach:
I wonder is this backwards compatible to databases already created?

ollekullberg:
MySQL can be case sensitive, and I saw at least one table name changing case: 

"the case sensitivity of the underlying operating system plays a part in the case sensitivity of database and table names. This means database and table names are not case sensitive in Windows, and case sensitive in most varieties of Unix." 

https://dev.mysql.com/doc/refman/5.0/en/identifier-case-sensitivity.html

mikehearn:
lol, mysql ....

schildbach:
I'm a bit unsure on how to deal with this PR. If column names can be case sensitive, this is a breaking change. And I can imagine a pretty nasty one.

If the "users" of the database-based block stores are ok with this change I'm happy to merge.

kparmar1:
Unfortunately this is potentially a breaking change - for some people depending on platform. I couldn't see an easy way to make everything consistent without this happening. It's something that needs to be done however as it's just too messy using case in table/column names *inconsistently*  and I can see this becoming an even bigger issue later if not rectified (if more people start to use the DB backed stores).

I'm hoping that users understand that the fullstores are experimental and until stable some of these issues will arise.

There are other options of course. Changes are managed-upgrade paths (schema versioning) and/or provide scripts for data migrations (of course this all could be automated to a certain degree).

schildbach:
I'm not aware of any coming enhancements. I can imagine they will be restricted to "add column" and "add table", so a version would not be strictly required.

I'll merge this for now. Migration is easy enough for existing users.

PR #1058, opened 2015-08-07
Name change: ProtobufParser -> ProtobufConnection

ollekullberg:
After working with ProtobufParser for 1/2 a year (we use it in the Stroem implementation) I find the name misleading. A name change would be breaking, but I still request it.

The responsibility of PP is to read raw data and chunk it so that complete protobuf messages comes out. The name is probably a twist on its super class: StreamParser, but PP also has a "write()" method, which indicates that it also can send, not only parse. 

When working with payment channels I see PP as the main gateway to the network; when I call "write()" the message will be sent to the server, and I don't have to care much how it is done. The code I write would make more sense if the name is changed to ProtobufConnection.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -18,6 +18,7 @@ package org.bitcoinj.core;
 
 import com.google.common.base.*;
 import com.google.common.base.Objects;
+import org.bitcoinj.net.StreamConnection;
 import org.bitcoinj.core.listeners.PeerConnectionEventListener;
 import org.bitcoinj.core.listeners.PeerDataEventListener;
 import org.bitcoinj.store.BlockStore;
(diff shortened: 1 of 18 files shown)


mikehearn:
Perhaps StreamParser -> StreamConnection also makes sense?

ollekullberg:
Yes. I fix.

ollekullberg:
Now its 18 changed files, but all the tests ran, and i made sure variable names were changed too.

schildbach:
Nickpick: You could use the unicode right arrow character (→) in the description.

ollekullberg:
Yeah, it also saves a char :-)

ollekullberg:
Now we got some conflicts. If we are good to go I can resolve the conflict and update this PR?

mikehearn:
We are good to go! If you fix the conflicts, let's merge.

ollekullberg:
Done.

schildbach:
You got a merge commit on your branch.

ollekullberg:
Yeah, rebaseing after merge conflict is boring, but it's done now :-D

schildbach:
Merged. Many thanks!

PR #1059, opened 2015-08-09
a transaction output may have multiple OP_RETURNs

peacekeeper:
E.g. see this:

https://testnet3.toshi.io/api/v0/transactions/f18045a1b7395a4b793bfd48d919c62d4a88ed5a0e3b7fbe0b12572b6696698a

Bitcoinj should recognize this TransactionOutput as OP_RETURN

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/Script.java
+++ b/core/src/main/java/org/bitcoinj/script/Script.java
@@ -750,7 +750,7 @@ public class Script {
     }
 
     public boolean isOpReturn() {
-        return chunks.size() == 2 && chunks.get(0).equalsOpCode(OP_RETURN);
+        return chunks.size() > 0 && chunks.get(0).equalsOpCode(OP_RETURN);
     }
 
     /**


mikehearn:
It looks OK but that's not a standard OP_RETURN output, is it? Why would anyone create an output like this?

peacekeeper:
To store more than 40 bytes in an output.. I created this output myself, using Bitcoinj. This is related to the discussions around storing arbitrary data in the blockchain, such as proof-of-existence or open assets.

Here's a discussion about this:
http://bitcoin.stackexchange.com/questions/37626/can-i-have-multiple-op-return-outputs-in-one-transaction

It sounds like it's not a "standard transaction", but it doesn't get rejected either, so I thought isOpReturn() should recognize it... But I also have to say I'm quite new to this, so not 100% sure.

luke-jr:
Uh, nothing stops you from pushing more than 40 bytes in a single pushdata...

peacekeeper:
Again I'm new to this, but I thought there's a well-known 40 byte limit, and therefore some people do multiple OP_RETURNs in a single output.

E.g. see here:
http://bitzuma.com/posts/op-return-and-the-future-of-bitcoin/

"The 0.9.0 release of Bitcoin Core added a new standard transaction type granting access to a previously disallowed script function, OP_RETURN. This function accepts a user-defined sequence of up to 40 bytes."

luke-jr:
The "40 byte limit" is mere node policy, *same as the one-pushdata limit*, except the 40 bytes can actually be changed easily...

mikehearn:
OK, this change is small and unlikely to affect anything negatively, even though it's kind of weird. Happy to merge this when github stops being DoSd.

schildbach:
Merged, thanks!

PR #1060, opened 2015-08-11
Wallet: Rename attribute "keychain" to "keyChainGroup"

oscarguindzberg:
"keychain" type is "KeyChainGroup" so being called "keychain" is misleading and could make the developer think it holds a "KeyChain".

I renamed locks to match this change.

I made the "C" letter uppercase in the few variable and method names it was not, so now you can read "keyChain" instead of "keychain" everywhere.

I also renamed Wallet method names that might led the developer to think it is updating or getting info from the Wallet's KeyChain when in fact it is the Wallet's KeyChainGroup. Eg getKeychainSize() was renamed to getKeyChainGroupSize(). 
I know changing public method names is a breaking change. This PR is for 0.14 so it will give developers time to update their code.
If this is controversial, I am open to:
a) re-add the old methods, deprecate them and make them call the new methods so we can remove the old methods after a while.
b) Keep the public method names as they were and just change the internal attributes and variable names.


schildbach:
Merged, thanks!

PR #1061, opened 2015-08-11
DeterministicKeyChain: Rename "externalKey" and "internalKey" 

oscarguindzberg:
DeterministicKeyChain: Rename "externalKey" and "internalKey" attribute names to "externalParentKey" and "internalParentKey"


schildbach:
Merged.

PR #1062, opened 2015-08-11
Remove redundant check

oscarguindzberg:
Since the code is written in MarriedKeyChain there is no way this check can fail.


schildbach:
Seems right. I'll merge.

PR #1063, opened 2015-08-11
DKC toString() fix

oscarguindzberg:
Removed "if" because I could not find a use case where the parent fingerprint is missing.
Parent key may be missing, but that does screwes up the xpub.

The "if" was making MarriedKeyChain.toString() to display the following keychains xpubs but not the  followed keychain xpub

Bellow is a piece of code that shows my point

```
public class WalletToStringTest {

        Context context = Context.getOrCreate(MainNetParams.get()); 
        Wallet wallet = new Wallet(context);
        System.out.println(wallet.toString());
        
        Wallet w2 = getWalletMultisig(context.getParams());
        System.out.println(w2.toString());

        Wallet w3 = getWalletWatchonly(context.getParams());
        System.out.println(w3.toString());

        Wallet w4 = getWalletWatchonly2(context.getParams());
        System.out.println(w4.toString());
        
    }
    
    private static String getNewRandomXpub(NetworkParameters params) {
        DeterministicKeyChain keyChain = new DeterministicKeyChain(new SecureRandom());
        return keyChain.getWatchingKey().serializePubB58(params);
    }
    
    private static Wallet getWalletMultisig(NetworkParameters params) {
        DeterministicKey key = DeterministicKey.deserializeB58(null, getNewRandomXpub(params), params);
        DeterministicKeyChain keyChain; 
        List<DeterministicKey> followingAccountKeys = new ArrayList<DeterministicKey>();
        DeterministicKey followingAccountKey = DeterministicKey.deserializeB58(null, getNewRandomXpub(params), params);
        followingAccountKeys.add(followingAccountKey);
        MarriedKeyChain marriedKeyChain = MarriedKeyChain.builder()
                 .seedCreationTimeSecs(new Date().getTime() / 1000)
                 .watchingKey(key)
                 .followingKeys(followingAccountKeys)
                 .threshold(2).build();
        keyChain = marriedKeyChain;
        KeyChainGroup kcg = new KeyChainGroup(params);
        kcg.addAndActivateHDChain(keyChain);        
        Wallet wallet = new Wallet(params, kcg);
        return wallet;
    }

    private static Wallet getWalletWatchonly(NetworkParameters params) {
        DeterministicKey key = DeterministicKey.deserializeB58(null, getNewRandomXpub(params), params);
        Wallet wallet = Wallet.fromWatchingKey(params, key);
        return wallet;
    }

    private static Wallet getWalletWatchonly2(NetworkParameters params) {
        DeterministicKey key = DeterministicKey.deserializeB58(null, getNewRandomXpub(params), params);
        DeterministicKeyChain keyChain = DeterministicKeyChain.watch(key);
        KeyChainGroup kcg = new KeyChainGroup(params);
        kcg.addAndActivateHDChain(keyChain);        
        Wallet wallet = new Wallet(params, kcg);
        return wallet;
    }
    
}
```

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
@@ -1340,13 +1340,7 @@ public class DeterministicKeyChain implements EncryptableKeyChain {
                     Utils.dateTimeFormat(seed.getCreationTimeSeconds() * 1000)));
         }
         final DeterministicKey watchingKey = getWatchingKey();
-        // Don't show if it's been imported from a watching wallet already, because it'd result in a weird/
-        // unintuitive result where the watching key in a watching wallet is not the one it was created with
-        // due to the parent fingerprint being missing/not stored. In future we could store the parent fingerprint
-        // optionally as well to fix this, but it seems unimportant for now.
-        if (watchingKey.getParent() != null) {
-            builder2.append(String.format("Key to watch:  %s%n", watchingKey.serializePubB58(params)));
-        }
+        builder2.append(String.format("Key to watch:  %s%n", watchingKey.serializePubB58(params)));
         formatAddresses(includePrivateKeys, params, builder2);
         return builder2.toString();
     }


schildbach:
@mikehearn What do you think on this one?
@oscarguindzberg Can you change your commit description to a more descriptive one? Also better not use abbreviations like DKC.

mikehearn:
@devrandom is doing more advanced things with this API than I am. I think the parent fingerprint can be missing sometimes when deserializing from certain forms?

devrandom:
@oscarguindzberg - we have a use case where we vend public keys to API users, but don't want to disclose to them how we derived it.  This is mostly to hide operational details about how we generate our keys from the API user.  That said, as long as this doesn't actually break (e.g. throw an exception), we don't care if that statement is removed.

oscarguindzberg:
@schildbach Commit message updated
@devrandom About your use case, the xpub of the following chains are being shown already. WalletToStringTest  (see the PR description for the source code) checks we don't get exceptions. Please, let me know if I missed any edge case.

devrandom:
@oscarguindzberg - no, I don't think you missed anything - thank you for checking.

mikehearn:
OK, the comment makes me think this code exists to work around an actual issue that was discovered. But it may be that the code evolved later so it's no longer needed, and as you say you checked, we can make this change. If there are issues they'll be discovered during testing of master. LGTM.

schildbach:
Merged.

PR #1068, opened 2015-08-23
Add BIP 34 enforcement

rnicoll:
This patch primarily enforces block height being present in the coinbase transaction input, altough it introduces a number of other fixes and changes to support this.

* VersionTally now returns the number of blocks at or above a version, rather than just at, to enable forward-compatible support (i.e. v3 blocks include all v2 block tests)
* Block version is now explicitely provided in most tests which generate blocks, in order to ensure correct tests are applied
* Block height is now used when generating coinbase transactions
* Added support for the chain parameters to determine which tests apply to a block, so altcoins can override the defaults if needed.
* Added structures to support full BIP 66 validation checks

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -436,14 +436,25 @@ public abstract class AbstractBlockChain {
                 return true;
             }
 
+            final StoredBlock storedPrev;
+            final Integer height;
+            final Set<VerificationFlags> flags;
+
             // Prove the block is internally valid: hash is lower than target, etc. This only checks the block contents
             // if there is a tx sending or receiving coins using an address in one of our wallets. And those transactions
             // are only lightly verified: presence in a valid connecting block is taken as proof of validity. See the
             // article here for more details: http://code.google.com/p/bitcoinj/wiki/SecurityModel
             try {
                 block.verifyHeader();
+                storedPrev = getStoredBlockInCurrentScope(block.getPrevBlockHash());
+                if (null != storedPrev) {
+                    height = storedPrev.getHeight() + 1;
+                } else {
+                    height = null;
+                }
+                flags = params.getValidationFlags(block, this.versionTally, height);
                 if (shouldVerifyTransactions())
-                    block.verifyTransactions();
+                    block.verifyTransactions(height, flags);
             } catch (VerificationException e) {
                 log.error("Failed to verify block: ", e);
                 log.error(block.getHashAsString());
@@ -451,7 +462,6 @@ public abstract class AbstractBlockChain {
             }
 
             // Try linking it to a place in the currently known blocks.
-            StoredBlock storedPrev = getStoredBlockInCurrentScope(block.getPrevBlockHash());
 
             if (storedPrev == null) {
                 // We can't find the previous block. Probably we are still in the process of downloading the chain and a
@@ -526,7 +536,7 @@ public abstract class AbstractBlockChain {
             // net, less on test) in order to be applied. It is also limited to
             // stopping addition of new v2 blocks to the tip of the chain.
             if (block.getVersion() == Block.BLOCK_VERSION_BIP34) {
-                final Integer count = versionTally.getCount(Block.BLOCK_VERSION_BIP66);
+                final Integer count = versionTally.getCountAtOrAbove(Block.BLOCK_VERSION_BIP66);
                 if (count != null
                     && count >= params.getMajorityRejectBlockOutdated()) {
                     throw new VerificationException.BlockVersionOutOfDate(block.getVersion());
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -437,8 +437,8 @@ public abstract class AbstractBlockChain {
             }
 
             final StoredBlock storedPrev;
-            final Integer height;
-            final Set<VerificationFlags> flags;
+            final int height;
+            final EnumSet<VerificationFlags> flags;
 
             // Prove the block is internally valid: hash is lower than target, etc. This only checks the block contents
             // if there is a tx sending or receiving coins using an address in one of our wallets. And those transactions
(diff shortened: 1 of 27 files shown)


rnicoll:
If there's a better solution for packing the height into a byte array for use in the script, please let me know. Otherwise will look at adding generic support into ScriptBuilder later.

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
```diff
@@ -436,14 +436,25 @@ private boolean add(Block block, boolean tryConnecting,
...
+            final Set<VerificationFlags> flags;
+
             // Prove the block is internally valid: hash is lower than target, etc. This only checks the block contents
             // if there is a tx sending or receiving coins using an address in one of our wallets. And those transactions
             // are only lightly verified: presence in a valid connecting block is taken as proof of validity. See the
             // article here for more details: http://code.google.com/p/bitcoinj/wiki/SecurityModel
             try {
                 block.verifyHeader();
+                storedPrev = getStoredBlockInCurrentScope(block.getPrevBlockHash());
+                if (null != storedPrev) {
```
mikehearn: Style: we use storedPrev != null

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
```diff
@@ -436,14 +436,25 @@ private boolean add(Block block, boolean tryConnecting,
...
+                    height = storedPrev.getHeight() + 1;
+                } else {
+                    height = null;
+                }
+                flags = params.getValidationFlags(block, this.versionTally, height);
```
mikehearn: 'this' here is redundant

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -610,10 +617,21 @@ private Sha256Hash calculateMerkleRoot() {
...
+     * @param height block height, if known. If provided, used to validate the
+     * coinbase input script of v2 and above blocks.
+     * @throws VerificationException if there was an error verifying the block.
+     */
+    private void checkTransactions(final Integer height, final Set<VerificationFlags> flags)
+            throws VerificationException {
         // The first transaction in a block must always be a coinbase transaction.
         if (!transactions.get(0).isCoinBase())
             throw new VerificationException("First tx is not coinbase");
+        if (flags.contains(VerificationFlags.HEIGHT_IN_COINBASE) && null != height) {
```
mikehearn: ditto

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -610,10 +617,21 @@ private Sha256Hash calculateMerkleRoot() {
...
 
-    private void checkTransactions() throws VerificationException {
+    /**
+     * Verify the transactions on a block.
+     *
+     * @param height block height, if known. If provided, used to validate the
+     * coinbase input script of v2 and above blocks.
+     * @throws VerificationException if there was an error verifying the block.
+     */
+    private void checkTransactions(final Integer height, final Set<VerificationFlags> flags)
```
mikehearn: use -1 here instead of boxing, as this is in a hot path

mikehearn: Set -> BitSet, yes Set is more general but BitSet is clearer and doesn't rely on the JVM to notice it can do type sharpening to avoid interface dispatch overhead.

mikehearn: Or just use a traditional int with a bitfield. Set<Flag> is quite a lot of overhead.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1164,6 +1167,50 @@ public int getSigOpCount() throws ScriptException {
     }
 
     /**
+     * Generate script data bytes to represent the given block height.
+     */
+    public static byte[] generateHeightScriptData(final int height) {
```
mikehearn: This probably should not be in Transaction

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1164,6 +1167,50 @@ public int getSigOpCount() throws ScriptException {
...
+        }
+    }
+
+    /**
+     * Check block height is in coinbase input script, for use after BIP 34
+     * enforcement is enabled.
+     */
+    public void checkCoinbaseHeight(final int height)
+            throws VerificationException {
+        // Check block height is in coinbase input script
```
mikehearn: Shouldn't there be an assert here on isCoinbase() ?

core/src/test/java/org/bitcoinj/core/AbstractFullPrunedBlockChainTest.java
```diff
@@ -340,4 +351,58 @@ public void testUTXOProviderWithWallet() throws Exception {
...
+            height++;
+
+            // Trying to add a broken v2 block should now result in rejection as
+            // we have a v2 supermajority
+            thrown.expect(VerificationException.CoinbaseHeightMismatch.class);
+            chainHead = chainHead.createNextBlockWithCoinbase(Block.BLOCK_VERSION_BIP34,
+                outKey.getPubKey(), height * 2);
+            chain.add(chainHead);
+        }  catch(final VerificationException ex) {
+                throw (Exception) ex.getCause();
```
mikehearn: Indentation.

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
rnicoll: Can't convince you to chance, can I? Putting the constant first means accidentally typing "=" instead of "!=" or "==" fails at compilation instead of runtime.

mikehearn: That's an old rule that's valid for C++ but not Java. Try it and you'll find it generates a compile error.

core/src/main/java/org/bitcoinj/core/Block.java
rnicoll: Given it's an EnumSet, and only called once per block, is it worth the saving?

core/src/main/java/org/bitcoinj/core/Transaction.java
rnicoll: Added. Also realised it's "CoinBase" elsewhere, changing the method name to match.

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
rnicoll: Huh, didn't even occur to check. Thanks, will get that updated.

rnicoll:
Think that covers all of the comments, please let me know if I've missed anything. I'll squash once you're happy with it.

mikehearn:
Merged it, thanks!

Issue #1064, opened 2015-08-13
Tor connections not being made


jim618:
In MultiBit HD we have had Tor support in our Labs section for a while but are seeing problems in it connecting in our error reports.  We see, repeatedly:

```
{"@timestamp":"2015-08-13T12:05:41.267+01:00","@version":1,"message":"Certificate read for unknown directory authority with identity: 27b6b5996c426270a5c95488aa5bceb6bcc86956","logger_name":"com.subgraph.orchid.directory.DirectoryImpl","thread_name":"Thread-9","level":"WARN","level_value":30000}
{"@timestamp":"2015-08-13T12:05:42.355+01:00","@version":1,"message":"Consensus signed by unrecognized directory authority: 27b6b5996c426270a5c95488aa5bceb6bcc86956","logger_name":"com.subgraph.orchid.directory.consensus.ConsensusDocumentImpl","thread_name":"Thread-9","level":"WARN","level_value":30000}
{"@timestamp":"2015-08-13T12:05:46.141+01:00","@version":1,"message":"Failed to download key certificates: Request /tor/keys/fp-sk/e8a9c45ede6d711294fadf8e7951f4de6ca56b58-86832be318b3775ac21b45d1896dcc92b27f3d8b+ed03bb616eb2f60bec80151114bb25cef515b226-cd7159a8de14bc6bdc7e5e1e51adc89e162fca08+585769c78764d58426b8b52b6651a5a71137189a-6b82b0ec44bd79cb0d1f1bb2a0c597e0fec71ae9+d586d18309ded4cd6d57c18fdb97efa96d330566-3a8218840c58f0f35b1eefaf3c39fe46fbac842b+80550987e1d626e3eba5e5e75a458de0626d088c-414e42cca5f51e7b0c020496ab0f4b7894fcd409+49015f787433103580e3b66a1707a00e60f2d15b-f98e385f2982778f50925f54f832e2fe744b5ed7+efcbe720ab3a82b99f9e953cd5bf50f7eefc7b97-193b238d2e74692fd16fd11404ac661399fc0a65+14c131dfc5c6f93646be72fa1401c02a8df2e8b4-34cabfa62e3f374d0d67f7bd3d104196baef18ba to directory (none) returned error code: 404 Not found","logger_name":"com.subgraph.orchid.directory.downloader.DirectoryDownloadTask","thread_name":"DirectoryDownloadTask worker-0","level":"WARN","level_value":30000}
```

[...]

Issue #633, opened 2015-06-11
IrcDiscovery reader and writer never closed


mikehearn:
Originally reported on Google Code with ID 267

```
I'd suggest closing these in the finally block.
```

Reported by `andreas.schildbach` on 2013-01-07 00:11:21

PR #1069, opened 2015-08-24
Fix Tor support does not use all available HTTP seeds.

schildbach:


mikehearn:
I think this should be using a MultiplexingDiscovery, no? Otherwise each one will be queried sequentially in a blocking manner with no mixing or parallelisation.

schildbach:
Ok, so I think I fixed it. How is this?

mikehearn:
Right, that looks better. Did you try it out? WalletTemplate can be made to use Tor just by uncommenting a line in Main.java

schildbach:
Yes I tried it and it appears to work. Merged.

gary-rowe:
Just a heads up on Tor - I thought I'd add it here since it may be related. 

Recently MultiBit HD users have been sending in error reports concerning Tor connectivity issues. After investigation we upgraded to Bitcoin 0.13.2/Orchid 1.1.1 but that didn't fix the problems so we've had to (hopefully temporarily) remove Tor support.

There's more information in [MultiBit HD Issue #633](https://github.com/bitcoin-solutions/multibit-hd/issues/633) which has a general overview and various links to error reports providing earlier detail. 

Do you think this fix will address the timeout problems in 1.1.1? We'd obviously like to have Tor support back in for our 0.1.4 release (expected late September) but for 0.1.3 it has to be out pending a Bitcoinj/Orchid release.

Anything we can do to help?

mikehearn:
It's unrelated. I've seen the "unable to download key certificates" error as well but haven't had time to investigate it yet. I probably won't for a while.

gary-rowe:
OK, I'll copy my comment into #1064 which is more specific to the issue.

PR #1070, opened 2015-08-24
Support for requiring specific services when connecting to peers.

schildbach:


mikehearn:
Looks good to me. Might want to add some javadocs to the PeerDiscovery interface noting that this is a bitmask matched against the nServices field of the version message peers send.

schildbach:
Merged.

PR #1072, opened 2015-08-25
Added TOTAL_RECEIVED as a balance type 

snakecharmer1024:
Added a new balance type to simplify keeping track of the amount of coin flowing thru the system. Chose to build off of ESTIMATE_SPENDABLE as opposed to AVAILABLE_SPENDABLE to avoid temporary reductions in the balance whenever coins are sent due to change not being included in the calculation.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/core/Wallet.java
@@ -3092,14 +3092,7 @@ public class Wallet extends BaseTaggableObject
         /** Same as ESTIMATED but only for outputs we have the private keys for and can sign ourselves. */
         ESTIMATED_SPENDABLE,
         /** Same as AVAILABLE but only for outputs we have the private keys for and can sign ourselves. */
-        AVAILABLE_SPENDABLE,
-
-        /**
-         * Amount of bitcoin ever received via output. If an output spends from a transactions whose inputs
-         * are also to our wallet, the input amounts are deducted from the outputs contribution, with a minimum of zero
-         * contribution.
-         */
-        TOTAL_RECEIVED
+        AVAILABLE_SPENDABLE
     }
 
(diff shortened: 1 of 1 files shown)


mikehearn:
What's the use case for this?

snakecharmer1024:
I was using it to keep statistics on how much was flowing thru the wallet. It's something some businesses, particularly exchanges will list as one of their metrics. It's also useful for an app that regularly pays out to some other address, but wants to be able to report how much the user has made from the app, which isn't possible with the existing balance types.

schildbach:
I think that usecase would be better dealt with a separate method. It has nothing to do with a balance. A balance is something you can spend.

snakecharmer1024:
Ah, that makes sense. Balance = debits - credits, not just debits... that's easy enough to change.

snakecharmer1024:
Ok now it's just a separate method, not a balance type. Is there anything else I should address?

snakecharmer1024:
Anything else for me to do here?

schildbach:
Can you squash your two commits into one? The "calculate" in the method name feels a bit redudant to me, what about just `totalReceived()`?
I think it would be good to have a unit test for your method. Something like: receiving 2 BTC, sending 1 BTC, receiving another 2 and the totalReceived() should be 3 BTC.

Did you consider to also implement a totalSpent() equivalent method? (Just asking, not needed for merging)

schildbach:
@snakecharmer1024 Did you see my last comments? Generally I'm happy to merge this if it's useful to you.

snakecharmer1024:
Hey Andreas, I did see your last comment. Been very busy, but going to work on this now. Thanks for the feedback!

snakecharmer1024:
Also, I think in the test you suggested, totalReceived should be 4 BTC. In any case, I'll write tests for these things.

Issue #1067, opened 2015-08-23
increasePaymentFuture does not complete when ERROR is received


ollekullberg:
I'm in the middle of testing the payment channel implementation, and I've fallen upon an interesting error case. In PaymentChannelClient we have a future:

    @GuardedBy("lock") SettableFuture<PaymentIncrementAck> increasePaymentFuture;

... and this future is used by incrementPayment(). When PAYMENT_ACK is returned from the serve the increasePaymentFuture is updated correctly. But what if the server responds with an ERROR instead of a PAYMENT_ACK? In this case the error message is written to debug and the (socket) connection is destroyed.

 case ERROR:
                        checkState(msg.hasError());
                        log.error("Server sent ERROR {} with explanation {}", msg.getError().getCode().name(),
                                msg.getError().hasExplanation() ? msg.getError().getExplanation() : "");
                        conn.destroyConnection(CloseReason.REMOTE_SENT_ERROR);
                        return;

What I failed to understand using this API, was that waiting for the increasePaymentFuture is not enough; in case of error I will wait until timeout. Is this by design? If so, I guess I could try to abort the "increasePaymentFuture.get()" in my destroyConnection() impl, but that will be a bit sticky.

If this is not by design we could try to set increasePaymentFuture to error as soon as the error has been received?

mikehearn:
Yes, just set increasePaymentFuture to contain an exception.

We should probably draw the state machine and verify that every branch is handled correctly.

ollekullberg:
I took a look at the PaymentChannelClient, and drew a state machine, see below. Didn't find anything strange, even though it is admittedly tricky logic :-) 
![pc_client_state_machine](https://cloud.githubusercontent.com/assets/201249/9451907/b2ce4858-4ab1-11e5-81f2-c1a18cac5021.png)

mikehearn:
Wow nice job.  That should be checked into the source tree and referenced
from the javadocs!

msgilligan:
Payment channels were removed by PR #1902.

PR #1075, opened 2015-08-29
Close increasePaymentFuture when receiving an ERROR from the server.

ollekullberg:
In PaymentChannelClient we have a future:

   @GuardedBy("lock") SettableFuture<PaymentIncrementAck> increasePaymentFuture;

... and this future is used by incrementPayment(). When PAYMENT_ACK is returned from the serve the increasePaymentFuture is updated correctly. This PR makes sure the increasePaymentFuture is set even if the server returns an error.

### DIFF
--- a/core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelClient.java
+++ b/core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelClient.java
@@ -338,12 +338,14 @@ public class PaymentChannelClient implements IPaymentChannelClient {
                         checkState(msg.hasError());
                         log.error("Server sent ERROR {} with explanation {}", msg.getError().getCode().name(),
                                 msg.getError().hasExplanation() ? msg.getError().getExplanation() : "");
+                        setIncreasePaymentFutureIfNeeded(CloseReason.REMOTE_SENT_ERROR, msg.getError().getCode().name());
                         conn.destroyConnection(CloseReason.REMOTE_SENT_ERROR);
                         return;
                     default:
                         log.error("Got unknown message type or type that doesn't apply to clients.");
                         errorBuilder = Protos.Error.newBuilder()
                                 .setCode(Protos.Error.ErrorCode.SYNTAX_ERROR);
+                        setIncreasePaymentFutureIfNeeded(CloseReason.REMOTE_SENT_INVALID_MESSAGE, "");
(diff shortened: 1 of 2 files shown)


ollekullberg:
Oh, I should probably link to this issue: https://github.com/bitcoinj/bitcoinj/issues/1067

mikehearn:
After this change testClientErrorHandlingIncreasePaymentError fails for me pretty reliably:

java.lang.AssertionError: Should not time out

	at org.junit.Assert.fail(Assert.java:88)
	at org.bitcoinj.protocols.channels.ChannelConnectionTest.testClientErrorHandlingIncreasePaymentError(ChannelConnectionTest.java:346)

ollekullberg:
Strange, works for me. I even did this TDD style, began with a failing test and got it to work by adding the fix. Now when re-reading the code I can only find one error, and that's that I should have used lock when updating the future: 

    private void setIncreasePaymentFutureIfNeeded(PaymentChannelCloseException.CloseReason reason, String message) {
        if (increasePaymentFuture != null && !increasePaymentFuture.isDone()) {
            lock.lock();
            try {
                increasePaymentFuture.setException(new PaymentChannelCloseException(message, reason));
            } finally {
                lock.unlock();
            }
        }
    }

I will add the lock, but I don't think it's the problem you're facing. When I run the test I get the following output at the end. The last log is telling me that bitcoinj actually received the ERROR. Do you get the same logs?
 
    10:37:03 1 PaymentChannelClientState.incrementPaymentBy: Signing new payment tx   eb9784450b91bf34f135dd6455df609f36df11d422b146a70c00be6bf8709361: Unknown confidence level.
     in    1.00 BTC
          outpoint:a7ac3929b3fd0d5f3d1aca8f52268ff87973e6081782eb64c8b9cb9528f0e314:0
     out  DUP HASH160 PUSHDATA(20)[8a1bb17e580ae94e631aa145e855dd4e81f04cd4] EQUALVERIFY CHECKSIG 0.98999 BTC
     fee  0.01001 BTC

    10:37:03 1 StoredPaymentChannelClientStates.updatedChannel: Stored client channel 31959752 was updated
    10:37:03 1 PaymentChannelClient.receiveMessage: Server sent ERROR CHANNEL_VALUE_TOO_LARGE with explanation

schildbach:
@mikehearn Do you still see the problem with @ollekullberg 's updated code?

mikehearn:
Nope, merged. Thanks!

PR #1076, opened 2015-09-01
Allowing settlement of unused channel

cyberzac:
The client should be able to settle a channel after it has payed the requested minimum payment specified in the Initiate message.


ollekullberg:
With the current impl, when a channel has been opened and we want to settle it immediately we are 1 satoshi short :-)

PR #1077, opened 2015-09-01
Rename getChangeAddress() to currentChangeAddress()

oscarguindzberg:
To match currentReceiveAddress()


schildbach:
LGTM. It would be interesting to know why the comment thinks the method will go away in future.

mikehearn:
LGTM

The method was scheduled for destruction because we knew that we'd do rotating addresses at some point :)

schildbach:
Merged, thanks. Of course the real question was: maybe it's time to remove the method entirely?

mikehearn:
Making it a deprecated alias seems OK to me.

PR #1079, opened 2015-09-04
READY FOR REVIEW - Fix #1078. Add ability to load Wallet without loading its transactions. ...

### LINKED ISSUE #1078: Loading bad transactions causes wallet loading failure, but loading wallet must succeed before one can remove offending transactions
*UPDATE: Fixed in #1079.*

`WalletProtobufSerializer.readWallet` always attempts to load transactions before returning a valid `Wallet`. Loading transaction can cause that process to fail. Unfortunately one needs that process to succeed before being able to call `Wallet.reset` to remove the offending transactions. :dizzy_face: :dizzy:

This vicious cycle was can be observed in situations like [this](https://github.com/bitcoinj/bitcoinj/issues/735#issuecomment-137327732). Unfortunately, there is no current work-around, aside from copy-paste-patching `WalletProtobufSerializer.readWallet`. `wallet-tool` is of no help because its `reset` and `sync` commands call `WalletProtobufSerializer.readWallet`.

posita:
… Use new methods with `wallet-tool reset` (where transactions are deleted anyway) and with `wallet-tool sync` (if the `--force` option is provided).

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/WalletProtobufSerializer.java
+++ b/core/src/main/java/org/bitcoinj/store/WalletProtobufSerializer.java
@@ -68,7 +68,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
  * You can extend the wallet format with additional fields specific to your application if you want, but make sure
  * to either put the extra data in the provided extension areas, or select tag numbers that are unlikely to be used
  * by anyone else.<p>
- * 
+ *
  * @author Miron Cuperman
  * @author Andreas Schildbach
  */
@@ -235,7 +235,7 @@ public class WalletProtobufSerializer {
     private static Protos.Transaction makeTxProto(WalletTransaction wtx) {
         Transaction tx = wtx.getTransaction();
         Protos.Transaction.Builder txBuilder = Protos.Transaction.newBuilder();
-        
+
         txBuilder.setPool(getProtoPool(wtx))
                  .setHash(hashToByteString(tx.getHash()))
                  .setVersion((int) tx.getVersion());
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/bitcoinj/store/WalletProtobufSerializer.java
```diff
@@ -450,33 +491,40 @@ public Wallet readWallet(NetworkParameters params, @Nullable WalletExtension[] e
...
-        // Update transaction outputs to point to inputs that spend them
-        for (Protos.Transaction txProto : walletProto.getTransactionList()) {
-            WalletTransaction wtx = connectTransactionOutputs(txProto);
-            wallet.addWalletTransaction(wtx);
-        }
-
-        // Update the lastBlockSeenHash.
-        if (!walletProto.hasLastSeenBlockHash()) {
+        if (forceReset) {
+            // Should mirror Wallet.reset()
```
schildbach: Why not actually *use* Wallet.reset()?

posita: Mainly because I was unclear on the side effects. `Wallet.reset` also calls [`Wallet.saveLater`](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/core/Wallet.java#L1403) and [`Wallet.maybeQueueOnWalletChanged`](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/core/Wallet.java#L2392). I don't know if these calls are safe to make while the Wallet is in the middle of being created.

schildbach: Fair point, ok.

posita: That being said, someone editing `Wallet.reset` may not know to edit this as well. I can refactor the common code out to a `private` or `protected` method if you think it's worth it?

schildbach:
Merged, thanks!

PR #1087, opened 2015-09-27
Dependencies in orchid and wallettemplate

peacekeeper:
It seems to me the following dependencies may be missing?

In orchid:
org.easymock:easymock:3.4
org.apache.xmlrpc:xmlrpc-client:3.1.3

In wallettemplate:
com.oracle:javafx:2.2

### DIFF
--- a/orchid/pom.xml
+++ b/orchid/pom.xml
@@ -151,6 +151,11 @@
             <artifactId>guava</artifactId>
             <version>18.0</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.xmlrpc</groupId>
+            <artifactId>xmlrpc-client</artifactId>
+            <version>3.1.3</version>
+        </dependency>
     </dependencies>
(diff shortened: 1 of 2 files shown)


schildbach:
Generally: any changes to Orchid (also to its pom.xml) need to be accompanied with an Orchid version increase. I can help with that.

peacekeeper:
hello thanks for looking at it, i'm sorry i may not have time to look into this further and update this pull request. maybe you want to close the pull request and instead just do any dependency updates yourself.

schildbach:
Ok thanks. Closing this for now. I'm happy to re-open if you have time to look into my concerns.

mikehearn:
WalletTemplate uses JavaFX 8 which is a part of Java 8 (and OpenJFX8 should be sufficient - they differ mostly in video codec support as far as a I know). So that change would be wrong.

PR #1088, opened 2015-09-27
Don't add Maven descriptor to build artifacts.

schildbach:
It's pretty useless and it stands in the way of reproducible builds.


mikehearn:
LGTM

PR #1090, opened 2015-10-01
Add script verification flags for DER format

rnicoll:
Add script verification flags for DER format encoding, low-S signatures and strict encoding.
Pass script verification flags through to signature parser.
Convert block verification flags to a sub-enum of Block.
Remove DER signature format verification flag from block flags.
Add test transaction with a non-standard DER signature, with the verify flag set/unset accordingly, to tx_invalid.json and tx_valid.json

Note this changes behaviour of Script.correctlySpends(Transaction, long, Script) as the method applies all flags, which now includes the new DER flag. The method version which takes in flags as an argument is unaffected.


rnicoll:
Have now fixed the excessive log spam that was upsetting Travis.

core/src/test/resources/org/bitcoinj/script/tx_invalid.json
```diff
@@ -109,5 +109,9 @@
 [[["ad503f72c18df5801ee64d76090afe4c607fb2b822e9b7b63c5826c50e22fc3b", 0, "0x21 0x027c3a97665bf283a102a587a62a30a0c102d4d3b141015e2cae6f64e2543113e5 CHECKSIG NOT"]],
 "01000000013bfc220ec526583cb6b7e922b8b27f604cfe0a09764de61e80f58dc1723f50ad0000000000ffffffff0101000000000000002321027c3a97665bf283a102a587a62a30a0c102d4d3b141015e2cae6f64e2543113e5ac00000000", "P2SH"],
 
+["A transaction with a non-standard DER signature."],
```
schildbach: I take it you created these test cases yourself? That's great!

There is only one issue with that: we sync these .json files from Bitcoin Core on a regular basis. So maybe you can submit your testcases to Core too?

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -1297,7 +1333,7 @@ private static void executeCheckSig(Transaction txContainingThis, int index, Scr
             // This RuntimeException occurs when signing as we run partial/invalid scripts to see if they need more
             // signing work to be done inside LocalTransactionSigner.signInputs.
             if (!e1.getMessage().contains("Reached past end of ASN.1 stream"))
-                log.warn("Signature checking failed! {}", e1.toString());
+                log.warn("Signature checking failed! {}", e1);
```
schildbach: I think in this case the "{}" stays unused, because the warn(String, Throwable) overload is used.

core/src/test/resources/org/bitcoinj/script/tx_invalid.json
rnicoll: Looks like Bitcoin Core has a number of DERSIG test cases I'd missed, so when you sync next it will replace my basic case with several tests instead.

rnicoll: Actually, I'll drop their tests in now.

rnicoll:
Okay, just cutting & pasting in the Bitcoin script tests isn't as trivial as I thought, I'll look at syncing both test data sets later on. Bitcoin doesn't have a transaction test (my mistake), so I have submitted the two test cases as bitcoin/bitcoin#6848.

Unused "{}" removed as well.

schildbach:
Merged. Thanks!

PR #1092, opened 2015-10-05
Fixed imported keys not being passed onto UTXOProvider

GlennMR:


core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -3971,7 +3971,8 @@ public boolean canSignFor(Script script) {
     protected List<UTXO> getStoredOutputsFromUTXOProvider() throws UTXOProviderException {
         UTXOProvider utxoProvider = checkNotNull(vUTXOProvider, "No UTXO provider has been set");
         List<UTXO> candidates = new ArrayList<UTXO>();
-        List<DeterministicKey> keys = getActiveKeyChain().getLeafKeys();
+        List<ECKey> keys = getImportedKeys();
+        keys.addAll(getActiveKeyChain().getLeafKeys());
```
schildbach: Looks reasonable, although this code assumes we are the owner of the imported keys list. Are we really?

GlennMR: @schildbach I would assume we are since we were in posession of the private key we imported earlier.

schildbach: Sorry, what I meant is: Are we allowed to modify the list we got from getImportedKeys()?

GlennMR: @schildbach ah I see, the getImportedKeys method calls the basic.getKeys() and this method returns a **copy** of the imported keys list:

BasicKeyChain.java getKeys() returns this:
```
return new ArrayList<ECKey>(hashToKeys.values());
```

So I think it's safe to modify this returned list.

schildbach: @mikehearn From the API design perspective, do you think it's ok to modify the list we got from Wallet.getImportedKeys()? If so, should we perhaps document the allowed ownership takeover?

mikehearn: Yeah, we're copying it deliberately and yes we could document that. In general we should aim for owned-by-caller by default, and if we return a pointer to mutable internal state for performance reasons or otherwise it should be that which is explicitly documented and called out.

schildbach:
I just merged this fix, thanks!

I think it would be nice to have a unit test that prevents this bug from happening in future.

core/src/main/java/org/bitcoinj/core/Wallet.java
schildbach: Thanks for the clarification. Owned-by-caller default should really be documented as a project convention.

PR #1094, opened 2015-10-14
Make transactions with signatures with "high" S component violate the…

schildbach:
… "canonical encoding" rule.

This makes those transactions non-standard and risky. By default, they will not be accepted into wallets unless confirmed by the block chain.

I think this change is needed because of the ongoing malleability attack. I'm aware there are still wallets that don't always create low S components but frankly I don't care. Their authors must feel the heat to move.


schildbach:
I just updated the PR to use a separate RuleViolation value. I think that's better, as the definition of a "signature encoding" violation is defined by Bitcoin Core.

schildbach:
Just minutes after I wrote this, Bitcoin Core released an updated version that indeed makes the low S rule part of the signature encoding rule. So I'll change it back.

mikehearn:
LGTM

PR #1095, opened 2015-10-15
Add hooks to override protocol version numbers

rnicoll:
Enables altcoin networks which may have out of sync protocol numbers, and also moves version numbers into a single enum rather than having them scattered through the code.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+++ b/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
@@ -47,11 +47,6 @@ import org.bitcoinj.utils.VersionTally;
  */
 public abstract class NetworkParameters {
     /**
-     * The protocol version this library implements.
-     */
-    public static final int PROTOCOL_VERSION = 70001;
-
-    /**
      * The alert signing key originally owned by Satoshi, and now passed on to Gavin along with a few others.
      */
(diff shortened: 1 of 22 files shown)


core/src/main/java/org/bitcoinj/core/NetworkParameters.java
```diff
@@ -505,4 +500,23 @@ public int getMajorityWindow() {
...
+
+    public static enum ProtocolVersion {
+        MINIMUM(70000),
+        PONG(60001),
+        BLOOM_FILTER(70000),
+        CURRENT(70001);
+
+        private int bitcoinProtocol;
+
+        ProtocolVersion(final int setNum) {
```
schildbach: "setNum" feels like a weird name for a method parameter. Maybe "bitcoinProtocol"?

schildbach:
Code looks ok.

I realize one of the PeerAddress constructors changed in a backwards incompatible way. I wonder if we should wait with this refactoring until the Context refactoring is complete (contexts contain network params). Though I'm not sure if the Context refactoring is planned to touch PeerAddress, maybe @mikehearn knows?

rnicoll:
Changed the constructor variable name, and fixed the backwards incompatible change in PeerAddress. May want to deprecate some of the PeerAddress constructors later, though.

rnicoll:
Unit tests run fine locally, and there's no obvious errors in the logs. Can you take a look and unless I've missed something, restart the build, see if that fixes it?

schildbach:
Merged!

PR #1096, opened 2015-10-16
Set minimum fee to 5000 satoshis, matching Bitcoin Core 0.11.1.

schildbach:
However, I'm not entirely convinced we need this. There will always be a certain percentage of nodes that use the old relay rules, e.g. Bitcoin XT. What do others think?

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -91,16 +91,15 @@ public class Transaction extends ChildMessage {
 
     /**
      * If fee is lower than this value (in satoshis), Bitcoin Core will treat it as if there were no fee.
-     * Currently this is 1000 satoshis.
      */
-    public static final Coin REFERENCE_DEFAULT_MIN_TX_FEE = Coin.valueOf(1000);
+    public static final Coin REFERENCE_DEFAULT_MIN_TX_FEE = Coin.valueOf(5000); // satoshis
 
     /**
      * Any standard (ie pay-to-address) output smaller than this value (in satoshis) will most likely be rejected by the network.
      * This is calculated by assuming a standard output will be 34 bytes, and then using the formula used in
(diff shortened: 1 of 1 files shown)


mikehearn:
The dust value needs to change too, no?

Unfortunately this change isn't optional. Miners will eventually all upgrade and get this "fix"

schildbach:
I was asking you about the dust on IRC a couple of days ago. Does it need to be multiplied by 5 too?

alexkuck:
https://github.com/bitcoin/bitcoin/pull/6722 implemented a floating min relay fee and https://github.com/bitcoin/bitcoin/commit/536766c903c8dc84e6709e4bd387aad5acf8e433#diff-e8db9b851adc2422aadfffca88f14c91 reverted DEFAULT_MIN_RELAY_TX_FEE back to 1000 satoshis.

Luckily, we no longer must follow Core's haphazard play.

schildbach:
Thanks, good to know.

mikehearn:
Well, not really. All it means is that now the min relay fee is entirely unpredictable.

alexkuck:
Outside of trusting an external source, how could a lightweight client form an opinion about the min relay fee before propagating its' own transaction?

The client could collect a sample set of recent transactions and reason the min relay fee from this (likely cursory) survey?

mikehearn:
You can determine confirmation-time/price buckets in an SPV manner with some protocol upgrades and very complicated code. I do not know a way you can determine min relay fee that way as you have nothing to measure it against.

schildbach:
I just asked on bitcoin-dev and they said the new dust limit is 2730 sat. I adjusted this PR.

Also they said we should not keep this as a constant – calculate it from the fee instead.

Anyhow, I fear we might need to merge this. Although the change is already undone for Bitcoin Core 0.12 it's not going to be released before February. The share of Core 0.11.1/0.11.2 is already quite high.

schildbach:
Merged.

PR #1098, opened 2015-10-18
Replace magic numbers for header sizes with constants

rnicoll:


schildbach:
Thanks for the PR. Replacing "80" by "HEADER_SIZE" is fine, but I would probably replace "81" by "HEADER_SIZE + 1" rather than introducing another constant for almost the same thing. What do you think?

rnicoll:
Actually that was what the first draft did, and then wondered if it would be better to have two constants. Updated and pushed

schildbach:
Yeah, I think that's better. Merged, thanks.

PR #1099, opened 2015-10-18
Clean up MessageSerializer to simplify subclasses

rnicoll:
By replacing MessageSerializer with an abstract class, methods which should just call other methods in the same class are now fixed, and subclasses only have to override a single abstract method instead of multiple methods.


schildbach:
Merged, thanks for the patch!

PR #1100, opened 2015-10-18
Correct length of block header when hashing at an offset

rnicoll:


core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -230,7 +230,7 @@ protected void parse() throws ProtocolException {
         time = readUint32();
         difficultyTarget = readUint32();
         nonce = readUint32();
-        hash = Sha256Hash.wrapReversed(Sha256Hash.hashTwice(payload, offset, cursor));
+        hash = Sha256Hash.wrapReversed(Sha256Hash.hashTwice(payload, offset, cursor - offset));
```
schildbach: This doesn't seem right. A couple of lines above, there is "cursor = offset", so that cursor-offset would always be 0.

rnicoll: The readUint32() and readHash() calls increment cursor as they go along, so at the end "cursor - offset" represents the length of the data that's just been parsed. Right now, offset is always 0, so just "cursor" has worked fine, but I have a later patch that uses non-zero offsets and breaks here.

schildbach:
You're right! Merged.

Issue #1097, opened 2015-10-17
Commit d3d11df for BIP34 has issue on Block 224430 


n1bor:
Issue is that the coinbase has a load of junk after the height.
https://blockchain.info/tx/e55cf5e0442aadf707d122282a26a134b145c3ce240a4a621d3f596ea140c6cc

So Transaction.java can not parse script and so throws ValidationException.

Infact script does not NEED to be valid just needs to have height set correctly.

rnicoll:
Now #1101 has been applied, can you confirm you're happy this is fixed and can be closed, please?

schildbach:
Assuming this is fixed on master.

PR #1101, opened 2015-10-18
Correct issue #1097

rnicoll:
Changes coinbase height validation to check for the bytes at the start of the coinbase script, and ignore everything after them. The validation then matches Bitcoin Core. Corrects #1097 

Also update error messages to more closely match latest Bitcoin Core.


mikehearn:
Merged, thanks!

PR #1102, opened 2015-10-18
Add support for block headers of variable length

rnicoll:
Add support for block headers of variable length, when parsing headers messages. Also removes a lot of double-handling of messages (i.e. copying subparts of the message out so they are parsed in isolation rather than within the context of the message)


rnicoll:
Failed test has a note "// TODO: This test sometimes fails due to relying on padding.", and it works locally. Could you restart the Travis build please?

schildbach:
I assume this is for Dogecoin? Because Bitcoin always has the same size of block headers…

rnicoll:
It is, but also removes the array copying that resulted from the old design.

schildbach:
Ok fair enough. I merged this.

rnicoll:
Thanks!

PR #1052, opened 2015-07-28
Remove lazy parsing of messages.

schildbach:
In particular:

* Message.parsed: removed
* Message.parseLite(): folded into Message.parse()
* Message.maybeParse(): removed
* Message.ensureParsed(): removed
* Message.isParsed(): removed
* Block.parseHeader(): folded into Block.parse()
* Block.parseTransactions(): folded into Block.parse()
* Block.maybeParseHeader(): removed
* Block.maybeParseTransactions(): removed
* Block.ensureParsedHeader(): removed
* Block.ensureParsedTransactions(): removed
* Block.isParsedHeader(): removed
* Block.isParsedTransasctions(): removed
* MessageSerializer.isParseLazyMode(): removed
* BitcoinSerializer.parseLazy: removed
* BitcoinSerializer.getSerializer(): parseLazy parameter removed
* LazyParseException: removed
* LazyParseByteCacheTest: renamed to ParseByteCacheTest

PR #1104, opened 2015-10-24
Remove redundant return statements

langerhans:
They have been redundant since #1052 removed the respective maybeParse...() calls.


schildbach:
Merged, thanks!

PR #1105, opened 2015-10-25
Add common superclass for block messages

rnicoll:
Opening this for initial discussion. This adds a common superclass (AbstractBlockHeader) for things that are block header-like. Specifically, Block and FilteredBlock both subclass this new class, and a new BlockHeader class has been added. The peer data event listener that previously took in a block and a filtered block now takes in just a single AbstractBlockHeader which encompasses the same data. This also makes it clearer whether a block object has transactions, as it's now indicated by the type.

I've added a cloneAsBlock() method as an interim solution for type coercion where a header needs to be used as a block, which is now just in AbstractBlockChain when adding a new block.

I'd appreciate feedback, especially on:

* Is AbstractBlockHeader the right name, would AbstractBlock, PartialBlock or something else be better?
* Would it be a good idea to have individual listener event methods for the three subclasses, or is it reasonable to expect applications receiving events to check type if they need specific subtypes?
* Block and FilteredBlock do not inherit from BlockHeader to enforce type safety, are you happy with that model? It's more verbose but IMHO safer.


rnicoll:
Rebased to correct conflicts

core/src/main/java/org/bitcoinj/core/AbstractBlockHeader.java
```diff
@@ -0,0 +1,630 @@
...
+import java.math.BigInteger;
+import java.util.Date;
+import java.util.EnumSet;
+import javax.annotation.Nullable;
+import org.bitcoinj.script.Script;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ *
```
schildbach: This could use a class level JavaDoc.

tools/src/main/java/org/bitcoinj/tools/WalletTool.java
```diff
@@ -761,7 +761,7 @@ public void onCoinsSent(Wallet wallet, Transaction tx, Coin prevBalance,
             case BLOCK:
                 peers.addDataEventListener(new AbstractPeerDataEventListener() {
                     @Override
-                    public void onBlocksDownloaded(Peer peer, Block block, @Nullable FilteredBlock filteredBlock, int blocksLeft) {
+                    public void onBlocksDownloaded(Peer peer, AbstractBlockHeader block, int blocksLeft) {
```
schildbach: I notice at several places, you drop the @Nullable. Why?

schildbach:
It feels like a pretty invasive change. Can you say what's the goal of the abstraction? For example, will it simplify the API at some point?

From your questions, I currently can only answer the first: I'd prefer AbstractBlock. Will look into the other questions soon.

Any feedback from others?

tools/src/main/java/org/bitcoinj/tools/WalletTool.java
rnicoll: The entire parameter is dropped. A major part of what this does is to clean up the API so Block and FilteredBlock have a common superclass.

rnicoll:
This primarily resolves the type hierarchy issues referred to in https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/core/listeners/PeerDataEventListener.java#L31, that FilteredBlock fundamentally is very similar to a block, but doesn't extend a common superclass with Block, resulting in a strange API when notifying listeners of new blocks downloaded, that provides different parameters depending on what has been downloaded.

It is a large change, though, and there's no rush, so it could be bumped back to 0.15 given we're now quite far into 0.14's development.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -251,45 +199,31 @@ protected void parseTransactions(final int transactionsOffset) throws ProtocolEx
         transactionBytesValid = serializer.isParseRetainMode();
     }
 
+    /** Returns a copy of the block, but without any transactions. */
```
schildbach: Really a copy? I see "return this" below.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -478,20 +344,7 @@ protected final void copyBitcoinHeaderTo(final Block block) {
...
-                isBIP65() ? "BIP65" : null, isBIP101() ? "BIP101" : null);
-        if (!bips.isEmpty())
-            s.append(" (").append(bips).append(')');
-        s.append('\n');
-        s.append("   previous block: ").append(getPrevBlockHash()).append("\n");
-        s.append("   merkle root: ").append(getMerkleRoot()).append("\n");
-        s.append("   time: [").append(time).append("] ").append(Utils.dateTimeFormat(time * 1000)).append("\n");
-        s.append("   difficulty target (nBits): ").append(difficultyTarget).append("\n");
-        s.append("   nonce: ").append(nonce).append("\n");
+        StringBuilder s = super.toStringBuilder();
```
schildbach: I think a better pattern would be a `void appendToString(StringBuilder b)` in the super class that appends to the given builder.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -572,15 +382,16 @@ private void checkSigOps() throws VerificationException {
         for (Transaction tx : transactions) {
             sigOps += tx.getSigOpCount();
         }
-        if (sigOps > MAX_BLOCK_SIGOPS)
+        if (sigOps > MAX_BLOCK_SIGOPS) {
```
schildbach: This is an unrelated change snuck in. Sorry for bitching, but it makes reviewing harder for big refactorings like this one.

core/src/main/java/org/bitcoinj/core/FilteredBlock.java
```diff
@@ -131,16 +151,11 @@ public boolean equals(Object o) {
...
-    }
-
-    @Override
-    public int hashCode() {
-        return Objects.hashCode(associatedTransactions, header, merkleTree);
+            && super.equals(other) && merkleTree.equals(other.merkleTree);
     }
 
-    @Override
+    /* @Override
```
schildbach: Why commenting out?

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
```diff
@@ -106,7 +106,7 @@
 
     // Holds a block header and, optionally, a list of tx hashes or block's transactions
     class OrphanBlock {
-        final Block block;
+        final AbstractBlock block;
```
schildbach: Why is this now an `AbstractBlock`? The constructor still can take only `Block`s.

schildbach:
> Would it be a good idea to have individual listener event methods for the three subclasses, or is it reasonable to expect applications receiving events to check type if they need specific subtypes?

I would think that especially with the listener refactoring we've already done, individual event methods would lead to an explosion of listeners? Am I right. If so, I vote for staying as-is (which is close to before the refactoring, right?).

> Block and FilteredBlock do not inherit from BlockHeader to enforce type safety, are you happy with that model? It's more verbose but IMHO safer.

Yes, that's fine for now. I'm scared of unintended side effects to be honest, so I vote for the safer model. We can switch to a full inheritance model later if we want.

core/src/main/java/org/bitcoinj/core/BlockHeader.java
```diff
@@ -0,0 +1,112 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bitcoinj.core;
+
+import javax.annotation.*;
+
+/**
+ * <p>A block is a group of transactions, and is one of the fundamental data structures of the Bitcoin system.
```
schildbach: This needs an updated JavaDoc: If I'm not mistaken, this is a "Message" that actually doesn't exist in the Bitcoin P2P protocol. So why does it exist?

rnicoll:
Don't review too closely right now, this is really more just a proof of concept for discussion. If you're happy it's a positive change (I'd like to remove the comment about refactoring the type hierarchy either way, but given the scope of this will understand if you decide it's not actually that good an idea), I'll clean it up and test it, which is likely to takes a few weeks given the scope

core/src/main/java/org/bitcoinj/core/FilteredBlock.java
rnicoll: Meant to come back and replace it - doing that now via appendToStringBuilder()

core/src/main/java/org/bitcoinj/core/BlockHeader.java
rnicoll: It's a child of HeadersMessage, but yes, it's not used as a stand-alone message. Not quite sure what to do with that, unless we're going to start reshuffling the Message heirarchy in a major way (i.e. make Message and ChildMessage interfaces, rename Message to AbstractMessage)

rnicoll:
Okay, I've given up on a separate BlockHeader class, this now just provides a common superclass for Block and FilteredBlock, which makes it much less invasive.

rnicoll:
Are you okay with the basic idea of what this does, now I've removed BlockHeader and it just adds a common superclass for Block and FilteredBlock?

schildbach:
I will defer this to after 0.14, but plan to come back to this as one of the first things in the cycle.

PR #1107, opened 2015-10-27
Enforce v4 block switchover

rnicoll:
This enforces the switch over to block v4 once a supermajority is met, as per BIP65. Does not add any handling of OP_CHECKLOCKTIMEVERIFY yet.


schildbach:
Merged, thanks!

PR #1108, opened 2015-10-27
Enforce the LOW_S script validation flag

rnicoll:
I've added in the code used to generate the test cases, as it might be an interesting example. Let me know if you'd rather I remove it.

I'll raise the two test cases to be added to Bitcoin Core, later in the week.


schildbach:
This somehow feels redundant to my earlier work on this. See a4c9ad68941c0f0335d6adae58e4dc52584ccd27 and 17071ce15ccd25505d048ccf2381afbf79959f03.

Can you explain how this adds enforcement that was not already there?

rnicoll:
This mostly moves the script validation options closer to matching Bitcoin Core, so if we wanted to adapt the script engine to use libconsensus later, this makes it easier to adapt.

Also means if BIP 62 is ever resurrected and full validation means rejecting low-S signatures in blocks (not just in relayed transactions), the support's there to do it.

schildbach:
Thanks for explanation. Merged.

PR #1110, opened 2015-11-03
Copy accessor for the channel map

cyberzac:


schildbach:
Can you describe what you need this for? The commit description could be a bit more verbose.

cyberzac:
In the server I've written on top of bitcoin I want to see which channels that are created. Typically for displaying them in an admin UI, I could use the information to manually close channels. Today it is possible to read the StoredServerChannel if I have the contract hash, i.e. the Sha256Hash. There is no way to list all contracts.

schildbach:
Thanks, merged.

PR #1112, opened 2015-11-07
Use maximum money value from network parameters

rnicoll:
Enables support for altcoins with different maximum money values.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Coin.java
+++ b/core/src/main/java/org/bitcoinj/core/Coin.java
@@ -83,15 +83,30 @@ public final class Coin implements Monetary, Comparable<Coin>, Serializable {
      */
     public final long value;
 
-    private Coin(final long satoshis) {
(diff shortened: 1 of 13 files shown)


schildbach:
That's a pretty hefty change just for getting rid of the MAX_COINS direct access in Coin.

I'm not sure why I put the range change into Coin at all. Coin is meant to be chain agnostic. The only thing it should be concerned about is not to overflow its own calculations. I'm pretty sure errors like trying to send more coins than MAX_COINS are caught in the higher layers of Bitcoin (-:

So what do you think about removing the check entirely? That should get rid of almost all your changes, no?

rnicoll:
Sounds good, will replace with a simpler patch later this week

schildbach:
I just opened PR https://github.com/bitcoinj/bitcoinj/pull/1130. Can you check if that solves your problem with Coin?

rnicoll:
Sorry, completely forgot about this! #1130  good, thanks.

PR #1113, opened 2015-11-08
Wallet-Tool: New action to modify the creation time of wallets.

schildbach:
This PR makes DeterministicSeed.creationTimeSeconds mutable. Since the creation time can be recovered by a full scan of the block chain, I don't consider this a risk.


rnicoll:
Not entirely sure of the use-case, but don't see any real risk either, or if there's real concern could have separate actual creation time and effective creation times stored as separate fields perhaps?

mikehearn:
LGTM, agree that the use case is unspecified though.

schildbach:
The usecase is repairing wallets that were accidently created with a clock in the future. Yes, it happens in real-life )-:  I'll add a comment before merging.

PR #1089, opened 2015-09-29
Ensure "OP_0" and "false" have the same represenstation on stack

dexX7:
During the script execution via [executeScript](https://github.com/bitcoinj/bitcoinj/blob/5dcf6439755ac5db6e151ba4403b5d3e8bf792eb/core/src/main/java/org/bitcoinj/script/Script.java#L763) certain operations add boolean values to the stack.

Using Bitcoin Core as reference, in particular:

- `OP_EQUAL`: [bitcoin/.../script/interpreter.cpp#L676](https://github.com/bitcoin/bitcoin/blob/48efbdbe986355bd2478f0fdd366b20952fbf30a/src/script/interpreter.cpp#L676)
- `OP_CHECKSIG`: [bitcoin/.../script/interpreter.cpp#L851](https://github.com/bitcoin/bitcoin/blob/48efbdbe986355bd2478f0fdd366b20952fbf30a/src/script/interpreter.cpp#L851)
- `OP_CHECKMULTISIG`: [bitcoin/.../script/interpreter.cpp#L947](https://github.com/bitcoin/bitcoin/blob/48efbdbe986355bd2478f0fdd366b20952fbf30a/src/script/interpreter.cpp#L947)

Boolean values, which are added to the stack as result, are represented as follows in Bitcoin Core:

- `true`: as vector with one element: `[1]`
- `false`: as empty vector with zero elements: `[]`

See:

- [bitcoin/.../script/interpreter.cpp#L244](https://github.com/bitcoin/bitcoin/blob/48efbdbe986355bd2478f0fdd366b20952fbf30a/src/script/interpreter.cpp#L244)
- [bitcoin/.../script/interpreter.cpp#L246](https://github.com/bitcoin/bitcoin/blob/48efbdbe986355bd2478f0fdd366b20952fbf30a/src/script/interpreter.cpp#L246)

However, in BitcoinJ the representation of `false` in these cases differs:

It is represented as Byte array with one element, which is zero: `[0]`

See:

- `OP_EQUAL`: [bitcoinj/.../script/Script.java#L1025](https://github.com/bitcoinj/bitcoinj/blob/5dcf6439755ac5db6e151ba4403b5d3e8bf792eb/core/src/main/java/org/bitcoinj/script/Script.java#L1025)
[...]

PR #1115, opened 2015-11-16
Fixed behaviour of OP_0 to not NullPointerException

w-shackleton:
`OP_0` was throwing a `NullPointerException` when chunk.data was attempted to access inside the if statement below. Since the behaviour of `OP_0` is to push an empty array to the stack I modified the behaviour as so.

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/Script.java
+++ b/core/src/main/java/org/bitcoinj/script/Script.java
@@ -822,8 +822,13 @@ public class Script {
         
         for (ScriptChunk chunk : script.chunks) {
             boolean shouldExecute = !ifStack.contains(false);
-            
-            if (!chunk.isOpCode()) {
+
+            if (chunk.opcode == OP_0) {
+                if (!shouldExecute)
+                    continue;
+
+                stack.add(new byte[] {});
+            } else if (!chunk.isOpCode()) {
                 if (chunk.data.length > MAX_SCRIPT_ELEMENT_SIZE)
                     throw new ScriptException("Attempted to push a data string larger than 520 bytes");
                 
--- a/core/src/test/java/org/bitcoinj/script/ScriptTest.java
+++ b/core/src/test/java/org/bitcoinj/script/ScriptTest.java
@@ -221,6 +221,18 @@ public class ScriptTest {
         }
     }
 
+    @Test
+    public void testOp0() {
+        // Check that OP_0 doesn't NPE and pushes an empty stack frame.
+        Transaction tx = new Transaction(params);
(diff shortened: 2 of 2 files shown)


schildbach:
I'm currently not confident enough in scripts to review this properly. Can anyone have a look, @TheBlueMatt maybe?

@w-shackleton It would be nice if your PR would include a test case that would fail with old code and pass with your fix applied. It's not a requirement for merging, but it would perhaps make understanding the issue easier.

w-shackleton:
Will do.

The purpose of `OP_0` is to push an "Empty array" of bytes onto the stack. While in C++ `vector` this is unambiguous, in java this could mean `null` or `new byte[] {}`. bitcoinj uses `null` and so when the regular instruction handler tries to read the stack value length we crash.

TheBlueMatt:
Honestly, my strong preference would be to remove all of the script execution stuff. If you want script execution functionality you should be building the consensus library for your platform and link it in... It is very unmaintained and people relying on it for consensus is rather dangerous :(. I had a patch for this a while back when I was testing the consensus library when I was working in it, it was pretty simple.

schildbach:
I squashed your two commits and merged. Thanks for spotting and fixing this.

@TheBlueMatt I guess an JNI adapter for the consensus library would be a valuable addition to bitcoinj.

dexX7:
> @w-shackleton: bitcoinj stores null as the data for an OP_0 and so when the interpreter tries to read its length it crashes.

Just to underline this point: actually BitcoinJ should use `new byte[] {}`, and not `null` or `new byte[] {0}`.

See also #1089.

> Since the behaviour of OP_0 is to push an empty array to the stack I modified the behaviour as so.

This seems like the right way to do.

schildbach:
@dexX7 This PR is already merged. If you know of occurences which use the wrong representation can you submit a PR to fix it, or open an issue?

dexX7:
@schildbach: yeah, this was just a mention, so no one gets the idea `null` may be the correct representation.

> If you know of occurences which use the wrong representation can you submit a PR to fix it, or open an issue?

Back then I tested BitcoinJ with the script unit tests of Bitcoin Core, and the failing tests I've found were all related to missing script flags, such as `DISCOURAGE_UPGRADABLE_NOPS` or `MINIMALDATA`.

Here is an older test report, which may provide an insight:

- http://bitwatch.co/uploads/reports/java-libbitcoinconsensus/149aa39248-bitcoinj/classes/co.bitwatch.libbitcoinconsensus.DataDrivenBitcoinJSpec.html

The report was generated before #1089 was merged, if I recall correctly. 

But other than that, I'm currently not aware of other wrong representations or issues. :)

PR #1116, opened 2015-11-18
Block: Convenience methods for BIP conformance

schildbach:

PR #1117, opened 2015-11-19
Fixes the coinbase height check.

kparmar1:
This fixes the height check for a coinbase tx. It appears that on the wire the height can be any number of bytes long, with the maximum being 4 bytes, but there doesn't appear to be any consistency, I.e a height that fits within 2 bytes is sometime received using 2 bytes, shortest representation possible, and sometime 3 (a trailing zeros byte).


schildbach:
If that's true it'd be a pretty good catch! Do you have any reference, e.g. a real block that fails the current check? Bonus points for adding a testcase (-:  See my latest PR #1116 for how to load blocks as a binary blob and perform tests on them.

kparmar1:
Someone should def check this.... I'm not at all familiar with the script part of bitcoinj.

(I may not have fully understood and there might be a different easier fix for this).

core/src/main/java/org/bitcoinj/script/ScriptBuilder.java
```diff
@@ -350,11 +350,12 @@ public static Script createOpReturnScript(byte[] data) {
...
-    public static byte[] createHeightScriptData(final int height) {
-        // TODO: Replace with something generic to any integer value
-        final byte[] int32Buffer = ByteBuffer.allocate(4).order(ByteOrder.LITTLE_ENDIAN).putInt(height).array();
-        if (int32Buffer[3] == 0) {
-            return Arrays.copyOf(int32Buffer, 3);
+    public static byte[] createHeightScriptData(final int height, final int numberOfBytes) {
+        final int maxNumberOfBytes = 4;
+        checkArgument(numberOfBytes <= maxNumberOfBytes);
+        final byte[] int32Buffer = ByteBuffer.allocate(maxNumberOfBytes).order(ByteOrder.LITTLE_ENDIAN).putInt(height).array();
+        if (numberOfBytes != maxNumberOfBytes) {
```
schildbach: I'd write '<' here, rather than '!='. It's a bit clearer imho.

kparmar1:
I encountered this on testnet ... Block 21066 (first), and then another block further on (should have noted that down, but I didn't- opposite problem, i.e with zeros).
I'll have a look at your PR and see if I can work out how to write a test :-) (I have little knowledge about the structure - might be a good time to learn I guess:-)

schildbach:
Basically you save the block to disk into the resources directory (same Java package as the unit test) and then you can use `Block block = TestNet3Params.get().getDefaultSerializer().makeBlock(ByteStreams.toByteArray(getClass().getResourceAsStream("block.dat")));` to load it and then you can assert on it.

kparmar1:
Ok thanks for the help :-)

schildbach:
Hmm, block 21066 on Testnet was on Aug 24, 2012. The BIP34 switch on Mainnet happened on Mar 24, 2013. I wonder when the Testbet BIP34 switch happened (if at all).

I'm wondering because before BIP34 is triggered, the coinbase could have any form afaik (it's an unused script). So maybe there is a problem with the BIP34 switchover logic. @rnicoll, What do you think?

schildbach:
@kparmar1 I also wonder how does this issue affect you? Do you use full validation? At least you don't bloom filter, I guess?

rnicoll:
I see the problem - on main net coinbase heights are always at least three bytes, while on test net BIP 34 blocks start a lot earlier, and can use just two bytes. There's an earlier patch that changes how the test data is generated so it doesn't care what's after the height, but it expects at least three bytes still.

Writing up a quick alternative patch that fixes this and adds the test block.

kparmar1:
Yes I'm using full validation (FullStore backend). If you are writing a patch shall I not do anything further then (finish off the test .....)?

kparmar1:
FYI: This isn't really true though...It's not just 2 bytes, a choice can be made between 2, 3, or even 4 (with 0's). Later (on testnet( I have seen 3 bytes used even though the height fits into 2 (initially I just removed all zero's but this doesn't work as it's inconsistent).

rnicoll:
Can you give me an example (by block height is great) that uses the wrong number of bytes? The Bitcoin Core code seems to be doing basically the same test I'm doing: https://github.com/bitcoin/bitcoin/blob/master/src/main.cpp#L2803

schildbach:
The BIP34 spec says:

> The format of the height is "serialized CScript" -- first byte is number of bytes in the number, following bytes are little-endian representation of the number.

So in theory the number could be up to 255 bytes long (-:  I guess we're good if we support 1 to 4 bytes.

kparmar1:
This is on testnet but unfortunately I can't very quickly.  I didn't write down the block number down :( I should have. I can create but it will be a little while to reproduce now.

rnicoll:
bitcoinj currently represents block height internally as a 32 bit signed integer, so at least we've got bigger problems when we overflow that!

kparmar1:
I don’t think you have to worry about that, 2³²-1 - I think it's unsigned also BTW.

You will be long gone by then :-)

kparmar1:
(I would have used 64 though :-)

rnicoll:
That's my draft fix in #1119 - if you can test it and if you do find any other blocks it breaks on, please do let me know.

kparmar1:
Wow you are quick (I feel like a little snail :-).

Shall I forget my PR then? I'll close it?

How did you get the block serialised so quickly? Is there a way to get them online (from either testnet/main)?

rnicoll:
It helps I do this a lot! I use a copy of Bitcoin Core on testnet, and then to get the hex dump of the block the RPC commands:

getblockhash <height>
getblock <hash from previous command> false

I then paste that into a text file, use "xxd -r -p block.hex > block.dat", replacing block.hex with your input and output filenames as appropriate. xxd is a UNIX command, but you can get it for Windows if needed under Cygwin.

Do check my patch fixes things before we abandon this one!

kparmar1:
Thanks that's really helpful.  I was doing it by debugging my java code and running evaluation - not to clever there :-)

schildbach:
I was using http://webbtc.com/ for this yesterday. However currently is seems down.

@kparmar1 Don't worry about coding speed. There are more important qualities. Let's keep this PR open until the issue is fixed.

kparmar1:
Ok I identified the block that caused the opposite problem- 32768 (this is the one that uses an extra byte at the end- 2 for number and 1 zeros).

rnicoll:
Great, loading it up now, thanks

kparmar1:
Thanks I'll bookmark the site.
I'll finish it off with the unit tests/plus updates- if it's used fine, if not NP :-)

rnicoll:
I'm currently trying to figure how on earth that gets past the Bitcoin checks, because I want to match them as closely as possible. Leading theory is that it's ignoring the length byte at the start, but still tracing back through the code.

kparmar1:
I have to admit I've never really looked at the code in core. I think I would be lost as my C++ is non-existent (not used since Uni). Hope you track it down.

rnicoll:
Got it - my method for encoding values into script sometimes produces shorter results than the C++ code does. Rewriting the method to closely match https://github.com/bitcoin/bitcoin/blob/master/src/script/script.h#L303

kparmar1:
Well I've finished this PR off anyway,but it looks like it may not be needed.  I did have a feeling this was something upstream, but didn't have the knowledge. Maybe I should make an effort to look at core sometimes :-)

rnicoll:
What I've done, if you want to take a look, is completely replaced the number encoding. Rather than just satisfying what we needed for coinbase, it now handles any number. It's also quirk-compatible with the Bitcoin Core encoder (which sometimes adds unnecessary space).

Also it now handles 64 bit integers, and I'm really not going to be here for when that's an issue :-D

luke-jr:
BIP34 requires the serialised height to be the minimum possible encoding. Are you overlooking the sign bit?

schildbach:
Closing this in favor of #1119. However @kparmar1 that was a very good catch, so thank you!

PR #1120, opened 2015-11-19
Use Guava Stopwatch for measuring code execution time.

schildbach:


mikehearn:
LGTM

Is it really an improvement though?

schildbach:
I think yes, because it increases readability of code. And the dedicated type for time measurement can prevent mistakes. Thanks for reviewing!

PR #1123, opened 2015-11-27
Added core support for BIP65 micropayment channels, added test tools for checking functionality

w-shackleton:
I added support to the core bitcoinj library for creating, spending and refunding BIP65-style micropayment channel transactions. I also added commands to `WalletTool` for using each of these behaviours and checking that they can be used in testnet.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/core/Wallet.java
@@ -3643,6 +3643,25 @@ public class Wallet extends BaseTaggableObject
             return req;
         }
 
+        public static SendRequest toLockTimeVerify(NetworkParameters params, Date releaseTime, ECKey from, ECKey to, Coin value) {
+            long time = releaseTime.getTime() / 1000L;
+            checkArgument(time >= Transaction.LOCKTIME_THRESHOLD, "Release time was too small");
+            return toLockTimeVerify(params, BigInteger.valueOf(time), from, to, value);
+        }
+
+        public static SendRequest toLockTimeVerify(NetworkParameters params, int releaseBlock, ECKey from, ECKey to, Coin value) {
+            checkArgument(0 <= releaseBlock && releaseBlock < Transaction.LOCKTIME_THRESHOLD, "Block number was too large");
+            return toLockTimeVerify(params, BigInteger.valueOf(releaseBlock), from, to, value);
+        }
+
+        private static SendRequest toLockTimeVerify(NetworkParameters params, BigInteger time, ECKey from, ECKey to, Coin value) {
+            SendRequest req = new SendRequest();
+            Script output = ScriptBuilder.createLockTimeVerifyOutput(time, from, to);
+            req.tx = new Transaction(params);
+            req.tx.addOutput(value, output);
+            return req;
+        }
+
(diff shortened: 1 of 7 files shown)


snakecharmer1024:
Hey great work on this!

A few small things:
1) There are lots of different types of CLTV transactions, but the method name createLockTimeVerifyOutput sounds like this is the only or at least canonical use of this opcode. Maybe we should make the method name more specific? Or maybe refactor CLTV-ness into a different function? 
2) Can we make the default signature type sighash.SINGLE? The reason for this is twofold: first that's how the existing bitcoinj MPC code works, second it allows for receiver to aggregate multiple MPCs into one receive transaction, which has a number of benefits (e.g. privacy and even lower lower bound on transaction size per person) and is a project I plan on taking up soon. There might be some delicate issues here I haven't considered.
3) What does this error message mean? fail("Spend before script failed to correctly spend");

Other than that it looks pretty good to me. An elegant solution.

w-shackleton:
I've renamed `LockTimeVerify` named things to `CLTVPaymentChannel`, and fixed the other things.

snakecharmer1024:
Looks good to me. Thank you for doing this. I don't have authority to pull, so I'll leave it here.

schildbach:
Code style looks good to me too, and the change seems not very invasive. I'm happy to merge this, if you think the code is ready for merge.

However it looks like we could squash the commits a bit, no?

Thanks for this contribution, it's much appreciated!

w-shackleton:
I think this is ready for merge - do you want me to do the commit squashing?

I've got some more changes stacked on top of this (actually adding payment channel proto v2) which I'll put up for PR once this is in.

On a side note, would you prefer those changes in one big PR or multiple small ones? The intermediate stages are a bit messy but I could clean them up for incremental review.

schildbach:
IMHO one commit should be one self-contained task. This could be adding a feature. Or fixing a bug (preferably in combination with a unit test showing the bug). Or be a refactoring.

Examples for things I'd squash:
* Adding a class in one commit and in a later commit rename it.
* Adding a feature and in a later commit add a unit test for the feature.
* Document something you did in an earlier commit.

If I look at your above PR, I'd perhaps split it into two: One commit implements the new feature (CLTV payment channel) and another one adds support for it in WalletTool.

Sometims I implement a feature and while implementing it I discover it would be good to refactor first. Then I do the refactoring as one commit, and the feature as a second.

I hope this makes things clearer. It's only nice-to-have, rather than being a blocker for merging.

schildbach:
@w-shackleton Do you want to do the squashing? Or would you like me to do it?

w-shackleton:
Ah sorry, I'll do it.

w-shackleton:
Right, done that. I've got other changes stacked on this that I'll squash as well before putting up a PR.

schildbach:
Looks great! Yes, the commits 3 and 4 look like you're just changing what you did in commit 1 and 2. Do you plan to squash that before merging?

w-shackleton:
Squashed. Sorry, I wasn't really familiar with the preferred commit semantics.

schildbach:
Well, still after the squash in commit 2 you were renaming of methods introduced in commit 1. So I just squashed both commits together and merged. Thanks again!

PR #1130, opened 2015-12-02
Coin: Remove chain-specific coin limit. Add some checks for arithmeti…

schildbach:
…c over/underflows instead.


schildbach:
Merged.

PR #1131, opened 2015-12-05
When testing for specific network, use network ID.

schildbach:


schildbach:
Merged, fixing https://github.com/schildbach/bitcoin-wallet/issues/159.

PR #1132, opened 2015-12-06
Move org.bitcoinj.testing package to test classpath.

schildbach:


schildbach:
Merged.

analytically:
This is breaking behaviour as this package is currently not available on Maven while BitcoinJ still is.

schildbach:
What is the specific problem?

analytically:
By moving these classes they now aren't packaged in the jar, while this was the case for 0.13. There is no bitcoinj-test artifact AFAICS

schildbach:
Yes, that's intended. Why is this a problem? It's my understanding of the classes to be used in bitcoinj's tests only.

analytically:
Ha, I have code that depended on org.bitcoinj.testing.FakeTxBuilder ...

schildbach:
Ok, I see. We could move that back (but perhaps into the .util package) if there is a usecase besides tests for that class.

lighthouseCash:
Hi Andreas,

Could you please move, these classes to util package, lighthouse is using them heavily.

PR #1134, opened 2015-12-08
Constructors should only call non-overridable methods

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S1699 - (Constructors should only call non-overridable methods) Calling an overridable method from a constructor could result in failures or strange behaviors when instantiating a subclass which overrides the method.

You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S1699

Please let me know if you have any questions.

Kirill.


schildbach:
After investigating a bit, I agree we should avoid this. I hope we don't break external code. Merged, thanks!

PR #1135, opened 2015-12-08
Strings literals should be placed on the left side when checking for equality

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S1132 - Strings literals should be placed on the left side when checking for equality. 

You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S1132

Please let me know if you have any questions.
Kirill.


schildbach:
We could do this, although I'm not sure if it makes any differences. Those strings should never be null.

schildbach:
I merged this anyway. I just don't agree with the "rule" stated in the commit message.

PR #1140, opened 2015-12-18
Print block hash in debug log message for unconnectable orphan.

jlopp:
Ran into this unhelpful message while testing.


schildbach:
Thanks, merged.

PR #1141, opened 2015-12-22
Support generating URIs for non-Bitcoin networks

rnicoll:
Adds in a new convertToBitcoinURI() method which takes in network parameters to determine the scheme from, for altcoin URI support.

### DIFF
--- a/core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
+++ b/core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
@@ -129,7 +129,6 @@ public class BitcoinURI {
      */
     public BitcoinURI(@Nullable NetworkParameters params, String input) throws BitcoinURIParseException {
         checkNotNull(input);
-        log.debug("Attempting to parse '{}' for {}", input, params == null ? "any" : params.getId());
 
         String scheme = null == params
             ? AbstractBitcoinNetParams.BITCOIN_SCHEME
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
```diff
@@ -360,13 +366,33 @@ public static String convertToBitcoinURI(Address address, Coin amount, String la
...
+     *
+     * @param params The network parameters that determine which network the URI
+     * is for, or null to use the default (Bitcoin).
+     * @param address The Bitcoin address
+     * @param amount The amount
+     * @param label A label
+     * @param message A message
+     * @return A String containing the Bitcoin URI
+     */
+    public static String convertToBitcoinURI(@Nullable NetworkParameters params,
```
schildbach: I'd prefer a non-nullable parameter here. Is it possible?

schildbach:
Generally I'm fine with the change.

rnicoll:
Changed it to require network parameters are passed to convertToBitcoinURI. Can you check that, please, and if it's okay I'll squash the commits.

schildbach:
I just realized that we already pass an Address, and that address already has the NetworkParameters. We could use this for now, without changing the method signature at all.

rnicoll:
Ah, had seen some bits of Address don't require parameters (i.e. https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/core/Address.java#L109 ), but looks like the constructor does. I'll clean this up, then raise a PR to make nullability of parameters going into an address consistent

rnicoll:
Oh, I see, Address always has parameters, just sometimes it looks them up. Cleaned, rebased and squashed.

schildbach:
I merged your PR with some cosmetic changes. Thanks!

PR #1142, opened 2015-12-30
Extract fee calculation from Wallet

rnicoll:
Extract the core fee calculation code from the wallet, so that it can be shared by other code (i.e. where transactions are built outside the wallet) and can be replaced (i.e. for altcoins).

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/core/Wallet.java
@@ -3535,6 +3535,11 @@ public class Wallet extends BaseTaggableObject
         public Coin feePerKb = DEFAULT_FEE_PER_KB;
 
         /**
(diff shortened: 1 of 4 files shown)


schildbach:
I have not forgotten your PR. Currently I'm working on fee calculation in order to get it on par with Core 0.12 and I'm sure this needs to be rebased afterwards. See #1219 and #1215 if interested, and more is coming.

I wonder what else is part of fee calculation and should be extracted to your new class. What about the minimum fee check (REFERENCE_DEFAULT_MIN_TX_FEE)? What about fee rate estimation (in future, we don't have any code for it yet)?

Do altcoins have different requirements? Do you have an example?

schildbach:
@jdojff Can you explain that need? Maybe you can write a unit test that shows how to use the interface in an alternative way?

But anyway, first this PR would need to be rebased.

PR #1143, opened 2015-12-30
Oracle Block Store

giles-saunders:
Ported from the MySQLFullPrunedBlockStore into the new OracleFullPrunedBlockStore.

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/OracleFullPrunedBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/OracleFullPrunedBlockStore.java
@@ -1,0 +1,137 @@
+package org.bitcoinj.store;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.bitcoinj.core.NetworkParameters;
+import org.bitcoinj.store.BlockStoreException;
+import org.bitcoinj.store.DatabaseFullPrunedBlockStore;
+
+public class OracleFullPrunedBlockStore extends DatabaseFullPrunedBlockStore {
+    private static final String DATABASE_DRIVER_CLASS = "oracle.jdbc.OracleDriver";
+    private static final String DATABASE_CONNECTION_URL_PREFIX = "jdbc:oracle:thin:@";
+    
+    // create table SQL
+    private static final String CREATE_SETTINGS_TABLE = "CREATE TABLE settings (\n" +
+            "    name VARCHAR2(32) NOT NULL,\n" +
+            "    value BLOB,\n" +
+            "    CONSTRAINT setting_pk PRIMARY KEY (name) \n" +
+            ")\n";
+
+    private static final String CREATE_HEADERS_TABLE = "CREATE TABLE headers (\n" +
+            "    hash RAW(28) NOT NULL,\n" +
+            "    chainwork RAW(12) NOT NULL,\n" +
+            "    height NUMBER(10,0) NOT NULL,\n" +
+            "    header RAW(80) NOT NULL,\n" +
+            "    wasundoable NUMBER(3,0) NOT NULL,\n" +
+            "    CONSTRAINT headers_pk PRIMARY KEY (hash) \n" +
+            ")\n";
+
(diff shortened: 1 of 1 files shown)


schildbach:
Thanks for your contribution! Generally it looks good. However I think we should follow the other database block stores and only use lowercase column names. If you look at the DatabaseFullPrunedBlockStore class, it's documented there.

giles-saunders:
Whilst I would normally try and keep to the existing project's conventions, I changed these to upper-case to avoid confusion and be consistent with Oracle. By default Oracle will create all columns with an upper-case name (even if it is defined using a lower-case identifier). It is possible to have columns with lower-case identifiers by enclosing the name in double-quotes, but this goes against the best practice for Oracle and extra care needs to be taken to ensure correct behaviour. An example where I disregarded Oracle's recommendation and kept with the project's existing convention was with the column "INDEX", which is a reserved word in Oracle and had to enclosed within double quotes to be acceptable.

schildbach:
If we want to go with the Oracle defaults, I think we should at least use uppercase column names in all the statements as well. Plus, there are camelcased column names in the select statements that are not consistent with the column names in the create statements.

Also, I think it should be noted in the DatabaseFullPrunedBlockStore comment that Oracle is an exception.

giles-saunders:
In Oracle case-sensitivity is only made when using quoted identifiers, unquoted identifiers are case insensitive (internally they are treated as upper-case), which is why those statements still work. I was attempting to change the minimum necessary to port this to Oracle, but if you feel the need for the case to be changed for clarity of the source, I will happily make those changes.

schildbach:
Yes, I think the case should match between statements and actual created column names.

giles-saunders:
I was looking at changing the case of the columns for the statements for Oracle, but it looks as though this will mean that the majority of DatabaseFullPrunedBlockStore would have to duplicated to OracleFullPrunedBlockStore. As unquoted identifiers are case-insensitive (technically Oracle upper-cases all the unquoted identifiers in both DML and DDL SQL statements) the existing statements in DatabaseFullPrunedBlockStore are perfectly valid and duplicating this code to upper-case them in OracleFullPrunedBlockStore seems unnecessary and would surely make maintenance going forward more arduous.

schildbach:
@kparmar1 What do you think about the case of column name for Oracle databases?

kparmar1:
Yes. Hey sry just seen this... I'll take a look today and get back to you.
K.

kparmar1:
My quick/initial thoughts....(out until this evening so a quick look over now)

1) I agree def no camel case in selects (update those). This was what was removed for consistency and compatibility, so there is no reason to reintroduce this. I mean it makes no difference here but why have it at all.
2) Maybe keep the name of the indexes the same. Is there a reason to change them? I know it's a little funny as you are looking at the other store impl - but consistent would be better unless there is a good reason not to do this?
3) Since Oracle doesn't care about case (well it upper-cases internally, unless you are explicit) I would stick with what's easiest/consistent (and make a doc note if required). The create/select lower case works fine as it is, and always should as we don't use mixed cases any longer (because of the issues it causes). I mean in Oracle you would write something like ...... SELECT column (lower case - even if internally this mean upper) FROM table (lower);
4) The issue of the index (reserved) word is handled just like with MySQL so that's fine (OK with the difference that it must be define in upper case, but that's not an issue as the lower cases (selects) are interpreted as upper in oracle- so any future selects/inserts created higher up will still work just fine). Make a doc note for this maybe. The thing is this column will always be specific to Oracle/MySQL so this column will always be handling differently. 

It's a little weird, but not really. All stores will consider all columns as lower case and in oracle it's opposite (internally). Since we are never going to mix cases this shouldn't ever cause an issue.

kparmar1:
These were just suggestions and not much to change really (hopefully simple). I hope I didn't confuse matters here.

schildbach:
@giles-saunders Do you think you could at least switch the camelcasing and use consistent index names?

kparmar1:
I can pick this up and finish it off? It's been a long while (I'll have time this weekend). I guess maybe it's in limbo at the moment (if I can remember anything :-).

schildbach:
Sure, feel free!

PR #1146, opened 2016-01-02
fix NPE in Wallet.calculateAllSpendCandidatesFromUTXOProvider

spicylobe:
to prevent the following which happens sometimes when receiving funds to a wallet associated with a UTXO provider such as the MySQLFullPrunedBlockStore
(pardon the out-of-date line numbers in this stacktrace)

java.lang.NullPointerException: null
    at org.bitcoinj.core.Wallet.calculateAllSpendCandidatesFromUTXOProvider(Wallet.java:3950) ~[classes/:na]
    at org.bitcoinj.core.Wallet.calculateAllSpendCandidates(Wallet.java:3892) ~[classes/:na]
    at org.bitcoinj.core.Wallet.calculateAllSpendCandidates(Wallet.java:3867) ~[classes/:na]

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/core/Wallet.java
@@ -4153,8 +4153,9 @@ public class Wallet extends BaseTaggableObject
         for (Transaction tx : pending.values()) {
             // Remove the spent outputs.
             for (TransactionInput input : tx.getInputs()) {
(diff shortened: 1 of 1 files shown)


schildbach:
That piece of code was written by @kparmar1 in commit 96a82800fd3b116e96e92f50b72e46998eb3d5cf. Kalpesh, do you think this is the right fix?

Issue #1121, opened 2015-11-24
Not receiving transactions for orphaned blocks that become best chain blocks


jlopp:
I reported this via email a few months ago but haven't made any progress on it. I'm currently running a service on 0.12.2 but when I upgrade to 0.13.3 my regression tests get stuck due to missing transactions. I've built a stripped down project that you can [download here](https://www.dropbox.com/s/f78b3kgi6iv04rg/bitcoinJMissingTxs.zip?dl=0) to reproduce this problem along with my regression test suite.

Upon running my regression test suite, after a few hundred blocks it starts to generate a high rate of blockchain forks, at which time you'll see errors like so:

```
11-24 13:53:32.881 [bitcoinj user thread] INFO  com.bitcoin.Indexer - notifyNewBestBlock 279: 4270db9bcff226b7cddeaa55873336f3a96cb27e838e2a6581870f5cdd2ab285
11-24 13:53:32.881 [bitcoinj user thread] ERROR com.bitcoin.Indexer - ALERT: 0 transactions received for block 4270db9bcff226b7cddeaa55873336f3a96cb27e838e2a6581870f5cdd2ab285
11-24 13:53:32.881 [bitcoinj user thread] INFO  com.bitcoin.Indexer - notifyNewBestBlock 280: 617114281282f97464f9c809f15d15ed8edd6ef087dbec8881792f711f679982
11-24 13:53:32.881 [bitcoinj user thread] ERROR com.bitcoin.Indexer - ALERT: 0 transactions received for block 617114281282f97464f9c809f15d15ed8edd6ef087dbec8881792f711f679982
11-24 13:53:32.881 [bitcoinj user thread] INFO  com.bitcoin.Indexer - notifyNewBestBlock 281: 13e9deb4e8f3bf50f4428c01c0fc36bf88bf7449451e6a128eef8b976c790275
11-24 13:53:32.881 [bitcoinj user thread] ERROR com.bitcoin.Indexer - ALERT: 0 transactions received for block 13e9deb4e8f3bf50f4428c01c0fc36bf88bf7449451e6a128eef8b976c790275
```
Grepping for the log entries related to a single one of the blocks that is missing transactions leads to output like so:
```
[...]

PR #1149, opened 2016-01-05
Don't throw away orphan block transactions before they are sent to listeners.

jlopp:
This is my proposed fix for https://github.com/bitcoinj/bitcoinj/issues/1121

I've checked that all unit tests pass and that my regression tests pass, both when my regression test suite continuously disconnects and reconnects nodes and when it keeps them all connected.

I'm sure there are other ways to fix the problem, though it's not clear to me why you'd ever want to throw away the transactions from the orphan block, so this seems like a simple solution.


schildbach:
@TheBlueMatt is the one who introduced this code in commit bd5c4e73e7cc05c0b072bb5bf51a442a06c30c11. Seems as if it was a performance optimization. Matt, do you think you could comment on this PR?

schildbach:
Merged. Correctness is more important than performance. Thanks Jameson!

Issue #924, opened 2015-06-11
Hammering on DNS seed


mikehearn:
Originally reported on Google Code with ID 558
```
Bitcoinj (master) still hammers on the DNS seed if not available. See attached log.

```

Reported by `andreas.schildbach` on 2014-05-15 09:57:29

<hr>
 * *Attachment: [logcat.txt](https://storage.googleapis.com/google-code-attachments/bitcoinj/issue-558/comment-0/logcat.txt)*

schildbach:
Likely fixed by cfb942a7fa99a017aefb0eef023ae0573e3e6210.

PR #1151, opened 2016-01-14
Rate limit peer discovery

rnicoll:
This changes PeerGroup to consider a peer discovery that doesn't return enough results as a failure, so it does exponential backoff. Otherwise it discovers constantly on a network with less than the maximum peers available.

Also added a second sanity check to ensure it's never done more than once a second.

Possibly related to #924 ?


schildbach:
Thanks! That original code was written by @mikehearn in commit 27bc229faba80b304711fe7b5e126844c58504cc. I'm not sure if Mike wan't to comment, but I thought I'd ping him in any case.

schildbach:
Merged.

PR #1153, opened 2016-01-16
TransactionInput: Add public clearScriptBytes() method, as this seems…

schildbach:
… to be a relevant use case.

It also saves some calls to the semi-private setScriptBytes().


rnicoll:
utACK

schildbach:
Merged.

PR #1154, opened 2016-01-16
Set connection channel in PaymentChannelServerListener

w-shackleton:
This method was never actually called with any argument but `null` - this initialises it with the correct value.


schildbach:
I wonder why this didn't show up in the tests?

w-shackleton:
Hmm - it only gets used when `closeConnection()` is called on the event handler. I'll see if there are tests of that.

schildbach:
Never mind, I just merged this. Thanks for catching!

PR #1157, opened 2016-01-16
Wallet: Make transaction pools private again, …

schildbach:
…but add a poolContainsTxHash() method that is visible for testing.


schildbach:
Merged.

PR #1159, opened 2016-01-16
TransactionTest: Move creation of fake transactions…

schildbach:
…over to FakeTxBuilder. Also, a couple of cleanups.


schildbach:
Merged.

PR #1163, opened 2016-01-17
Update to okhttp 2.7.2

schildbach:
This is the latest release in the 2.x series. Unlike 3.x, it doesn't break any
API. It has lots of security and Android-related fixes which could affect us.

Also see: https://github.com/square/okhttp/blob/master/CHANGELOG.md


schildbach:
Merged.

PR #1165, opened 2016-01-17
ScriptChunk: JavaDoc for public fields.

schildbach:


schildbach:
@troggy Am I correct with these descriptions? There seems to be confusion about OP_0, so I'd like to clarify.

troggy:
I'm not sure either. Afaik OP_0 meant to be anything just to overcome off-by-one error of CHECKMULTISIG.

schildbach:
The CHECKMULTISIG bug pops one item too much from the stack. Initially it ignored the value of the item, later imho it was changed to require the empty vector. OP_0 pushes the empty vector, so that's in-line with the JavaDoc.

PR #1166, opened 2016-01-18
Implemented version 2 payment channels using CLTV micropayments

w-shackleton:
Right, it's a big one.

Key changes:
* `PaymentChannelServerState` is now `PaymentChannelV1ServerState extends PaymentChannelServerState` and `PaymentChannelV2ServerState extends PaymentChannelServerState` (and the same for `PaymentChannelClientState`)
* Protocol V2 uses P2SH CLTV micropayments, since the created script would otherwise be considered 'non-standard' by the Satoshi client.
* There is now a `StateMachine` class for more easily tracking which state transitions are allowed.
* All the tests are parameterised for version 1 and version 2
* The protocol can negotiate version 1 and 2 to some extent; `VersionSelector` can be used to indicate which versions should be allowed by default.

@schildbach is this squashed enough? I can squash further if need be. There are a few changes later on that were very hard to move around in the commit history.


core/src/main/java/org/bitcoinj/protocols/channels/StoredServerChannel.java
```diff
@@ -105,13 +123,14 @@ public synchronized PaymentChannelServerState getOrCreateState(Wallet wallet, Tr
     @Override
     public synchronized String toString() {
         final String newline = String.format(Locale.US, "%n");
-        return String.format(Locale.US, "Stored server channel (%s)%n" +
```
schildbach: If you use String.format, can you always specify a locale?

schildbach:
I'm a bit concerned about backwards compatibility for the old channels, since a few classes were renamed and method signatures changed.

schildbach:
Travis chokes because the 4 MB log limit is exceeded.

core/src/main/java/org/bitcoinj/protocols/channels/StoredServerChannel.java
w-shackleton: :+1:

w-shackleton:
I tried to keep the backwards compatibility - the only breaking change could be `PaymentChannel*State -> PaymentChannelV{x}*State`.
We could solve this by naming the V1 classes `PaymentChannel*State` and having an explicit V2 State, and renaming the common subclass?

Re. Travis, is it possible to have junit only output debug logs for tests that fail?

schildbach:
I just pushed a commit that disables log output for the unit tests if run via Maven. Can you rebase this PR on current master?

w-shackleton:
Updated with rebase and fixed `Locale.US`, but Travis hasn't run.

schildbach:
I restarted the build manually and it passed. Not sure why the connection to Travis is broken on this PR.

core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelV2ServerState.java
```diff
@@ -0,0 +1,221 @@
...
+        Futures.addCallback(future, new FutureCallback<Transaction>() {
+            @Override public void onSuccess(Transaction transaction) {
+                log.info("TX {} propagated, channel successfully closed.", transaction.getHash());
+                stateMachine.transition(State.CLOSED);
+                closedFuture.set(transaction);
+            }
+
+            @Override public void onFailure(Throwable throwable) {
+                log.error("Failed to settle channel, could not broadcast: {}", throwable.toString());
+                throwable.printStackTrace();
```
schildbach: Instead of printing exceptions to the console, can we log them to the log file? In this instance I suggest removing the `.printStackTrace()` line and use the `log.error()` (2 lines above) with a throwable as a parameter so that the stack trace is logged.

I implemented this in 593d92390b6b71edd7644f385ff65ff1ff8e690c for all occurances of `.printStackTrace()`.

schildbach:
Ok I think the one class rename is fine. We can mention it in the release notes.

Apart from the one in-line comment I made, I'm happy. Again I can't really review the functionality, I trust you and your unit tests (thanks for making the effort). If anyone else – proficient in payment channels – wants to comment, please do.

Will the feature be finished with this PR? I'd like to get it into a releasable state so we can enter the stabilization phase for bitcoinj 0.14 soonish.

About the squashing: Ideally it would have been 1 commit, implementing the entire new feature. I guess we missed that goal, since half of the feature is already merged (-:  So I tend to recommend squashing this PR into 1 commit. Except if you see a certain "sub-task" that makes sense on its own and could be merged separately. Example: if there would be a refactoring that prepares for your feature but also cleans up the code a but.

w-shackleton:
Fixed the log messages, squashed commits.
I think this feature is finished - I'll do some further testing over the weekend. I'd feel happier if someone else also had a look at this code.

schildbach:
Great, I'll merge this in a couple of days to allow for comments of others.

w-shackleton:
I've found a problem with the protobuf - it's not backwards compatible, meaning a wallet can't be upgraded at the moment. I'll set default values for all added fields.

schildbach:
Oh yes that's important. Thanks for catching this early!

w-shackleton:
I'm also having problems getting `TransactionBroadcaster.broadcastTransaction`'s future to ever return anything on testnet, even without any of my modifications (ie. on `master`). When should this future fire? I can't get it to fire in `ExamplePaymentChannelServer` and friends (waited 30 mins) and hence can't test channels at the moment. I'll do some more investigation.

schildbach:
Broadcasting a transaction tries to estabilish a good number of connections first before it actually broadcasts to half of the peers, hearing back from the other half. I assume the needed number of connections is never made in your case? Are you connecting to maybe only one local peer?

w-shackleton:
I had 13 connections - I debugged it and watched it broadcast to all of the peers selected, but the future never completed. When I use `WalletTool` to broadcast the future returns within a few seconds, but similar code running in `ExamplePaymentChannelServer` doesn't. My first guess is it's something to do with threading - I'll have an investigate.

w-shackleton:
The tx broadcasting issue is fixed in #1183 and I've added default version numbers to the stored channel extensions (I've tested upgrading a wallet from v1 to v2 and it works)

schildbach:
I merged this. Many thanks!

PR #1167, opened 2016-01-19
Make sure we don't use the deprecated listener classes or methods any…

schildbach:
… more.

At least I tried (-:  This PR isn't finished, I just put it up for discussing how to progress.

As you can see, we need to add a lot of empty methods. What's our take on a set of default implementations for the new interfaces, e.g. "SimpleWalletCoinEventListener"? Asking everyone, but pinging @rnicoll specifically.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/ChainSplitTest.java
+++ b/core/src/test/java/org/bitcoinj/core/ChainSplitTest.java
@@ -17,8 +17,8 @@
 
 package org.bitcoinj.core;
 
-import org.bitcoinj.core.listeners.AbstractWalletEventListener;
 import org.bitcoinj.core.TransactionConfidence.ConfidenceType;
+import org.bitcoinj.core.listeners.WalletCoinEventListener;
 import org.bitcoinj.params.UnitTestParams;
 import org.bitcoinj.store.MemoryBlockStore;
 import org.bitcoinj.testing.FakeTxBuilder;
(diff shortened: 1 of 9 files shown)


rnicoll:
I'd be inclined to leave this for now, and split the listener interfaces into fine grain (1-2 method each) in 1.5, and replace these calls then. Could even take the deprecation tags off the abstract listeners, I don't think we need to rush to remove them.

schildbach:
Hmm, it feels a bit weird to offer all these fine granular listeners and then not provide examples for how to use them (in an efficient way).

Also, if you think of splitting the interfaces even more we should discuss this for 0.14 soon. I'd prefer to not touch the listeners again so soon after 0.14.

rnicoll:
Sorry about the delay getting back to you - looking at further splitting it into single-method interfaces, which means they'll actually be compatible with lambda functions. Obviously this would give us an insane number of listeners to track in each class, so looking at adding a supporting class similar to https://docs.oracle.com/javase/7/docs/api/javax/swing/event/EventListenerList.html (but with the ability to track intended thread to execute on) to hold the listeners. Does that seem like a sensible next step?

schildbach:
Sounds reasonable. I take it reaching lambda function compatibility is the holy grail… I can't judge myself – will be stuck with Java 6 (Android) for some time to come. Can the EventListenerList be made typesafe? I wonder why they took this "generic" approach.

rnicoll:
That seemed to be the intent Mike had, and I didn't complete it first time because the patch was getting incredibly large already.

Having started trying to add EventListenerList, I'm reconsidering it, it makes very little actual difference to the classes that exist, and adds a complex and brittle new data store.

rnicoll:
Refactoring the peer data events down to single method interfaces is in progress now under https://github.com/rnicoll/bitcoinj/tree/listener-refactor-3 but won't be raised as a PR until at least next weekend. Mostly thought I'd let you know it's happening at least.

schildbach:
Closing this, Ross has already taken care of most (if not all) of it.

PR #1168, opened 2016-01-25
Be aware of opt-in full replace-by-fee.

schildbach:


core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -46,9 +48,7 @@
     // Magic outpoint index that indicates the input is in fact unconnected.
     private static final long UNCONNECTED = 0xFFFFFFFFL;
 
-    // Allows for altering transactions after they were broadcast. Tx replacement is currently disabled in the C++
-    // client so this is always the UINT_MAX.
-    // TODO: Document this in more detail and build features that use it.
+    // Allows for altering transactions after they were broadcast. Values below NO_SEQUENCE mean it can be altered.
```
jlopp: values below NO_SEQUENCE - 1, right?

jlopp:
utACK

core/src/main/java/org/bitcoinj/core/TransactionInput.java
schildbach: Yes, that's right. I was tying to describe the original meaning of the sequence. I'll amend.

Do you know why BIP 125 leaves this "gap" at NO_SEQUENCE-1?

jlopp: I believe that the idea is that:

nSequence == 0xffffffff disables nLockTime field values & finalizes a txn
nSequence == 0xffffffff - 1 explicitly disables RBF while still allowing nLockTime field values

schildbach: Makes sense. Thanks for the review!

PR #1171, opened 2016-01-25
Can serialize and deserialize transaction with witnesses

NicolasDorier:
First step for implementation of segwit, transaction serialization and deserialization

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -114,6 +114,7 @@ public class Transaction extends ChildMessage {
     private long version;
     private ArrayList<TransactionInput> inputs;
     private ArrayList<TransactionOutput> outputs;
+    private ArrayList<TransactionWitness> witnesses;
 
     private long lockTime;
 
@@ -189,6 +190,7 @@ public class Transaction extends ChildMessage {
         version = 1;
         inputs = new ArrayList<TransactionInput>();
         outputs = new ArrayList<TransactionOutput>();
+        witnesses = new ArrayList<TransactionWitness>();
         // We don't initialize appearsIn deliberately as it's only useful for transactions stored in the wallet.
         length = 8; // 8 for std fields
     }
@@ -239,7 +241,14 @@ public class Transaction extends ChildMessage {
     @Override
     public Sha256Hash getHash() {
         if (hash == null) {
-            hash = Sha256Hash.wrapReversed(Sha256Hash.hashTwice(unsafeBitcoinSerialize()));
+            ByteArrayOutputStream stream = new UnsafeByteArrayOutputStream(length < 32 ? 32 : length + 32);
+            try {
+                bitcoinSerializeToStream(stream, TransactionOptions.NONE);
+            } catch (IOException e) {
+                // Cannot happen, we are serializing to a memory stream.
+            }
+            byte[] bits = stream.toByteArray();
+            hash = Sha256Hash.wrapReversed(Sha256Hash.hashTwice(bits));
         }
         return hash;
     }
@@ -383,6 +392,47 @@ public class Transaction extends ChildMessage {
         }
 
         return totalOut;
+    }
+
+    /**
+     * Get the transaction witness of an input
+     * @return the witness of the input
+     * @throws java.lang.IndexOutOfBoundsException
+     */
+    public TransactionWitness getWitness(int inputIndex) {
+        if (!(0 <= inputIndex && inputIndex < inputs.size()))
+            throw new java.lang.IndexOutOfBoundsException();
+        if (inputIndex >= witnesses.size())
+            return TransactionWitness.getEmpty();
+        return witnesses.get(inputIndex);
+    }
+
+    /**
+     * Set the transaction witness of an input
+     * @throws java.lang.IndexOutOfBoundsException
+     */
+    public void setWitness(int inputIndex, TransactionWitness witness) {
+        if (!(0 <= inputIndex && inputIndex < inputs.size()))
+            throw new java.lang.IndexOutOfBoundsException();
+        witness = witness == null ? TransactionWitness.getEmpty() : witness;
+        while (inputIndex >= witnesses.size()) {
+            witnesses.add(TransactionWitness.getEmpty());
+        }
+        witnesses.set(inputIndex, witness);
+    }
+
+    /**
+     * Returns true if the transaction has witnesses
+     * @return true if the transaction has witnesses
+     */
+    public boolean hasWitness() {
+        for (int i = 0; i < inputs.size(); i++) {
+            if (i >= Math.min(inputs.size(), witnesses.size()))
+                break;
+            if (witnesses.get(i).getPushCount() != 0)
+                return true;
+        }
+        return false;
     }
 
     @Nullable private Coin cachedValue;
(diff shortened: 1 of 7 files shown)


NicolasDorier:
You can cross check transactions on https://blockchainprogramming.azurewebsites.net/checktx

There is also another problem which I bumped into with NBitcoin: It is impossible to deserialize which those 2 conditions : No inputs, One or More Output.

In practice this is not a big deal because transaction without input does not exist... But sometimes, during internal processing, you may serialize and deserialize for cloning your transaction for example. And I've encountered case in NBitcoin where it happened without input.

The way I dealt with it is by fitting a bit in the tx version which means "I really have no witness, don't read the next byte as flag". 

I have not implemented that here.

schildbach:
Thanks, that's great! I think this PR misses the code for protobuf-serialization of witnesses. So a wallet that saves to disk and back will miss the witnesses.

Also please watch the code style. Though I can fix that while merging.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -230,7 +231,13 @@ public Transaction(NetworkParameters params, byte[] payload, @Nullable Message p
     @Override
     public Sha256Hash getHash() {
         if (hash == null) {
-            byte[] bits = bitcoinSerialize();
+            ByteArrayOutputStream stream = new UnsafeByteArrayOutputStream(length < 32 ? 32 : length + 32);
+            try{
+            bitcoinSerializeToStream(stream, TransactionOptions.NONE);
```
schildbach: indentation

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -536,23 +543,62 @@ protected static int calcLength(byte[] buf, int offset) {
...
-        inputs = new ArrayList<TransactionInput>((int) numInputs);
-        for (long i = 0; i < numInputs; i++) {
-            TransactionInput input = new TransactionInput(params, this, payload, cursor, serializer);
-            inputs.add(input);
-            long scriptLen = readVarInt(TransactionOutPoint.MESSAGE_LENGTH);
-            optimalEncodingMessageSize += TransactionOutPoint.MESSAGE_LENGTH + VarInt.sizeOf(scriptLen) + scriptLen + 4;
-            cursor += scriptLen + 4;
+        readInputs();
+        byte flags = 0;
+        if(inputs.size() == 0 && witSupported) {
```
schildbach: space before '('

NicolasDorier:
I tried to be careful with code style did I missed something ?
Where is handled protobuf serialization ? (I think it is better in a separate PR though)

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -536,23 +543,62 @@ protected static int calcLength(byte[] buf, int offset) {
...
+            throw new ProtocolException("Unknown transaction optional data");
+        }
+
+        lockTime = readUint32();
+        optimalEncodingMessageSize += 4;
+        length = cursor - offset;
+    }
+
+    private void readWitness()
+    {
```
schildbach: '{' at end of line

core/src/main/java/org/bitcoinj/core/Transaction.java
NicolasDorier: ok I fix it

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -278,6 +278,33 @@ public void testCLTVPaymentChannelTransactionRefund() {
     }
 
     @Test
+    public void canParseTransaction() {
+        byte[] hex = null;
+        byte[] hex2 = null;
+        Transaction tx = null;
+
+        //Roundtrip without witness
```
schildbach: space between '//' and comment

NicolasDorier:
Fixing formatting, I did not knew how to do it on android studio :p

NicolasDorier:
@schildbach ok I reformatted, some other things has been catched up that were not properly formatted.

NicolasDorier:
@schildbach Actually, there is a second way of doing, which I did with NBitcoin.

In this PR, each transaction have "outputs", "inputs" and "witnesses" where witnesses is either null OR of exactly the size of inputs. It matches the serialization model on a bitcoin transaction. (witnesses are not serialized in the "TxInput").

In NBitcoin I initially took this approach, but found it bothersome to develop against. So I changed and added an accessor "Witness" into the "TxInput" class. This does not match the serialization model, however, it is conceptually more correct.

schildbach:
Protobuf serialization is done in the WalletProtobufSerializer class and the structure will be extended in wallet.proto. Don't forget to generate fresh bindings – see comment in wallet.proto. It's fine to do it in a separate commit. Separate PR feels weird though, at least it depends on your changes already done.

About the witnesses structure I have no preference right now. Usually it makes sense to look at how Bitcoin Core does things. Also, maybe it makes the decision easier if we anticipate there can be multiple witnesses in future.

If you use a formatter, please make sure to format only *your* code. Correcting already existing format problems should go into a different PR (don't bother about it now). It may sound a bit finicky, but keeping changes to the minimum helps with reviewing and backporting your code. If you're unsure with formatting, I can do it at merge time.

core/src/main/java/org/bitcoinj/core/TransactionWitness.java
```diff
@@ -0,0 +1,22 @@
+package org.bitcoinj.core;
+
+public class TransactionWitness {
+    public TransactionWitness(int pushCount)
+    {
```
schildbach: '{' to the EOL please

core/src/main/java/org/bitcoinj/core/TransactionWitness.java
```diff
@@ -0,0 +1,22 @@
...
+        pushes = new byte[pushCount][];
+    }
+    byte[][] pushes;
+
+    public byte[] getPush(int i)
+    {
+        return pushes[i];
+    }
+    public int getPushCount()
```
schildbach: one empty line of separation between methods

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -278,6 +278,33 @@ public void testCLTVPaymentChannelTransactionRefund() {
...
+        tx = new Transaction(NetworkParameters.fromID(NetworkParameters.ID_MAINNET), hex);
+        assertEquals(3, tx.getOutputs().size());
+        assertEquals(3, tx.getInputs().size());
+        assertNull(tx.getWitnesses());
+        hex2 = tx.bitcoinSerialize();
+        assertArrayEquals(hex, hex2);
+        assertEquals("Uncorrect hash","38d4cfeb57d6685753b7a3b3534c3cb576c34ca7344cd4582f9613ebf0c2b02a", tx.getHash().toString());
```
schildbach: space after comma

NicolasDorier:
Ok now it is perfectly formatted, I'll be more careful next commit. 
I am wondering if serialization of the witness is really useful for the wallet. The advantage of segwit is precisely that you don't need to store the witness part. Do you prefer I do it still ?

There is other things to do for segwit, one of which is the new hash function for signature, do I do that in the same PR as well ?

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -278,6 +278,33 @@ public void testCLTVPaymentChannelTransactionRefund() {
     }
 
     @Test
+    public void canParseTransaction() {
```
schildbach: Maybe rename this test to something more meaningful, e.g. `witnessTransaction()`?

schildbach:
The formatting looks very good now. Thanks for making the effort.

You're right about witnesses not needed by the wallet (unless there will be future witnesses that are useful), so yes let's skip protobuf persistence for now.

Yes if you have further work for segwit feel free to put a commit on top. Ideally we'd work towards a usable feature (like those I pictured in my "rough roadmap" posted to the mailing list) and if one is finished, and the BIPs are finished/committed to by Bitcoin Core, we merge.

NicolasDorier:
Can you advise also what you would prefer: Currently if you serialize then deserialize a transaction with no input BUT with output, it will failed, due to how segwit format has been designed.
In practice this is not a problem, a transaction without input does not really exist. But in some internal processing with NBitcoin I needed to clone the transaction, which I did by serializing/deserializing, and ran into problems. Also my tests broke because I often used input less transaction for expediency.

My fix to that was to put a flag in the version during serialization if those 3 conditions were true : Segwit supported, Input size is 0, Outputs size is above 0.
Do you want I do that here ?

schildbach:
I prefer to do without that flag workaround. And you said this PR comes without – still it passes all the unit tests so I guess it's no problem for bitcoinj.

NicolasDorier:
New hashSignature scheme implemented, I borrowed from @GreenAddress   https://github.com/greenaddress/bitcoinj/commit/891e349410599f743124e62da3af53e6a3c6e9b5

NicolasDorier:
@schildbach I think I finished all is needed for basic segwit support. To my mind there is still some nice to have method to make, like a way to go back and forth between Script and TransactionWitness class. (A TransactionWitness is basically a Script with only pushes encoded more efficiently)

Let me know if you would like something more or if you'd like any fix. (or if you have any question on segwit)

NicolasDorier:
@schildbach it can be useful to merge that before it get merged to core so people can start testing on segnet. Let me know if you want I add something in this PR.

schildbach:
I'd rather wait with merging to master until the spec is committed (will most likely not change any more).

I can merge it to a branch on our repo, but then again this is only useful if we have multiple developers working on segregated witness.

NicolasDorier:
Ok no problem, I'll squash the commits as you said, and I will also add the segnet params. (I forgot about that)

NicolasDorier:
Ok @schildbach segnet is now pushed and the commit you mentioned squashed. Ping @GreenAddress I know you need that for your tests.

NicolasDorier:
Added ref to the BIP, as well as the BIP's sample in tests

greenaddress:
We tested on segnet3 this [patch](https://github.com/greenaddress/bitcoinj/commit/891e349410599f743124e62da3af53e6a3c6e9b5) as our code doesn't use the rest.

what is the plan for this PR? I believe the spec is not changing anymore.

@NicolasDorier it probably needs rebase (and maybe squash)  :)

schildbach:
The plan is to let SegWit mature (as a whole, so also the work on Core). As soon as Core SegWit merged, I'd merge all parts of this PR that form a viable usecase.

Until then, if multiple people start contributing code to this PR I can merge it to a feature branch.

greenaddress:
sounds good, thanks @schildbach!

NicolasDorier:
rebased

matsjj:
Thanks a lot for starting to work on the heavy lifting required to utilise SegWit!

I started using this branch for (thunder)[https://github.com/blockchain/thunder/commit/e9132c9f30719f430385e88cc1c04d7611b5dfd0], and added a few utility methods in my code that probably rather belong over here (like constructing a TransactionWitness object out of a P2SH output and input). 

I think a feature branch makes sense at this point

schildbach:
Feature branch done: https://github.com/bitcoinj/bitcoinj/tree/segwit
I'm closing this PR as future work on segwit should go on new PRs against the segwit branch.

@matsjj Great to know you're working on thunder/lightning!

btcdrak:
A note abotu segwit, the witness program size was extended to 40 bytes in BIP141 c.f. https://github.com/bitcoin/bips/commit/d1b52cb198066d4e515e8a50fc3928c5397c3d9b

PR #1172, opened 2016-01-26
Transaction: modify hashForSignature() to be thread-safe.

jlopp:
I've been using hashForSignature more often recently to generate normalized hashes and I ran into a multithreading issue. I noticed that different threads that were operating on the same transaction would suddenly think that the regular hash for it had changed.

I think this is a fairly straightforward fix to create a new transaction object and then perform all of the operations upon it instead of upon the original transaction object that may be referenced by other threads.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -872,7 +872,7 @@ public class Transaction extends ChildMessage {
      * @param anyoneCanPay Signing mode, see the SigHash enum for documentation.
      * @return A newly calculated signature object that wraps the r, s and sighash components.
      */
-    public synchronized TransactionSignature calculateSignature(int inputIndex, ECKey key,
+    public TransactionSignature calculateSignature(int inputIndex, ECKey key,
                                                                 byte[] redeemScript,
                                                                 SigHash hashType, boolean anyoneCanPay) {
         Sha256Hash hash = hashForSignature(inputIndex, redeemScript, hashType, anyoneCanPay);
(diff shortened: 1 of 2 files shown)


schildbach:
Hmm, the `hashForSignature()` methods are synchronized, so the race condition shouldn't happen so easily. But I agree your solution feels more safe. And duplicating the transaction shouldn't add much overhead, compared to the relatively expensive signing operation, right?

The original code has been written by @mikehearn and @TheBlueMatt , maybe they'd like to comment?

jlopp:
I think the synchronization gives us a false sense of safety here - sure, it will prevent other threads from calling the hashForSignature() method or other synchonized methods at the same time, but that's not the fundamental problem. The problem is that you can still call other nonsynchonized methods such as getHash() that will serialize the (current) transaction data and hash it. If we set the synchronized keyword on all of the other accessor functions that could be affected by hashForSignature, that might also fix this issue... but I'd still argue that my approach here is more future-proof against it being accidentally reintroduced when new accessor methods are written.

schildbach:
I agree. Can we perhaps remove the synchronized keywords, given that the methods should now be inherently thread safe?

Your PR contains a merge commit – can you remove that and squash the rest together?

jlopp:
Removed synchronization and squashed commits.

jlopp:
Not sure why Travis failed; it looks like it fails for a number of PRs... does the code look acceptable?

schildbach:
Your code is fine, and I plan to merge it. Operating on a temporary clone of the tx is certainly safer than modifying it forth and back. However, the whole Message hierarchy (with Transaction being a subclass) seems to not be thread safe. With or without the "synchonized" we removed, the code can stil race inside `bitcoinSerialize()`. I'm not sure what the original idea is. Probably synchronization is the duty of the caller?

schildbach:
Btw. the failing unit test is consequence of this PR, with or without synchronized. Is there anything else you changed between now and a week ago?

jlopp:
You're right - it looks like the bug was that I was still calling bitcoinSerialize() on the original object rather than the cloned object. While I was taking a look at the code I also found some more code that I was able to remove because it is no longer necessary.

schildbach:
Darn, and I just reviewed the changes for exactly this kind of mistake and didn't notice…

I think we should change the commit description away from hashForSignature being thread safe... it isn't as far as I understand. What about: "Transaction: Build hash for signature from a clone, rather than tweak the original."?

I this sense I also wonder if we should merge the unit test, because it tests for this property that doesn't exist currently.

jlopp:
Well I suppose it depends upon your perspective of thread safety. My goal with this PR is that now you can have any number of threads calling hashForSignature() on the same Transaction object and be assured that the properties of the object itself will not be modified. As you noted, the Transaction / Message classes are not completely thread safe because other functions can modify an object's properties across threads, but with this change at least the hashForSignature() function will be thread safe.

schildbach:
Fair enough. I merged your PR, many thanks!

PR #1174, opened 2016-02-02
Split wallet events into single method interfaces

rnicoll:
Provides per-event methods for wallet listeners, so they're then compatible with lambda functions, and updates all call to use these methods.

Still need to look at whether network events can be split up, but wanted to raise this as its own PR.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/listeners/ScriptsChangeEventListener.java
+++ b/core/src/main/java/org/bitcoinj/core/listeners/ScriptsChangeEventListener.java
@@ -16,16 +16,20 @@
 
 package org.bitcoinj.core.listeners;
 
-import org.bitcoinj.wallet.KeyChainEventListener;
+import org.bitcoinj.core.Wallet;
+import org.bitcoinj.script.Script;
+
+import java.util.List;
 
 /**
- * <p>Common interface for wallet changes and transactions.</p>
- * @deprecated Use the superinterfaces directly instead.
+ * <p>Implementors are called when the contents of the wallet changes, for instance due to receiving/sending money
+ * or a block chain re-organize. It may be convenient to derive from {@link AbstractWalletEventListener} instead.</p>
  */
-@Deprecated
-public interface WalletEventListener extends 
-        KeyChainEventListener, WalletChangeEventListener,
-        WalletCoinsReceivedEventListener, WalletCoinsSentEventListener,
-        WalletReorganizeEventListener, ScriptsChangeEventListener,
-        TransactionConfidenceEventListener {
+public interface ScriptsChangeEventListener {
+    /**
+     * Called whenever a new watched script is added to the wallet.
+     *
+     * @param isAddingScripts will be true if added scripts, false if removed scripts.
+     */
+    void onScriptsChanged(Wallet wallet, List<Script> scripts, boolean isAddingScripts);
 }
--- a/core/src/main/java/org/bitcoinj/jni/NativeKeyChainEventListener.java
+++ b/core/src/main/java/org/bitcoinj/jni/NativeKeyChainEventListener.java
@@ -14,18 +14,21 @@
  * limitations under the License.
  */
 
-package org.bitcoinj.core.listeners;
+package org.bitcoinj.jni;
 
 import org.bitcoinj.wallet.KeyChainEventListener;
+import org.bitcoinj.core.ECKey;
+
+import java.util.List;
 
 /**
- * <p>Common interface for wallet changes and transactions.</p>
- * @deprecated Use the superinterfaces directly instead.
+ * An event listener that relays events to a native C++ object. A pointer to that object is stored in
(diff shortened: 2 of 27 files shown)


core/src/main/java/org/bitcoinj/core/listeners/ScriptsChangeEventListener.java
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright 2013 Google Inc.
```
schildbach: Why copyright Google and why 2013?

core/src/main/java/org/bitcoinj/jni/NativeKeyChainEventListener.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.List;
+
+/**
+ * An event listener that relays events to a native C++ object. A pointer to that object is stored in
+ * this class using JNI on the native side, thus several instances of this can point to different actual
+ * native implementations.
+ */
+public class NativeKeyChainEventListener implements KeyChainEventListener {
+    public long ptr;
```
schildbach: What are these ptr fields good for?

schildbach:
So far looks good to me (-:

core/src/main/java/org/bitcoinj/core/listeners/ScriptsChangeEventListener.java
rnicoll: Copy & paste direct from the original interface (WalletEventListener.java). I can add a new copyright notice if you think it needs it, but given it doesn't make any real change apart from splitting it down, figured best to leave as-is.

core/src/main/java/org/bitcoinj/jni/NativeKeyChainEventListener.java
rnicoll: Honestly not a clue, again copy & paste from the original (NativeWalletEventListener.java). Want me to take out the new JNDI interfaces until I have an answer one way or another?

rnicoll:
In the end reconsidered what to do with AbstractWalletEventListener, and it's now deprecated again, and I've replaced everywhere it's used

core/src/main/java/org/bitcoinj/core/listeners/ScriptsChangeEventListener.java
schildbach: Ok.

core/src/main/java/org/bitcoinj/jni/NativeKeyChainEventListener.java
schildbach: No, it's ok. For reference, the ptr fields were introduced in commit 87ad8f2e5b59ed457ed187fece00922c7e76844e.

schildbach:
Travis dies because of resource exhaust – most likely memory.

rnicoll:
Can you restart the Travis job, check it's temporary? Did run all the tests here before submitting, but you never know...

schildbach:
I'm on the problem.

schildbach:
Hrmpf, both builds (Oracle JDK and OpenJDK) exhibit the problem. I tried switching Travis to the [trusty beta VM](https://docs.travis-ci.com/user/trusty-ci-environment/) which would double the amount of memory from 4 GB to 7.5 GB, but it runs into [this Travis issue](https://github.com/travis-ci/travis-ci/issues/5227). At the moment I'm out of ideas.

schildbach:
Fixed the Travis issue by [limiting heap size of unit tests to 1 GB](af769fe708a654988898e3fa61aee5b0cef80d25). Can you rebase your PR on current master?

rnicoll:
Done, running now

schildbach:
I just verified this PR by porting Bitcoin Wallet to the changes and it works nicely. One last review comment: I remember from PR #1034 that the goal is to put executors first in the method signatures of the various add*Listener methods. Is there a reason you put the last, especially for the new methods where backwards compatibility is of no concern?

core/src/test/java/org/bitcoinj/core/ChainSplitTest.java
```diff
@@ -558,7 +562,7 @@ public void coinbaseDeath() throws Exception {
         // transactions would be. Also check that a dead coinbase on a sidechain is resurrected if the sidechain
         // becomes the best chain once more. Finally, check that dependent transactions are killed recursively.
         final ArrayList<Transaction> txns = new ArrayList<Transaction>(3);
-        wallet.addEventListener(new AbstractWalletEventListener() {
+        wallet. addCoinsReceivedEventListener(new WalletCoinsReceivedEventListener() {
```
schildbach: Stray space (-:

rnicoll:
That's me cutting & pasting again. Patched and pushed.

schildbach:
This looks really great! One last check I did was compile my app with just the package moves, so yes we're quite backwards compatible.

I just merged.

PR #1175, opened 2016-02-02
Disable max money check on networks with no supply cap

rnicoll:
More altcoin support stuff

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -1239,8 +1239,7 @@ public class Transaction extends ChildMessage {
                 if (output.getValue().signum() < 0)    // getValue() can throw IllegalStateException
                     throw new VerificationException.NegativeValueOutput();
                 valueOut = valueOut.add(output.getValue());
-                // Duplicate the MAX_MONEY check from Coin.add() in case someone accidentally removes it.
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1256,7 +1256,7 @@ public void verify() throws VerificationException {
                     throw new VerificationException.NegativeValueOutput();
                 valueOut = valueOut.add(output.getValue());
                 // Duplicate the MAX_MONEY check from Coin.add() in case someone accidentally removes it.
```
schildbach: This comment can go, too.

schildbach:
There is some more of these checks in the code, maybe you want to tackle more of them at the same time?

rnicoll:
I've now replaced everything except the case where it's used to send all from a wallet. I'll do that one too, later, but for now I'd rather give this time to settle first.

schildbach:
I just discovered that some of the checks you're replacing in this PR are actually off-by-one.

Assuming that my fixes in PR 1177 are correct, do you think you could rebase this PR on PR 1177 and fix the checks in your replacement too?

rnicoll:
#1177 looks good, and I've rebased this now

schildbach:
Merged, thanks!

PR #1177, opened 2016-02-05
Fix a few max coins checks.

schildbach:
This is largely esoteric, but was brought up by https://github.com/schildbach/bitcoin-wallet/issues/299.


schildbach:
Merged.

PR #1178, opened 2016-02-05
Another max coins check, involving a series of transactions

schildbach:
Transactions that would cause estimated balance to exceed the maximum allowed are considered risky. There is a risk that the culprit is actually a previous transaction and we're punishing the wrong one, but this situation will clear up as transactions are confirmed. Most important is that the balance will not rise too high.


schildbach:
I decided to not merge this for now. Rather, 44ca39bfd7df0068a7aedb1780a8ad954e1cd50c removes the max coins check in Coin. So balances can (theoretically) rise beyond max coins temporarly. It will be sorted out as transactions confirm.

PR #1180, opened 2016-02-06
Move Transaction.isConsistent() to Wallet.isTxConsistent(), …

schildbach:
…as the wallet was the only consumer of that method.


schildbach:
Merged.

PR #1182, opened 2016-02-07
TransactionConfidence: Add lastBroadcastedAt, the time a transaction…

schildbach:
…was last announced to us.

The idea of this new field is that if a pending transaction was not announced for a longer time,
it is unlikely to confirm. Thus, the user should be warned and maybe the transaction removed from
the wallet at some point.


schildbach:
Merged.

PR #1183, opened 2016-02-07
Added broadcast multisig contract to wallet bloom filter. Fixes #1181

### LINKED ISSUE #1181: Transaction broadcast future never completes in payment channel demo
I [modified the payment channel demo to run on testnet3](https://github.com/w-shackleton/bitcoinj/tree/hang) to demonstrate this - if the demos are run and the client wallet is topped up with test coins then the server reaches `broadcastTransaction` in `PaymentChannelServerState`, and then that future never completes. I've tracked this down to the fact that the `InventoryMessage` detailing the transaction are never received in response from the peers (`Peer.processInv` never gets called post broadcast).

w-shackleton:
The payment channel contract broadcast future was never completing, since the client was never receiving notification of acceptance into memory pool. This adds the contract to the wallet's watch list so it receives notification.

### DIFF
--- a/core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelServerState.java
+++ b/core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelServerState.java
@@ -235,6 +235,7 @@ public class PaymentChannelServerState {
             throw e;
         }
         log.info("Broadcasting multisig contract: {}", multisigContract);
+        wallet.addWatchedScripts(ImmutableList.of(multisigContract.getOutput(0).getScriptPubKey()));
         state = State.WAITING_FOR_MULTISIG_ACCEPTANCE;
         final SettableFuture<PaymentChannelServerState> future = SettableFuture.create();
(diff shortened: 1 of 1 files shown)


w-shackleton:
I've checked and this code fixes payment channels in master, but is this the preferred way to add the tx to the bloom filter?

schildbach:
Thanks for debugging into this! @devrandom added the concept of watched scripts in commit da2e3e6c98281f251d46272cdfc717c4e34f3b0c, so maybe he can tell if your fix is the right thing to do.

devrandom:
Are none of the inputs in this transaction expected to be from addresses in the wallet?

w-shackleton:
No - the transaction comes from another wallet (the client in the payment channel). This wallet simply receives a serialised transaction whose inputs come from another wallet and then broadcasts it.

devrandom:
Then it's a reasonable solution.  The transaction is relevant, but the wallet would not know, so letting it know in this way is appropriate.

schildbach:
Great, thanks for clearing this up. So I'll merge this tomorrow.

schildbach:
Merged. Will, do you think we should backport this to 0.13.5?

PR #1184, opened 2016-02-09
Peer: Configure maximum recursion level when requesting dependent…

schildbach:
…transactions.

The idea is to limit download to a sane amount, rather than disabling it completely.


schildbach:
Merged.

Issue #464, opened 2015-06-11
Replace downloadBlockChain() by setMode(mode)


mikehearn:
Originally reported on Google Code with ID 98

```
Mode being one of DOWNLOAD or STANDBY.

Rationale: the concept of a "complete download of blockchain" is not meaningful in
Bitcoin, where we might be behind or on a losing fork of the chain and not know it.
 We should either continuously try to download, or pause.

Suggested by andreas.schildbach in issue 90.
```

Reported by `miron@google.com` on 2011-10-28 18:53:04

PR #1186, opened 2016-02-10
Split peer event listeners into single method interfaces

rnicoll:


schildbach:
Great, thanks! I'll test this soon.

core/src/main/java/org/bitcoinj/core/PeerGroup.java
```diff
@@ -1380,13 +1476,13 @@ public void setConnectTimeoutMillis(int connectTimeoutMillis) {
      *
      * @param listener a listener for chain download events, may not be null
      */
-    public void startBlockChainDownload(PeerDataEventListener listener) {
+    public void startBlockChainDownload(ChainDownloadStartedEventListener listener) {
```
schildbach: I fear some incompatibility here. Previously I passed my PeerDataEventListener and listened onBlocksDownloaded(). This doesn't seem possible any more at this point.

Reminds me we should get rid of this notion of "starting a block chain download". Or even the word "download". It's not a single process that starts at a specific time. It's an ongoing process, more like a "mode". There used to be a stone old ticket open for this.

rnicoll: Do you want me to try fixing the compatibility break, or just charge forwards and rip the entire event out? Or fix the break and mark it deprecated?

schildbach: In the context of this PR, we should stay as compatible as we can. However, we could start a new PR (which perhaps gets merged before this one) that removes PeerGroup.startBlockChainDownload() and adds a listener pair instead. I found the original ticket here: https://github.com/bitcoinj/bitcoinj/issues/464

rnicoll:
I've added a halfway solution that uses a common interface for the chain download starting, and a new block downloading. Avoids adding the event listener to a lot of lists it generally won't be involved with. Is that okay, or want me to go all the way back to PeerDataEventListener?

Can you assign #464 to me and I'll look at it later, but I think it's going to be simpler to solve this first.

rnicoll:
Rebuilt around registering/deregistering all four listeners

core/src/main/java/org/bitcoinj/core/listeners/PeerConnectionEventListener.java
```diff
@@ -21,21 +21,24 @@
 import java.util.Set;
```
schildbach: This file seems to be unchanged. If I'm right, can you remove the JavaDoc reformatting from this PR?

schildbach:
Yes, this looks good. We can change the startBlockChainDownload() in a separate commit.

rnicoll:
Removed the Javadoc changes

schildbach:
With this PR the onBlocksDownloaded() events are not thrown any more. Just checked, on master it still works. Investigating…

core/src/main/java/org/bitcoinj/core/PeerGroup.java
```diff
@@ -687,15 +699,74 @@ public void setUserAgent(String name, String version) {
...
+    /**
+     * <p>Adds a listener that will be notified on the given executor when
+     * blocks are downloaded by the download peer.</p>
+     * @see Peer#addBlocksDownloadedEventListener(Executor, BlocksDownloadedEventListener)
+     */
+    public void addBlocksDownloadedEventListener(Executor executor, BlocksDownloadedEventListener listener) {
+        peersBlocksDownloadedEventListeners.add(new ListenerRegistration<BlocksDownloadedEventListener>(checkNotNull(listener), executor));
+        for (Peer peer : getConnectedPeers())
+            peer.addBlocksDownloadedEventListener(executor, listener);
+        for (Peer peer: getPendingPeers())
```
schildbach: Tiny formatting: Space before ':'

core/src/main/java/org/bitcoinj/core/PeerGroup.java
```diff
@@ -1506,14 +1620,14 @@ private void setDownloadPeer(@Nullable Peer peer) {
             if (downloadPeer != null) {
                 log.info("Unsetting download peer: {}", downloadPeer);
                 if (downloadListener != null)
-                    downloadPeer.removeDataEventListener(downloadListener);
+                    downloadPeer.removeChainDownloadStartedEventListener(downloadListener);
```
schildbach: Do we need to remove it from the other lists as well?

core/src/main/java/org/bitcoinj/core/PeerGroup.java
```diff
@@ -1506,14 +1620,14 @@ private void setDownloadPeer(@Nullable Peer peer) {
...
                 downloadPeer.setDownloadData(false);
             }
             downloadPeer = peer;
             if (downloadPeer != null) {
                 log.info("Setting download peer: {}", downloadPeer);
                 if (downloadListener != null)
-                    peer.addDataEventListener(Threading.SAME_THREAD, downloadListener);
+                    peer.addChainDownloadStartedEventListener(Threading.SAME_THREAD, downloadListener);
```
schildbach: …and add it?

schildbach:
Yeah, that fixes my problem. I'll fix the remaining problem and merge.

rnicoll:
If you give me a second I've got a patch I'm testing, and refactors the add/remove into a common method so I can't get it wrong again :-D

schildbach:
Oh sorry I just merged. But the refactoring you're preparing is perhaps better kept in a separate commit anyway.

rnicoll:
Okay, closing then

schildbach:
Regarding listener splitting: Do you plan to split the remaining PeerConnectionEventListener too?

rnicoll:
Yup, on that now

PR #1187, opened 2016-02-11
Fixed stored channel timers failing, fixed v2 server deserialisation

w-shackleton:
* I added catch-all exception handlers to the `TimerTask` scheduled methods in the stored client and server channels - if something does fail during broadcast of transactions in either end of the payment channel then it won't take down the entire `TimerTask` with it now.
* I noticed that deserialised V2 servers weren't setting the expiry time correctly.


schildbach:
It would be nice to have one bugfix per commit. You can keep both commits in one PR if you like, but the separation between independent fixes would be nice to have.

w-shackleton:
:+1:

schildbach:
Merged, thanks.

PR #1188, opened 2016-02-13
Fix typo in PeerConnectionEventListener

rnicoll:
Correct "procesesed" to "processed"


rnicoll:
Then noticed same typo on PeerDataEventListener, so have added that here too

schildbach:
Merged.

PR #1190, opened 2016-02-13
Refactor adding/removing block download listeners

rnicoll:
Something that becomes apparent in doing this, is that when we first register listeners they have an executor on the user thread, and if the peer dies they're then added with an executor on the same thread. Is that intentional?


schildbach:
Merged.

PR #1191, opened 2016-02-13
Add listeners before starting peer group

rnicoll:
Okay, so it can't do anything until later, but still good practice to add the listeners first.


schildbach:
I think this commit's description could be a bit more precise. What parts of the framework does it affect? In this case, it's only one test class so I suggest prefixing the description with "PeerGroupTest:"

rnicoll:
How's that?

schildbach:
Great!

schildbach:
Merged.

PR #1192, opened 2016-02-13
Split PeerConnectionEventListener

rnicoll:
Split PeerConnectionEventListener into individual connect, disconnect and discovery listeners.
Remove custom listener registration from Peer, as now it's possible to register a connect listener only, without a disconnect listener.


schildbach:
Works fine! I wonder if we can remove the `PeerConnectionListenerRegistration` class? It's not referenced any more, and it's not public either.

rnicoll:
Done

PR #1195, opened 2016-02-16
Peer: Disconnect peers without copy of the block chain gracefully.

schildbach:
Previously, a protocol exception was thrown, spamming error reporting facilities with useless messages.


schildbach:
Merged.

PR #1196, opened 2016-02-16
Wallet.isConsistentOrThrow(): New variant of .isConsistent()…

schildbach:
…that throws an exception describing the inconsistency.


schildbach:
Merged.

PR #1197, opened 2016-02-17
Added getFromWallet method to StoredPaymentChannelServerStates

w-shackleton:
This matches the same method in `StoredPaymentChannelClientStates`. I also added a `getChannelMap` method to the client states as per the server states and made the `getChannel` method public as per the server states.

### DIFF
--- a/core/src/main/java/org/bitcoinj/protocols/channels/StoredPaymentChannelClientStates.java
+++ b/core/src/main/java/org/bitcoinj/protocols/channels/StoredPaymentChannelClientStates.java
@@ -16,6 +16,8 @@
 
 package org.bitcoinj.protocols.channels;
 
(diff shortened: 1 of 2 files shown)


schildbach:
Do you plan future commits that use these added methods?

w-shackleton:
I'm working on some things in [bitcoin-wallet](https://github.com/schildbach/bitcoin-wallet), and one thing I'd like to be able to do is list the current open channels in a `ListView`. Adding these methods allows me to call `toString` on each open channel to simply dump that channel. This is fine for diagnostics purposes, but a more full solution would be to provide a proper read-only API for retrieving useful information about stored channels. Should I add such things as well?

schildbach:
Thanks, merged. If you think it's useful for payment channels, go ahead. Right now, you're pretty much the payment channel expert.

PR #1199, opened 2016-02-24
Print transaction fee per kilobyte besides total fee in wallet dump.

schildbach:


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -693,9 +693,10 @@ public String toString(@Nullable AbstractBlockChain chain) {
             }
             s.append(String.format(Locale.US, "%n"));
         }
-        Coin fee = getFee();
+        final Coin fee = getFee();
         if (fee != null)
-            s.append("     fee  ").append(fee.toFriendlyString()).append(String.format(Locale.US, "%n"));
+            s.append("     fee  ").append(fee.multiply(1000).divide(unsafeBitcoinSerialize().length).toFriendlyString())
```
jlopp: "fee" is satoshis, right? Shouldn't it instead be: 

`s.append("     fee  ").append(fee.divide(unsafeBitcoinSerialize().length*1024).toFriendlyString())`

core/src/main/java/org/bitcoinj/core/Wallet.java
```diff
@@ -4196,13 +4196,14 @@ public void completeTx(SendRequest req) throws InsufficientMoneyException {
...
-            int size = req.tx.bitcoinSerialize().length;
+            final int size = req.tx.unsafeBitcoinSerialize().length;
             if (size > Transaction.MAX_STANDARD_TX_SIZE)
                 throw new ExceededMaxTransactionSize();
 
             final Coin calculatedFee = req.tx.getFee();
             if (calculatedFee != null)
-                log.info("  with a fee of {}", calculatedFee.toFriendlyString());
+                log.info("  with a fee of {}/kB, {} total",
+                        calculatedFee.multiply(1000).divide(size).toFriendlyString(), calculatedFee.toFriendlyString());
```
jlopp: Similarly:

`calculatedFee.divide(size*1024).toFriendlyString(), calculatedFee.toFriendlyString());`

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: Thanks for your review! First of all, I believe fee is always calculated per 1000 bytes, rather than 1024 bytes. Please correct me if I'm wrong.

The Coin class internally stores satoshis, but `.toFriendlyString()` prints BTC values currently. So in the end you'll see values like '0.00044 BTC/kB'. That's consistent with block explorers, e.g. https://blockexplorer.com/tx/05e1b533e63b8c93d0944e3b024a52a22677664e9c2b63364e336208d56db093

Lastly, I think you meant `/1024` rather than `*1024`. I didn't run your code but I believe it will output much too small values. Also, if you do it like that you'll run into rounding problems so I prefer to multiply first, then divide. Any thoughts?

luke-jr: Actually, for a long time now, feerate has been calculated per byte. (It used to be per 1000 bytes though.)

schildbach: @luke-jr If you calculate per byte, what precision do you use? Indevisible satoshis do not seem up to the task.

jlopp: Yeah, I meant length/1024 but for simplicity and @luke-jr's note it's probably best to just drop the conversion to KB and show the full precision of fees in satoshis per byte. That's what the fee estimators tend to recommend anyway.

jlopp: @schildbach that's a good point - I checked on the fee estimates from Core and they are returned as satoshis per KB: https://statoshi.info/dashboard/db/fee-estimates

schildbach: @jlopp My point is that per byte has not enough precision, as the number of satoshis per byte would be in the 10-50 range (and even has been lower in the past). Unless you print fractions of a satoshi, but I'm not sure this is what you want.

Issue #627, opened 2015-06-11
Peer.getDownloadData() -> Peer.isDownloadData()


mikehearn:
Originally reported on Google Code with ID 261
```
Boolean properties usually use "is" rather than "get" for the getter.
```

Reported by `andreas.schildbach` on 2012-12-15 17:22:02

schildbach:
Fixed via 4217a5c23168e88b75b6ddb9609edd445f5585a8.

PR #1201, opened 2016-02-26
Rename boolean Peer getters to use 'is' prefix.

jlopp:
This should allow us to close out #627


schildbach:
Looks good to me.

schildbach:
Merged.

Issue #1200, opened 2016-02-26
If I start peer group, I keep getting error stack printed into log every second, ip address does not change


PavelNiedoba:
[PeerGroup Thread] WARN org.bitcoinj.net.ConnectionHandler - Error writing message to connection, closing connection
java.nio.channels.CancelledKeyException
    at sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:73)
    at sun.nio.ch.SelectionKeyImpl.interestOps(SelectionKeyImpl.java:77)
    at org.bitcoinj.net.ConnectionHandler.setWriteOps(ConnectionHandler.java:106)
    at org.bitcoinj.net.ConnectionHandler.writeBytes(ConnectionHandler.java:152)
    at org.bitcoinj.core.PeerSocketHandler.sendMessage(PeerSocketHandler.java:90)
    at org.bitcoinj.core.Peer.ping(Peer.java:1535)
    at org.bitcoinj.core.Peer.ping(Peer.java:1526)
    at org.bitcoinj.core.PeerGroup$12.run(PeerGroup.java:1679)
    at com.google.common.util.concurrent.MoreExecutors$ScheduledListeningDecorator$NeverSuccessfulListenableFutureTask.run(MoreExecutors.java:648)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
    at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:304)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:178)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    at org.bitcoinj.utils.ContextPropagatingThreadFactory$1.run(ContextPropagatingThreadFactory.java:35)
    at java.lang.Thread.run(Thread.java:744)
[PeerGroup Thread] INFO org.bitcoinj.core.PeerSocketHandler - [188.226.202.220]:18333 - java.nio.channels.CancelledKeyException

schildbach:
Can you attach your whole log? What bitcoinj commit are you on?

[...]

PR #1202, opened 2016-02-27
Peer: Fix version handshake.

schildbach:
This PR fixes two issues with the version handshake:

- Handshake is only complete when both halves (incoming and outgoing version+verack) are complete. Both of these halves run in parallel. Previously, in some cases we started sending messages even if the handshake wasn't complete.
- I the remote Peer sends main protocol messages while handshake is still in progress, that's a protocol violation. TODO investigate if we should pardon such violations and just log them.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -153,11 +153,13 @@ public class Peer extends PeerSocketHandler {
     private static final int PING_MOVING_AVERAGE_WINDOW = 20;
 
     private volatile VersionMessage vPeerVersionMessage;
-    private boolean isAcked;
 
     // A settable future which completes (with this) when the connection is open
     private final SettableFuture<Peer> connectionOpenFuture = SettableFuture.create();
-    private final SettableFuture<Peer> versionHandshakeFuture = SettableFuture.create();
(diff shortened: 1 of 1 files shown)


schildbach:
This hopefully fixes #1200.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -451,6 +459,12 @@ protected void processMessage(Message m) throws Exception {
             currentFilteredBlock = null;
         }
 
+        // No further communication is possible until version handshake is complete.
+        if (!(m instanceof VersionMessage || m instanceof VersionAck
+                || (versionHandshakeFuture.isDone() && !versionHandshakeFuture.isCancelled())))
```
jlopp: Interesting, if you call isDone() or isCancelled() on a list of futures, it's equivalent to checking all of the futures and ANDing their responses?

schildbach: Yes, that's how I understand the [documentation of Futures.allAsList()](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/util/concurrent/Futures.html#allAsList%28com.google.common.util.concurrent.ListenableFuture...%29).

jlopp:
Looks good to me

schildbach:
Thanks for the review! I also tested the change for a while in the app and did not notice any regressions. So I just merged.

PR #1204, opened 2016-02-27
Fixed NullPointerException in creating new StoredPaymentChannelClientStates

w-shackleton:
cc @rnicoll - I think you added this extra check, just letting you know that I'm modifying this. I found a boundary case where this code can be reached without setting the wallet (causing a NPE in `getNetworkParameters()`
Is it OK to not perform the check when the stored client states object doesn't have an attached wallet? I could alternatively investigate the issue further to find why it's not being set.

### DIFF
--- a/core/src/main/java/org/bitcoinj/protocols/channels/StoredPaymentChannelClientStates.java
+++ b/core/src/main/java/org/bitcoinj/protocols/channels/StoredPaymentChannelClientStates.java
@@ -288,13 +288,16 @@ public class StoredPaymentChannelClientStates implements WalletExtension {
         lock.lock();
         try {
             final NetworkParameters params = getNetworkParameters();
+            // If we haven't attached to a wallet yet we can't check against network parameters
+            final boolean hasMaxMoney = params != null ? params.hasMaxMoney() : true;
+            final Coin networkMaxMoney = params != null ? params.getMaxMoney() : NetworkParameters.MAX_MONEY;
             ClientState.StoredClientPaymentChannels.Builder builder = ClientState.StoredClientPaymentChannels.newBuilder();
             for (StoredClientChannel channel : mapChannels.values()) {
                 // First a few asserts to make sure things won't break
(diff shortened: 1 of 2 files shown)


rnicoll:
My change was taking it from a static maximum (for Bitcoin) and replacing it with a network-dependent value, rather than the original change. Might be better to have it behave as previously (i,e, apply Bitcoin limit) where no wallet is available, and I'll look into why the wallet is sometimes not set, later. It's not something I expect anyone to use with Lite/Doge, so difficult to make a case for changing behaviour to be compatible with those networks.

albrechta:
The wallet is not set yet because of the way the `WalletAppKit` sets up the extensions. The WalletAppKit first serializes and then immediately deserializes the extensions again during the create procedure of the wallet. During the serialization, the wallet is not set and the NPE occurs. The wallet may not exist yet when you create the extension instances (using `provideWalletExtensions`), i.e. you cannot pass the wallet to the constructor.

See this comment: https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/kits/WalletAppKit.java#L372-L374

w-shackleton:
Ohh it's *that* bug! :+1: 
I've switched the behaviour to match @rnicoll's suggestion (use `NetworkParamters.MAX_MONEY` when no wallet is available)

schildbach:
Somehow this change feels a bit like "curing the symptom". Though I must admit I'm not confident in the payment channels code.

Is it actually a valid usecase to have a channel without wallet? If not, perhaps the wiring up should be fixed so that this state can't exist? Just a thought…

w-shackleton:
I think this is because on creation of a wallet, the extensions are created and then serialised into a fresh wallet, but are created before it. The wallet is then serialised and deserialised to yield the correctly linked wallet.

w-shackleton:
I've applied this fix to `StoredPaymentChannelServerStates` as well.

schildbach:
Ok, thanks for the explanation. Merged.

Issue #1203, opened 2016-02-27
Wallet.saveToFile throws java.io.IOException: Invalid argument


rickheiden:
If the APP_NAME in wallettemplate's main.java contains a special character, like an asterisk, the Wallet.saveToFile throws java.io.IOException: Invalid argument. The reason for this is that the wallet's filename includes the "public static String APP_NAME" assignment, so when file creation is attempted Windows returns an error. I think filenames should be checked for invalid characters.

schildbach:
Or maybe construct a valid filename by filtering special characters? Would you like to send a pull request?

rickheiden:
Would you suggest in main.java or an enhancement to saveToFile method? I'm not familiar with a lot of the project. I'm assuming most code for doing file saving uses saveToFile.

schildbach:
I propose #1206 as a fix. Does it help in your case?

rickheiden:
Nicely done. Works well.

On Mon, Feb 29, 2016 at 6:47 AM, Andreas Schildbach <
notifications@github.com> wrote:

> I propose #1206 <https://github.com/bitcoinj/bitcoinj/pull/1206> as a
> fix. Does it help in your case?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitcoinj/bitcoinj/issues/1203#issuecomment-190175453>.
>



-- 

Rick

PR #1206, opened 2016-02-29
wallet-template: When constructing the wallet file name from…

schildbach:
… the app's name, replace problematic characters so the filesystem doesn't throw up.


schildbach:
Fix for #1203.

schildbach:
Merged.

PR #1208, opened 2016-03-01
Change fee calculation in WalletadjustOutputDownwardsForFee():

schildbach:
- Cent rule is removed
- Fee based on feePerKb is calculated on byte precision

This change affects the special case of emptying and rekeying wallets. Fee calculation of standard transactions is unaffected for now.


schildbach:
Merged.

Issue #1209, opened 2016-03-03
Protobuf serialization fails for watching wallets


schildbach:
Consider this test case, to be pasted into WalletProtobufSerializerTest:

```
    @Test
    public void testRoundTripWatchingWallet() throws Exception {
        final String xpub = "tpubD9LrDvFDrB6wYNhbR2XcRRaT4yCa37TjBR3YthBQvrtEwEq6CKeEXUs3TppQd38rfxmxD1qLkC99iP3vKcKwLESSSYdFAftbrpuhSnsw6XM";
        final long creationTimeSeconds = 1457019819;
        Wallet wallet = Wallet.fromWatchingKey(PARAMS, DeterministicKey.deserializeB58(null, xpub, PARAMS),
                creationTimeSeconds);
        Wallet wallet2 = roundTrip(wallet);
        assertEquals(xpub, wallet.getWatchingKey().serializePubB58(PARAMS));
        assertEquals(xpub, wallet2.getWatchingKey().serializePubB58(PARAMS)); // fails!
        assertEquals(creationTimeSeconds, wallet.getWatchingKey().getCreationTimeSeconds()); // fails!
        assertEquals(creationTimeSeconds, wallet2.getWatchingKey().getCreationTimeSeconds()); // fails!
        assertEquals(creationTimeSeconds, wallet.getEarliestKeyCreationTime()); // fails!
        assertEquals(creationTimeSeconds, wallet2.getEarliestKeyCreationTime()); // fails!
    }
```

As you can see, bitcoinj fails to persist both the birth date and the xpub.

Pinging @devrandom and @oscarguindzberg as they were the last to touch the DKC serialization code.

schildbach:
I think I fixed the issue in #1210. The root issue is a lot of the API puts the creation time into fields that can't get persisted to protobuf. See the fix for details, review is much appreciated.

schildbach:
Fixed.

PR #1210, opened 2016-03-04
For watching wallets, store the key creation time…

schildbach:
…always in the watching key rather than the DeterministicKeyChain. Creation times in the DeterministicKeyChain can't be persisted to protobuf, as that structure has no full-blown protobuf equivalent.

This means a couple of DeterministicKeyChain, KeyChainGroup and Wallet factory method variants that take creation dates have been removed. On the other hand, a convenient Wallet.fromWatchingKeyB58(params, xpub, creationDate) has been added.

Also adds a test for protobuf-roundtripping watching wallets.

Supposed to fix issue #1209.


schildbach:
Merged.

PR #1215, opened 2016-03-07
Wallet: Get rid of the concept of a "base fee".

schildbach:
Fee is always expressed as a rate in Coin per 1000 bytes.

SendRequest.fee was removed; use SendRequest.feePerKb for requesting a certain fee rate. If the field was
read for knowing the calculated fee, use SendRequest.tx.getFee() instead.

Also makes a couple of unit tests independent of fee.


schildbach:
Merged.

PR #1221, opened 2016-03-13
Added documentation which would help developers run the RefreshWallet example

krisrr3:
Added documentation which would help developers run the RefreshWallet example. Also ignoring the generated wallet in the commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ target
 *.iml
 *.chain
 *.spvchain
+test.wallet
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,8 +1,9 @@
...
-*.spvchain
+target
+.project
+.classpath
+.settings
+.idea
+*.iml
+*.chain
+*.spvchain
+test.wallet
```
schildbach: Please don't add any .wallet files to .gitignore, even if it's called test.wallet.

examples/src/main/java/org/bitcoinj/examples/RefreshWallet.java
```diff
@@ -27,6 +27,15 @@
 
 /**
  * RefreshWallet loads a wallet, then processes the block chain to update the transaction pools within it.
+ * 
+ * To get a test wallet run the following command against the tools subproject: <br>
```
schildbach: I think rather than providing the exact command line it would be enough to just refer to wallet-tool.

schildbach:
I added a small comment via commit 840df06b79beac1b984e6e247e90fcdedc4ad6e0.

PR #1222, opened 2016-03-13
complete SigHash enum and make updates to stop using ordinal

mruddy:
Complete the SigHash enum and make updates to stop using the enum's ordinal while preserving the original ordinal value for any existing code that might use it.

In my own code, I found that I was using an enum defined like below. The changes in this PR, deviate slightly from my original code in order to preserve the ordinal values of the original bitcoinj enum should any external code depend on it still. All in all, I think these changes make the SigHash enum more complete and make the code a little easier to read (less brittle with the removal of the ordinal dependency in the core code) with fewer casts. Let me know if I missed something, but I think my references checks were pretty complete and all the bitcoinj projects still compile and pass tests.

```
public enum Sighash {
  UNSET(0), ALL(1), NONE(2), SINGLE(3), ANYONECANPAY(0x80), ANYONECANPAY_ALL(0x81), ANYONECANPAY_NONE(0x82), ANYONECANPAY_SINGLE(0x83);
  private final int type;

  /**
   * @param type
   */
  private Sighash(final int type) {
    this.type = type;
  }

  /**
   * @return the type as a byte
   */
  public byte byteValue() {
    return (byte) this.type;
  }

  /**
   * @return the type
   */
  public int intValue() {
    return this.type;
  }
}
```

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -460,16 +460,37 @@ public class Transaction extends ChildMessage {
     /**
      * These constants are a part of a scriptSig signature on the inputs. They define the details of how a
      * transaction can be redeemed, specifically, they control how the hash of the transaction is calculated.
-     * <p/>
-     * In Bitcoin Core, this enum also has another flag, SIGHASH_ANYONECANPAY. In this implementation,
-     * that's kept separate. Only SIGHASH_ALL is actually used in Bitcoin Core today. The other flags
-     * exist to allow for distributed contracts.
      */
     public enum SigHash {
-        ALL,         // 1
-        NONE,        // 2
-        SINGLE,      // 3
+        ALL(1),
+        NONE(2),
+        SINGLE(3),
+        ANYONECANPAY(0x80), // Caution: Using this type in isolation is non-standard. Treated similar to ANYONECANPAY_ALL.
+        ANYONECANPAY_ALL(0x81),
+        ANYONECANPAY_NONE(0x82),
+        ANYONECANPAY_SINGLE(0x83),
+        UNSET(0); // Caution: Using this type in isolation is non-standard. Treated similar to ALL.
+
+        public final int value;
+
+        /**
+         * @param value
+         */
+        private SigHash(final int value) {
+            this.value = value;
+        }
+
+        /**
+         * @return the value as a byte
+         */
+        public byte byteValue() {
+            return (byte) this.value;
+        }
     }
+
+    /**
+     * @deprecated Instead use SigHash.ANYONECANPAY.value or SigHash.ANYONECANPAY.byteValue() as appropriate.
+     */
     public static final byte SIGHASH_ANYONECANPAY_VALUE = (byte) 0x80;
 
     @Override
@@ -958,14 +979,14 @@ public class Transaction extends ChildMessage {
             TransactionInput input = tx.inputs.get(inputIndex);
             input.setScriptBytes(connectedScript);
 
-            if ((sigHashType & 0x1f) == (SigHash.NONE.ordinal() + 1)) {
+            if ((sigHashType & 0x1f) == SigHash.NONE.value) {
                 // SIGHASH_NONE means no outputs are signed at all - the signature is effectively for a "blank cheque".
                 tx.outputs = new ArrayList<TransactionOutput>(0);
                 // The signature isn't broken by new versions of the transaction issued by other parties.
                 for (int i = 0; i < tx.inputs.size(); i++)
                     if (i != inputIndex)
                         tx.inputs.get(i).setSequenceNumber(0);
-            } else if ((sigHashType & 0x1f) == (SigHash.SINGLE.ordinal() + 1)) {
+            } else if ((sigHashType & 0x1f) == SigHash.SINGLE.value) {
                 // SIGHASH_SINGLE means only sign the output at the same index as the input (ie, my output).
                 if (inputIndex >= tx.outputs.size()) {
                     // The input index is beyond the number of outputs, it's a buggy signature made by a broken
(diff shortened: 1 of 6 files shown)


schildbach:
I'm fine with embedding the value into the enum. I'm not so sure about the new enum values, isn't ANYONECANPAY more like a flag? Finally, I'd make `private final int type;` public, so the `intValue()` method isn't needed.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -466,9 +466,37 @@ public void setUpdateTime(Date updatedAt) {
...
+        ANYONECANPAY_SINGLE(0x83),
+        UNSET(0);
+
+        private final int type;
+
+        /**
+         * @param type
+         */
+        private SigHash(final int type) {
+          this.type = type;
```
schildbach: Mind the spacing here. 4 spaces per level.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -466,9 +466,37 @@ public void setUpdateTime(Date updatedAt) {
...
-        NONE,        // 2
-        SINGLE,      // 3
+        ALL(1),
+        NONE(2),
+        SINGLE(3),
+        ANYONECANPAY(0x80),
+        ANYONECANPAY_ALL(0x81),
+        ANYONECANPAY_NONE(0x82),
+        ANYONECANPAY_SINGLE(0x83),
+        UNSET(0);
```
schildbach: Do we need an UNSET value? In the Java world, this would typically be represented as `null`.

mruddy:
@schildbach OK, updated. I made the type publicly accessible. I had it private and used intValue as a stylistic thing to parallel how things are done in the java.lang wrapper classes. But, no big deal, I changed it.

About the enum values:
ANYONECANPAY, SINGLE, and NONE are all really treated like flags, reference https://github.com/bitcoin/bitcoin/blob/a38107643f3f01cd92a4e6da9e7a9d025770ff37/src/script/interpreter.cpp#L1016-L1105

ALL is default behavior which is why 0x00 and 0x80 are valid also.

0x00 is equivalent to SIGHASH_ALL and 0x80 is like SIGHASH_ANYONECANPAY|SIGHASH_ALL.

Using 0x00 and 0x80 are treated as non-standard, but I don't see a consensus rule prohibiting their use.
In fact, that's why this transaction mentioned in a comment in the bitcoinj code is valid (it uses a 0x00 sighash type): https://blockchain.info/tx/c99c49da4c38af669dea436d3e73780dfdb6c1ecf9958baa52960e8baee30e73?show_adv=true

That's why I added UNSET as an enum value. I didn't have a better name. With this explanation though, it might make more sense why I named it that.

I updated some comments too. It's all fundamentally pretty much the same as the previous commit though.

core/src/test/java/org/bitcoinj/core/FullBlockTestGenerator.java
```diff
@@ -759,7 +759,7 @@ public boolean add(Rule element) {
                     try {
                         ByteArrayOutputStream bos = new UnsafeByteArrayOutputStream(73);
                         bos.write(coinbaseOutKey.sign(hash).encodeToDER());
-                        bos.write(SigHash.SINGLE.ordinal() + 1);
```
schildbach: Really byteValue() here? ordinal()+1 was an int, no?

core/src/test/java/org/bitcoinj/core/FullBlockTestGenerator.java
```diff
@@ -830,7 +830,7 @@ public boolean add(Rule element) {
                             ByteArrayOutputStream bos = new UnsafeByteArrayOutputStream(
                                     73);
                             bos.write(coinbaseOutKey.sign(hash).encodeToDER());
-                            bos.write(SigHash.SINGLE.ordinal() + 1);
```
schildbach: And here?

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -460,16 +460,34 @@ public void setUpdateTime(Date updatedAt) {
...
+
+        /**
+         * @return the type as a byte
+         */
+        public byte byteValue() {
+            return (byte) this.type;
+        }
     }
+
     public static final byte SIGHASH_ANYONECANPAY_VALUE = (byte) 0x80;
```
schildbach: Is this still needed? One could write ANYONECANPAY.type too.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -460,16 +460,34 @@ public void setUpdateTime(Date updatedAt) {
...
+        ALL(1),
+        NONE(2),
+        SINGLE(3),
+        ANYONECANPAY(0x80), // Caution: Using this type in isolation is non-standard. Treated similar to ANYONECANPAY_ALL.
+        ANYONECANPAY_ALL(0x81),
+        ANYONECANPAY_NONE(0x82),
+        ANYONECANPAY_SINGLE(0x83),
+        UNSET(0); // Caution: Using this type in isolation is non-standard. Treated similar to ALL.
+
+        public final int type;
```
schildbach: Maybe rename this to "value"?

schildbach:
Thanks for the clarification. It feels weird an enum as a bitfield, but you're right it was used like that already before.

core/src/main/java/org/bitcoinj/core/Transaction.java
mruddy: sure, will do in my next commit. the same thing crossed my mind before. i left it named "type" because people sometimes refer to the "sighash type". but "value" makes sense when reading the code that uses it.

core/src/main/java/org/bitcoinj/core/Transaction.java
mruddy: sure, i'll update references to that in my next commit and get rid of it.

core/src/test/java/org/bitcoinj/core/FullBlockTestGenerator.java
mruddy: yep, definitely. both have the same result in this case. but, not using byteValue() avoids some unnecessary casting so i'll update this in my next commit. i only used byteValue() because i liked the readability.

core/src/main/java/org/bitcoinj/core/Transaction.java
mruddy: on second thought, i remember that left that around because it was a public member that existing code could depend on. i think i'll leave it as a public member, but remove references to it. it effectively becomes deprecated. i'll probably add a comment that it's now deprecated and to use the enum.

schildbach:
Looks good! I merged this. Thanks very much!

PR #1223, opened 2016-03-15
Added source and target versions for compilation

deb-sandeep:
The compilation was failing when run under JDK 1.8. Added the build source and target versions.

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -30,6 +30,18 @@
     <name>bitcoinj Examples</name>
     <description>A collection of examples using the bitcoinj library</description>
 
+    <build>
+        <plugins>
+            <plugin>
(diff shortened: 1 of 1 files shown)


schildbach:
The bitcoinj parent POM already contains:

```
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-compiler-plugin</artifactId>
        <configuration>
          <source>1.8</source>
          <target>1.8</target>
          <showDeprecation>true</showDeprecation>
          <showWarnings>true</showWarnings>
        </configuration>
      </plugin>
```

Only the core POMs override this with source/target 1.6.

Regarding your compilation problem, I suggest asking on the mailing list or opening an issue. Please include the build log.

PR #1224, opened 2016-03-15
H2FullPrunedBlockStore constructor with H2 database credentials

kushti:
I've added H2FullPrunedBlockStore constructor with H2 database credentials. H2 database viewers(e.g. default launched with java -jar h2[...].jar) require credentials, so the constructor could be useful.


schildbach:
Merged. Thanks for your contribution!

PR #1226, opened 2016-03-16
Fixing reading of sequence to big endian.

Xekyo:
Sequence was incorrectly read as little endian, but is big endian. Should fix issue where `nLockTime` transactions get falsely labeled as OptInRBF transactions.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Message.java
+++ b/core/src/main/java/org/bitcoinj/core/Message.java
@@ -286,10 +286,20 @@ public abstract class Message {
             checkState(false, "Length field has not been set in %s.", getClass().getSimpleName());
         return length;
     }
-
+    
     long readUint32() throws ProtocolException {
         try {
             long u = Utils.readUint32(payload, cursor);
+            cursor += 4;
+            return u;
+        } catch (ArrayIndexOutOfBoundsException e) {
+            throw new ProtocolException(e);
+        }
+    }
+    
+    long readUint32BE() throws ProtocolException {
+        try {
+            long u = Utils.readUint32BE(payload, cursor);
             cursor += 4;
             return u;
         } catch (ArrayIndexOutOfBoundsException e) {
--- a/core/src/main/java/org/bitcoinj/core/TransactionInput.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionInput.java
@@ -136,7 +136,7 @@ public class TransactionInput extends ChildMessage {
         int scriptLen = (int) readVarInt();
         length = cursor - offset + scriptLen + 4;
         scriptBytes = readBytes(scriptLen);
-        sequence = readUint32();
+        sequence = readUint32BE();
     }
 
     @Override


instagibbs:
I think you need to implement the equivalent of the call in: https://github.com/bitcoinj/bitcoinj/blob/becd8641c92817e85029ddd63d1f424cdd626278/core/src/main/java/org/bitcoinj/core/Message.java#L290

schildbach:
Thanks! If this is true, the parsing bug existed since the first version of bitcoinj. Do you have any reference?

https://en.bitcoin.it/wiki/Protocol_documentation says:

> Almost all integers are encoded in little endian. Only IP or port number are encoded big endian.

Do you have a transaction hash that shows the issue? Ideally we would add a test for one of those transactions.

core/src/main/java/org/bitcoinj/core/Message.java
```diff
@@ -286,7 +286,7 @@ public final int getMessageSize() {
             checkState(false, "Length field has not been set in %s.", getClass().getSimpleName());
         return length;
     }
-
```
schildbach: Stray whitespace. Can you remove this change?

instagibbs:
After some further code-reading, I think this is incorrect. It appears to be serialized as LE in Core, and correctly deserialized in the app itself.

The concern was the wallet was marking standard Core transactions as RBF. I ran my own test on testnet, and the wallet handled it just fine. 

Still a bit confusing why this user was having issues: http://bitcoin.stackexchange.com/questions/43367/what-is-an-insecure-transaction-type-as-mentioned-by-bitcoin-wallet-for-andro

schildbach:
So this means the parsing in bitcoinj is correct? (Little endian)

To me, it looks like some bug in how the transaction was created. Or maybe the 0xfeffffff locktime was intentional somehow?

I just scrolled through all my past transactions (hundreds!), and not one is classified as RBF. I guess a good amount of them has been created by Bitcoin Core.

schildbach:
Also I just noticed there is no `gd26f951` commit in the bitcoin repo (the user claims to be using `Bitcoin Core v0.11.0.0-gd26f951 (64-bit)`). I wonder if this is custom in some way?

instagibbs:
That feffffff is correct 0.12 Core behavior. It's max-1 for nlocktime to
prevent fee sniping.

schildbach:
Can someone paste the raw transaction data of the transactions mentioned on stack exchange? I want to reproduce this issue in a unit test, but do not have an updated Bitcoin Core node at hand.

instagibbs:
./src/bitcoin-cli -testnet getrawtransaction 0e5a00dfdfca9e7a00dd828fa00573b5a81a0b6e783fed399b469cc8a9b2e4ec

0100000001f5b23d01a20f3a1a5477ed7f727d992d88e934536943ed0ff51029bed9ca76e5010000006a47304402201b10317e6b8e6b585acdef9926e7af85c6291d13e7ddef52c4a2a74200e648e702205fb1c90ba7c0db812d485349f90debf0cc4b1ab2926f87c18525c89496e748a80121036b7fdbd327b3088f357640699c9f8f8d9f6eb80676267fb8775675ea4a9e1587feffffff0280f0fa02000000001976a914b97c6f6c86158bb6c4ec15cf4efbdaa962b7d6fa88ac1f0127ad010000001976a914936cd0d271b08a1a1fa006323728602b8db4a03988aca9160b00

Notice the fefffff's.

schildbach:
How can I cut and paste just the raw data on that page? The formatting seems to prevent that.

instagibbs:
I don't run txindex on mainnet sorry(pruned) :(  

But throw a dart on a block explorer and you'll probably hit that FEFFFFF pattern.

instagibbs:
Here we go: http://btc.blockr.io/api/v1/tx/raw/168ad57b8f99d4e966249d5f6332805e13e53c7e579f2e17fca22fe9df0430cd  

http://btc.blockr.io/api/v1/tx/raw/69a0d15abaae43f5e5e05d3c679b73986f9cd7f00add1c307e507754bf935d83

schildbach:
Ok, I took the first posted transaction, but so far everything works as expected:

```
    @Test
    public void optInFullRBF_notRBF() {
        // a standard transaction as created by Bitcoin Core 0.12
        Transaction tx = new Transaction(MainNetParams.get(), Utils.HEX.decode(
                "0100000001f5b23d01a20f3a1a5477ed7f727d992d88e934536943ed0ff51029bed9ca76e5010000006a47304402201b10317e6b8e6b585acdef9926e7af85c6291d13e7ddef52c4a2a74200e648e702205fb1c90ba7c0db812d485349f90debf0cc4b1ab2926f87c18525c89496e748a80121036b7fdbd327b3088f357640699c9f8f8d9f6eb80676267fb8775675ea4a9e1587feffffff0280f0fa02000000001976a914b97c6f6c86158bb6c4ec15cf4efbdaa962b7d6fa88ac1f0127ad010000001976a914936cd0d271b08a1a1fa006323728602b8db4a03988aca9160b00"));
        assertEquals("0e5a00dfdfca9e7a00dd828fa00573b5a81a0b6e783fed399b469cc8a9b2e4ec", tx.getHashAsString());
        assertEquals(1, tx.getInputs().size());
        assertEquals(TransactionInput.NO_SEQUENCE - 1, tx.getInput(0).getSequenceNumber());
        assertFalse(tx.isOptInFullRBF());
    }
```

I tried both `master` and `release-0.13`.

Xekyo:
Hi @schildbach,
Yeah, sorry, I messed up the endianness in the first post. Endianness is confusing! :( Thank you @instagibbs for clearing up the confusion.

@schildbach: I am kinda confused though that the message apparently showed up for the linked transaction for the user that asked on Bitcoin.Stackexchange.com. I understand that you couldn't reproduce the behavior? I assume that you have added the test to another branch already, so probably this pull-request is obsolete?

schildbach:
@Xekyo The above test is for another transaction which obviously has the NO_SEQUENCE-1 sequence value. It would still be useful to get the two transactions that were linked in the stackexchange in its raw form.

Xekyo:
@schildbach That's the thing. The two transactions that  [instagibbs posted](https://github.com/bitcoinj/bitcoinj/pull/1226#issuecomment-199344895) are the ones that the [asker on Bitcoin.SE cited (see the links "First, Second")](http://bitcoin.stackexchange.com/q/43367/5406)! I just checked again the original comment, they match. I've also requested confirmation by the asker.

schildbach:
Sorry, I missed the comment with the blockr links. Too bad I still can't reproduce the problem. It must be some different issue.

Xekyo:
@schildbach: The [asker confirmed](http://bitcoin.stackexchange.com/questions/43367/what-is-an-insecure-transaction-type-as-mentioned-by-bitcoin-wallet-for-andro?noredirect=1#comment50970_43367) that the above two transactions did indeed produce the message "This payment was delayed because the sender used an insecure transaction type". He added a screenshot:

Opt-In RBF

The address from the screenshot has no other incoming transactions: [1MBPZt3BuCGG694aBYUauZ4o4jwo4fV8p1](https://blockexplorer.com/address/1MBPZt3BuCGG694aBYUauZ4o4jwo4fV8p1)

schildbach:
Closing, as the reason for this bug was a different one (fixed via 5aa4b66c7ed07275a0c396d2acee46f5b943a176). Still: Thanks for your contribution, it's much appreciated!

PR #1230, opened 2016-03-24
Fix Integer overflow in BitcoinSerializer.BitcoinPacketHeader

WNoort:
Casting readUint32(...) to int could result in negative size. Causes crash at BitcoinSerializer.java:164

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
+++ b/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
@@ -367,7 +367,7 @@ public class BitcoinSerializer extends MessageSerializer {
             size = (int) readUint32(header, cursor);
             cursor += 4;
 
-            if (size > Message.MAX_SIZE)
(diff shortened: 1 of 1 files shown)


schildbach:
Just out of curiosity: Is there any implementation sending such huge packet sizes?

schildbach:
Merged, thanks!

WNoort:
@schildbach I hope not. It triggered once with me though. Nice DOS possibility.

schildbach:
How can that be a DOS vector? It was throwing an IndexOutOfBoundsException before, right?

WNoort:
No. It causes a NegativeArraySizeException if used as argument of deserializePayload:
```public Message deserializePayload(BitcoinPacketHeader header, ByteBuffer in) throws 
ProtocolException, BufferUnderflowException {
        byte[] payloadBytes = new byte[header.size];```

schildbach:
Ah yes, of course. But it would not cause excessive resource consumption, or do you think it could?

Issue #1211, opened 2016-03-05
Private Keys Vulnerable to New Android Attack


HashEngineering:
There is a new article out concerning this matter.

https://news.bitcoin.com/private-keys-vulnerable-to-new-android-ios-attack/

The above article refers to these two papers:
1. https://www.cs.tau.ac.il/~tromer/mobilesc/
2. https://eprint.iacr.org/2016/231

Paper 1 says:
"BouncyCastle: Android's version of BouncyCastle is vulnerable, as discovered by a concurrent and independent work, Side-Channel Analysis of Weierstrass and Koblitz Curve ECDSA on Android Smartphones."

Is this a matter to be concerned about?  Is there a plan for code changes?

schildbach:
I'm quite sure any countermeasures will need to be implemented in [BouncyCastle](http://bouncycastle.org/)/[SpongyCastle](https://github.com/rtyley/spongycastle).

However, the paper says

> The attack requires measuring a few thousand ECDSA signatures

…presumably by the same key. This means massive address re-use, which bitcoinj already disregards by using HD wallets etc. So in a way, we implemented the countermeasure already years ago.

greenaddress:
As of early February secp256k1 comes with tested [JNI bindings](https://github.com/bitcoin/secp256k1/commit/e4570184ffba0794dee0dd6f5d76d4edeb50fa0d)

Both openssl an BC are likely to patch the code against this vulnerability; even so, the objective-C library Core Bitcoin decided to move to libsecp256k1 presumably because they like that it wasn't affected and that it has shown good testing and performance.

bitcoinj has some now old interface to libsecp256k1 but that diverged since; given Java bindings are now tested and available it may be a good idea to resume optionally supporting it in bitcoinj. @schildbach I've been toying with it and if you are interested could work on a PR.

> …presumably by the same key. This means massive address re-use, which bitcoinj already disregards by using HD wallets etc. So in a way, we implemented the countermeasure already years ago.

@schildbach True however with opt-in rbf and lightning we are likely to make multiple signatures with the same private key and the methods used to extract the key may be improved to the point where even low reuse is enough to extract a private key. If the private key is part of a bip32 wallet and public derived (non hardened, non private) and the attack has access to your public master then it is game over the attacker can reconstruct your private master and all your private keys, including privately derived are compromised.

[...]

PR #1233, opened 2016-03-26
Native ECDSA implementation updated to latest libsecp256k1

greenaddress:
This should help improve #1211 

The bindings are enabled if the secp256k1 so file is found, otherwise falls back to normal.
For now we hooked up bindings for verify and sign (tested on GreenBits) but in a new PR people can also use secp256k1 for BIP32 or other things.

helper script to build the secp256k1 library for Android [here](https://github.com/greenaddress/GreenBits/commit/b58b926bedcb8ce1e3701f95900beaa10f2b0f86) but you should be able to test it on a normal linux box, windows or osx.

### DIFF
--- a/core/src/main/java/org/bitcoin/NativeSecp256k1.java
+++ b/core/src/main/java/org/bitcoin/NativeSecp256k1.java
@@ -1,5 +1,6 @@
 /*
  * Copyright 2013 Google Inc.
+ * Copyright 2014-2016 the libsecp256k1 contributors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


schildbach:
Thanks for the contribution! I wonder is this code unit tested in some way? I know it's probably difficult due to the native code…

The indenting is a bit off at certain parts of the commit. Remember we indent by 4 spaces per indent level. But I can do the indenting for you when merging, so don't worry too much.

schildbach:
Another thing I noticed the code from this PR is more or less a copy of https://github.com/bitcoin/secp256k1/tree/master/src/java. What the plan for the future? Should the JNI wrappers exist at both places?

greenaddress:
we can fix the indenting and then i can do a PR to secp to keep it in sync (i already did here https://github.com/bitcoin/secp256k1/pull/384 for the copyrights/license bit).

there is some travis testing of the jni bindings within the secp repo

I think it is good for bitcoinj to auto detect and use the native secp256k1 library if available and likewise i think it is good that libsecp256k1 runs the tests for the jni bindings in travis.

I guess ideally bitcoinj would have an optional dependency on a javasecp256k1 package but i don't think anyone is planning to package that for maven/gradle; until then, seems like having a couple of files in both repos could be a reasonable compromise.

schildbach:
I fixed the formatting and also rewrite the statements that printed to the console to use proper logging. Merged, thanks!

Yes, an optional dependency containing the wrappers and the native code would be perfect. Ideally it could be built from source, including the native code. The gradle-android plugin 2.0 is said to be able to built native libraries.

FWIW, we have the same "problem" with scrypt.

PR #1241, opened 2016-04-13
Allow to specify buffer size for saving wallet file via OutputStream

jdojff:
Now it's possible to specify system dependent output buffer size for better IO performance writing wallet to disk.

```java
public class MyWallet extends Wallet {

	public static final int WALLET_IO_BUFFER_SIZE = 512 * 1024; // 512KB

	// ...

	@Override
	public void saveToFileStream(OutputStream f) throws IOException {
		lock.lock();
		try {
			new WalletProtobufSerializer().writeWallet(this, f, WALLET_IO_BUFFER_SIZE);
		} finally {
			lock.unlock();
		}
	}
}
```

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/WalletProtobufSerializer.java
+++ b/core/src/main/java/org/bitcoinj/wallet/WalletProtobufSerializer.java
@@ -131,10 +131,7 @@ public class WalletProtobufSerializer {
      */
     public void writeWallet(Wallet wallet, OutputStream output) throws IOException {
         Protos.Wallet walletProto = walletToProto(wallet);
-
-        int serializedSize = walletProto.getSerializedSize();
-        int bufferSize = serializedSize > this.walletWriteBufferSize ? this.walletWriteBufferSize : serializedSize;
-        final CodedOutputStream codedOutput = CodedOutputStream.newInstance(output, bufferSize);
+        final CodedOutputStream codedOutput = CodedOutputStream.newInstance(output, this.walletWriteBufferSize);
         walletProto.writeTo(codedOutput);
         codedOutput.flush();
     }
--- a/core/src/main/java/org/bitcoinj/wallet/WalletProtobufSerializer.java
+++ b/core/src/main/java/org/bitcoinj/wallet/WalletProtobufSerializer.java
@@ -31,6 +31,7 @@ import org.bitcoinj.wallet.Protos.Wallet.EncryptionType;
 import com.google.common.collect.Lists;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.CodedInputStream;
+import com.google.protobuf.CodedOutputStream;
 import com.google.protobuf.TextFormat;
 import com.google.protobuf.WireFormat;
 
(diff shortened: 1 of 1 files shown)


schildbach:
Thanks for the contribution!

Do you have some benchmarks that show an advantage?

API-wise I think rather than specifying the buffer size as a parameter of writeWallet() why not make it a settable property of WalletProtobufSerializer? It's not like you want a different buffer size for each operation.

jdojff:
Updated API, now usage would be:

```java
public class MyWallet extends Wallet {

	public static final int WALLET_IO_BUFFER_SIZE = 512 * 1024; // 512KB

	// ...

	@Override
	public void saveToFileStream(OutputStream f) throws IOException {
		lock.lock();
		try {
			WalletProtobufSerializer walletProtobufSerializer = new WalletProtobufSerializer();
			walletProtobufSerializer.setWalletWriteBufferSize(WALLET_IO_BUFFER_SIZE);
			walletProtobufSerializer.writeWallet(this, f);
		} finally {
			lock.unlock();
		}
	}
}
```

I don't have exact benchmarks at the moment. IO performance is highly whole stack dependent, so this may not give you better performance, need to experiment on your own environment. For example, AWS (http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-io-characteristics.html) provisioned storage block size is 256KiB. Default buffer size is 4KiB, so saving 2MiB wallet with my example relatively would generate 8 operations instead of 512.

core/src/main/java/org/bitcoinj/wallet/WalletProtobufSerializer.java
```diff
@@ -114,14 +116,27 @@ public void setRequireMandatoryExtensions(boolean value) {
...
     /**
      * Formats the given wallet (transactions and keys) to the given output stream in protocol buffer format.<p>
      *
      * Equivalent to <tt>walletToProto(wallet).writeTo(output);</tt>
      */
     public void writeWallet(Wallet wallet, OutputStream output) throws IOException {
         Protos.Wallet walletProto = walletToProto(wallet);
-        walletProto.writeTo(output);
+
+        int serializedSize = walletProto.getSerializedSize();
```
schildbach: Wouldn't this forced call slow down saving considerably? Do we need the minimum calculation? It doesn't hurt to pick a larger buffer, no?

core/src/main/java/org/bitcoinj/wallet/WalletProtobufSerializer.java
```diff
@@ -114,14 +116,27 @@ public void setRequireMandatoryExtensions(boolean value) {
...
+        int bufferSize = serializedSize > this.walletWriteBufferSize ? this.walletWriteBufferSize : serializedSize;
```
schildbach: If this line stays, I'd prefer `Math.min(...)`.

schildbach:
Thanks, looks much better now. I added some inline comments.

core/src/main/java/org/bitcoinj/wallet/WalletProtobufSerializer.java
jdojff: I took all the code from original method which was called in ```com.google.protobuf.AbstractMessageLite```. Now tried with fixed buffer size, and it works as well. Will commit update

jdojff:
Updated

schildbach:
I squashed your commits and merged. Thanks!

We could add a second property like "autoAdjustWalletWriteBufferSize" which after a successful write sets the buffer size to the size of the protobuf (plus a little bit of space for one of two additional transactions). So all the repetitive writes that only adjust the confidences would use just the right amount of buffer. What do you think?

PR #1244, opened 2016-04-20
Payment channels: ClientChannelProperties ServerChannelProperties

cyberzac:
To allow for configurations of the payment channels without adding more parameters to the constructor, one can use the ClientChannelProperties and ServerChannelProperties. 
This change is breaking code using the non default constructors of the PaymentChannelClient, PaymentChannelClientConnection, PaymentChannelServer. 
To ensure that the interfaces can evolve without breaking future code, there are two default implementations DefaultClientChannelProperties, DefaultServerChanelProperties that uses the current default. See ExamplePaymentChannelClient for an example of how to use the DefaultClientChannelProperties.

### DIFF
--- a/core/src/main/java/org/bitcoinj/protocols/channels/IPaymentChannelClient.java
+++ b/core/src/main/java/org/bitcoinj/protocols/channels/IPaymentChannelClient.java
@@ -23,6 +23,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 
 import com.google.protobuf.ByteString;
 import org.bitcoin.paymentchannel.Protos;
+import org.bitcoinj.wallet.SendRequest;
 import org.spongycastle.crypto.params.KeyParameter;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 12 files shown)


schildbach:
Looks reasonable. @w-shackleton , what do you think about this change?

cyberzac:
Any progress?

cyberzac:
Ping...

schildbach:
Do you think you could rebase this on current master? Would be nice to get a comment of @w-shackleton but I will merge without too.

cyberzac:
Will do.
tors 11 aug. 2016 kl. 22:34 skrev Andreas Schildbach <
notifications@github.com>:

w-shackleton:
This looks good to me - I'm mobile-only this week but I can have a deeper
look next week if required. Otherwise this definitely cleans things up, and
removes the need for subclassing the state objects which is now more
complicated with two protocol versions.
Looks good from mobile!

cyberzac:
Rebased.

cyberzac:
Rebased.

Regards

Martin

--
Martin Zachrison
Prinsessvägen 3A        callto:+46 8 556 333 03
131 46 Nacka              mobile:+46 768 04 4747

schildbach:
Merged.

PR #1249, opened 2016-04-26
Copyright harmonization

schildbach:
Add a generic copyright line to all license headers that don't have one. Note it is not legally necessary to state copyright, but we think it makes things easier.

Replace "The bitcoinj developers" copyright statement by the generic statement. There is no such legal entitiy.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Context.java
+++ b/core/src/main/java/org/bitcoinj/core/Context.java
@@ -1,4 +1,6 @@
 /*
+ * Copyright by the original author or authors.
+ * 
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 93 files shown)


TheBlueMatt:
I actually think some of those files are not copyright the original author(s), but are instead copyright Google/their employers.

schildbach:
@TheBlueMatt Thanks for the hint. It's hard to tell which ones since neither the respective author nor Google claimed the copyright. By default copyright goes to the author, which is why I'm proposing these changes.

TheBlueMatt:
I mean it's technically a factually inaccurate statement... I have seen vacuous statements like "copyright the respective copyright holders" used before.

Matt

Issue #1260, opened 2016-05-18
H2FullPrunedBlockStore constructor creates file with incorrect path


dcw312:
The following constructor:
H2FullPrunedBlockStore(NetworkParameters params, String dbName, int fullStoreDepth, int cacheSize)

... will prepend "jdbc:h2:" twice to while creating the connection URL.

I will prepare a pull request.

Steps to reproduce:
1. Call the constructor in an new Unit Test with dbName = "/tmp/foo.h2"

Expected Result: new file with absolute path /tmp/foo.h2
Actual Result: new file in current directory with name "jdbc:h2"

schildbach:
Fixed.

PR #1261, opened 2016-05-18
Issue 1260: the h2 prefix is prepended twice using this constructor. …

dcw312:
#1260 …Only prepend when calling the super constructor which happens in line 92.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -429,7 +429,7 @@
         <dependency>
             <groupId>org.bitcoinj</groupId>
             <artifactId>orchid</artifactId>
-            <version>1.2</version>
+            <version>1.2.1</version>
         </dependency>
         <dependency>
             <groupId>com.squareup.okhttp</groupId>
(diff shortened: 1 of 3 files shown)


dcw312:
The travis-ci build seems to have failed for an independent reason. It could not obtain the orchid jar.

schildbach:
Thanks for finding this. The connection parameters seem to be duplicated too, would you like to fix that as well?

I just fixed the build error, if you rebase your commit on current master it should be good.

dcw312:
Regarding duplicate params, I won't have time for a few days to address that.  Please merge this request because the file name issue is a more significant bug and I'd like to prevent anyone else from duplicating the work. Thanks!

I just rebased. Incidentally, these instructions were very helpful: http://stackoverflow.com/a/7244456/205736

schildbach:
Don't worry, I just merged (and fixed the duplicate params in the same go).

PR #1266, opened 2016-05-25
Fix wrong balance calculation if identical (apart from the index)…

schildbach:
…outputs are involved.

This regression was introduced in a53b508049d0febf59dfd3f7e5d75776a0b3f7a8.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionOutput.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionOutput.java
@@ -419,8 +419,8 @@ public class TransactionOutput extends ChildMessage {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         TransactionOutput other = (TransactionOutput) o;
-        return value == other.value && (parent == null || parent == other.parent)
-            && Arrays.equals(scriptBytes, other.scriptBytes);
+        return value == other.value && (parent == null || (parent == other.parent && getIndex() == other.getIndex()))
+                && Arrays.equals(scriptBytes, other.scriptBytes);
     }
 
     @Override
--- a/core/src/test/java/org/bitcoinj/wallet/WalletTest.java
+++ b/core/src/test/java/org/bitcoinj/wallet/WalletTest.java
@@ -545,6 +545,16 @@ public class WalletTest extends TestWithWallet {
         assertEquals(v4, wallet.getBalance(Wallet.BalanceType.AVAILABLE));
     }
 
+    @Test
+    public void balanceWithIdenticalOutputs() {
+        assertEquals(Coin.ZERO, wallet.getBalance(BalanceType.ESTIMATED));
+        Transaction tx = new Transaction(PARAMS);
+        tx.addOutput(Coin.COIN, myAddress);
+        tx.addOutput(Coin.COIN, myAddress); // identical to the above
+        wallet.addWalletTransaction(new WalletTransaction(Pool.UNSPENT, tx));
+        assertEquals(Coin.COIN.plus(Coin.COIN), wallet.getBalance(BalanceType.ESTIMATED));
+    }
+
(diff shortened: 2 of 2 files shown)


schildbach:
@amichair What do you think of this one?

amichair:
Are you sure this regression was introduced in a53b508?

The Object.equals/hashCode contract is that equal objects must have an equal hashCode (whereas unequal objects may or may not have equal hashCodes). That commit removed the index from TransactionOutput's hashCode, since it was not used in determining equality, i.e. instances that are identical other than the index would be equal but with different hashCode, which violates the contract.

Also, making a hashCode less specific (i.e. incorporating fewer instance fields) should always remain correct. It can even return a constant value and still be correct (although inefficient). So even if equals() did check the index, removing the index from hashCode would still be correct.

I haven't checked but I suspect either this is not the source of the bug, or some other code somewhere is dependent on a broken hashCode implementation and should be fixed. Or I missed something :-)

In any case, a regression unit test would be useful in testing and preventing such issues...

schildbach:
Yes, I'm pretty sure. I tried simply undoing the change and it fixed the issue. What you say about equals/hashCode is not entirely true: HashSet doesn't care about equals afaik, which is why it worked before. However I agree with you that the equals/hashCode contract should be respected, which is why I added the index to equals as well in this PR.

You're also right about the regression unit test, I'll try to come up with one.

schildbach:
I added a small unit test showing the issue.

amichair:
I'm pretty sure all collections use equals - how else would they find and identify an element? I actually just checked and it's mentioned explicitly in the Collection interface javadoc. There is IdentityHashMap which uses reference equality instead of equals, but that's a special case.

btw did you revert the whole commit when you tested this, or only this class? Perhaps a change in some other class in the same commit was involved? (not likely since the fix is in this class, but still... I don't get what exactly got broken here).

amichair:
Ok, a git bisect based on your test method indeed pointed at the commit, and a manual change of only the "+ getIndex()" confirms the test passes with it.

Here's my analysis:

The code had a logical bug, which is that equals() did not account for the index, so two outputs with the same value and address (as created by Transaction.addOutput()) would be considered equal. So in theory (as far as the Collection/Set contract guarantees), only one of them could be added to a Set.

The code had another technical bug, which is breaking the equals/hashcode contract byt using the index in hashCode. By coincidence (?), since they are added to a HashSet, backed by a HashMap, whose standard implementation happens to check hash equality before instance equality (for performance reasons), this hid the first bug. If at some point the HashSet was replaced with a different kind of collection, or its internal implementation would change, or it would run on a different JRE... it could have broken in various subtle ways. Interestingly, the Set.add() method contract is broken if equals/hashCode contract for the elements is broken, as happens in this case, i.e. its documented behavior no longer works (see javadoc). Thus the general warning not to break the equals/hashCode contract :-)

The proper fix is indeed to add the index to equals, like in this PR. Whether it is added also to hashCode doesn't matter, as I explained earlier (if it keeps the code simpler, you can remove it). Also, after the change, you have to make sure that in all places where TransactionOutputs are compared for equality (including all uses in other collections), that the new semantics are correct, i.e. transaction outputs who have a parent should always be considered different if they are at a different index.

I hope this helps :-)

schildbach:
Thanks for your review! Ok, I took out the getIndex() from hashCode(). I also changed the commit comment a bit to clarify that the referred commit didn't really break the code.

If there are no objections, I plan to merge this on Monday.

dcw312:
Did you consider using the apache commons utilities (EqualsBuilder, HashCodeBuilder)?

https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/builder/EqualsBuilder.html

schildbach:
@dcw312 We're using what comes with Guava instead.

schildbach:
Merged.

PR #1267, opened 2016-05-27
Improve BuildCheckpoints tool to use a DNS peer group

dcw312:
The tool did not work as expected for these parameters
--net=TEST --days=1

Expected behavior: Check point files are created from the test network
Actual behavior: main method hangs trying to connect to localhost

This change will use DNS discovery for the MainNetParams and TestNet3Params. With this change, two successful checkpoint files are created for TEST.

With this change two files are created for PROD as well.

I could not test REGTEST because I don't know the required setup for that. The new code should not run when a peer is specified. 

Please let me know if you have any questions.

PR #1268, opened 2016-06-03
Use a peer group for the Prod and Test networks.

dcw312:
Squashed earlier pull request into one commit. This improves the BuildCheckpoints tool to use a peer group for prod and test.

The earlier pull request was https://github.com/bitcoinj/bitcoinj/pull/1267

### DIFF
--- a/tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
+++ b/tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
@@ -19,6 +19,7 @@ package org.bitcoinj.tools;
 
 import org.bitcoinj.core.listeners.NewBestBlockListener;
 import org.bitcoinj.core.*;
+import org.bitcoinj.net.discovery.DnsDiscovery;
 import org.bitcoinj.params.MainNetParams;
 import org.bitcoinj.params.RegTestParams;
 import org.bitcoinj.params.TestNet3Params;
@@ -44,10 +45,11 @@ import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.security.DigestOutputStream;
 import java.security.MessageDigest;
-import java.util.Date;
-import java.util.TreeMap;
+import java.util.*;
+import java.util.concurrent.Future;
 
 import static com.google.common.base.Preconditions.checkState;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
 /**
  * Downloads and verifies a full chain from your local peer, emitting checkpoints at each difficulty transition period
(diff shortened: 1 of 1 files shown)


tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
```diff
@@ -90,31 +98,44 @@ public static void main(String[] args) throws Exception {
...
             String peerFlag = (String) options.valueOf("peer");
             try {
                 ipAddress = InetAddress.getByName(peerFlag);
+                startPeerGroup(peerGroup, ipAddress);
             } catch (UnknownHostException e) {
                 System.err.println("Could not understand peer domain name/IP address: " + peerFlag + ": " + e.getMessage());
                 System.exit(1);
                 return;
             }
+        } else if (dnsNetworks.contains(params.getClass())) {
```
schildbach: If you want to know if a given network has DNS seeds, why don't you use params.getDnsSeeds(). It will return null if it doesn't support DNS.

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
```diff
@@ -90,31 +98,44 @@ public static void main(String[] args) throws Exception {
...
+            //for PROD and TEST use a peer group discovered with dns
+            peerGroup.setUserAgent("PeerMonitor", "1.0");
+            peerGroup.setMaxConnections(20);
+            peerGroup.addPeerDiscovery(new DnsDiscovery(params));
+            peerGroup.start();
+            Future<List<Peer>> future = peerGroup.waitForPeers(4);
```
schildbach: Do we really need 4 peers? Block chain download happens only on 1 anyway.

dcw312: My assumption is that all peers may not support download. Connecting to 4 peers increases the probability that one supports download, correct?

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
dcw312: I'll change it, thanks!

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
schildbach: Ok, fair enough.

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
```diff
@@ -90,31 +92,50 @@ public static void main(String[] args) throws Exception {
...
+        // Configure bitcoinj to fetch only headers, not save them to disk, connect to a local fully synced/validated
+        // node and to save block headers that are on interval boundaries, as long as they are <1 month old.
+        final BlockStore store = new MemoryBlockStore(params);
+        final BlockChain chain = new BlockChain(params, store);
+        final PeerGroup peerGroup = new PeerGroup(params, chain);
+
         final InetAddress ipAddress;
+
+        // DNS discovery can be used for some networks
+        boolean networkHasDnsSeeds = params.getDnsSeeds() != null;
```
dcw312: @schildbach As requested, the method now uses params.getDnsSeeds(). I used a local variable to make the code readable and added comments to all the if/else conditions.

schildbach:
Merged, thanks!

PR #1270, opened 2016-06-06
Making OP_RETURN to have 80 bytes.

johnygeorge:
Making OP_RETURN to have 80 bytes instead of the previous 40 bytes.

This is to be consistent with the bitcoin-core, which have made it back to 80 bytes.


schildbach:
Thanks for your contribution! I think it would be nice to have at least one of the tests try using all available bytes.

luke-jr:
Really shouldn't have any arbitrary limit here, as it's node policy what each node will accept.

schildbach:
@luke-jr It's not arbitrary. Exceeding the limit won't relay.

PR #1271, opened 2016-06-08
Wallet: Fix precondition for transaction confidence of transactions…

schildbach:
…in the pending pool. They can now also be IN_CONFLICT.


schildbach:
@oscarguindzberg Do you agree with this bugfix?

oscarguindzberg:
Yes, I agree.

schildbach:
Thanks for the review! I merged.

PR #1281, opened 2016-08-06
Check if open transaction exists before inserting (Postgresql-specific)

paulminer:
Postgresql invalidates the entire transaction if a duplicate key exception is thrown, so an extra check to see if the open output exists is done prior to inserting. The put(...) method already does this, but addUnspentTransactionOutput(...) did not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,3 +56,4 @@ more info on this technique.
 ### Where next?
 
 Now you are ready to [follow the tutorial](https://bitcoinj.github.io/getting-started).
+
--- a/core/src/main/java/org/bitcoinj/store/DatabaseFullPrunedBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/DatabaseFullPrunedBlockStore.java
@@ -521,35 +521,47 @@ public abstract class DatabaseFullPrunedBlockStore implements FullPrunedBlockSto
      * @throws BlockStoreException If the block store could not be created.
      */
     private void createTables() throws SQLException, BlockStoreException {
-        Statement s = conn.get().createStatement();
-        // create all the database tables
-        for (String sql : getCreateTablesSQL()) {
-            if (log.isDebugEnabled()) {
-                log.debug("DatabaseFullPrunedBlockStore : CREATE table [SQL= {0}]", sql);
+        Statement s = null;
+        try {
+            s = conn.get().createStatement();
+            // create all the database tables
+            for (String sql : getCreateTablesSQL()) {
+                if (log.isDebugEnabled()) {
+                    log.debug("DatabaseFullPrunedBlockStore : CREATE table [SQL= {0}]", sql);
+                }
+                s.executeUpdate(sql);
             }
-            s.executeUpdate(sql);
-        }
-        // create all the database indexes
-        for (String sql : getCreateIndexesSQL()) {
-            if (log.isDebugEnabled()) {
-                log.debug("DatabaseFullPrunedBlockStore : CREATE index [SQL= {0}]", sql);
+            // create all the database indexes
+            for (String sql : getCreateIndexesSQL()) {
+                if (log.isDebugEnabled()) {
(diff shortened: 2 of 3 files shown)


schildbach:
@kparmar1 You were the last to work on the block stores, would you like to comment on this PR?

schildbach:
Thanks for your contributions. Just one request, can you keep separate issues in separate PRs?

core/src/main/java/org/bitcoinj/core/ECKey.java
```diff
@@ -705,7 +705,7 @@ public static boolean verify(byte[] data, ECDSASignature signature, byte[] pub)
 
         if (Secp256k1Context.isEnabled()) {
             try {
-                return NativeSecp256k1.verify(data, signature.encodeToDER(), pub);
```
schildbach: @greenaddress You were the last who touched native code signature verification. What do you think of these changes?

core/src/main/java/org/bitcoinj/core/ECKey.java
```diff
@@ -733,14 +733,6 @@ public static boolean verify(byte[] data, ECDSASignature signature, byte[] pub)
      * @param pub       The public key bytes to use.
      */
     public static boolean verify(byte[] data, byte[] signature, byte[] pub) {
-        if (Secp256k1Context.isEnabled()) {
```
schildbach: @greenaddress …and this?

paulminer:
> Just one request, can you keep separate issues in separate PRs?

Sure, but I could use a little guidance here (I've not really used GitHub before). I've been searching around on the topic, and it seems that whenever I want to make a change, I should create a branch for it first so that the branch can be used in a PR?

Is there anything you suggest I do to clean this PR up, or is it too late?

schildbach:
You can clean up pull requests by force-pushing over their github branch,

I'd say the ECKey change belongs into a different PR. Create a new branch off master, cherry-pick the commit there and create a new PR of the new branch. Then we can discuss and merge that PR independently – it's a change that needs to be reviewed by a couple of people ideally.

If you've done with that, I can help you clean up this PR (hint: interactive rebase is your friend).

paulminer:
I think I have the ECKey commit pulled into a separate branch, but I don't really understand what I need to do next. I also merged from this repository at some point, which might make things uglier.

schildbach:
The branch for the ECKey change should really just be master + 1 commit on top. You simply push it to GitHub and create the PR from the web interface, just like you did with _this_ PR.

schildbach:
Ok the next easy thing you can do is get rid of the merge commit. I think you can do an interactive rebase (git rebase -i master) and remove the line, then force-push the branch again.

Generally if you work with the bitcoinj codebase you should never merge, always use rebase. We have a policy to not allow merge commits into our master and release branches, because they make it hard to follow history and in the end we don't care about where changes stem from, we only care what has been changed.

paulminer:
Okay, I think I've got it? (this is so much different than SVN...)

I used interactive rebase to remove all the stuff merged from bitcoinj/bitcoinj, added an upstream, fetched the upstream, rebased from upstream/master, and finished with a push -f. I can't figure out how to view commit history on github, but git log seems to show what I think should be there?

paulminer:
Btw, is interactive rebase like using vim? I didn't quite understand what was going on, but it felt just like using vi. I also head to specify HEAD~20 to get anything to show up.

paulminer:
This last commit was to fix a bug I introduced when I added Postgresql ON CONFLICT support. The root problem is that the superclass constructor calls an overridden method in a subclass, before the subclass constructor has executed and initialized a final field that I had added. I didn't see a good way of correcting that since the constructor wants to insert the genesis block if the database is new, so the workaround was to check if the final field had been initialized yet and default to not supporting ON CONFLICT.

msgilligan:
Closing as we have removed the Postgresql classes.

PR #1282, opened 2016-08-09
Check that getCountAtOrAbove is not null, to avoid null pointer exception

greglaun:
tally.getCountAtOrAbove(Block.BLOCK_VERSION_BIP65) is compared against getMajorityEnforceBlockUpgrade() without checking for nullity. This is an issue since it is initialized to null in utils/VersionTally.java if the version window is not full. This resulted in a null pointer exception in regtest mode that prevented downloading of the blockchain.


schildbach:
@greglaun I think it would be nicer to cache the result of tally.getCountAtOrAbove() in a local variable.

@rnicoll What do you think about this PR?

rnicoll:
Ah, I'd assumed live/test networks (where this could never happen), hadn't thought of regtest. Seconded on the caching the result, but otherwise looks good.

PR #1283, opened 2016-08-12
Bugfixes for signature verification with native secp256k1 library

paulminer:
Calling verify(byte[], ECDSASignature, byte[]) with the native secp256k1 library would fail on non-canonicalized (not low-S) signatures. The Java counterpart ECDSASigner.verifySignature() allows non-canonical signatures. When relaying and mining, non-canonical signatures are not allowed, but this is checked for in TransactionSignature.decodeFromBitcoin(). The fix is to canonicalize the signature.

Calling verify(byte[], byte[], byte[]) with the native secp256k1 library would ignore the FAKE_SIGNATURES flag. Since this method also needs to allow low-S signatures, it has been modified to instead chain to verify(byte[], ECDSASignature, byte[]).

To improve native performance, it would be best if the canonicalizing happened in the native library via secp256k1_ecdsa_signature_normalize(ctx, &sig, &sig). However, it would require an API change in both bitcoinj and secp256k1 because the expected behavior in secp256k1 is to fail non-normalized signatures (this is evident in their testing suite). The API change could be either a normalization flag passed to the verify method, or a new method (normalize a byte[] signature).

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/core/ECKey.java
@@ -705,7 +705,7 @@ public class ECKey implements EncryptableItem {
 
         if (Secp256k1Context.isEnabled()) {
             try {
-                return NativeSecp256k1.verify(data, signature.encodeToDER(), pub);
+                return NativeSecp256k1.verify(data, signature.toCanonicalised().encodeToDER(), pub);
             } catch (NativeSecp256k1Util.AssertFailException e) {
                 log.error("Caught AssertFailException inside secp256k1", e);
                 return false;
@@ -733,14 +733,6 @@ public class ECKey implements EncryptableItem {
      * @param pub       The public key bytes to use.
      */
     public static boolean verify(byte[] data, byte[] signature, byte[] pub) {
-        if (Secp256k1Context.isEnabled()) {
-            try {
-                return NativeSecp256k1.verify(data, signature, pub);
-            } catch (NativeSecp256k1Util.AssertFailException e) {
-                log.error("Caught AssertFailException inside secp256k1", e);
-                return false;
-            }
-        }
         return verify(data, ECDSASignature.decodeFromDER(signature), pub);
     }
 


schildbach:
Thanks, this PR looks good from the branching point of view. @greenaddress What do you think? I never used the native code adapter so I'd love to hear your opinion.

greenaddress:
@schildbach Thanks for the ping

if there is asymmetry between using bitcoinj with and without secp256k1 for non low-s signatures then I agree something should be done because it wouldn't be correct.

However I am not convinced a conversion from bytes to ECDSASignature and then canonicalised and back to bytes is a performance penalty worth taking, I think the fix should be in https://github.com/bitcoin-core/secp256k1/blob/master/src/java/org_bitcoin_NativeSecp256k1.c and it should be quite cheap to do all the time but i still think it's worthwhile supporting both options (this including current behaviour, with a flag or a different method for the new behaviour that bitcoinj will call into)

@paulminer I'll be happy to review the changes if you want to update the bindings in the other repo

This link may be relevant re low S: https://github.com/bitcoin/bips/pull/433

paulminer:
So the two options would be:
1. Update org_bitcoin_NativeSecp256k1.c to normalize the signature via secp256k1_ecdsa_signature_normalize(...).
2. OR, create a new native binding, modifying both bitcoinj and secp256k1. Specifically:
* secp256k1_ecdsa_verify(ByteBuffer byteBuff, long context, int sigLen, int pubLen**, boolean normalizeSignature**)
* The original method would then call the new method with a default value for normalizeSignature.

The second option allows for both behaviors, but requires updating both libraries. @greenaddress  What do you think should be done?

Also regarding low-s: https://bitcoin.org/en/release/v0.11.1
>Make the node require the canonical ‘low-s’ encoding for ECDSA signatures when relaying or mining. This removes a nuisance malleability vector.
>
>Consensus behavior is unchanged.

As I understand it, this means at present low-S is valid in a block, but the clients don't have to relay such transactions or include them in the blocks they build (the behavior is heavily discouraged but technically allowed). In bitcoinj, I believe TransactionSignature.decodeFromBitcoin(...) is what currently checks for low-S, but it throws an exception if a violation is found, since ECDSASigner.verifySignature(...) allows low-S. I don't know that it matches core's beahvior.

In order to handle BIP-146, it looks like MainNetParams.getBlockVerificationFlags(...) would need to be updated to include VerifyFlag.LOW_S depending on the block number at which BIP-146 was enforced.

paulminer:
I should also add, normalizing the signature in Java amounts to a BigInteger comparison and subtraction, nothing too heavy (comparison to check for high-S, subtraction to normalize). I chose to modify ECKey because if the bindings present in NativeSecp256k1 are supposed to reflect secp256k1, then they shouldn't do normalization themselves (which in my opinion, pushes things towards option 2, because option 1 would create an asymmetry between NativeSecp256k1.secp256k1_ecdsa_verify(...) and secp256k1's secp256k1_ecdsa_verify(...)). Option 2 would at least create a new function so as to preserve the behavior of the existing function.

msgilligan:
We have removed the JNI implementation of secp256k1 and will be replacing it with the new [secp256k1-jdk](https://github.com/bitcoinj/secp256k1-jdk) -- See Issue #3389.

PR #1284, opened 2016-08-15
Expose the DeterministicKey ctors used by serialisation

jgriffiths:
These ctors are the only way to construct a key without a parent
but with a depth that matches the path passed in. The ctors taking
parents set the depth to 0 when parent is null. Whether this is
desirable is not clear; exposing the other ctors seems a lower risk
change given the comments in this file that detail support for depth
not matching the path length.

The use case for this change is interfacing with APIs that allow
derivation of children from a path without returning the intermediate
keys. In this case we have no parent to pass to the ctor but we want
to create a correct DeterministicKey for interop with the rest of bitcoinj.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -130,7 +130,7 @@ public class DeterministicKey extends ECKey {
      * information about its parent key.  Invoked when deserializing, but otherwise not something that
      * you normally should use.
      */
-    private DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
+    public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
                             byte[] chainCode,
                             LazyECPoint publicAsPoint,
                             @Nullable DeterministicKey parent,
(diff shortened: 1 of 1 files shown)


schildbach:
Do you think you could provide a unit test for your use case?

@mackler Do you have an opinion on this PR?

jgriffiths:
A realistic unit test of this specific use-case isn't possible without importing a library that generates keys via paths. The ctors I'm exposing should already be tested in any case, as they are used by the key serialisation code. All this change does is make them available to external bitcoinj users.

mackler:
All we're doing is publicizing two private `DeterministicKey` constructors?  It seems unlikely to break anything unless there are other classes with those names somewhere, which is information I lack.

Other than that, the appropriateness of merging seems to depend more on the project's expectation that those constructors will continue to be available, since once they're public others may start depending on them.  Again, information I lack.

Those are just my immediate thoughts that should apply to the making of any private methods public and are hardly rare.

jgriffiths:
@mackler There are already public ctors available for DeterministicKey so there is no risk of name clashes. The ctors are used for serialisation which stores depth in addition to the keys path since they are now allowed to diverge (depth may be greater than path length if the user wants to hide part of the derivation path). Since this invariant is encoded in the serialised format it is difficult for me to imagine a scenario in which the ctors could be removed in the future - this code will always need to construct keys with an explicit depth AFAICS.

An alternative approach would have been to change the existing public ctors that take parents but not depth to compute the depth as the length of the passed chain instead of 0 when the parent passed is null. This would provide the same functionality for my use case since you need the full path to derive the key using an external path based api. In general the invariant that path length == depth holds true everywhere except for the special cases mentioned in the source. However there are no guarantees that current users correctly pass an empty path when passing a null parent. For this reason I perceive changing the existing ctor as risky whereas exposing the ctors that take explicit depth as safe.

mackler:
Clearly @jgriffiths has thought about this more than I have or will in the near future.  I have little of value to add.

schildbach:
Merged.

PR #1285, opened 2016-08-16
Remove Context requirement from Transaction.toString()

ewiner:
For me, by far the most common cause of `Content.get()`'s "Performing thread fixup" warnings is from when I call `Transaction.toString()` from a thread outside of BitcoinJ's control.

It doesn't seem like it's really necessary for the `toString()` to look something up in global state, so this change just avoids printing out the transaction confidence if it can't be looked up without a warning.

If you prefer, I'm happy to also do the simpler change of just removing confidence entirely from this `toString()`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -636,8 +636,6 @@ public class Transaction extends ChildMessage {
     public String toString(@Nullable AbstractBlockChain chain) {
         StringBuilder s = new StringBuilder();
         s.append("  ").append(getHashAsString()).append('\n');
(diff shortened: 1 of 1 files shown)


schildbach:
Actually I think we should indeed remove confidence from toString() entirely.

ewiner:
Great - this is now a two-line PR.

schildbach:
Thanks! I merged, but added a commit on top that adds confidence back to Wallet.toString(). I think when working with a Wallet, it's clear you need to use the Context object as well. I hope this doesn't contradict with your usecase.

ewiner:
Yep, that makes sense.  Thanks for the merge.

PR #1286, opened 2016-08-24
Fix download progress tracking

jgriffiths:
Currently progress() can be called after doneDownload() which should not
happen. Also make sure that we call progress() with 100% progress before we
call doneDownload() - this is in case any existing users have given up on
downDownload() being called at the right time and instead use 100% progress
as their signal that downloading is complete.


schildbach:
Thanks for catching this. Could you maybe add a unit test that catches this case?

jgriffiths:
Sure, I'll look at adding a test this weekend.

jparcus34:
I need my bit coins

schildbach:
I just merged this finally.

PR #1287, opened 2016-08-24
Make UTXO serializable

oscarguindzberg:
I see https://github.com/bitcoinj/bitcoinj/commit/783dfc58d1e7802da9b46daa3756a7bb57dc25eb removed java serialization from UTXO.
I fixed UTXO serialization and added a test

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/UTXO.java
+++ b/core/src/main/java/org/bitcoinj/core/UTXO.java
@@ -30,7 +30,9 @@ import java.util.Locale;
  * It avoids having to store the entire parentTransaction just to get the hash and index.
  * Useful when working with free standing outputs.
  */
-public class UTXO {
+public class UTXO implements Serializable {
+
+    private static final long serialVersionUID = 4736241649298988166L;
 
     private Coin value;
     private Script script;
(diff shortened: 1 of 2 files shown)


schildbach:
As you can see from the linked commit, the decision to remove support for Java serialization was deliberate. Is there a specific reason why you want UTXOs to be serializable?

oscarguindzberg:
I need to serialize them on a piece of code I am writing.

At RSK we are writing a bitcoin sidechain.
In order to implement it we use a "wallet" with just 1 address (the lock address).
We used to use the standard bitcoinj wallet (with standard protobuf serialization) but as more txs came in and out, serialization (and deserialization) was slower and produced a bigger byte[]. 
We need to deserialize and serialize the wallet very often (probably several times per second). Also we had to do some hacks to make serialization deterministic. 
Then we switched to a wallet with a UTXO provider and we persist just the UTXOs.
We are using java serialization to persist them. Protobuf serialization is not ready for UTXOs, so java serialization was the easiest thing to do. We may change that in the future for protubuf or something else.
I read in the commit log the reason to remove java serialization was because it was not tested, not working and not used. I think all that has changed for UTXO serialization. 

I might send other PR related to UTXO provider based wallets in the near future.

schildbach:
Thanks for the explanation. I merged. But be aware Java serialization remains kind of on the hotseat…

Issue #1259, opened 2016-05-17
More reasons Wallet stops adding new blocks to chain on Testnet


jarlfr:
Here below are some more reasons I found when analyzing that will cause bitcoinj wallet get stuck and stop following the correct block chain on Testnet.  We experienced a lot of problems like this on testnet and my wish was to fix them or rule out that these things could happen in production on mainnet.

I present some ideas for action but more thoughts are needed. Testnet forks can be very long, to the degree it will run beyond what can be handled with SPVBlockStore, so we need to think about this. A solution for a very low spec (memory/disk) wallet could maybe do a two-pass download from peers.  (I am unsure but does bitcoinj make use getheaders in normal catch up) 

##### Peers do not recognize hashes in `getblock`-message

The hashes sent in the getblock message should help a peer to decide which blocks the client needs to catch up with the chain.  I noticed that Peers response to getblocks was to send the 500 first blocks of the block chain. These are not to much use of course. It will not change the state of of the chain head in the wallet's store so the wallet will be stuck. 

Analysis: The bitcoinj seams to fill in 100 hashes starting from its chain head in a linear fashion, and if all are on a fork that was discarded, the Peer cannot find any common block except for the genesis block, thus it starts there in the reply inv-message.

Action: Use better set of hashes from the known blocks in the store (5000 for SPVBlockStore). A better selection is proposed on the bitcoin wiki: "dense to start, but then sparse".  This helps but I ran into the next problem:

##### `getdata` 500 blocks do not trigger re-organize despite head on a dead fork

Requesting blocks with better hashes can still leave the the wallet store chain head unchanged. This will result in the same request for blocks again, and the store head is effectively stuck. 

Analysis: This happens when the downloaded blocks, despite belong to the correct chain, will not trigger a re-organize despite the head is a dead chain.  Why? It seems the special difficulty jumps on testnet can make a branch of blocks have less total work despite being very much longer, e.g. several 100 blocks longer than head.  For some reason the network selected this longer chain for many blocks.

[...]

PR #1292, opened 2016-08-27
Delegate the process of creating the block locator related to 1259

dcw312:
This is a refactor to abstract the logic of creating the block locator from the peer. This will allow for development of a "dense at first, then sparse" implementation. This does not directly address issue #1259 but removes a "todo" from the code.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -19,9 +19,9 @@ package org.bitcoinj.core;
 import com.google.common.base.*;
 import com.google.common.base.Objects;
 import org.bitcoinj.core.listeners.*;
+import org.bitcoinj.core.strategies.BlockLocatorStrategy;
+import org.bitcoinj.core.strategies.LinearBlockLocatorStrategy;
 import org.bitcoinj.net.StreamConnection;
-import org.bitcoinj.store.BlockStore;
-import org.bitcoinj.store.BlockStoreException;
 import org.bitcoinj.utils.ListenerRegistration;
 import org.bitcoinj.utils.Threading;
 import org.bitcoinj.wallet.Wallet;
@@ -78,6 +78,7 @@ public class Peer extends PeerSocketHandler {
         = new CopyOnWriteArrayList<ListenerRegistration<PreMessageReceivedEventListener>>();
     private final CopyOnWriteArrayList<ListenerRegistration<OnTransactionBroadcastListener>> onTransactionEventListeners
         = new CopyOnWriteArrayList<ListenerRegistration<OnTransactionBroadcastListener>>();
+    private BlockLocatorStrategy blockLocatorStrategy;
     // Whether to try and download blocks and transactions from this peer. Set to false by PeerGroup if not the
     // primary peer. This is to avoid redundant work and concurrency problems with downloading the same chain
     // in parallel.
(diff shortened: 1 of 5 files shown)


schildbach:
Concept looks good to me. I'll leave this open for a few days so that others can comment.

If you do plan to implement a "dense first, then sparse" implementation, feel free to go ahead and add it as a second commit. It's always good to see what changes are good for.

core/src/test/java/org/bitcoinj/core/strategies/DenseThenSparseStrategyTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+        Block b1 = createFakeBlock(blockStore, Block.BLOCK_HEIGHT_GENESIS).block;
+        blockChain.add(b1);
+
+        for (int i=blockChain.getBestChainHeight() + 1; i<100; i++) {
+            try {
+                Block b = createFakeBlock(blockStore, i).block;
+                blockChain.add(b);
+            } catch (Exception e) {
+                log.error("fails at i {}", i, e);
```
dcw312: The test set-up does not complete. What am I doing wrong?

SEVERE: fails at i 10
org.bitcoinj.core.VerificationException: Could not verify block:
Caused by: org.bitcoinj.core.VerificationException: Network provided difficulty bits do not match what was calculated: 201fffff vs 207fffff
	at org.bitcoinj.params.AbstractBitcoinNetParams.checkDifficultyTransitions(AbstractBitcoinNetParams.java:135)
	at org.bitcoinj.core.AbstractBlockChain.add(AbstractBlockChain.java:488)

dcw312:
This is not ready for merge after my last commit. I need help with the test set-up in the last commit. Can you please look at my comment on setting up the test?

core/src/test/java/org/bitcoinj/core/strategies/DenseThenSparseStrategyTest.java
schildbach: The unit test params have a difficulty interval of 10. It looks like Block.createNextBlock() doesn't take that into account (yet).

If it's an option to you, try running the test with TestNetParams/MainNetParams.

core/src/main/java/org/bitcoinj/core/strategies/DenseThenSparseStrategy.java
```diff
@@ -0,0 +1,82 @@
...
+    public List<Sha256Hash> createBlockLocator(AbstractBlockChain chain) {
+
+        //make a list of the block heights that we want to send
+        int height = chain.getBestChainHeight();
+        Set<Integer> indexes = new HashSet<Integer>();
+        int step = 1;
+        int minIndex = 0;
+        int maxDepth = getMaxDepth();
+        // stop once we reach a maximum depth
+        for (Integer index = height;index > 0 && (height - index) < maxDepth; index = index - step) {
```
dcw312: currently, the locator list will not go any deeper than the current implementation (100 deep).  Can be overridden in a sub class.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -226,6 +227,7 @@ public Peer(NetworkParameters params, VersionMessage ver, PeerAddress remoteAddr
                 @Nullable AbstractBlockChain chain, int downloadTxDependencyDepth) {
         super(params, remoteAddress);
         this.params = Preconditions.checkNotNull(params);
+        this.setBlockLocatorStrategy(new LinearBlockLocatorStrategy());
```
dcw312: No functional change yet because this strategy uses the existing code for constructing the block locator.

PR #1303, opened 2016-10-13
Add setter for wallet factory to WalletAppKit.

javgh:
While playing with the WalletAppKit I noticed that it optionally makes uses of a WalletFactory, but provides no way of setting one. This commit adds a setter for it. This makes it possible to use the WalletAppKit for things like a watching wallet, without having to extend WalletAppKit.

I hope this can be included - cheers!


schildbach:
Merged.

PR #1304, opened 2016-10-14
Make getFileSize a static method to compute size before creating instance

jarlfr:
This change makes the method useful for testing a given chain file and know if an upgrade is needed.


schildbach:
Merged, thanks.

PR #1306, opened 2016-10-16
Make Context members final.

jarlfr:
This change makes all fields in a Context instance final to simplify and make the rules clearer. The fields are not meant to be updated as the Context object is propagated in different threads as Thread Local Storage. Even if that mechanism gets removed later, different set of values (in the same process) should be represented as different Context instances.


schildbach:
Yes, that's great! Merged.

PR #1308, opened 2016-10-17
Add property to require all found wallet extensions to be known when loading wallet.

jarlfr:
(Updated) This change allows wallets to set a property to enforce that wallet files that contain unknown extensions fail to load. This ensures wallets with unknown extensions will not get overwritten with those extensions removed.

Default behaviour is not changed, i.e. this property is default set to false.


schildbach:
Looks good to me.

schildbach:
Thanks! Merged.

PR #1310, opened 2016-10-26
Fix violation of equals/hashCode contract in ECKey

dwhjames:
ECKey violates the equals/hashCode contract for
Java objects. Objects that are equivalent via
the equals method, _must_ have the same hash code.

This is not the case for ECKeys when comparing
the compressed and uncompressed forms of a public key.
The implementation of the `equals` method defines
these objects to be equivalent, but the hashCode
method defined them to be distinct!

Original implementation comes from commit
640db52cf48416db8e2b24b502b3775243ad5162
which contains this bug. _1_

Note that the comment identifies the correct intent:

> Public keys are random already so we can just use a part of them as the hashcode. Read from the start to
> avoid picking up the type code (compressed vs uncompressed) which is tacked on the end.

But the second sentence is incorrect. The first byte (index 0) is the type code
(compressed vs. uncompressed), not “the end”.

The code has since been refactored in commit
9219d8a9b5714cf4e65dc046c70930c86416e65d
but the implementation is effectively
identical. _2_

The fix is simple: use the most-significant four
bytes of the X-coordinate of the public key.

1: [640db52](https://github.com/bitcoinj/bitcoinj/commit/640db52cf48416db8e2b24b502b3775243ad5162#diff-b59ef8be77b9148b27a14be59762c0c5R353)
2: [9219d8a](https://github.com/bitcoinj/bitcoinj/commit/9219d8a9b5714cf4e65dc046c70930c86416e65d#diff-1849449aac05f7e59de7ebd56efd7f43R1201)

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/core/ECKey.java
@@ -1219,10 +1219,11 @@ public class ECKey implements EncryptableItem {
 
     @Override
     public int hashCode() {
-        // Public keys are random already so we can just use a part of them as the hashcode. Read from the start to
-        // avoid picking up the type code (compressed vs uncompressed) which is tacked on the end.
+        // Public keys are random already so we can just use a part of them as the hashcode.
+        // Read from the X-coord (byte indexes [1-33)) to avoid pick up the
+        // compression indicator, which is the first byte.
         byte[] bits = getPubKey();
-        return Ints.fromBytes(bits[0], bits[1], bits[2], bits[3]);
+        return Ints.fromBytes(bits[1], bits[2], bits[3], bits[4]);
     }
 
     @Override
--- a/core/src/main/java/org/bitcoinj/core/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/core/ECKey.java
@@ -1219,11 +1219,7 @@ public class ECKey implements EncryptableItem {
 
     @Override
     public int hashCode() {
-        // Public keys are random already so we can just use a part of them as the hashcode.
-        // Read from the X-coord (byte indexes [1-33)) to avoid pick up the
-        // compression indicator, which is the first byte.
-        byte[] bits = getPubKey();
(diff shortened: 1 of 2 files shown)


schildbach:
Good catch! Your fix looks good to me. Just curious: have you had a concrete problem caused by this bug?

I'm a bit scared by the thought of accidently looking user keys (e.g. by putting them into a HashSet). @amichair Would you like to take a second look at this fix?

amichair:
Good catch indeed!

As for the fix, I think it might be safer and simpler to just delegate to `pub.hashCode()` and delete the elaborate explanation and implementation in `ECKey`. If you then want to optimize or change the implementation... well that's a separate issue from which both classes will benefit  :-)

dwhjames:
@schildbach 
> Just curious: have you had a concrete problem caused by this bug?

Yes, I had two scenarios:

1. a whitelist of individual public keys, defined to be `Set<ECKey>`
2. a whitelist of key sets (multi-sig related), defined to be `Set<Set<ECKey>>`

Given the behavior of `equals`, I was anticipating the whitelist would work no matter if my instance of `ECKey` that I was testing against the whitelist was compressed or not.

If the concrete type of  `Set` is `HashSet`, this this breaks.
(Thinking further, it looks like if I had used `TreeSet` with `ECKey.PUBKEY_COMPARATOR`, that would also fail.)
But a linear scan in a list would have been fine.

> I'm a bit scared by the thought of accidently looking user keys (e.g. by putting them into a HashSet).

I’m not following you here… “accidently looking user keys”? Maybe with your reference to HashSet, you are alluding to what I described above?

dwhjames:
@amichair 
> As for the fix, I think it might be safer and simpler to just delegate to pub.hashCode() and delete the elaborate explanation and implementation in ECKey.

Good point, the definition of hashCode there seems good to me:
https://github.com/rtyley/spongycastle/blob/sc-v1.51.0.0/core/src/main/java/org/spongycastle/math/ec/ECPoint.java#L399-L415

schildbach:
Merged. Thanks to both of you!

PR #1314, opened 2016-11-14
Remove Orchid forked subproject and support for connecting via Tor.

schildbach:
I propose to remove Orchid and with it support for connecting via Tor from bitcoinj. The feature was never really used. Those who tried reported issues and Orchid hasn't been maintained for ~3 years. On the other hand, the way how bitcoinj depends on Orchid has repeatedly caused build/dependency errors and bloats the binary size for those who do not need Tor.

If anyone wants to volunteer for the job of maintaining the Tor feature, we should find a way to plug in Orchid in a way to that the depencency is optional. I would be happy to move Orchid to a separate repository under the bitcoinj GitHub organization in this case so it can keep its home.


Polve:
I'm using bitcoinj as a library and I'm happy to get rid of orchid and related dependencies :+1:

greenaddress:
Given Orchid has little maintenance and hasn't kept up with Tor updates I agree.

It also had problems on Android and I think it stopped working with Android 6 or 7 not sure if it that was caused by some Android update and if it was then fixed as we moved to support tor via socks5 proxy and Orbot rather than using Orchid.

schildbach:
Thanks for your feedback. Merged.

PR #1316, opened 2016-11-19
Update to Bouncy/Spongy Castle 1.52.

schildbach:
Release notes can be found here: https://github.com/rtyley/spongycastle/blob/sc-v1.52.0.0/docs/releasenotes.html


Polve:
isn't possible to upgrade directly to latest version 1.55?

schildbach:
When I checked yesterday, only version 1.52 was available. Can you link the changelog of 1.55?

Polve:
I got confused with Bouncy Castle: I thought the two would be kept in sync.

http://www.bouncycastle.org/releasenotes.html

PR #1318, opened 2016-11-25
Add possibility to pause blockchain download and than return to this …

nxtpool:
…process.


schildbach:
Thanks! A couple of questions: Why do you make it a subclass? How does it work? Can you provide unit tests that show how it works? Also, should the ability to reliably pause the blockchain download not also be built into PeerGroup? (PeerGroup is coordinating the download and switches between peers if one download stalls.)

PR #1321, opened 2016-11-26
ForwardingService: Fix hardcoded fee…

schildbach:
…by simply emptying the entire wallet to the forwarding address. This will make sure the library is calculating the correct fee.


ar-:
As I mentioned before this is wrong. And this hides an existing severe bug. But if you just want to have ANY example running, then maybe you should rename it or at least change the intro comment. Vor now it still says "ForwardingService demonstrates basic usage of the library. It sits on the network and when it receives coins, simply sends them onwards to an address given on the command line."

PR #1322, opened 2016-11-30
Bugfix on javascript forward example (wrong package).

wardialer:
Fixed the "TypeError: Java.extend needs at least one type argument" bug (wrong package)


schildbach:
Merged, thanks!

PR #1327, opened 2016-12-12
Add a jdbc connection isValid check to make sure the db connection is…

brimchan:
… still available

before re-using again in the Thread.
Closes the connection and removes from the ThreadLocal otherwise


brimchan:
Motivation: For longer lived Threads or Threads that might have been re-cycled from use in a thread pool, JDBC connections could have timed out in between successive calls to the database. Using isValid() sends a small query to the DB to make sure the connection is really still available; otherwise closes the connection and removes from the threadlocal.

PR #1333, opened 2017-01-20
Keychain

MIGUELWAXX:
ECKey: more specific exception types for missing/encrypted privkeys, and use them to avoid an indirect test of whether we can sign for a transaction, which is now not quite right in the HD world. Instead of trying to test whether we can sign, just try and use a dummy sig if the key is unavailable.
HD wallets: Add completed test plan.


schildbach:
Seems like you're trying to requesting the merge of the wrong branch here. The "keychain" branch is from 2013/2014. Can you open a new PR from the correct (your) branch?

PR #1331, opened 2017-01-11
[master branch] Improvement on bitcoinj for managing many account number (arbitrary accout path)

gmagnotta:
Dear Andreas @schildbach ,

please find the new PR that replace #1329 . This contains the same changes applied to master branch, plus the space fix.

I will add tests in following days, but if you prefer to start testing by yourself, you are welcome!!!

Regards,
Giuseppe

PR #1341, opened 2017-02-20
Allow to create wallets with arbitrary path (e.g m/44'/0'/0/0) and sync/receive/send coins correctly

gmagnotta:
Hi Guys,

starting from my past work in PR #1331, I enhanced the logic to create a wallet with a custom account path.

Our requirement was to be able to create  deterministic keys under M/44' (with arbitrary depth) and allow the Wallet to synchronize propertly from the bockchain, send and receive coins in the standard/expectery way.

Well, this change allows to create a wallet with an arbitrary account path.

For example:

<pre>public static ImmutableList<ChildNumber> BIP44_ACCOUNT_TEST = ImmutableList.of(new ChildNumber(44, true),
			new ChildNumber(0, true), new ChildNumber(0, false), new ChildNumber(0, false));

Wallet.fromSeed(networkParameters, detSeed, BIP44_ACCOUNT_TEST);</pre>

Will create a Wallet with m/44'/0'/0/0 account path.

I had to change also a little the protobuf for persisting the wallet on disk in order to save the original account path when restoring.

Please include upstream so everybody can benefit from this change.

Regards,
Giuseppe

giuseppe.magnotta@gmail.com


schildbach:
Is there any reason you closed this?

gmagnotta:
Sorry, I did not closed this issue. Don't know why it is marked as closed!

DaLN:
Hi humans, any estimated time on when this would be merged ?

Kind regards.

core/src/main/java/org/bitcoinj/wallet/DefaultKeyChainFactory.java
```diff
@@ -33,16 +37,25 @@ public DeterministicKeyChain makeKeyChain(Protos.Key key, Protos.Key firstSubKey
     }
 
     @Override
```
schildbach: New public method requires a JavaDoc.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -307,17 +308,29 @@ public DeterministicKeyChain(byte[] entropy, String passphrase, long seedCreatio
      * if the starting seed is the same.
      */
     protected DeterministicKeyChain(DeterministicSeed seed) {
-        this(seed, null);
+        this(seed, null, ACCOUNT_ZERO_PATH);
+    }
+
+    public DeterministicKeyChain(DeterministicSeed seed, ImmutableList<ChildNumber> accountPath) {
```
schildbach: Public method requires a JavaDoc.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -349,13 +371,26 @@ public static DeterministicKeyChain watchAndFollow(DeterministicKey watchKey) {
...
     }
 
     /**
      * For use in {@link KeyChainFactory} during deserialization.
      */
     protected DeterministicKeyChain(DeterministicSeed seed, @Nullable KeyCrypter crypter) {
+        this(seed, crypter, ACCOUNT_ZERO_PATH);
+    }
+
+    protected DeterministicKeyChain(DeterministicSeed seed, @Nullable KeyCrypter crypter, ImmutableList<ChildNumber> accountPath) {
```
schildbach: Should have a JavaDoc too.

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
```diff
@@ -84,6 +84,10 @@ public KeyChainGroup(NetworkParameters params, DeterministicSeed seed) {
         this(params, null, ImmutableList.of(new DeterministicKeyChain(seed)), null, null);
     }
 
+    public KeyChainGroup(NetworkParameters params, DeterministicSeed seed, ImmutableList<ChildNumber> accountPath) {
```
schildbach: JavaDoc

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -271,6 +271,10 @@ public static Wallet fromSeed(NetworkParameters params, DeterministicSeed seed)
         return new Wallet(params, new KeyChainGroup(params, seed));
     }
 
+    public static Wallet fromSeed(NetworkParameters params, DeterministicSeed seed, ImmutableList<ChildNumber> accountPath) {
```
schildbach: JavaDoc

core/src/wallet.proto
```diff
@@ -132,6 +132,9 @@ message Key {
 
   // Encrypted version of the seed
   optional EncryptedData encrypted_deterministic_seed = 9;
+
+  // The original account path
+  optional string account_path = 10;
```
schildbach: Hmmm, isn't this redundant? Deterministic keys already have their path stored into the protobuf.

HashEngineering:
@schildbach Are there any plans to merge this particular pull request or something similar that would allow a `DeterministicKeyChain` to be created that supports a different path besides that of BIP32 (m/0').  

I am involved in a few projects that want to support BIP44 with bitcoinj or bitcoinj forks.  In the first project, I created a class that extends `DeterministicKeyChain` that specifies a BIP44 path instead (m/44/0'/0') for an app that will only support 1 account.  For a second project, we are interested in supporting multiple accounts using a BIP44 path and this was achieved by modifying the `DeterminisiticKeyChain` class in a way similar to this pull request.

I agree that the protobuf format need not be changed.

"sync/receive/send coins correctly" probably means that bitcoinj would still work as far as syncing the blockchain, sending and receiving coins as it does with the default BIP32 setup.  My own limited testing thus far shows that changing to a BIP44 path still results in working software (an android app in my case).

schildbach:
Yes, I'm still very much interested in this. I would prefer a solution that doesn't change the protobuf format though, if this is possible. As I said above, the new `account_path` field seems redundant to me.

DaLN:
I pushed some changes for this PR. Added some javadoc in https://github.com/uniquid/bitcoinj/pull/2

Let me know if there is anything that need improvement or is still missing.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -729,6 +784,8 @@ public boolean isFollowing() {
             Protos.Key.Builder mnemonicEntry = BasicKeyChain.serializeEncryptableItem(seed);
             mnemonicEntry.setType(Protos.Key.Type.DETERMINISTIC_MNEMONIC);
             serializeSeedEncryptableItem(seed, mnemonicEntry);
+            // TODO this is a current workaround!!! Please move it to a dedicated entry
+            mnemonicEntry.setLabel(HDUtils.formatPath(getAccountPath()));
```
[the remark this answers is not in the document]
DaLN: I originally wanted to change as little as possible and keep the code as it was done in previous commits.
Happy to make changes towards a more sensible architecture for this.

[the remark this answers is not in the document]
schildbach: Thanks, it's much appreciated. Before you continue: What do you think? Can we do without persisting the path for mnemonic keys, because we can derive it on the fly? Or is it mandatory to store it?

Giszmo:
@schildbach said:

> I hope the GitHub "files changed" is accurate. There are a lot of merge commits in your PR, according to GitHub. Maybe squash it all together?

As the PR has several authors, squashing commits wipes attribution. Rebasing could help a bit on that but would still bare the risk of wrong attribution as committer A could attribute his wallet-stealing underhanded code to committer B in the rebasing process. Therefore I prefer not messing with history and rather compare stuff on your end, if you don't trust github. Ideally commits were signed (for attribution) and builds deterministic and signed (for approval).

schildbach:
@Giszmo Generally I'm all for attributing as precisely as possible, but imho a clean history is also important (means: no merge commits, and no reverts within a PR). I propose we can add authors in the commit message, or we can squash the PR into two commits, one for each author. (Meanwhile, git should allow multiple author lines per commit.)

But I think this is a side discussion anyway. Most importantly, we should finish this PR. @DaLN, did you see my comment from 4 days ago?

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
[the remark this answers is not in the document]
DaLN: I actually think it's already stored in that part of the code (last lines)
```java
protected List<Protos.Key> serializeMyselfToProtobuf() {
        // Most of the serialization work is delegated to the basic key chain, which will serialize the bulk of the
        // data (handling encryption along the way), and letting us patch it up with the extra data we care about.
        LinkedList<Protos.Key> entries = newLinkedList();
        if (seed != null) {
            Protos.Key.Builder mnemonicEntry = BasicKeyChain.serializeEncryptableItem(seed);
            mnemonicEntry.setType(Protos.Key.Type.DETERMINISTIC_MNEMONIC);
            serializeSeedEncryptableItem(seed, mnemonicEntry);
            entries.add(mnemonicEntry.build());
        }
        Map<ECKey, Protos.Key.Builder> keys = basicKeyChain.serializeToEditableProtobufs();
        for (Map.Entry<ECKey, Protos.Key.Builder> entry : keys.entrySet()) {
            DeterministicKey key = (DeterministicKey) entry.getKey();
            Protos.Key.Builder proto = entry.getValue();
            proto.setType(Protos.Key.Type.DETERMINISTIC_KEY);
            final Protos.DeterministicKey.Builder detKey = proto.getDeterministicKeyBuilder();
            detKey.setChainCode(ByteString.copyFrom(key.getChainCode()));
            for (ChildNumber num : key.getPath())
                detKey.addPath(num.i());
```

But it seems it's actually never read which is quite puzzling to me. Any idea why ?

DaLN:
@schildbach I encountered an issue with several tests broken in my last commit with very minimal changes, i only remove the label. I will keep working on it but do you have a pointer towards where i went wrong ?

https://github.com/bitcoinj/bitcoinj/pull/1341#ref-commit-45b1e5f

 * DeterministicKeyChainTest
  * DeterministicKeyChainTest.encryption
  * DeterministicKeyChainTest.serializeUnencrypted
 * KeyChainGroupTest
  * KeyChainGroupTest.serialization
 * WalletTest
  * WalletTest.basicSpendingWithEncryptedWallet

schildbach:
@DaLN Can you post one of the test failures?

DaLN:
@schildbach org.bitcoinj.wallet.DeterministicKeyChainTest encryption

    18:34:20 1 MnemonicCode.toSeed: PBKDF2 took 36.34 ms
    18:34:20 1 DeterministicKeyChain.maybeLookAhead: 1 keys needed for M/0H/0 = 1 issued + 0 lookahead size + 0 lookahead threshold - 0 num children
    18:34:20 1 DeterministicKeyChain.maybeLookAhead: Took 147.3 μs
    18:34:20 1 KeyCrypterScrypt.deriveKey: Deriving key took 211.4 ms for 16384 scrypt iterations.
    18:34:20 1 DeterministicKeyChain.maybeLookAhead: 1 keys needed for M/0H/0 = 2 issued + 0             lookahead size + 0 lookahead threshold - 1 num children
    18:34:20 1 DeterministicKeyChain.maybeLookAhead: Took 84.97 μs
    18:34:21 1 KeyCrypterScrypt.deriveKey: Deriving key took 152.4 ms for 16384 scrypt iterations.
    18:34:21 1 KeyCrypterScrypt.deriveKey: Deriving key took 154.5 ms for 16384 scrypt iterations.
    18:34:21 1 KeyCrypterScrypt.deriveKey: Deriving key took 153.7 ms for 16384 scrypt iterations.

    java.lang.NullPointerException
	at org.bitcoinj.wallet.DeterministicKeyChain.maybeLookAhead(DeterministicKeyChain.java:1212)
	at org.bitcoinj.wallet.DeterministicKeyChain.maybeLookAhead(DeterministicKeyChain.java:1201)
	at org.bitcoinj.wallet.DeterministicKeyChain.maybeLookAhead(DeterministicKeyChain.java:1186)
	at org.bitcoinj.wallet.DeterministicKeyChain.fromProtobuf(DeterministicKeyChain.java:850)
	at org.bitcoinj.wallet.DeterministicKeyChain.fromProtobuf(DeterministicKeyChain.java:821)
	at org.bitcoinj.wallet.DeterministicKeyChainTest.encryption(DeterministicKeyChainTest.java:294)



org.bitcoinj.wallet.KeyChainGroupTest serialization

    18:34:22 1 MnemonicCode.toSeed: PBKDF2 took 26.12 ms
    18:34:22 1 KeyChainGroup.addAndActivateHDChain: Creating and activating a new HD chain: org.bitcoinj.wallet.DeterministicKeyChain@47fbc56
    18:34:22 1 DeterministicKeyChain.maybeLookAhead: 1 keys needed for M/1 = 1 issued + 0 lookahead size + 0 lookahead threshold - 0 num children
    18:34:22 1 DeterministicKeyChain.maybeLookAhead: Took 77.31 μs

    java.lang.AssertionError: 
    Expected :18
    Actual   :8

DaLN:
@schildbach @gmagnotta As far as i understood, the original account path can not be derived properly without adding it as an optional (repeated in protobuf is optional by design) field which i added in Key.

https://github.com/uniquid/bitcoinj/pull/3

DaLN:
Previous unit test failures came from misunderstandings on my side.

HashEngineering:
The original account path is stored in the protobuf format already:
read:  https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java#L827
write:  https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java#L741

While this pull request is adding the path optionally under `Protos.Key.Type.DETERMINISTIC_MNEMONIC`, the original code has it added as part of this data: `Protos.Key.Type.DETERMINISTIC_KEY`

Couldn't you change this line:  https://github.com/uniquid/bitcoinj/blob/9211be5a34830ae5e8dbda4a17fdad36b2aa97f3/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java#L925

which currently says:
`chain = factory.makeKeyChain(key, iter.peek(), seed, crypter, isMarried,
                                ImmutableList.<ChildNumber>builder().addAll(originalAccountPath).build());`

to
`chain = factory.makeKeyChain(key, iter.peek(), seed, crypter, isMarried,
                                ImmutableList.<ChildNumber>builder().addAll(path).build());`

`path` is determined in line 895 of DeterministicKeyChain of this pull request https://github.com/uniquid/bitcoinj/blob/9211be5a34830ae5e8dbda4a17fdad36b2aa97f3/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java#L895

DaLN:
@HashEngineering In the path field, it is not the original account path (for example 44H 1H 0H) It is the full absolute path of each key.
Path values for each Deterministic Key in the case of a Testnet wallet : 
* 44H
* 44H 1H
* 44H 1H 0H
* 44H 1H 0H 0
* 44H 1H 0H 0 0
* 44H 1H 0H 0 1
* 44H 1H 0H 0 2
* 44H 1H 0H 0 3
* 44H 1H 0H 0 4
* etc...

core/src/wallet.proto
```diff
@@ -132,6 +132,9 @@ message Key {
 
   // Encrypted version of the seed
   optional EncryptedData encrypted_deterministic_seed = 9;
+
+  // The path to the root.
+  repeated uint32 originalAccountPath = 10 [packed=true];
```
schildbach: For consistency, I think this should be named "original_account_path". What does the "original" mean here, isn't it just an "account_path"? Should it perhaps be optional, because existing wallets don't have this?

core/src/main/java/org/bitcoinj/crypto/DeterministicHierarchy.java
```diff
@@ -58,7 +58,7 @@ public DeterministicHierarchy(DeterministicKey rootKey) {
     }
 
     /**
-     * Inserts a key into the heirarchy. Used during deserialization: you normally don't need this. Keys must be
+     * Inserts a key into the hierarchy. Used during deserialization: you normally don't need this. Keys must be
```
schildbach: Since this is a common misspellung, and also to not touch this file in this PR, I just fixed this misspelling in the entire project. If you rebase on master, this change will disappear.

core/src/main/java/org/bitcoinj/wallet/DefaultKeyChainFactory.java
```diff
@@ -32,17 +36,26 @@ public DeterministicKeyChain makeKeyChain(Protos.Key key, Protos.Key firstSubKey
         return chain;
     }
 
+    @Override
+    public DeterministicKeyChain makeKeyChain(Protos.Key key, Protos.Key firstSubKey, DeterministicSeed seed,
+                                              KeyCrypter crypter, boolean isMarried, ImmutableList<ChildNumber> originalAccountPath) {
+        DeterministicKeyChain chain;
+        if (isMarried)
+            chain = new MarriedKeyChain(seed, crypter);
+        else {
```
schildbach: These braces are unnecessary (and inconsistent with the if branch).

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -738,6 +796,7 @@ public boolean isFollowing() {
             proto.setType(Protos.Key.Type.DETERMINISTIC_KEY);
             final Protos.DeterministicKey.Builder detKey = proto.getDeterministicKeyBuilder();
             detKey.setChainCode(ByteString.copyFrom(key.getChainCode()));
+            // key.getPath() is the path relative from the root.
```
schildbach: Is this comment still relevant?

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -896,19 +965,25 @@ public boolean isFollowing() {
...
                         if (detkey.getChildNumber().num() == 0) {
+                            // External chain is used for addresses that are meant to be visible outside of the wallet
+                            // (e.g. for receiving payments).
                             chain.externalParentKey = detkey;
                             chain.issuedExternalKeys = key.getDeterministicKey().getIssuedSubkeys();
                             lookaheadSize = Math.max(lookaheadSize, key.getDeterministicKey().getLookaheadSize());
                             sigsRequiredToSpend = key.getDeterministicKey().getSigsRequiredToSpend();
                         } else if (detkey.getChildNumber().num() == 1) {
+                            // Internal chain is used for addresses which are not meant to be visible outside of the
+                            // wallet and is used for return transaction change.
```
schildbach: FYI: Also for wallet maintenance.

schildbach:
Generally I'm quite happy now, I've added some minor comments. I'd be even happier if we could do without the protobuf change, and I think it should be possible. If I look at the new deterministic-walletbip44-serialization.txt file I can read the account path from it even if I ignore the new field. So a parser should be able to do the same thing I can (-:  But if it's really easier using the field way (and @HashEngineering agrees) I don't want to stand in the way of this PR any longer. Just one more thought: if bitcoinj were ever able to support multiple accounts per wallet (not saying it should), what meaning would the new field then have?

core/src/wallet.proto
DaLN: I actually changed the name of the field because it was not clear (i originally had the name used previously).
https://github.com/uniquid/bitcoinj/pull/4

What we want is the cointype path (44H 1H 0H), which is what we save. 
https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#Path_levels

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
DaLN: It's not, you are right. Deleting it.

core/src/wallet.proto
schildbach: From the perspective of bitcoinj and BIP32, there is no cointype. "account path" is the more correct term, and consistent with what we use in bitcoinj already. I was only wondering about the "original" prefix.

core/src/main/java/org/bitcoinj/wallet/DefaultKeyChainFactory.java
DaLN: Corrected.

core/src/wallet.proto
schildbach: ...and even if I look at the spec you're quoting, the level the current field points at seems to be "account" rather than "cointype".

DaLN: Just saw that i misread. It is indeed the accountPath that we want and have. original is not needed.

I will update my commits to reflect that.

schildbach:
@DaLN Theoretically we could scan the protobuf, look for the longest path and remove the last two elements. But ok I agree that's perhaps not a good way to go.

DaLN:
@schildbach So simple i have not even thought about it. Would you prefer i try this change so the protobuf does not change at all ? Should not be very complicated.

schildbach:
@DaLN Yes, please :smile_cat:

DaLN:
@schildbach in KeyChainGroupTest.serialization
i get 5 keys (1 DETERMINISTIC_MNEMONIC, 4 DETERMINISTIC_KEY) from the Protobuf
* Empty path
* Empty path
* 0H
* 0H 0
* 0H 1

`    private static List<ChildNumber> getAccountPathFromProtobuf(List<Protos.Key> keys) {
        List<ChildNumber> accountPath = Collections.emptyList();
        for (Protos.Key key : keys) {
            List<ChildNumber> keyFullPath = newArrayList();
            for (int i : key.getDeterministicKey().getPathList())
                keyFullPath.add(new ChildNumber(i));
            if (keyFullPath.size() > 2 && keyFullPath.size() - 2 > accountPath.size())
                accountPath = keyFullPath.subList(0, keyFullPath.size() - 2);
        }
        if (accountPath.isEmpty())
            return ACCOUNT_ZERO_PATH;
        return accountPath;
    }`

So i end up with an empty account path and return ACCOUNT_ZERO_PATH in that case.

Issue seems to be that we don't do the lookahead, so address_index level keys are not generated. So should i assume that i can remove the if condition, add "group.getBloomFilterElementCount();" in the test and that this case would most likely never happen or should i leave it there ?

schildbach:
Hmmm yeah, maybe the risk of deriving the account path wrong is too high. There is always the risk a wallet is not fully "inflated".

Maybe my idea was not so good and we should continue with the additional field approach.

schildbach:
My idea also fails in the case of a mixed wallet (e.g. one BIP44 chain and one BIP32 chain). Let's stay with the field approach, ok?

schildbach:
…Hmm, on the other hand the field approach would fail too with a mixed wallet. Too bad the protobuf format doesn't have an entity for KeyChainGroup – that would have made things much easier.

HashEngineering:
Having a BIP32 and a BIP44 keychain in the same wallet is something that I am interested in also.  In the Bitcoin Wallet fork (another coin) that I do most of my work on I want to add BIP44 support while maintaining backward compatibility with current users.  Both keychains have the same mnemonic seed.

It is still a work in progress, but it looks at the first ChildNumber of the path.  If it is a 44', then it creats a BIP44 DeterministicKeyChain, but if it is not a 44', then it created a normal DeterministicKeyChain (BIP32).  My current scheme will see that the path is m/44' and then use the seed to create a keychain that is for m/44'/05'/0' without reading the rest of the path that comes in subsequent DETERMINISTIC_KEYS.  I can see that this needs some more work to make it less error prone.

Couldn't some logic be added to look at the path and then go from there?:
if the first child is 0' - then this is a BIP32 Account 0 key chain (default for bitcoinj)
if the first child is 44' - then continue and get the information from the next two DETERMINISTIC_KEY's to get the coin type (0' or 1') and then the account number.  This is a BIP44 path. 

If the path is not inflated then the account path could be set from the optional field stored in the DETERMINISTIC_SEED section of the protobuf.  But if one had two key chains in the protobuf, there would only be one path.

schildbach:
We should add a test case for the mixed case as soon as possible. Thinking this over again, I believe a wallet with multiple key chains will be serialized to multiple DETERMINISTIC_MNEMONIC entities. The reason is multiple (BIP32) key chains are already in use today – for key rollovers. So we should indeed have an account_path field _per key chain_.  

@HashEngineering It would be nice if we could do without hardcoding the layouts, as this PR is about arbitrary paths. For example, for SegWit I believe 1' will be added to supplement the (BIP32) 0' path.

DaLN:
@schildbach @HashEngineering @gmagnotta 

Should we merge this PR and move the discussion on the mixed case extra feature elsewhere ?

schildbach:
Ok. Then we'll go with the additional field for now. Can you push your changes from 3 days ago?

About merging: I will squash everything together in one commit, getting rid of the merge commits. Authored by @DaLN. @gmagnotta seems to have contributed code to via merge commits; I'll mention him in the commit description.

DaLN:
@schildbach It's in https://github.com/uniquid/bitcoinj/pull/4 already. Waiting for @gmagnotta to merge.

gmagnotta:
@DaLN @schildbach  Merged changes! Thanks

DaLN:
@schildbach Where is the best to discuss when 0.15 will be released ? 
Same question about segwit.

core/src/wallet.proto
```diff
@@ -132,6 +132,9 @@ message Key {
 
   // Encrypted version of the seed
   optional EncryptedData encrypted_deterministic_seed = 9;
+
+  // The path to the root.
+  repeated uint32 accountPath = 10 [packed=true];
```
schildbach: Shouldn't this field be optional? We have many existing wallets without this field, I guess it will throw a protobuf error if it is missing?

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -778,11 +836,19 @@ public boolean isFollowing() {
...
         int sigsRequiredToSpend = 1;
 
+        List<ChildNumber> accountPath = newArrayList();
         PeekingIterator<Protos.Key> iter = Iterators.peekingIterator(keys.iterator());
         while (iter.hasNext()) {
             Protos.Key key = iter.next();
             final Protos.Key.Type t = key.getType();
             if (t == Protos.Key.Type.DETERMINISTIC_MNEMONIC) {
+                accountPath = newArrayList();
+                for (int i : key.getAccountPathList()) {
```
schildbach: Related to my other comment: I guess we need to guard for null here for if an old wallet is read?

schildbach:
@DaLN Such topics are perfect for the mailing list :smile_cat:

core/src/wallet.proto
DaLN: I addressed this here https://github.com/bitcoinj/bitcoinj/pull/1341#issuecomment-346439720

https://developers.google.com/protocol-buffers/docs/proto

"repeated: this field can be repeated any number of times (including zero) in a well-formed message. The order of the repeated values will be preserved."

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
DaLN: getAccountPathList() would just return an empty list in the case.

DaLN: Repeated Fields

For this field definition:

repeated int32 foo = 1;
The compiler will generate the following accessor methods in both the message class and its builder:

int getFooCount(): Returns the number of elements currently in the field.

int getFoo(int index): Returns the element at the given zero-based index.

List<Integer> getFooList(): 
Returns the entire field as a list. If the field is not set, returns an empty list. 
The returned list is immutable for message classes and unmodifiable for message builder classes.

https://developers.google.com/protocol-buffers/docs/reference/java-generated#repeated-fields

schildbach:
Merged with small changes. Thanks a lot for your work and for your patience! I'm sure this enhancement is very useful for a lot of usecases.

Serguncheouss:
Hi @gmagnotta,

please tell me why after creating the wallet after the value "index" is the value "change"?

(m/44H/0H/0H/0/23, account)
(m/44H/0H/0H/0/23/0, external)
(m/44H/0H/0H/0/23/1, internal)

The documentation specifies a different sequence in the path:
https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#registered-coin-types
`<html><body>

Bitcoin | first | external | first | m / 44' / 0' / 0' / 0 / 0
-- | -- | -- | -- | --
Bitcoin | first | external | second | m / 44' / 0' / 0' / 0 / 1
Bitcoin | first | change | first | m / 44' / 0' / 0' / 1 / 0
Bitcoin | first | change | second | m / 44' / 0' / 0' / 1 / 1

</body>
</html>`

PR #1344, opened 2017-02-28
Revert REFERENCE_DEFAULT_MIN_TX_FEE to 1000 and  MIN_NONDUST_OUTPUT t…

oscarguindzberg:
…o 546


schildbach:
I wonder if we should fix this together with #1345. I mean reverting this change increases the chance of bitcoinj creating a payment that won't be relayed (if the sender pays minimum fee), no? The goal of REFERENCE_DEFAULT_MIN_TX_FEE to prevent creating unrelayable transactions.

Issue #1345, opened 2017-02-28
Transaction.REFERENCE_DEFAULT_MIN_TX_FEE should be a fee/kb


oscarguindzberg:
To match bitcoind, Transaction.REFERENCE_DEFAULT_MIN_TX_FEE should be a fee/kb
This requires changes in Wallet.calculateFee()

oscarguindzberg:
Fixed by https://github.com/bitcoinj/bitcoinj/pull/1352

PR #1352, opened 2017-03-22
Refactor Wallet.calculateFee() with recipientsPayFees

oscarguindzberg:
I ported Wallet.calculateFee() from bitcoin core (no small fixes, this is a complete rewrite).

This is the code I am copying from
https://github.com/bitcoin/bitcoin/blob/9c7b7cf0bb579714b35f7c88db20a407dfc716bd/src/wallet/wallet.cpp#L2302-L2629

I still have to fix PaymentChannelStateTest

Other improvements that can be made:

   - Fix emptyWallet feature.
     Use Transaction.REFERENCE_DEFAULT_MIN_TX_FEE per kb for the empty wallet feature (as opposed to a fixed value).
     See Wallet.adjustOutputDownwardsForFee()

   - Rewrite from scratch unit tests for Wallet.calculateFee()
     Write tests just for uses cases I know and remove most of existent tests that may be testing several times the same thing.

   - WalletTest: remove hardcoding estimated tx size: use max/min as in some sample tests

   - Take changes from latest core version. My implementation is based on core as in September 2016
     (There are some improvements, but the logic is roughly the same)

   - If SendRequest.feePerKb > Transaction.REFERENCE_DEFAULT_MIN_TX_FEE 
      - When selecting coins to spend, don't select coins that are "dust to me" given the feePerKb: 
        If to spend an UTXO we have to add an input whose size makes the fee increase bigger than the UTXO value, don't pick that UTXO.
      - A change output may not be dust for the network, but may be "dust to me".
        If adding a change output (not dust) makes the tx to pay an extra fee bigger than the change output, throw the change to fees (or make recipients pay the difference to take it out of dust if SendRequest.recipientsPayFees is true).
        Also, a P2SH output may require a big 10 of 15 multisig script sig, so the size of the script sig required to spend it may be taken into account.
        Create a boolean SendRequest.avoidDustToMe to implement this feature
   - Currently, if we create a change output which causes the tx size to be bigger (ie pay a biggger fee) and we don't have enough coins for pay for it, we throw InsufficientMoneyException
     There are cases where just by discarding the change output, the tx size is smaller, then we have enough funds to pay it.

    - When change is dust we throw change to fees (or reduce user output values if SendRequest.recipientsPayFees is true).
      Try to create another tx with an extra input which may not have a dust output and compare the 2 txs to see what is cheaper
      (throwing away dust output vs pay extra fee for the added input)
      This is similar to the category system that was in place before this implementation

    - If change output is dust, split the cost of getting out of dust between all the recipients' outputs, not just the first output.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/SendRequest.java
+++ b/core/src/main/java/org/bitcoinj/wallet/SendRequest.java
@@ -145,6 +145,12 @@ public class SendRequest {
      * of the payment request that initiated the transaction.
      */
     public String memo = null;
+    
+    /**
+    * If false (default value), tx fee is paid by the sender
+    * If true, tx fee is paid by the recipient/s. If there is more than one recipient, the tx fee is split equally between them regardless of output value and size.
+    */
+    public boolean recipientsPayFees = false;    
 
     // Tracks if this has been passed to wallet.completeTx already: just a safety check.
     boolean completed;
@@ -262,6 +268,7 @@ public class SendRequest {
         helper.add("aesKey", aesKey != null ? "set" : null); // careful to not leak the key
         helper.add("coinSelector", coinSelector);
         helper.add("shuffleOutputs", shuffleOutputs);
+        helper.add("recipientsPayFees", recipientsPayFees);
         return helper.toString();
     }
 }
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -3964,11 +3964,13 @@ public class Wallet extends BaseTaggableObject
 
             CoinSelection bestCoinSelection;
             TransactionOutput bestChangeOutput = null;
+            List<Coin> updatedOutputValues = null;
             if (!req.emptyWallet) {
                 // This can throw InsufficientMoneyException.
                 FeeCalculation feeCalculation = calculateFee(req, value, originalInputs, req.ensureMinRequiredFee, candidates);
                 bestCoinSelection = feeCalculation.bestCoinSelection;
                 bestChangeOutput = feeCalculation.bestChangeOutput;
+                updatedOutputValues = feeCalculation.updatedOutputValues;
             } else {
                 // We're being asked to empty the wallet. What this means is ensuring "tx" has only a single output
                 // of the total value we can currently spend as determined by the selector, and then subtracting the fee.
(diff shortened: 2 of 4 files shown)


oscarguindzberg:
Another TODO:
- Change min relay fee to 1000 (currently 5000)

oscarguindzberg:
Fixes #1345

schildbach:
Thanks a lot! Yes, the unit tests for fee calculation (and coin selection) are hard to maintain. I came to realize that last year when I migrated everything to a fee rate. If you see the chance to sync them with core feel free to do so.

I think we could drop the SendRequest.emptyWallet field entirely. And the SendRequest.empytWallet() method would then use the new .recipientsPayFees field. I think that would fix your first todo, right?

Just out of curiosity, why Core from Sept 2016 and not 0.14?

oscarguindzberg:
Ok, I will drop the emptyWallet feature.

I used core as of September 2016 because I made a mistake pulling the latest core version hahaha.

I noticed the payment channels code / tests assume Transaction.REFERENCE_DEFAULT_MIN_TX_FEE to be a fixed value, not a fee per kb.
Still have to decided how to deal with that.

Still working on the TODO list. 

Happy to receive more feedback.

oscarguindzberg:
Another open issue: 
Wallet.estimateBytesForSigning() bug: It does not take into account supplied inputs in the SendRequest.tx.

oscarguindzberg:
After merging this PR, we can fix and merge #1344

oscarguindzberg:
I fixed PaymentChannelStateTest.
Now all tests pass.

I have not implemented any of the improvements mentioned above by me and @schildbach .

If there are no other comments/warnings, my suggestion is:
- Merge this PR
- Create a Github issue for the unimplemented improvements mentioned above. Once I find some spare time I will work on some of those improvements.

oscarguindzberg:
Just squashed my commits and updated the commit message.

oscarguindzberg:
The code of this PR was written by @m-sossich and me.

schildbach:
Merged! I'll add some issues like you suggested.

schildbach:
Regarding your last 4 points from the original issue description, shouldn't that all be handled by the logic from Core?

oscarguindzberg:
2 weird things here:

I am wondering why I still see a message here: Closed with unmerged commits
This pull request is closed, but the recipientsPayFees branch has unmerged commits.

Also... I can't see the commit in the master branch.

schildbach:
Sorry I forgot to push to the correct GitHub repo. I amended your commit by a bit of reformatting, but the code is the same.

oscarguindzberg:
About the last 4 points... no, the core implementation does not solve them.

PR #1357, opened 2017-03-24
Update CouldNotAdjustDownwards comment

oscarguindzberg:


schildbach:
Couldn't this commit go together with migrating emptyWallet() to use recipientsPaysFee nicely? I mean, for now emptyWallet() is still implemented the old way.

PR #1362, opened 2017-04-05
fix number encoding in ScriptBuilder

thokon00:
The number 16 must be encoded with smallNum().


schildbach:
Good catch! Do you think you could add a test case?

thokon00:
Should I squash?

thokon00:
Squashed.

schildbach:
Thanks, merged! I moved to your testcase to ScriptTest because there already are some similar tests there.

Issue #414, opened 2015-06-11
base58-encode private key [includes source]


mikehearn:
Originally reported on Google Code with ID 48

```
After getting some advice on the group, I propose to add the this method to ECKey.java:


    /**
     * Output this ECKey as an base58 encoded private key.
     */
    public String toBase58() {
    return Base58.encode(priv.toByteArray());
    }

```

Reported by `andreas.schildbach` on 2011-07-13 14:47:22

PR #1373, opened 2017-05-16
ECKey: Check that constructed private key doesn't exceed 32 bytes.

schildbach:


schildbach:
@natzei Would you like to review this?

natzei:
That's exactly what I would have done. I think it's fine.

natzei:
Well, I think the method `Utils.bigIntegerToBytes` should be discussed.
```java
public static byte[] bigIntegerToBytes(BigInteger b, int numBytes) {
    byte[] bytes = new byte[numBytes];
    byte[] biBytes = b.toByteArray();
    int start = (biBytes.length == numBytes + 1) ? 1 : 0;
    int length = Math.min(biBytes.length, numBytes);
    System.arraycopy(biBytes, start, bytes, numBytes - length, length);
    return bytes;
}
```
What's the reason of cutting 1 byte?

schildbach:
Merged.

schildbach:
@natzei Oh, sorry, I didn't see your last comment. Actually I don't know, the code is ages old. But it has to do with the weird internal format for BigInteger. I wonder why we're using BigInteger at all for storing private keys?

schildbach:
It used to be a byte array, but was switched to BigInteger in 1a4acc18be3a8923aafcd7d3f46a08e2b4c2e6a2. But the byte cutting logic was there even before. The method was extracted from getPrivKeyBytes() with 1ff81a3243d28337f616b8cf8f985d12b3b0ed55.

Finally, here is where the logic stems from: 098671ffbaf962c27e9498ae50134630225d828f. It refers to Google Code issue 48, which is archived on GitHub as #414.

natzei:
Thanks for the references.

Commit https://github.com/bitcoinj/bitcoinj/commit/098671ffbaf962c27e9498ae50134630225d828f explains where the problem arose. As you can see in the comments, they thought that BigInteger.toByteArray() adds an extra byte, when it actually adds an extra **bit**.

natzei:
I think the method should just fail if toByteArray() exceeds the expected length, without cutting off nothing.

schildbach:
Do you want to submit a PR? Then all discussions can happen in that PR. I admit I'm a bit scared of changing handling of private key data…

natzei:
I'll do it in the evening. You shouldn't be scared, because the most common
case will remain the same.

Even now, if someone asks for a byte representation of a key larger or
equal than 33 bytes, the behaviour is unexpected. Probably it never
happened before.

On 12 Jul 2017 11:29 am, "Andreas Schildbach" <notifications@github.com>
wrote:

Do you want to submit a PR? Then all discussions can happen in that PR. I
admit I'm a bit scared of changing handling of private key data…

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/bitcoinj/bitcoinj/pull/1373#issuecomment-314707925>,
or mute
the thread

natzei:
@schildbach from `ECKey.fromPrivate` seems that private keys are always treated as positive numbers, isn't it?

schildbach:
@natzei I believe so, yes.

Yesterday I spoke to a friend who stumbled upon the same problem in his own project. He seemed to be confident BigInteger adds an extra zero *byte* (rather than bit).

natzei:
That's not really true in general. It adds an extra bit, that could result in adding an entire byte.
For example:
```
N     bit (unsigned)          2-comp                    byte array
127         111_1111       0111_1111                   [0111 1111]
255        1111_1111     0 1111_1111        [0000_0000, 1111_1111]
```
I've done some tests. This behaviour is in fact what I expected by reading the javadoc.

natzei:
Assuming only positive numbers, it's possible to discriminate if an extra byte is added by checking if the first element (`b.toByteArray()[0]`) is 0 (0000_0000).

Due to the minimal representation provided by BigInteger, it means that the bit sign is 0000_000 **0** .
Otherwise the representation is not minimal.

For example, if the sign bit is 0000_00 **0** 0, then the representation is not minimal due to the rightmost zero.

natzei:
Please continue the discussion in #1404.

Issue #1374, opened 2017-05-19
Irrelevant double spend sometimes leaves wallet in an inconsistent state


saulbein:
# Situation

Wallet receives 2 relevant transactions from the network: tx2 & tx3 (spending tx0 & tx1).
Then tx4 is received, which double spends tx3 and also spends an irrelevant output of tx2.
Since tx4 is irrelevant it does not get saved to the wallet, however, tx2's output is still marked as spent by tx4. When reloading the wallet this causes an inconsistent state as tx2's output cannot point spentBy back to tx4.

# Solutions

2 likely solutions could be implemented for this problem: 
1. Saving the irrelevant double spend (tx4)
2. Disconnecting the irrelevant output (tx2 -> tx4).

# Replication

I've added this test to WalletTest in order to have a (somewhat) minimal way of replicating the issue:

```
    @Test
    public void testIrrelevantDoubleSpend() throws Exception {
        Transaction tx0 = createFakeTx(PARAMS);
        Transaction tx1 = createFakeTx(PARAMS);

        Transaction tx2 = new Transaction(PARAMS);
        tx2.addInput(tx0.getOutput(0));
        tx2.addOutput(COIN, myAddress);
        tx2.addOutput(COIN, OTHER_ADDRESS);

        sendMoneyToWallet(AbstractBlockChain.NewBlockType.BEST_CHAIN, tx2, tx1, tx0);

        // tx3 and tx4 double spend each other
        Transaction tx3 = new Transaction(PARAMS);
        tx3.addInput(tx1.getOutput(0));
        tx3.addOutput(COIN, myAddress);
        tx3.addOutput(COIN, OTHER_ADDRESS);
        wallet.receivePending(tx3, null);

        // tx4 also spends irrelevant output from tx2
        Transaction tx4 = new Transaction(PARAMS);
        tx4.addInput(tx1.getOutput(0)); // spends same output
        tx4.addInput(tx2.getOutput(1));
        tx4.addOutput(COIN, OTHER_ADDRESS);

        // tx4 does not actually get added to wallet here since it by itself is irrelevant
        sendMoneyToWallet(AbstractBlockChain.NewBlockType.BEST_CHAIN, tx4);

        // since tx4 is not saved, tx2 output 1 will have bad spentBy
        wallet = roundTrip(wallet);

        assertTrue(wallet.isConsistent());
    }
```
I also have a workaround/solution implementation by disconnecting irrelevant outputs, will put a PR soon.

[...]

PR #1375, opened 2017-05-19
Irrelevant double spend sometimes leaves wallet in an inconsistent state

saulbein:
This PR solves the issue by disconnecting inputs of the double spend transaction, which are not relevant to the wallet (solution 2).

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -2188,6 +2188,7 @@ public class Wallet extends BaseTaggableObject
         // against our pending transactions. Note that a tx may double spend our pending transactions and also send
         // us money/spend our money.
         boolean hasOutputsToMe = tx.getValueSentToMe(this).signum() > 0;
+        boolean hasOutputsFromMe = false;
         if (hasOutputsToMe) {
             // Needs to go into either unspent or spent (if the outputs were already spent by a pending tx).
             if (tx.isEveryOwnedOutputSpent(this)) {
@@ -2198,6 +2199,7 @@ public class Wallet extends BaseTaggableObject
                 addWalletTransaction(Pool.UNSPENT, tx);
             }
         } else if (tx.getValueSentFromMe(this).signum() > 0) {
+            hasOutputsFromMe = true;
             // Didn't send us any money, but did spend some. Keep it around for record keeping purposes.
             log.info("  tx {} ->spent", tx.getHashAsString());
             addWalletTransaction(Pool.SPENT, tx);
@@ -2207,16 +2209,20 @@ public class Wallet extends BaseTaggableObject
             addWalletTransaction(Pool.SPENT, tx);
         }
 
-        if(!findDoubleSpendsAgainst(tx, transactions).isEmpty()) {
-            // Kill txns in conflict with this tx
-            Set<Transaction> doubleSpendTxns = findDoubleSpendsAgainst(tx, pending);
-            if (!doubleSpendTxns.isEmpty()) {
-                // no need to addTransactionsDependingOn(doubleSpendTxns) because killTxns() already kills dependencies;
-                killTxns(doubleSpendTxns, tx);
-            }
+        // Kill txns in conflict with this tx
+        Set<Transaction> doubleSpendTxns = findDoubleSpendsAgainst(tx, pending);
+        if (!doubleSpendTxns.isEmpty()) {
+            // no need to addTransactionsDependingOn(doubleSpendTxns) because killTxns() already kills dependencies;
+            killTxns(doubleSpendTxns, tx);
+        }
+        if( !hasOutputsToMe
+            && !hasOutputsFromMe
+            && !forceAddToPool
+            && !findDoubleSpendsAgainst(tx, transactions).isEmpty())
+        {
             // disconnect irrelevant inputs (otherwise might cause serialization issue)
-            for(TransactionInput input : tx.getInputs()) {
-                TransactionOutput output = input.getOutpoint().getConnectedOutput();
+            for (TransactionInput input : tx.getInputs()) {
+                TransactionOutput output = input.getConnectedOutput();
                 if(output != null && !output.isMineOrWatched(this)) {
                     input.disconnect();
                 }
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -1820,6 +1820,9 @@ public class Wallet extends BaseTaggableObject
         // Now for each pending transaction, see if it shares any outpoints with this tx.
         Set<Transaction> doubleSpendTxns = Sets.newHashSet();
         for (Transaction p : candidates.values()) {
+            if(p.getHash().equals(tx.getHash())) {
+                continue;
+            }
             for (TransactionInput input : p.getInputs()) {
                 // This relies on the fact that TransactionOutPoint equality is defined at the protocol not object
                 // level - outpoints from two different inputs that point to the same output compare the same.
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/bitcoinj/wallet/WalletTest.java
```diff
@@ -3441,4 +3442,39 @@ public void totalReceivedSent() throws Exception {
 
         // TODO: test shared wallet calculation here
     }
+
+    @Test
+    public void testIrrelevantDoubleSpend() throws Exception {
+        // 1 spends 0
+        Transaction tx0 = createFakeTx(PARAMS, COIN.multiply(5), OTHER_ADDRESS, OTHER_ADDRESS2)[1];
```
schildbach: This version of createFakeTx also creates feeder transactions which you throw away. Is this needed? Probably we could use the "one address" version here?

core/src/test/java/org/bitcoinj/wallet/WalletTest.java
```diff
@@ -3441,4 +3442,39 @@ public void totalReceivedSent() throws Exception {
...
+        tx4.addInput(tx2.getOutput(1));
+        tx4.addOutput(COIN, OTHER_ADDRESS);
+
+        // tx4 does not actually get added to wallet here since it by itself is irrelevant
+        sendMoneyToWallet(AbstractBlockChain.NewBlockType.BEST_CHAIN, tx4);
+
+        // since tx4 is not saved, tx2 output 1 will have bad spentBy
+        wallet = roundTrip(wallet);
+
+        log.info(wallet.toString()); // will throw exception if state invalid
```
schildbach: We shouldn't log.info() this, it will spam the log.

Actually we should be able to use Wallet.isConsistent() here. If isConsistent currently isn't sensitive to this problem, it should be changed so it detects such issues.

saulbein:
Cleaned up the tests - thanks for the notes!

saulbein:
Added another test and fix for the case mentioned by @jdojff in #1374

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -2209,6 +2209,14 @@ private void processTxFromBestChain(Transaction tx, boolean forceAddToPool) thro
         if (!doubleSpendTxns.isEmpty()) {
             // no need to addTransactionsDependingOn(doubleSpendTxns) because killTxns() already kills dependencies;
             killTxns(doubleSpendTxns, tx);
+
+            // disconnect irrelevant inputs (otherwise might cause serialization issue)
+            for(TransactionInput input : tx.getInputs()) {
```
schildbach: Just one more thought about this first fix: what if the double spending transaction *is* relevant to the wallet (either directly by the double-spent UTXO or indirectly somehow). Is it still ok to disconnect the inputs? Do we already have a test for that?

saulbein: Well, it is only disconnecting the irrelevant outputs, although I'm not sure about the implication of that.
I've also checked that it can disconnect an irrelevant link between 2 relevant transactions.

saulbein: I'll add another commit to only disconnect when transaction is irrelevant.

schildbach: Hmm, I'm not sure but I think there is a slight risk in doing this. Generally if two transactions are stored in the wallet and one spends the other it is expected that spentBy is set accordingly… but again I'm not sure. Would that be difficult to achieve with your patch?

saulbein: Yeah, I'll just change it to only diconnect when the transaction is not going to get saved anyways (doesn't send/give money and is not force added).

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -2214,11 +2216,13 @@ private void processTxFromBestChain(Transaction tx, boolean forceAddToPool) thro
...
-            // disconnect irrelevant inputs (otherwise might cause serialization issue)
-            for(TransactionInput input : tx.getInputs()) {
-                TransactionOutput output = input.getOutpoint().getConnectedOutput();
-                if(output != null && !output.isMineOrWatched(this)) {
-                    input.disconnect();
+            if(!hasOutputsToMe && !hasOutputsFromMe && !forceAddToPool) {
+                // disconnect irrelevant inputs (otherwise might cause serialization issue)
+                for (TransactionInput input : tx.getInputs()) {
+                    TransactionOutput output = input.getConnectedOutput();
+                    if(output != null && !output.isMineOrWatched(this)) {
```
saulbein: Apparently my home setup failed me and I didn't notice that test fail because of missing `!`

saulbein:
So is this a good enough solution or should there be anything else added?

schildbach:
The first and the last commit are one unit, right? Then I will squash those two and merge.

The middle commit seems to be a separate thing?

wrehas:
Well we discussed them separately, but in a way it's the same issue since it's caused by leaving connected outputs with transaction which is not saved, so I guess you could actually squash them all.

saulbein:
Well we discussed them separately, but in a way it's the same issue since it's caused by leaving connected outputs with transaction which is not saved, so I guess you could actually squash them all.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -2204,11 +2209,22 @@ private void processTxFromBestChain(Transaction tx, boolean forceAddToPool) thro
             addWalletTransaction(Pool.SPENT, tx);
         }
 
-        // Kill txns in conflict with this tx
-        Set<Transaction> doubleSpendTxns = findDoubleSpendsAgainst(tx, pending);
-        if (!doubleSpendTxns.isEmpty()) {
-            // no need to addTransactionsDependingOn(doubleSpendTxns) because killTxns() already kills dependencies;
-            killTxns(doubleSpendTxns, tx);
+        if(!findDoubleSpendsAgainst(tx, transactions).isEmpty()) {
```
saulbein: I realised now that this check might not be necessary anymore, since the transaction at this point is either a double spend or has outputs to/from wallet, in which case it will be ignored below.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -1820,6 +1820,9 @@ public boolean isTransactionRelevant(Transaction tx) throws ScriptException {
         // Now for each pending transaction, see if it shares any outpoints with this tx.
         Set<Transaction> doubleSpendTxns = Sets.newHashSet();
         for (Transaction p : candidates.values()) {
+            if(p.getHash().equals(tx.getHash())) {
```
saulbein: If the check for double spends is removed, this would also be uneccessary, should I leave this or not?

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: If unsure, maybe add a unit test for this case?

saulbein: The only time this could cause potential problems is if we get a transaction with no outputs to/from wallet which is not a double spend, any idea if that is possible?

schildbach: Generally peers could sent any transaction, since bloom filters are designed to have false positives (or the wallet may not use bloom filters). The more robust we are against receiving unexpected transactions, the better.

saulbein: Alright, I'm leaving this in then.

schildbach:
So this is ready to merge then?

saulbein:
I optimized it by a bit so it should check for double spend last since it is quite expensive. It should be ready for merging now.

schildbach:
I simplified the transaction equality check (no need for getHash()) and merged. Many thanks again!

PR #1383, opened 2017-06-12
Update Script.java

natzei:


schildbach:
Merged, thanks! Do you use fully verifying mode of bitcoinj? I'm asking because I expect it to be quite buggy and there is discussion to remove it. (It's not yet decided.)

natzei:
@schildbach yes I'm using fully verifying mode. I'm building up a tool upon bitcoinj. Can you please link that discussion (if any)? thanks

PR #1390, opened 2017-06-30
Upgrade to spongycastle 1.56

btcontract:
Lightning cryptography uses some classes which are only available since bouncycastle v1.55 so it would be nice to have them in bitcoinj, too. v1.56 introduced an additional DER validation which is not compatible with some old existing bitcoin transactions so I've copied a workaround as seen here: https://github.com/ACINQ/bitcoin-lib/commit/98acda788679ce9dc215fd4e9ad3ce4b8ccc1166 and `ScriptTest.dataDrivenValidTransactions` now passes.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -192,7 +192,7 @@
                                         <urn>com.google.protobuf:protobuf-java:2.6.1:jar:null:compile:d9521f2aecb909835746b7a5facf612af5e890e8</urn>
                                         <urn>com.h2database:h2:1.3.167:jar:null:compile:d3867d586f087e53eb12fc65e5693d8ee9a5da17</urn>
                                         <urn>com.lambdaworks:scrypt:1.4.0:jar:null:compile:906506b74f30c8c20bccd9ed4a11112d8941fe87</urn>
-                                        <urn>com.madgag.spongycastle:core:1.56.0.0:jar:null:compile:871d2315bc3a171c080684ac7fbc55ecd85b6548</urn>
+                                        <urn>com.madgag.spongycastle:core:1.56.0.0:jar:null:compile:9622d6de1407dd3506254fb9b0292eb1206f6991</urn>
                                         <urn>junit:junit:4.12:jar:null:test:2973d150c0dc1fefe998f834810d68f278ea58ec</urn>
                                         <urn>mysql:mysql-connector-java:5.1.33:jar:null:compile:8af455a9a3267e6664cafc87ace71a4e4ef02837</urn>
                                         <urn>net.jcip:jcip-annotations:1.0:jar:null:compile:afba4942caaeaf46aab0b976afd57cc7c181467e</urn>
--- a/core/src/main/java/org/bitcoinj/core/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/core/ECKey.java
@@ -610,7 +610,7 @@ public class ECKey implements EncryptableItem {
                 return new ECDSASignature(new BigInteger(1, r), new BigInteger(1, s));
 
             } catch (IOException e) {
-                throw new RuntimeException(e);
+                throw new IllegalArgumentException(e);
             }
         }
 


core/src/main/java/org/bitcoinj/core/ECKey.java
```diff
@@ -574,31 +575,49 @@ public ECDSASignature toCanonicalised() {
...
                 }
-                // OpenSSL deviates from the DER spec by interpreting these values as unsigned, though they should not be
-                // Thus, we always use the positive versions. See: http://r6.ca/blog/20111119T211504Z.html
-                return new ECDSASignature(r.getPositiveValue(), s.getPositiveValue());
+
+                return len1;
+            }
+        }
+
+        private static ECDSASignature decodeSignatureLax(ByteArrayInputStream input) {
```
schildbach: Can you add a Javadoc that documents why this is needed, and why BC/SC doesn't suit us in this case.

core/src/main/java/org/bitcoinj/core/ECKey.java
```diff
@@ -574,31 +575,49 @@ public ECDSASignature toCanonicalised() {
...
+                input.read(s);
+
+                return new ECDSASignature(new BigInteger(1, r), new BigInteger(1, s));
+
             } catch (IOException e) {
-                throw new IllegalArgumentException(e);
-            } finally {
-                if (decoder != null)
-                    try { decoder.close(); } catch (IOException x) {}
+                throw new RuntimeException(e);
```
schildbach: Please don't throw RuntimeException(). The previous code threw IllegalArgumentException, right?

core/pom.xml
```diff
@@ -192,7 +192,7 @@
                                         <urn>com.google.protobuf:protobuf-java:2.6.1:jar:null:compile:d9521f2aecb909835746b7a5facf612af5e890e8</urn>
                                         <urn>com.h2database:h2:1.3.167:jar:null:compile:d3867d586f087e53eb12fc65e5693d8ee9a5da17</urn>
                                         <urn>com.lambdaworks:scrypt:1.4.0:jar:null:compile:906506b74f30c8c20bccd9ed4a11112d8941fe87</urn>
-                                        <urn>com.madgag.spongycastle:core:1.52.0.0:jar:null:compile:99976d02054732e513eb2d884d9e20bec539e283</urn>
+                                        <urn>com.madgag.spongycastle:core:1.56.0.0:jar:null:compile:871d2315bc3a171c080684ac7fbc55ecd85b6548</urn>
```
schildbach: Can you double-check your artifact? The sha1sum for me is

```
$ sha1sum ~/.m2/repository/com/madgag/spongycastle/core/1.56.0.0/core-1.56.0.0.jar
9622d6de1407dd3506254fb9b0292eb1206f6991  /home/aschildbach/.m2/repository/com/madgag/spongycastle/core/1.56.0.0/core-1.56.0.0.jar
```

core/src/main/java/org/bitcoinj/core/ECKey.java
btcontract: Sorry, I don't exactly know, just copied a working solution to make a failing test pass.

btcontract: Well, after some research I probably do know: we get `r` and `s` from `ASN1InputStream` as instance of `ASN1Integer` and `1.56` release notes specifically say that "Strict encoding enforcement has been introduced for ASN1Integer" which is probably incompatible with "lax DER signatures" which were allowed in old Bitcoin transactions. Should I put all this into description?

schildbach:
I'm currently in the process of understanding the failing test. See https://github.com/bitcoin/bitcoin/pull/10679. A bit I wonder if we should add a BIP66 verify flag for this. Also I'll try to get some guidance from the BC project to see if they think our workaround is a good idea. They must somehow deal with "lax" signatures as well.

schildbach:
According to the "Strict Public Key checking leads to broken certificates" thread of the BouncyCastle crypto-dev mailing list, a backwards-compatibility-switch has been added to the current beta of BC 1.58.

I think I'd rather wait for this version. I realize this might take a couple of weeks. @btcontract, will that block you on your project? I think we could upgrade to SC 1.54 without running into the issue; would that help you already?

btcontract:
SC 1.54 won't help, at least 1.55 is needed but it's totally fine to wait a few weeks for 1.58. I'll just continue using a fork with all these changes for development and testing.

schildbach:
BC 1.58 is out, now we have to wait for SC 1.58 (see https://github.com/rtyley/spongycastle/issues/36). I wonder if we should switch back to BC given that Android is repackaging BC since Android 3.0 (see [discussion here](https://github.com/rtyley/spongycastle/issues/34)).

Polve:
As a server-side user of bitcoinj I would very much welcome the switch back to BC.

schildbach:
Closing in favor of #1440.

PR #1391, opened 2017-06-30
SigOps for Segwit

xenog:


core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -466,6 +482,55 @@ public TransactionInput duplicateDetached() {
         return DefaultRiskAnalysis.isInputStandard(this);
     }
 
+    public int countSigOps() {
```
schildbach: Would you like to add a description for this and the following method?

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -179,11 +182,26 @@ public void testCLTVPaymentChannelTransactionSpending() {
         tx.getInput(0).setSequenceNumber(0);
         tx.setLockTime(time.subtract(BigInteger.ONE).longValue());
         TransactionSignature fromSig =
-                tx.calculateSignature(0, from, outputScript, Transaction.SigHash.SINGLE, false);
+                tx.calculateSignature(
```
schildbach: Looks like you re-formatted quite a bit of existing tests. Generally, I would like to avoid this, as it bloats the final "segwit" commit. For reformatting existing code, I would prefer a separate PR for that (against master).

schildbach: Just to make the point clear, I cannot see what you changedin TransactionTest.java (if at all), because all I see is reformatting.

core/src/test/java/org/bitcoinj/core/SigOpCountTest.java
```diff
@@ -0,0 +1,289 @@
+/*
```
schildbach: Thanks for these tests, it's really good to see them.

PR #1392, opened 2017-06-30
Reformat TransactionTest.java

xenog:
Separate these code reformattings from Segwit patch.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/TransactionTest.java
+++ b/core/src/test/java/org/bitcoinj/core/TransactionTest.java
@@ -16,6 +16,7 @@
  */
 
 package org.bitcoinj.core;
+import static org.bitcoinj.script.ScriptOpCodes.*;
 
 import org.bitcoinj.core.TransactionConfidence.*;
 import org.bitcoinj.crypto.TransactionSignature;
@@ -25,6 +26,8 @@ import org.bitcoinj.testing.*;
 import org.easymock.*;
 import org.junit.*;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.math.BigInteger;
 import java.util.*;
 import static org.bitcoinj.core.Utils.HEX;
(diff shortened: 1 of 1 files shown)


schildbach:
I'm not sure, is this format really better? Imho the binary blobs can be as small as possible because you cannot read them anyway, so the overlong one line format seems appropriate. And the mehod invocations stretched over one line per argument surely take a lot of vertical space…

What do others think?

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -16,6 +16,7 @@
  */
 
 package org.bitcoinj.core;
+import static org.bitcoinj.script.ScriptOpCodes.*;
```
schildbach: The imports seem to be unnecessary.

xenog:
This short-versus-long-line business is a matter of readability. Our screens have become wider, and with that our code too. This is one of the reasons why some websites limit the width of their content to a section in the center of the screen; why a site like Medium is more readable than Wikipedia on full-screen.

The eyes focus on a tiny section of a line when reading, when the end of the current line is reached they have to return, but the distance to retrace is long, and the eyes move quickly. It is likely that when focus has returned it will not be in the start of the intended target line, but on the line before, on the same one that was just read. It will take a bit of reading into the wrong line to realise that and correct the problem, which breaks flow.

Code and text are kind of the same thing when it comes to reading, but it is actually a bit more complicated with code, because it is important whether an argument (word?) is in the fifth or sixth position of a function call (phrase?). When a function takes many arguments, this stops being obvious in a long line of arguments. This is relevant in languages that are not in practice as typeful, such as Java, where there can be multiple arguments in a row of types such as `String`, `int` or `null`. Mistakes are made sometimes because of poor legibility.

I would not really consider vertical space to be something to preserve as much as possible. I consider it something rather strategic. The most relevant stuff should be closer to the left and the least relevant to the right. Arguments to a function are each as relevant as the other, and except for trivial functions that fit nicely in 70 columns or so with all their arguments, or functions that take few arguments.

A good rule of thumb is to keep a function within a screenful of code, but that is meant to keep the code simple and the functions/methods doing concise easy-to-understand things, I do not think it is a rule that is meant to be hacked by writing long lines to cram more logic in that lot. After all, a widescreenful of code is quite a lot.

Another practical reason to make code more vertical is GitHub. They have been nudging programmers in the right direction by limiting their code display area to about 160 columns. If code does not respect that horizontal scrolling is the punishment. Bitcoin Core is a good (bad) example of that.

An example:

```java
objectWithAnnoyinglyLongName.aMethodThatDescribesEverythingInItsName().enterpriseyBeanComputationalResource().withExtraMethods().andThenSome(thatItself, justTakesLots, ofArgumentsWithLongNames, letsSeeIfYouCan, figureOut(whichArgumentIsFifth));
```

Compare:

```java
objectWithAnnoyinglyLongName
        .aMethodThatDescribesEverythingInItsName()
        .enterpriseyBeanComputationalResource()
        .withExtraMethods()
        .andThenSome(
                thatItself,
                justTakesLots,
                ofArgumentsWithLongNames,
                letsSeeIfYouCan,
                figureOut(whichArgumentIsFifth));
```

schildbach:
Ok thanks. Does your reasoning apply to the binary blobs too?

From time to time, I copy & paste such a blob from the source into some tool or website. For these cases, the additional formatting is super annoying. Do we really need to be able to read them on one screen? Personally I'd stuff them away in one character, if this were possible.

xenog:
The blobs are fine in one line. In fact, that would make it easier to just double-click to select and then copy to another place, such as an interpreter, in order to do manipulations. I based my changes on a subjective compulsive desire to get rid of horizontal scrolling entirely. I'll make an exception for blobs in test files and revert those to their single-line form.

schildbach:
Merged.

schildbach:
Let me know if I should rebase segwit on master.

xenog:
Please, do so, that would be great.

schildbach:
Done.

PR #1394, opened 2017-07-04
Merge script_valid.json and script_invalid.json into one script_tests.json.

schildbach:
This also synchs script_tests.json to Bitcoin Core's version as of dde46d3ae1de18700e089bc1861cf1e18649dc5d, minus some of the newer 'invalid' tests which we can't handle yet.

We also need to check 'invalid' tests more specificly for the expected error; the current behaviour mimics how dataDrivenInvalidScripts() used to work.


natzei:
@schildbach did you pass all the scripts? I was doing the merge on my own, refactoring the way scripts are checked (like in Bitcoin Core, creating a tx crediting and tx spending).

In my case, one script fails (maybe there are more):
`["0x4c 0x00","DROP 1","MINIMALDATA","MINIMALDATA","Empty vector minimally represented by OP_0"]`

Is it the same for you?

schildbach:
Yes, the "minimaldata" tests I removed for now. Still the number of tests increased by ~20%, so I think it's an improvement.

Can you PR your merge (against master) so I can compare the two?

natzei:
here you go #1395. Obviously it's not meant to be merged.

schildbach:
I've updated this PR with your (@natzei) code to create the creditiing and spending transactions. I think will merge this soon, so that the structural change of the script tests is done.

Note this only synchs the script tests to the date where the merge happened in Bitcoin Core (April 2016). And not even that, because as described above some 'invalid' tests are missing and the errors are currently not checked specifically. I'll add both as issues once this is merged.

natzei:
Nice! I may work on it after you'll create the issues. Thanks.

schildbach:
Merged.

Issue #1397, opened 2017-07-05
ScriptTests: Check for specific errors


schildbach:
As noted in #1394, currently no specific errors are tested for in the script tests. So the tests will not catch the case where a specific error is expected, but another VerificationException is thrown instead.

The possible errors are defined in https://github.com/bitcoin/bitcoin/blob/master/src/test/script_tests.cpp#L58, though some of them refer to not implemented features as of now (e.g. `WITNESS*`).

Maybe we can use a subclass of VerficationException for each of these errors? However, there are already some subclasses defined, can we re-use or re-arrange them? Using such subclasses it should be easy to fix the test logic.

natzei:
I think we could just arrange ScriptException, allowing to pass the error code in the constructor.

natzei:
I've started working on it. See PR #1399

schildbach:
Implemented in d8af6b4be7525459ad0f806bd1969151fdb3c0f6.

Issue #1396, opened 2017-07-05
ScriptTests: Fix and sync MINIMALDATA tests.


schildbach:
As noted in #1394, `scripts_tests.json` is missing all the `MINIMALDATA` tests from Bitcoin Core. This is because for some reason, the tests pass though they're expected to fail with a specific error.

The job would be to investigate why these tests don't behave as expected and fix this. I guess a VerifyException (or a subclass of, see next issue) should be thrown. Actually I believe there is some code that tries to get the data opcodes into minimal form. I just don't know why the tests don't trigger this.

And as soon as the issue is fixed, re-add all the `MINIMALDATA` tests from Bitcoin Core.

natzei:
PR #1399 should solve this issue.

schildbach:
Fixed in d8af6b4be7525459ad0f806bd1969151fdb3c0f6.

PR #1399, opened 2017-07-06
Introduced script specific errors

natzei:
I've also removed some redundant cases of opcode and refactored a bit CHECKLOCKTIMEVERIFY (added a new method like in BitcoinCore) and CHECKSEQUENCEVERIFY (allowing better specific messages).

I'll wait a feedback before proceeding, i.e. before check that script tests fail with the right error code.


natzei:
I've completed the implementation :)
All tests passed.

natzei:
Added all the Bitcoin Core MINIMALDATA tests.
This PR solves #1396 and #1397.

core/src/main/java/org/bitcoinj/script/ScriptError.java
```diff
@@ -0,0 +1,90 @@
+package org.bitcoinj.script;
```
schildbach: For new java files, we need the license.

natzei: done

natzei:
Regarding separating the commits, can you suggest how?

core/src/main/java/org/bitcoinj/script/ScriptError.java
```diff
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2013 Google Inc.
```
schildbach: Oh, please don't donate your copyright to Google...

You can simply use "Copyright by the original author or authors." here.

natzei: oh... I've copy/pasted without reading.
I've seen that also Script.java contains your copyright. Could I add my name too?

schildbach:
Ok, let's take this example: If I understand your change right, you removed all the case statements that lead into throwing ScriptException("Attempted to use disabled Script Op."). All these cases now run into the default case and will throw ScriptException("Script used a reserved opcode " + opcode).

That change is fine by me, but I think it's unrelated to this PR and should be discussed individually and -- if merged -- end up as a separate commit in the history.

The usage of static imports for VerifyFlags is also an unrelated refactoring, and actually now that I think of it I'd rather not do it because the context of the flags gets lost. (I personally use static imports very rarely, as it tends to obfuscate the code.)

Imho the ultimate goal should be to get the impact of this PR as small as possible, so reviewing is as easy as possible.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -17,32 +17,44 @@
 
 package org.bitcoinj.core;
 
+import static com.google.common.base.Preconditions.checkArgument;
```
schildbach: This "organize imports" seems unnecessary.

natzei:
I agree with you. When refactoring I basically viewed the Bitcoin Core implementation and made them much more similar.
But I know, the impact should be as small as possible. I'll try to fix all the things you mentioned.

core/src/main/java/org/bitcoinj/script/ScriptError.java
schildbach: If you want you can do that.

natzei:
Done. Please let me know if it's fine now. I'll propose the other changes of Script.java in another PR when this will be merged. Thanks

schildbach:
I removed the two unused constants and merged. You might want to add them in a followup PR. Thanks!

PR #1403, opened 2017-07-12
Refactoring of Script class to follow Bitcoin Core implementation.

natzei:
Removed unused cases of opcode.
Created some constants.
Removed OP_0 "special treatment".


schildbach:
Merged.

PR #1406, opened 2017-07-17
Verify segwit commitment structure

xenog:


core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -590,12 +593,59 @@ private void checkMerkleRoot() throws VerificationException {
...
 
+    private void checkSegwitCommit() throws VerificationException {
+        if (transactions == null)
+            throw new VerificationException("No transactions in block");
+        final Transaction coinBase = transactions.get(0);
+        byte[] rootHash = null;
+        for (TransactionOutput out : coinBase.getOutputs()) {
+            final byte[] pkScript = out.getScriptBytes();
+            if (pkScript.length < 38) continue;
+            final byte[] expected = HEX.decode("6a24aa21a9ed");
```
schildbach: What's this magic value? Perhaps extract it as a constant and add a comment?

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -264,7 +264,21 @@ public Transaction(
...
+    }
+
+    public Sha256Hash getHash(boolean segwit) {
+        if (segwit) {
+            if (isCoinBase()) return Sha256Hash.ZERO_HASH;
+            ByteArrayOutputStream stream = new UnsafeByteArrayOutputStream(length < 32 ? 32 : length + 32);
+            try {
+                bitcoinSerializeToStream(stream, TransactionOptions.WITNESS);
+            } catch (IOException e) {
+                // Cannot happen, we are serializing to a memory stream.
```
schildbach: I know you copied this from below, but shouldn't we throw an error here?

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1811,7 +1835,6 @@ public void setMemo(String memo) {
      * traditional manner without witness data, and t is the size of a transaction serialized in the segwit format
      * with witness data.
      */
-    // TODO: write tests for this
```
schildbach: Do we have a test for getWeight() now?

core/src/main/java/org/bitcoinj/core/Block.java
xenog: It is less magic than it seems:

```
6a - OP_RETURN
24 - push 36 bytes to stack
aa21a9ed - commitment header
```

In the pull request for Bitcoin Core these constants are set in code as opposed to symbols.

https://github.com/bitcoin/bitcoin/pull/8149/files#diff-7ec3c68a81efff79b6ca22ac1f1eabbaR3511

It's always the same no matter which network.

core/src/main/java/org/bitcoinj/core/Transaction.java
xenog: When serializing to a memory stream `IOException` cannot occur. Nonetheless the code for `bitcoinSerializeToStream` is generic enough to be able to take a stream that may serialize to disk or network, where `IOException` can happen.

core/src/main/java/org/bitcoinj/core/Transaction.java
xenog: I mistook weight for cost. We do not have a test yet for weight. I reinstated the TODO.

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: Ok thanks, but can we add that as a comment? And I still think a constant would make sense, currently you're using the rather expensive HEX.decode() in a loop.

xenog: I'm a Haskell programmer and I'm used to this kind of statement getting memoized by the runtime or removed by the compiler. My bad. I corrected and defined a constant as suggested.

schildbach:
I rebased our branch against master. Had to assign ScriptErrors to all ScriptException and probably got several wrong. I'll try to merge the script_tests.json from Core.

schildbach:
Damn, we have to implement these testing infrastructure changes from Core before we can run Core's tests: 06d3805c1a44aae39ad95538063e6882249b7633, 66cca79130a204277e61088b7844b405dc3868a5, 330b0f31ee5719d94f9e52dfc83c5d82168241f9 and d846e0237256a4105199673d2eb90ed1dbb55b35.

xenog:
The architecture of the library is making it difficult to implement BIP-9. Challenges are described below. These problems really only affect validating nodes, since for SPV nodes the whole BIP-9 verification can be turned off.

BIP-9 is a stateful system. For any deployment the state starts `DEFINED` from the genesis block up to the first state transition whose median 11-parent timestamp is after that deployment's start time or timeout. To initialize the state of a deployment it is necessary to skip backwards through the blockchain looking at difficulty transition blocks until one is found whose median timestamp is below the start time of the deployment. The performance of this operation is bad if one can only find difficulty-transition blocks by walking down the blockchain following parents. If the headers have been pruned it may be the case that there is not enough history to compute a deployment's state.

After locating the last `DEFINED` difficulty transition block, one must revisit difficulty-transition blocks in reverse setting their states accordingly. This operation is a bit slow only when the state is `STARTED` and the timeout for the deployment has not been reached, since it becomes necessary to count all the signals during the difficulty period to determine if a transition to `LOCKED_IN` should happen. This is fine though as Bitcoin Core has the same constraint, although with *block headers stored in memory* instead of disk.

The fun starts when verifying individual blocks. For every new block it is necessary to find its state. That is computed by looking at the parent's state (normally) or computing a state transition (with every difficulty transition). Since state transitions only happen every 2016 blocks, its performance can be lower without much consequence. For regular non-state-transition blocks the last difficulty transition must be located in order to determine the state, which is slow because parents must be followed.

Problems can be somewhat mitigated if an arbitrary ancestor for a block could be quickly located. In Bitcoin Core there is an extra field stored with every block header that points to an ancestor at a somewhat-arbitrary position in the chain. This allows to sometimes jump past a number of blocks when locating an ancestor, speeding the process. BitcoinJ has various disadvantages here. The first and most important being that blocks are stored in disk instead of memory. The second is that adding a block skip like Bitcoin Core's would require to change various implementations of block store that are offered by the library, so it is a bit of work.

In my opinion the most efficient way to implement BIP-9 in bitcoinJ is by having the state of all deployments be recorded and stored for every block. This because even with a block skip record, looking up in disk is orders of magnitude more expensive than skipping in memory. But this solution is still far from ideal because it would bloat the size of every block record.

My choice for now is not to implement BIP-9 support and to set the block heights at which certain deployments get activated on each network manually.

schildbach:
@xenog My idea for BIP9 is to start by implementing it only for nodes that read the blockchain (maybe just headers) from genesis. Then we could extend the BuildCheckpoints tool, CheckpointManager and the checkpoint file format to record the activation state of known BIPs for each checkpoint. The checkpoints are always at the difficulty transitions, so this should be straightforward. Then nodes that inititialize their blockchain state from a checkpoint can initialize the corresponding BIP activation state too, and continue from there.

The problem I see with this approach is with BIPs that we don't know yet, or have been too lazy to implement in time. Those activations have been missed by old and current clients, e.g. BIP68. Unfortunately also rollbacks are not feasible, the current SPV block store only stores 5000 blocks. A replay would be necessary, leading to an unpleasant user experience.

PR #1408, opened 2017-07-19
Add mainnet and testnet seeds from Bloq.

schildbach:


ManfredKarrer:
Isn't that problematic to add a seed node from a company (Bloq) which has acquired a chain analysis company (Skry)? That renders BitcoinJ's privacy reputation even worse, even if the bloom filter issue is much more concerning.

PR #1409, opened 2017-07-22
Avoid NPE on preallocated exception without message

zkejid:
Hello.

I'm trying to make my own research on bitconj codebase. During test execution I ran into NPE in code added by commit 83995e9284c1f76b07fb2cc587555296b7f61dec. There is check, which doesn't expect exception message to be null. So I just made additional null-check, so it will not throw NPE and will not log noisy messages.

The cause is an optimisation in HotSpot. Here is an article about it: http://jawspeak.com/2010/05/26/hotspot-caused-exceptions-to-lose-their-stack-traces-in-production-and-the-fix/

My solution is not so good. The best solution would be to completely avoid 'legal' exceptions in this case. However, my knowledge is not enough for it. 

Hope this helps.

### DIFF
--- a/core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelServer.java
+++ b/core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelServer.java
@@ -526,8 +526,10 @@ public class PaymentChannelServer {
         log.error(message);
         Protos.Error.Builder errorBuilder;
         errorBuilder = Protos.Error.newBuilder()
-                .setCode(errorCode)
-                .setExplanation(message);
+                .setCode(errorCode);
+        if (message != null) {
+            errorBuilder.setExplanation(message);
+        }
         conn.sendToClient(Protos.TwoWayChannelMessage.newBuilder()
                 .setError(errorBuilder)
                 .setType(Protos.TwoWayChannelMessage.MessageType.ERROR)
--- a/core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelClient.java
+++ b/core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelClient.java
@@ -475,8 +475,10 @@ public class PaymentChannelClient implements IPaymentChannelClient {
             } catch (VerificationException e) {
                 log.error("Caught verification exception handling message from server", e);
                 errorBuilder = Protos.Error.newBuilder()
-                        .setCode(Protos.Error.ErrorCode.BAD_TRANSACTION)
-                        .setExplanation(e.getMessage());
+                        .setCode(Protos.Error.ErrorCode.BAD_TRANSACTION);
+                if (e.getMessage() != null) {
+                    errorBuilder.setExplanation(e.getMessage());
+                }
                 closeReason = CloseReason.REMOTE_SENT_INVALID_MESSAGE;
             } catch (IllegalStateException e) {
                 log.error("Caught illegal state exception handling message from server", e);
--- a/core/src/main/java/org/bitcoinj/script/Script.java
+++ b/core/src/main/java/org/bitcoinj/script/Script.java
@@ -1541,7 +1541,7 @@ public class Script {
 
             // This RuntimeException occurs when signing as we run partial/invalid scripts to see if they need more
             // signing work to be done inside LocalTransactionSigner.signInputs.
-            if (!e1.getMessage().contains("Reached past end of ASN.1 stream"))
+            if (e1.getMessage() != null && !e1.getMessage().contains("Reached past end of ASN.1 stream"))
                 log.warn("Signature checking failed!", e1);
         }
 


Polve:
Doe this problem arise only on very old versions of java?

zkejid:
> Doe this problem arise only on very old versions of java?

No. Preallocated exception (NPE, ClassCastException, ArrayIndexOutOfBoundsException, etc) is the feature of HotSpot JVM. If your code produces a lot of exceptions, like it does during tests execution, JVM starts throwing exception without message and stacktrace. It improves performance. But code in catch clause should work properly with such empty exception instance.

I reproduced the issue running org.bitcoinj.script.ScriptTest#dataDrivenInvalidTransactions until failure. This test breaks on second hundred of executions.

here is my environment:

```
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) Server VM (build 25.131-b11, mixed mode)
```

zkejid:
Oops. Looks like there is one more "until failure" issue in org.bitcoinj.wallet.WalletTest#basicSpendingWithEncryptedWallet. Test fails on ~175 sweep with exception
```
java.lang.IllegalArgumentException: private key exceeds 32 bytes: {} bits [376]

	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:180)
	at org.bitcoinj.core.ECKey.<init>(ECKey.java:193)
	at org.bitcoinj.crypto.DeterministicKey.<init>(DeterministicKey.java:69)
	at org.bitcoinj.crypto.DeterministicKey.derivePrivateKeyDownwards(DeterministicKey.java:427)
	at org.bitcoinj.crypto.DeterministicKey.findOrDeriveEncryptedPrivateKey(DeterministicKey.java:406)
	at org.bitcoinj.crypto.DeterministicKey.decrypt(DeterministicKey.java:377)
	at org.bitcoinj.crypto.DeterministicKey.decrypt(DeterministicKey.java:43)
	at org.bitcoinj.core.ECKey.decrypt(ECKey.java:1113)
	at org.bitcoinj.crypto.DeterministicKey.decrypt(DeterministicKey.java:388)
	at org.bitcoinj.crypto.DeterministicKey.decrypt(DeterministicKey.java:43)
	at org.bitcoinj.wallet.DecryptingKeyBag.maybeDecrypt(DecryptingKeyBag.java:50)
	at org.bitcoinj.wallet.DecryptingKeyBag.findKeyFromPubHash(DecryptingKeyBag.java:67)
	at org.bitcoinj.core.TransactionOutPoint.getConnectedRedeemData(TransactionOutPoint.java:169)
	at org.bitcoinj.core.TransactionInput.getConnectedRedeemData(TransactionInput.java:295)
	at org.bitcoinj.wallet.Wallet.signTransaction(Wallet.java:4091)
	at org.bitcoinj.wallet.Wallet.completeTx(Wallet.java:4028)
	at org.bitcoinj.wallet.WalletTest.basicSpendingCommon(WalletTest.java:340)
	at org.bitcoinj.wallet.WalletTest.basicSpendingWithEncryptedWallet(WalletTest.java:164)
```

schildbach:
Thanks for finding this! I was already trying to investigate into these sporadic unit test failures, but failed and would have searched forever.

I wonder if we should perhaps disable this optimization for the tests. According to the article you linked, the `-XX:-OmitStackTraceInFastThrow` flag would accomplish this. Any thoughts?

zkejid:
PR addresses issue in production code, and user can face the same problem while using bitconj. Also, this code goes against javadoc for java.lang.Throwable#getMessage:

>      @return  the detail message string of this {@code Throwable} instance
>               (which may be {@code null}).

So I think it is better to change code than Travis configuration. 

Flag should help. As a temporary solution it is OK.

zkejid:
Also I can look for more use cases of getMessage, which can lead to the same NPE.

schildbach:
Ok I see your point about production code. Ideally production code should never depend on the exception message. However in this case I guess the upstream API just didn't think about this case, and upstream APIs are difficult to fix/enhance. I'm willing to merge your patch, but I think we should document it a bit in code.

As far as I could see, there are no other dependencies of production code on the exception message (except printing it to a log, wrapping it a domain specific exception, etc.). However, we have tons of unit tests that assert for a specific exception message. In this case imho we have no other choice but to fail the test, be it through an (unexpected) NullPointerException or by properly failing the assert – the outcome will be the same, the test will fail. At the same time, running the tests will likely be the only "usecase" that raises enough exceptions to trigger the HotSpot optimization. Therefore, I still suggest disabling the optimization for the tests.

schildbach:
Yes, please feel free to look for other instances of production code depending on the value of or the non-null-ness of getMessage().

zkejid:
> I'm willing to merge your patch, but I think we should document it a bit in code.

Should I change my PR somehow?

schildbach:
If you find other similar cases in production code, feel free to add more commits. I will write a line or two of documentation when merging.

Meanwhile, I've proposed a patch for disabling the optimization in unit tests: #1412

zkejid:
There is one more case potentially leading to NPE. Proto.Error builder doesn't allow null arguments and throws NPE. I've put an empty string for null messages.

core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelServer.java
```diff
@@ -527,7 +527,7 @@ private void error(String message, Protos.Error.ErrorCode errorCode, CloseReason
         Protos.Error.Builder errorBuilder;
         errorBuilder = Protos.Error.newBuilder()
                 .setCode(errorCode)
-                .setExplanation(message);
```
schildbach: Hmm, explanation is optional in the protobuf. This means the better way would be to not set the explanation at all rather than set it to the empty string.

core/src/main/java/org/bitcoinj/protocols/channels/PaymentChannelServer.java
```diff
@@ -527,7 +527,7 @@ private void error(String message, Protos.Error.ErrorCode errorCode, CloseReason
...
+                .setExplanation(message != null ? message : "");
```
schildbach: …same here.

schildbach:
I squashed, added some comments and merged. Again, thanks a lot for investing your time into this issue! I'm glad this is now fixed.

PR #1411, opened 2017-07-22
Destruction of temporary byte arrays containing private keys

eaglerainbow:
There are multiple cases around, where a private key is being copied over (temporarily) in a `byte[]`. Due to the nature of Java, the memory of these arrays is only disposed during garbage collection, if just the reference is being lost. This may pose a security threat, as even though the operation, which required the private key, has already been completed. Yet, the credential-like piece of information still is kept in memory and thus could be spied out by a potential attacker.
A better approach is to delete the content of the array as soon as the private key data is no longer required. By this, even if an attacker could spy on the array later on, no valuable data would be found and thus security isn't compromised.
To facilitate this, a new class `DestructionUtils` is being provided, which provides a utility method for cleaning up byte arrays. Futhermore, several code locations, where a temporary `byte[]` is being used containing private key data, is being cleaned once the array is no longer required.
Furthermore, Java itself also provides an approach for destructing security-sensitive data by its `Destroyable` interface. For ease of consumption, this pattern is applied to `VersionedChecksummedBytes`, which is an ancestor of `DumpedPrivateKey`.

NB: This PR does not solve the much bigger problem of the fact that `ECKey` is not destroyable. For a discussion on this matter, also refer to https://groups.google.com/forum/#!topic/bitcoinj/s-SCDB1QYjY

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/DumpedPrivateKey.java
+++ b/core/src/main/java/org/bitcoinj/core/DumpedPrivateKey.java
@@ -29,7 +29,6 @@ import javax.annotation.Nullable;
  * bytes with a header byte and 4 checksum bytes at the end. If there are 33 private key bytes instead of 32, then
  * the last byte is a discriminator value for the compressed pubkey.
  */
-/* TODO: to be checked: all consumers need to call destroy(), if no longer needed */
 public class DumpedPrivateKey extends VersionedChecksummedBytes {
     private boolean compressed;
 
--- a/core/src/main/java/org/bitcoinj/core/DumpedPrivateKey.java
+++ b/core/src/main/java/org/bitcoinj/core/DumpedPrivateKey.java
@@ -29,7 +29,7 @@ import javax.annotation.Nullable;
  * bytes with a header byte and 4 checksum bytes at the end. If there are 33 private key bytes instead of 32, then
  * the last byte is a discriminator value for the compressed pubkey.
  */
-/* TODO: Class must implement the Destroyable interface + consumer need to call it*/
+/* TODO: to be checked: all consumers need to call destroy(), if no longer needed */
 public class DumpedPrivateKey extends VersionedChecksummedBytes {
     private boolean compressed;
 
--- a/core/src/main/java/org/bitcoinj/core/DumpedPrivateKey.java
+++ b/core/src/main/java/org/bitcoinj/core/DumpedPrivateKey.java
@@ -29,6 +29,7 @@ import javax.annotation.Nullable;
  * bytes with a header byte and 4 checksum bytes at the end. If there are 33 private key bytes instead of 32, then
  * the last byte is a discriminator value for the compressed pubkey.
  */
+/* TODO: Class must implement the Destroyable interface + consumer need to call it*/
 public class DumpedPrivateKey extends VersionedChecksummedBytes {
     private boolean compressed;
 
--- a/core/src/main/java/org/bitcoinj/core/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/core/ECKey.java
@@ -19,6 +19,8 @@
 package org.bitcoinj.core;
 
 import org.bitcoinj.crypto.*;
+import org.bitcoinj.utils.DestructionUtils;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
@@ -423,8 +425,9 @@ public class ECKey implements EncryptableItem {
      * @throws org.bitcoinj.core.ECKey.MissingPrivateKeyException if the private key is missing or encrypted.
      */
     public byte[] toASN1() {
+        byte[] privKeyBytes = null;
         try {
-            byte[] privKeyBytes = getPrivKeyBytes();
+            privKeyBytes = getPrivKeyBytes();
             ByteArrayOutputStream baos = new ByteArrayOutputStream(400);
 
             // ASN1_SEQUENCE(EC_PRIVATEKEY) = {
@@ -442,6 +445,8 @@ public class ECKey implements EncryptableItem {
             return baos.toByteArray();
         } catch (IOException e) {
             throw new RuntimeException(e);  // Cannot happen, writing to memory stream.
+        } finally {
+            DestructionUtils.destroyByteArray(privKeyBytes);
         }
     }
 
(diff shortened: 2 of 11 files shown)


eaglerainbow:
Partially, apparently unit tests are failing due to introduction of destruction - further analysis required from my side (please don't review yet)

schildbach:
Just one initial comment: We discussed years ago if erasing private key bytes is useful in terms of security. The outcome of the discussion was as far as you let go of the reference, it should make no difference. Because if you've got low-level access to the VM/memory, you can already do anything you want, including stealing of private keys.

eaglerainbow:
> The outcome of the discussion was as far as you let go of the reference, it should make no difference. 

Do you have any reference on the discussion? I'd be interested in reading it...

> Because if you've got low-level access to the VM/memory, you can already do anything you want, including stealing of private keys. 

True point. However, I am also concerned about the effect that we don't have the almighty attacker being in control of the VM, but that the JVM is just buggy (or old) and partially reveals some (random) data (from the heap) based on a tweaked request. The more of the stuff is lying around, the larger is the likelihood that something is leaked.
So, yes, my PR can't solve the entire problem for sure. But my hope would be that it at least reduces the attack surface a little with not using too much effort...

Or don't you consider this proposal to be worthwhile anyway?

eaglerainbow:
> Partially, apparently unit tests are failing due to introduction of destruction - further analysis required from my side (please don't review yet)

That's fixed now (it had been just a small glitch, which I had overlooked before).

schildbach:
I'll see if I can find the discussion but it was 5+ years ago and I suspect it was on IRC.

We also talked about a buggy JVM, and buggy hardware and I think the outcome was we have to rely on them. The code would become a mess if we tested against all potential failure modes (probably an infinite number).

Now, to get closer to the idea of your PR: As you noted, we're keeping private keys (or representations of it) in byte[], BigInteger, String and maybe even char[]. I wonder if we could introduce a class ConfidentialBytes, probably wrapping a byte[]. It would be immutable except for a destroy() method which works similar to what you implemented, but it would also make any further usage of the reference throw an exception. Also, the class could contain our sanity length checks and would not have to deal with the inefficiency of BigInteger. It would help in a type-safe way that confidential data doesn't accidently get leaked. We could perhaps also use this class for passwords or seeds (in the 12 words form).

Of course, we would still need to convert confidential data to byte[] or BigInteger, unless BouncyCastle decides to provide such class. It might make sense to discuss this idea with the BC people.

eaglerainbow:
> We also talked about a buggy JVM, and buggy hardware and I think the outcome was we have to rely on them. The code would become a mess if we tested against all potential failure modes (probably an infinite number).

Yes, I agree that this is not promoting readability of the code - but security often behaves like this, unfortunately 😢 

> I wonder if we could introduce a class ConfidentialBytes, probably wrapping a byte[].

Whilst I very much agree to your idea, I abandoned it quite early when writing the PR, as it has one major drawback: To be consistent in its usage (and only then it makes sense to my mind), it would mean that multiple APIs would become deprecated, as the `byte[]` is being used in many prototypes of methods. Even worse, they are partially also `public`, thus also having an impact on consumers (who would then have to make use of the new data type instead of `byte[]`). 
That is why I did not dare to go that way, as to my mind this would be a "bigger bang", which - based on the benefit to gain - would not balance in the end.

> Of course, we would still need to convert confidential data to byte[] or BigInteger, unless BouncyCastle decides to provide such class. 

Yes, for `byte[]` you could still destroy them (as you are in control of them as consumer of BC). For `BigInteger`, we are back to the original discussion.

> It might make sense to discuss this idea with the BC people.

You are right. Now that you say it, it also feels strange, that they are not taking care on this themselves, but use `BigInteger` all around. I wonder whether they did have the same kind of question discussed already...

schildbach:
I found an old discussion, but I think there was more: https://groups.google.com/forum/#!msg/bitcoinj/Gycr-QbY7dE/vpXnM5F4PNkJ

One more interesting aspect is that even zeroing over an array doesn't guarantee the data is erased.

eaglerainbow:
> One more interesting aspect is that even zeroing over an array doesn't guarantee the data is erased.

To my mind, there are two aspects to that matter, which should not be left unmentioned:
1. simply zeroing memory isn't suitable - that's correct. On hardware level, still some rest of charges of the memory cell may be available, which - if the attacker is under control of the hardware - can be used for reading the data. But that kind of attack really requires some brute force methods. 
2. If the GC did a reorg before, an old copy of the array may still be around, which you cannot touch. 

Concerning item 1, AFAIK there is also some quite simple trick, which you may apply to reduce the risk even further: you do not just overwrite it with zero, but with `0x5a` and another time with `0xa5`, before you zero it. Still, this approach does not really discard all the charges from the memory cell, but with that overwriting, things become even much more ugly for the attacker (and I think that's the maximal thing we can do on software level anyway). If you want to have further protection than this, I think that is the point where the consumer of the app should really start thinking about cryptographic devices, such as smartcards or similar.
We can't save the world with software, but I think we can raise the bar for an attacker as much as possible.

> I found an old discussion, but I think there was more: https://groups.google.com/forum/#!msg/bitcoinj/Gycr-QbY7dE/vpXnM5F4PNkJ

Interesting, though. Especially, this `Unsafe` discussion would perfectly fit to your `ConfidentialBytes` idea. Complexity could be hidden behind the `ConfidentialBytes`'s API. Whilst also this wouldn't solve the challenge on swapping (because that still could be done by the operating system without us letting it know), it would properly solve the "the GC reorganizes the heap" issue (see item 2 above - in a nutshell for tl&dr; people: `Unsafe` allows direct memory allocation on native memory, outside of the Java's help and thus is outside of the reach of garbage collection). 
The only thing which makes me feel unhappy is the compatibility aspect of it: Only with JDK 9 a certain amount of compatibility across platforms is guaranteed. This means, if you go for the `Unsafe` approach, you would have to provide two implementations: one with the `Unsafe ` class, if it is available, and a less-secure (buah! I even don't like to write that) one, which runs everywhere.

For other readers being interested more on the `Unsafe` approach, I found these two pages quite helpful:
* https://dzone.com/articles/understanding-sunmiscunsafe
* http://hg.openjdk.java.net/jdk7/jdk7/jdk/file/9b8c96f96a0f/src/share/classes/sun/misc/Unsafe.java

zkejid:
> you do not just overwrite it with zero, but with 0x5a and another time with 0xa5, before you zero it

What is the difference between "fill with null" and "fill with 0xa5"?

eaglerainbow:
> What is the difference between "fill with null" and "fill with 0xa5"?

It's a long time back, when I last had worked on that topic. IIRC the approach to reverse-engineer the original values in the memory cells with this kind of attack was based on the fact that it counted how often the bit in the cell was flipped. With the sequence 0xa5 --> 0x5a --> 0x0, you always flip the bit in the cell, plus it has the maximal hamming distance (in average) from the original value. 
I wouldn't take an oath on the details, but I just kept in mind back then, that this sequence was "the best you could do"...
You now may also argue whether with all this caching that we have today, all values also really reach the memory chips... But that's yet another story...

Issue #1328, opened 2016-12-14
IllegalArgumentException when calling isSentToMultiSig


alecalve:
Trying to decode the output script of [this transaction](https://blockchain.info/tx/514c46f0b61714092f15c8dfcb576c9f79b3f959989b98de3944b19d98832b58) with :
```java
String hex = "210378d430274f8c5ec1321338151e9f27f4c676a008bdf8638d07c0b6be9ab35c717c8bae";
Script script = new Script(Hex.decode(hex));
script.isSentToMultiSig();
```

results in the following exception:

```
java.lang.IllegalArgumentException: decodeFromOpN called on non OP_N opcode

	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:122)
	at org.bitcoinj.script.Script.decodeFromOpN(Script.java:577)
	at org.bitcoinj.script.Script.isSentToMultiSig(Script.java:707)
```

natzei:
The script is not a check-multisig. However, `isSentToMultiSig` was handling the wrong exception. #1413 should fix it.

schildbach:
#1413 was merged.

PR #1413, opened 2017-07-24
Script: fixed exception handling within isSentToMultiSig() method

natzei:
It fixes issue #1328

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/Script.java
+++ b/core/src/main/java/org/bitcoinj/script/Script.java
@@ -717,7 +717,7 @@ public class Script {
             }
             // First chunk must be an OP_N opcode too.
             if (decodeFromOpN(chunks.get(0).opcode) < 1) return false;
-        } catch (IllegalStateException e) {
+        } catch (IllegalArgumentException e) {
             return false;   // Not an OP_N opcode.
         }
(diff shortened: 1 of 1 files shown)


schildbach:
Do you know where this IllegalArgumentException is thrown? Maybe we should mark this in a code comment or narrow the try block down to the line?

alecalve:
@schildbach A partial stacktrace is available at #1328

natzei:
`decodeFromOpN` throws the exception: the method uses `Preconditions.checkArgument`.

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -717,7 +717,7 @@ public boolean isSentToMultiSig() {
             }
             // First chunk must be an OP_N opcode too.
             if (decodeFromOpN(chunks.get(0).opcode) < 1) return false;
-        } catch (IllegalStateException e) {
+        } catch (IllegalArgumentException e) {
             return false;   // Not an OP_N opcode.
```
natzei: If you consider this comment, the intended behaviour was to catch that exception.

schildbach:
Thanks a lot, merged!

PR #1417, opened 2017-07-26
Add witnesses to Transaction.toString()

oscarguindzberg:


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -857,10 +858,22 @@ public String toString(@Nullable AbstractBlockChain chain) {
                         if (in.isOptInFullRBF())
                             s.append(", opts into full RBF");
                     }
+                    if (this.hasWitness() && witnesses.get(i).getPushCount() > 0) {
+                        s.append("\n          ");
+                        s.append("witness  ");
+                        TransactionWitness witness = witnesses.get(i);
```
schildbach: Hmm, why not delegate this code to a TransactionWitness.toString(). And there, you can use Utils.SPACE_JOINER to print the pushes.

oscarguindzberg: Done + add test

schildbach:
Thanks, merged!

PR #1420, opened 2017-07-27
Support P2P SegWit

xenog:
Testing this is a bit more challenging than it is for previous changes. The code though is obviously correct and does not make any existing tests fail. I will work on tests but decided to open this request earlier so it may be merged sooner at your discretion.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/VersionMessage.java
+++ b/core/src/main/java/org/bitcoinj/core/VersionMessage.java
@@ -98,7 +98,7 @@ public class VersionMessage extends Message {
     public VersionMessage(NetworkParameters params, int newBestHeight) {
         super(params);
         clientVersion = params.getProtocolVersionNum(NetworkParameters.ProtocolVersion.CURRENT);
-        localServices = 0;
+        localServices = NODE_WITNESS;
         time = System.currentTimeMillis() / 1000;
         // Note that the Bitcoin Core doesn't do anything with these, and finding out your own external IP address
         // is kind of tricky anyway, so we just put nonsense here for now.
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/bitcoinj/core/VersionMessage.java
```diff
@@ -98,7 +98,7 @@ public VersionMessage(NetworkParameters params, byte[] payload) throws ProtocolE
     public VersionMessage(NetworkParameters params, int newBestHeight) {
         super(params);
         clientVersion = params.getProtocolVersionNum(NetworkParameters.ProtocolVersion.CURRENT);
-        localServices = 0;
+        localServices = NODE_WITNESS;
```
schildbach: I think this is wrong, we're not a full node in the sense that we can answer queries for blocks and transactions.

xenog: I haven't checked the corresponding code that announces newly-created transactions from the wallet to connected peers, but if these transactions are announced with an `inv` message including the transaction id, instead of a `tx` message, then the peer will follow up with a `getdata` message requesting the transaction. If the peer does not see that we are capable of providing witness data, it will request the transaction with an `type` field set to `MSG_TX`, which will be replied with the transaction serialised without witness data, consequently failing validation on a SegWit-capable node. The requesting node checks that we have `NODE_WITNESS` on our services field and requests `MSG_WITNESS_TX` when it wants to receive witness data for validation. This service flag is unrelated to being able to serve blocks or otherwise behaving as a full node.

schildbach: Broadcasting of txns simply sends them as a tx message, unannounced. See the comment in PeerGroup.broadcastTransaction(). But if you're sure announcing the flag is correct I'll of course trust you and merge this soon.

PR #1422, opened 2017-07-31
support bitcoin cash (BCC) signature

woodywang:
Add support for Bitcoin UAHF (Bitcoin Cash or BCC) signature algorithm, according to https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/doc/abc/replay-protected-sighash.md

While signing a transaction of BCC, you should provide the input value (unit satoshi). Example: 

```
// Sign an input whose value is 5 BCCs
Sha256Hash hash = tx.hashForSignature(i, scriptPubKey, Transaction.SigHash.ALL, false, 500000000L);
ECKey.ECDSASignature sig = ecKey.sign(hash);

// the last boolean 'true' argument indicates it uses replay-protected signature hash for BCC
TransactionSignature signature = new TransactionSignature(sig, Transaction.SigHash.ALL, false, true);
Script inputScript = ScriptBuilder.createInputScript(signature, ecKey);
```


Zurvarian:
Hello, 

I've been testing this patch against the branch 1.14.4 and found couple of issues, both of them related:

- When creating a transaction to be compatible with BCC blockchain (Bitcoin-ABC in this case), the transaction is considered as valid by the blockchain but Script.correctlySpends() fails due to the transaction hash not being canonical.
- When signing a transaction of P2SH type, Script.getSigInsertionIndex(), fails due to the transaction hash not being canonical.

I couldn't check yet if the issue affects master branch (1.15-Snapshot), but my initial guess is that it does.

schildbach:
@woodywang Thanks for your contribution. I admit I'm a bit skeptical this belongs into bitcoinj. A "bitcoincashj" fork would certainly be more appropriate.

There is however one usecase that could be supported within bitcoinj: spending all transactions outputs that were unspent before the fork to an address, signed with a BCH signature. Would you, or anyone else, implement something like this?

Also, we need tests for this PR.

amyodov:
As a potential user of bitcoin-cash supporting codebase, just curious – won't splitting it into something like a “bitcoincashj” fork make it harder to support and have about twice the same, mostly repeating, codebase, for the projects which opt to support both Bitcoin and Bitcoin-Cash simultaneously?
Also it would be just great to have the author hints/shortcuts how would you use this code to use bitcoinj in Bitcoin-Cash-supporting mode, what is needed to switch between the supported modes/forks.

woodywang:
Hi @schildbach , I've tried to write unit test cases for BCC supporting. But I met a problem since BCC signature hash algorithm needs the transaction input value, which is not included in the transaction object itself. When signing a raw transaction with with private keys offline, it seems impossible to get the input values, unless someone pass the value as an argument. Also, it means org.bitcoinj.script.Script.executeCheckSig and org.bitcoinj.script.Script.executeMultiSig need the input value to verify the  signed input. I have not a good idea where and how to store the input value. Could anyone provide ideas or advices?

woodywang:
@amyodov Thanks for your advice. Now I tend to favor amyodov's idea, so I renamed my fork repo to '[bitcoincashj](https://github.com/woodywang/bitcoincashj)'. Even though more than 99% codes are the same, BCC and original BTC should be treated as 2 different coins. We couldn't predict how BCC will change in the future. So I think a separate code repo is preferred, just like the relationship between bitcoin-abc and bitcoin.

Zurvarian:
@woodywang If that's the preferred approach, what about changing the base package of the product to be bitcoinjcash instead of bitcoinj? It would help to anybody who wants to have support for both, otherwise importing both libraries would cause class clash.

woodywang:
@Zurvarian That's right. The base package should be renamed to 'bitcoincashj' according to the name convention. (not 'bitcoinjcash', just a typo?)

schildbach:
@woodywang Regarding your problem with the input value I suggest looking at how the segwit branch deals with this problem. Segwit transactions sign input values too afaik.

schildbach:
Yes, for full Bitcoin Cash support (including peer discovery and difficulty adjustment changes) a separate fork in its own package is preferred. Bitcoinj has never supported multiple networks; it's kind of a basic design decision.

The usecase from my above comment would be useful to support in this branch however, imho.

Zurvarian:
@woodywang (not 'bitcoinjcash', just a typo?)

Yes, I just copied the name of current package and append cash at the end :)

amyodov:
> Bitcoinj has never supported multiple networks; it's kind of a basic design decision.

I have some, rather good, success in running Bitcoinj with mainnet and testnet simultaneously. Are some problems to be expected, on the same grade as it would be with running Bitcoinj in -Cash and non-Cash modes simultaneously?

magmahindenburg:
I think it would be reasonable to have support for verifying and signing transactions for Bitcoin Cash in bitcoinj, but leave the support for the Bitcoin Cash network to the fork.

woodywang:
Hi @Zurvarian , I've renamed the base package name. Would you like to join this project?

Zurvarian:
Hi @woodywang ,

Unfortunately I have too much in my plate to help right now, but if I get some spare time I'll consider joining.

Thanks for the offer, anyway.

AndyKD:
Can I get the bitcoinJ version, In which version BCC support is added?

Thanks,

Polve:
Fork an entire project when with a few lines of code it could handle both chains it seems a little excessive to me.

The only differences should be the different difficulty adjustment and the signatures (and, in the future, network magic).

Is there any way to extend NetworkParameters to handle such differences?

hegjon:
It seems like there is an active fork at https://github.com/bitcoinj-cash/bitcoinj

Danconnolly:
Yes, we (https://github.com/bitcoinj-cash/bitcoinj) aim to provide long term support for Bitcoin Cash.

Issue #1424, opened 2017-08-02
Implement sendheaders message (was: Segwit branch can't connect to peers)


btcontract:
Trying out the latest commit https://github.com/bitcoinj/bitcoinj/commit/5fa0262243b213e62a235b3a98dda5d7f2454707 and it drops peer connections, getting the following log output:

[https://pastebin.com/raw/xGgHWuZ1](https://pastebin.com/raw/xGgHWuZ1)

schildbach:
In the segwit branch we incremented our own version to 70012, causing core to send us that `sendheaders` message. I think we should – on the master branch – step by step implement the changes necessary for incrementing the version 70001→70012.

For this case, I think should just implement the SendHeaders message including serialization and deserialization, but then ignore the message in PeerGroup because we don't advertize new blocks.

globaltoken:
Same in #1429

schildbach:
Fixed via #1425.

PR #1425, opened 2017-08-02
Add "sendheaders" message

btcontract:
Fixes https://github.com/bitcoinj/bitcoinj/issues/1424
Should I also add some test for this?

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/SendHeadersMessage.java
+++ b/core/src/main/java/org/bitcoinj/core/SendHeadersMessage.java
@@ -1,0 +1,16 @@
+package org.bitcoinj.core;
+
+/**
+ * <p>A new message, "sendheaders", which indicates that a node prefers to receive new block announcements
+ * via a "headers" message rather than an "inv".</p>
+ *
+ * <p>See https://github.com/bitcoin/bips/blob/master/bip-0130.mediawiki</p>
+ */
+public class SendHeadersMessage extends EmptyMessage {
+    public SendHeadersMessage() {
+    }
+
+    // this is needed by the BitcoinSerializer
+    public SendHeadersMessage(NetworkParameters params, byte[] payload) {
+    }
+}
--- a/core/src/main/java/org/bitcoinj/core/SendHeadersMessage.java
+++ b/core/src/main/java/org/bitcoinj/core/SendHeadersMessage.java
@@ -1,3 +1,21 @@
+/*
+ * Copyright 2011 Google Inc.
+ * Copyright 2014 Andreas Schildbach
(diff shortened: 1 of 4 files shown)


schildbach:
Yes, please. At least a test for parsing one known sendheaders message would be nice.

Also, can you actually increase the CURRENT protocol version in this PR (to 70012 according to BIP130)? Otherwise the message isn't even tried to be used.

btcontract:
A raw message is extracted from Core in regtest mode, hope I did everything right. As to CURRENT version, is this it: https://github.com/bitcoinj/bitcoinj/blob/segwit/core/src/main/java/org/bitcoinj/core/NetworkParameters.java#L531 ? It's 70012 already.

globaltoken:
Hey,

I tried this Code from you and now I receive instead of this sendheaders issue the Issue that sendcmpct is missing :( 

More on #1429

Thank you.

btcontract:
@globaltoken it seems to be a kind of continuation of this issue, since 70012 Core sends "sendheaders" and since 70014 it also sends "sendcmpct" according to https://github.com/bitcoin/bips/blob/master/bip-0152.mediawiki. Perhaps this should be solved similarly but I'm not sure.

globaltoken:
@btcontract Okay thanks in Anyway. I used this Code to add sendcmpct and then I received 1 more error, I received that feefilter caused an error also. So sendheaders, sendcmpct and feefilter should be fixxed at once. ;)

Regards.

schildbach:
Oh sorry, I was under the impression this PR is against master. Actually I'd like to merge this way ahead of segwit.

core/src/test/java/org/bitcoinj/core/PeerTest.java
```diff
@@ -970,4 +973,20 @@ public void bitcoinSerializeToStream(OutputStream stream) throws IOException {
                     || (e instanceof SocketException && e.getMessage().equals("Socket is closed")));
         }
     }
+
```
schildbach: This could go into its own SendHeadersMessageTest class, to be consistent with the other *MessageTest classes.

core/src/main/java/org/bitcoinj/core/SendHeadersMessage.java
```diff
@@ -0,0 +1,16 @@
+package org.bitcoinj.core;
```
schildbach: Can you add the license here?

schildbach:
Thanks! I merged this against master with minor changes. The protocol version is now 70012.

schildbach:
ps. If anyone wants to go for even higher protocol versions, "feefilter" (BIP 133) would be next in line.

PR #1432, opened 2017-08-07
[segwit] Serialize transactions sent without using inv

xenog:
Extra SegWit-related improvements.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
+++ b/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
@@ -231,7 +231,7 @@ public class BitcoinSerializer extends MessageSerializer {
         } else if (command.equals("getutxos")) {
             return new GetUTXOsMessage(params, payloadBytes);
         } else if (command.equals("sendheaders")) {
-            return new SendHeadersMessage();
+            return new SendHeadersMessage(params, payloadBytes);
         } else {
             log.warn("No support for deserializing message with name {}", command);
             return new UnknownMessage(params, command, payloadBytes);
--- a/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
+++ b/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
@@ -231,7 +231,7 @@ public class BitcoinSerializer extends MessageSerializer {
         } else if (command.equals("getutxos")) {
             return new GetUTXOsMessage(params, payloadBytes);
         } else if (command.equals("sendheaders")) {
-            return new SendHeadersMessage(params, payloadBytes);
+            return new SendHeadersMessage();
         } else {
             log.warn("No support for deserializing message with name {}", command);
             return new UnknownMessage(params, command, payloadBytes);
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -1469,6 +1469,7 @@ public class Peer extends PeerSocketHandler {
                     this, toHash, chainHead.getHeader().getHashAsString());
         StoredBlock cursor = chainHead;
         for (int i = 100; cursor != null && i > 0; i--) {
+            //
             blockLocator.add(cursor.getHeader().getHash());
             try {
                 cursor = cursor.getPrev(store);
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -1469,7 +1469,6 @@ public class Peer extends PeerSocketHandler {
                     this, toHash, chainHead.getHeader().getHashAsString());
         StoredBlock cursor = chainHead;
         for (int i = 100; cursor != null && i > 0; i--) {
-            //
             blockLocator.add(cursor.getHeader().getHash());
             try {
                 cursor = cursor.getPrev(store);
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -277,6 +277,10 @@ public class Transaction extends ChildMessage {
         super(params, payload, 0, parent, setSerializer, length);
     }
 
+    /**
+     * Create a new version of this transaction that serializes using the pre-SegWit method,
+     * excluding witness data.
+     */
     public Transaction disableWitnessSerialization() {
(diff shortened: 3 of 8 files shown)


xenog:
Test is failing on bloom filter membership. Since bloom filters are probabilistic data structures and false positives are possible, this test may end up passing if rescheduled.

.gitignore
```diff
@@ -6,3 +6,4 @@ target
 *.iml
 *.chain
 *.spvchain
+*.wallet
```
schildbach: This change seems unrelated.

schildbach:
Thanks for implementing Bech32, it's great to see the puzzle being completed piece by piece!

I think the SegwitAddress class should somehow fit into the VersionChecksummedBytes hierarchy. My idea would be to separate the "legacy base58"-aspects of Address into a subclass LegacyAddress and then make SegwitAddress a subclass of Address. The reason is we will have many cases where we want to use different types of addresses in a list, or want to use the same UI elements, etc.

Also I think perhaps the Bech32 + SegwitAddress changes could go into a separate PR so we don't mix too many different changes.

xenog:
I agree that it would be nice to have to have them in the same hierarchy and behaving in a similar manner. I am not sure how to go about it since it seems that `VersionChecksummedBytes` is very tightly-coupled with base58 semantics. If you factor out the base58 stuff, I'll do the necessary changes to `SegWitAddress`.

core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
```diff
@@ -205,7 +205,7 @@ private Message makeMessage(String command, int length, byte[] payloadBytes, byt
         } else if (command.equals("getheaders")) {
             message = new GetHeadersMessage(params, payloadBytes);
         } else if (command.equals("tx")) {
-            message = makeTransaction(payloadBytes, 0, length, hash);
```
schildbach: I think this will be fixed once #1438 is merged. Would you like to review that PR?

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -276,6 +276,25 @@ public Transaction(
         super(params, payload, 0, parent, setSerializer, length);
     }
 
+    public Transaction disableWitnessSerialization() {
```
schildbach: I think this method is missing a JavaDoc.

schildbach:
Can you rebase this on the segwit branch? I lost track on what is left of this PR, and the title/description doesn't really tell.

xenog:
I rebased this branch as requested.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -1469,6 +1469,7 @@ private void blockChainDownloadLocked(Sha256Hash toHash) {
                     this, toHash, chainHead.getHeader().getHashAsString());
         StoredBlock cursor = chainHead;
         for (int i = 100; cursor != null && i > 0; i--) {
+            //
```
schildbach: This change seems unrelated.

core/src/test/java/org/bitcoinj/core/PeerTest.java
```diff
@@ -94,7 +94,7 @@ public void tearDown() throws Exception {
     }
 
     private void connect() throws Exception {
-        connectWithVersion(70001, VersionMessage.NODE_NETWORK);
+        connectWithVersion(70012, VersionMessage.NODE_NETWORK | VersionMessage.NODE_WITNESS);
```
schildbach: This change changes all the tests. I wonder if we should configure this class such that tests are run twice, once with witness/WitnessBlock and once without?

core/src/test/java/org/bitcoinj/core/PeerTest.java
```diff
@@ -373,7 +373,7 @@ public synchronized void onBlocksDownloaded(Peer p, Block block, @Nullable Filte
         List<InventoryItem> items = getdata.getItems();
         assertEquals(1, items.size());
         assertEquals(b2.getHash(), items.get(0).hash);
-        assertEquals(InventoryItem.Type.Block, items.get(0).type);
+        assertEquals(InventoryItem.Type.WitnessBlock, items.get(0).type);
     }
```
schildbach: Same as above.

core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
```diff
@@ -231,7 +231,7 @@ private Message makeMessage(String command, int length, byte[] payloadBytes, byt
         } else if (command.equals("getutxos")) {
             return new GetUTXOsMessage(params, payloadBytes);
         } else if (command.equals("sendheaders")) {
-            return new SendHeadersMessage(params, payloadBytes);
+            return new SendHeadersMessage();
```
schildbach: This change seems unrelated.

core/src/test/java/org/bitcoinj/core/PeerTest.java
xenog: Is the node itself ever expected to have the `NODE_WITNESS` flag disabled after the fork?

core/src/main/java/org/bitcoinj/core/Peer.java
xenog: One that fell through the cracks.

core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
xenog: Another one that fell through the cracks.

xenog:
I removed unrelated changes, but I think we should not disable the `NODE_WITNESS` flag anymore in the P2P code. SegWit is active in the main network, and it would be pointless to disable this flag until a protocol upgrade is done where the flag becomes implicit, or if we want to maintain Bitcoin Cash compatibility in the same code branch, in which case I wolud recommend adding Bitcoin Cash tests separately anyway.

core/src/test/java/org/bitcoinj/core/PeerTest.java
schildbach: I can't answer this question for sure, but I assume yes. Bitcoinj is used in networks that don't have segwit (yet). Also even in the Bitcoin P2P network peer versions <70012 will exist for quite some years.

One also could argue that simply because we have the code/ability to do so we should continue to have tests for it. It seems to me that we now have quite some `if (segwit) { … } else { … }` in code, and it feels wrong to me to drop tests we already have for the else branches.

schildbach:
Ok, don't worry. I merged. Thanks!

PR #1433, opened 2017-08-11
Option to decrypt private keys and seed on the fly…

### LINKED ISSUE #1431: Dumping private keys of a wallet with PIN
Having used [Schildbach's Bitcoin Wallet for Android](https://github.com/bitcoin-wallet/bitcoin-wallet), I wanted to dump my private keys, and so set out to do so using `wallet-tool` in this project. The wallet that was backed up had a "spending PIN" set, which, following the instructions set out [here](https://plus.google.com/101256420499771441772/posts/7ygWZgHLnFK), should be specified using the `--password=<SPENDING PIN>` flag when doing `wallet-tool dump --dump-privkeys`. However, even if the correct PIN is provided, the private keys are not dumped. In fact, output is no different than if the `--password` flag is omitted completely.

This was circumvented by removing the spending PIN on the Android device itself, and then creating a new backup from which to dump the private keys. While fine in my case, this is not an option if the device has been lost/stolen/etc., and an already existing backup is the only one to hand.

schildbach:
…if printing a wallet dump of an encrypted wallet.

Fixes #1431.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/core/ECKey.java
@@ -1224,15 +1224,15 @@ public class ECKey implements EncryptableItem {
 
     @Override
     public String toString() {
-        return toString(false, null);
+        return toString(false, null, null);
     }
 
     /**
      * Produce a string rendering of the ECKey INCLUDING the private key.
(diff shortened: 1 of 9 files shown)


esifoo:
Please help me .. I did not set the spending PIN. I cannot move my coins right now. Because it doesn't accept my PIN. Help me recover the PIN

PR #1435, opened 2017-08-12
Fix spelling of Tor

wodry:
[*'Tor is not spelled "TOR". Only the first letter is capitalized.'*](https://www.torproject.org/docs/faq.html.en#WhyCalledTor)


schildbach:
Thanks! Actually we don't use Tor any more (for now), so I just entirely removed that line.

PR #1437, opened 2017-08-16
wallet store: include witness data

sstone:


core/src/wallet.proto
```diff
@@ -398,9 +403,9 @@ message Wallet {
 
 /** An exchange rate between Bitcoin and some fiat currency. */
 message ExchangeRate {
-  // This much of satoshis (1E-8 fractions)…
+  // This much of satoshis (1E-8 fractions)?
```
schildbach: This change and the next one is unrelated.

core/src/wallet.proto
```diff
@@ -153,6 +156,8 @@ message TransactionInput {
   optional uint32 sequence = 4;
   // Value of connected output, if known
   optional int64 value = 5;
+  // script witness
+  optional ScriptWitness witness = 10;
```
schildbach: Any reason for 10? Why not continue with 6?

core/src/wallet.proto
```diff
@@ -22,7 +22,6 @@
  * - Endianness: All byte arrays that represent numbers (such as hashes and private keys) are Big Endian
  * - To regenerate after editing, run: mvn generate-sources -DupdateProtobuf
  */
-
```
schildbach: Unrelated change.

core/src/main/java/org/bitcoinj/core/TransactionWitness.java
```diff
@@ -64,4 +64,19 @@ public String toString() {
         }
         return Utils.SPACE_JOINER.join(stringPushes);
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        TransactionWitness that = (TransactionWitness) o;
```
schildbach: Nitpick: We usually use the variable name "other" at that point in equals() impls.

core/src/wallet.proto
sstone: I'll restore the original formatting

core/src/wallet.proto
sstone: Old habit when editing protobuf files, there's no real reason here I'll change it to 6

core/src/wallet.proto
sstone: I'll restore the original formatting

schildbach:
Merged. Thanks a bunch!

PR #1438, opened 2017-08-16
don't use the payload hash to set the tx hash

sstone:
when we receive a tx message from a peer, the hash is the hash of the payload.
if the tx is a segwit tx it is not equal to its bitcoin hash (which uses the
pre-segwit serialization format) so we don't set it, it will be computed later.


schildbach:
I wonder is this version of BitcoinSerializer.makeTransaction() actually being used with hash != null? At least in bitcoinj itself I see no direct usages besides the other makeTransaction() overloads.

I'm asking because probably we could also just remove the hash from that method. Being curious, how did you actually trip over this problem?

sstone:
I found this while testing my segwit-enabled wallet with bitcoind, I don't recall exactly what type of tests but I could see that segwit tx hashes of txs sent by bitcoind were wrong and that the problem came from makeTransaction().

schildbach:
Ok, let me reiterate. Please correct me if I'm wrong:

- Transaction.hash continues to be the transaction hash that is referenced in transaction inputs. (Is there a new more distinctive name for this hash with segwit?)
- BitcoinSerializer.makeTransaction() gets passed the messageHash, which is not equal to the transaction hash for segwit because of the witnesses attached to the message. If this is true, I think we could take the opportunity and rename the `hash` parameter accordingly.

sstone:
Yes, Transaction.hash is still used to reference tx outputs in tx inputs, I dont' think that there is a new name for it (but there is a new wtxhash that includes witness data and is used to compute a merkle root that is included in the coinbase tx of segwit blocks).
The `hash` parameter could be renamed to `payloadHash` ?  or maybe removed as you suggested above ?

schildbach:
Alright, I changed the comments a git, renamed the parameter and merged this to segwit.

I kept the optimization because I thought it might be necessary. Maybe we need a similar optimization for the witness hash in future, if segwit transactions become the norm.

I also rebased segwit on current master.

PR #1439, opened 2017-08-22
Peer: Drop Bitcoin Cash peers early, based on service bit 5.

schildbach:
They will fail the difficulty check anyway.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -564,6 +564,13 @@ public class Peer extends PeerSocketHandler {
             close();
             return;
         }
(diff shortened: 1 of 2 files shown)


luke-jr:
This seems inadvisable, as service bit 5 is unassigned, but may be used for a legitimate service in the future.

Bitcoin Core is doing something similar, but only until August 1st, 2018: https://github.com/bitcoin/bitcoin/pull/10982/files#diff-eff7adeaec73a769788bb78858815c91R1264

That seems like a sensible way to deal with the issue.

You might also consider including bit 7, which is similar.

schildbach:
I don't think the time when this bit will be reused can be predicted, so I prefer to not include a potential timebomb in the code.

PR #1440, opened 2017-08-22
Migrate from SpongyCastle back to BouncyCastle.

schildbach:
Google moved their broken BouncyCastle out of the way with Android 3.0
(see https://android.googlesource.com/platform/libcore/+/0ac85ead96f1ba7d35f3acadd154de4ef0a8fd87%5E%21/).


schildbach:
Just added a second commit that updates BC to 1.58. It also uses the ASN.1 integer parsing workaround.

schildbach:
Finally merged this one.

PR #1441, opened 2017-08-24
wallet: add optional 'segwit' mode

sstone:
In segwit mode, wallet will: 

- generate and manage p2sh-of-p2wpkh outputs instead of p2pkh outputs
- use a BIP39 path of m/1'/... instead of m/0'/...


schildbach:
Wow, this is a big PR! Thanks for starting on this what is probably the most difficult task with implementing segwit.

I read your PR twice over the past week, trying to understand. Here are my first comments:

1. The Wallet seems to save a "useSegwit" state in a tag. (Side note: tags are more meant for library users. The library itself could add a field.) Now, why should the wallet keep that state? Ideally, all key management should be delegated to its KeyChainGroup.

2. And KeyChainGroup doesn't need a "useSegwit" field either. It's feels actually misleading, since KeyChainGroups can have many DeterministicKeyChains, potentially with different address types. Afaik KeyChainGroup also delegates most key management to its various key chains.

3. The key chains themselves seem to not save its address type, at least I do not see any changes to the protobuf. I would expect an optional address type field in wallet.proto / DeterministicKey, a little bit like the existing isFollowing for married wallets.

schildbach:
By the way, the first commit of this PR could be dropped.

schildbach:
@troggy I think this PR touches married wallets, because (if I understand right) married wallets will also be extended to use segwit addresses (nested for now). Do you want to have a married wallets-focused look at the changes?

sstone:
I've tought about it and maybe I've missed something, but my point of view is that a wallet is either using segwit tx for all is outputs (including change output), or is not using segwit at all. I don't see a use case for wallets that would mix segwit/non-segwit utxos (uses would be much better off creating 2 different wallets). Hence the 'global' segwit flag. I don't see how KeyChainGroup could manage different types of DeterministicKeyChains since it seems that it is always the most recent chain that is active ?

Sorry what do you mean by "first commit of this PR could be dropped" ? Is it 400af1c ? (I picked it up when I rebased against segwit)

schildbach:
Yes, 400af1c can be dropped. If you look at it, it just inserts an empty comment line.

schildbach:
I understand your point of view, but I believe it's not realistic. We have tens of millions of non-segwit wallets created with bitcoinj, and lots of them would like a seamless migration to segwit. If segwit support were a boolean for the entire wallet, the only way to migrate is create a new segwit wallet, backup that wallet, spend all outputs to that new wallet (eating up a lot of fees) and remove the old wallet, hoping it will never receive any coins any more.

In contrast, if we would just add a segwit KeyChainGroup to the old wallet (re-using the seed from the previously active KeyChainGroup), coins can be migrated "on demand". This means as you make payments, non-segwit outputs will be spent first but any change sent to segwit outputs. Addresses exposed to other wallets (e.g. via payment requests / QR codes), will be segwit, while the wallet can still receive coins on the non-segwit addresses for those sender who still reuse addresses.

This discussion ties in to the "HD wallet format for Segwit" on the mailing list, your PR basically proposing another alternative I didn't cover there.

schildbach:
@sstone What do you think about my last comment?

sstone:
I think it makes sense but I'll need more time to look into it (I could not find a simple way to implement it this way and needed a "segwit-only" wallet quickly).
Do you mean adding a new KeyChainGroup to the wallet, or a new KeyChain to the wallet's existing KeyChainGroup ?

SamouraiDev:
I've tried signing mixed segwit/non-segwit utxos using the segwit branch. I was not able to manage it despite trying various approaches. Perhaps I missed something. If I understand correctly, this PR doesn't support it or foresee the need for it. Is mixed input signing planned for when segwit is merged into bitcoinj for production release ?

sstone:
mixing segwit/non-segwit utxos is indeed not supported (yet) by this PR. We needed a "non-malleable" wallet and this is what I thought was the simplest solution (all utxos are p2sh-of-p2wpkh). 
A strategy has been proposed (see above and the ML discussion Andreas refers to) for a smoother "switch to segwit" migration path but I haven't had time to look into the details yet.

schildbach:
@sstone I was thinking about adding a new KeyChainGroup for segwit.

schildbach:
I have rebased our segwit branch on master. This allows us to make use of arbitrary derivation paths as this has been implemented by Nelson Melina and Giuseppe Magnotta.

schildbach:
In two comments I used the term KeyChainGroup instead of DeterministicKeyChain. I edited the comments.

sneurlax:
P2SH-P2WKPH (3...) is not native segwit, but it's a good first step!

sneurlax:
Re: your latest comment @sstone and https://github.com/bitcoinj/bitcoinj/pull/1441#issuecomment-327469614 from @schildbach: perhaps bitcoinj could support generating addresses from legacy derivation paths like 44' and 49', while also supporting the best standard, 173'.

It's very curious that there's no reference implementation listed for Java in https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki !  Do you need any help with the Java implementation?

sneurlax:
Re: BIP173 lacking a Java implementation: see See https://github.com/sipa/bech32/pull/19 and https://github.com/sipa/bech32/pull/40 !  Two open pull requests with working an reviewed Java code.

ZhiyuHelloWorld:
This needs a rebase.

jonathancross:
@sstone can you please rebase / resolve conflicts?  Or maybe @sneurlax can reviv this?

jonathancross:
Ping @sstone :-)

rblaine95:
> Guys? I don't mean to be rude, but is there any reason for this to not be merged? There is several projects, including Bisq (decentralized exchange), which would really use this.

I think it's because there's conflicts with files.
OP needs to rebase and fix conflicts.
This may be an abandoned PR.

schildbach:
The original purpose of this PR (as described in at the top) is also covered by https://github.com/bitcoinj/bitcoinj/pull/1563. And that one is mainly missing the signing/spending the be ready for merging.

PR #1445, opened 2017-08-31
DeterministicSeed: decodeMnemonicCode() weaker access

shkschneider:
While developing a new mnemonic-related feature for the Android bitcoin-wallet,
I found that this method would expose a standard process of decoding mnemonic codes
and avoid repeating code from the clients than implement such logic.


schildbach:
The new public methods require a JavaDoc. Also, unit tests would be nice.

shkschneider:
Not sure what more would be required as javadoc for this. I also don't know how to add unit tests with this PR. Just a simple change really, just to unify decoding.

Giszmo:
`decodeMnemonicCode()` implies it decodes something and that that had to do with Mnemonic Codes. If the bitcoinj api exposed this, I would assume it would do something more than just split a string. Match the words to the word lists, fail if the checksum doesn't match, ... something along these lines.

PR #1446, opened 2017-09-05
Refactor script pattern matching

erasmospunk:
Standard script patterns are moved to a separate class. Additionally the Script class can handle customized patterns.

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/Script.java
+++ b/core/src/main/java/org/bitcoinj/script/Script.java
@@ -233,8 +233,7 @@ public class Script {
      * useful more exotic types of transaction, but today most payments are to addresses.
      */
     public boolean isSentToRawPubKey() {
-        return chunks.size() == 2 && chunks.get(1).equalsOpCode(OP_CHECKSIG) &&
-               !chunks.get(0).isOpCode() && chunks.get(0).data.length > 1;
+        return ScriptPattern.isSentToRawPubKey(chunks);
     }
 
     /**
@@ -244,12 +243,7 @@ public class Script {
      * way to make payments due to the short and recognizable base58 form addresses come in.
      */
     public boolean isSentToAddress() {
-        return chunks.size() == 5 &&
-               chunks.get(0).equalsOpCode(OP_DUP) &&
-               chunks.get(1).equalsOpCode(OP_HASH160) &&
-               chunks.get(2).data.length == Address.LENGTH &&
-               chunks.get(3).equalsOpCode(OP_EQUALVERIFY) &&
-               chunks.get(4).equalsOpCode(OP_CHECKSIG);
+        return ScriptPattern.isSentToAddress(chunks);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


schildbach:
Thanks for your contribution!

What's the idea behind this class? Are you planning to subclass it or something like that, and for what use?

Your ScriptPattern class is missing unit tests.

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -227,14 +257,20 @@ private void parse(byte[] program) throws ScriptException {
     }
 
     /**
+     * Returns true if the provided pattern matches this script
+     */
+    public boolean chucksMatch(ScriptPattern.ScriptPatternMatcher pattern) {
```
schildbach: Typo in method name?

core/src/main/java/org/bitcoinj/script/ScriptChunk.java
```diff
@@ -70,6 +71,10 @@ public boolean isPushData() {
         return opcode <= OP_16;
     }
 
+    public int getDataLength() {
```
schildbach: This public method is missing a JavaDoc.

Is it a good idea to return 0 if the chunk is no data? I'd expect `null`.

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -102,27 +104,55 @@
...
     }
 
     /**
      * Construct a Script that copies and wraps the programBytes array. The array is parsed and checked for syntactic
      * validity.
+     * @param pattern The pattern used form methods like isSentToAddress
      * @param programBytes Array of program bytes from a transaction.
      */
-    public Script(byte[] programBytes) throws ScriptException {
+    public Script(ScriptPattern pattern, byte[] programBytes) throws ScriptException {
```
natzei: Why not call `this(pattern, programBytes, 0)`?

erasmospunk: Good point.

core/src/main/java/org/bitcoinj/script/Script.java
erasmospunk: Yes, will change it.

core/src/main/java/org/bitcoinj/script/ScriptChunk.java
erasmospunk: The idea was to avoid checking for null data when pattern matching. In order to absolutely avoid any future problems, I will remove this and just change the pattern matching code.

erasmospunk:
Amended code.

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -93,6 +94,7 @@
 
     // The program is a set of chunks where each element is either [opcode] or [data, data, data ...]
     protected List<ScriptChunk> chunks;
+    private ScriptPattern pattern;
```
schildbach: Can this be final?

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -117,12 +118,38 @@ private Script() {
...
+     * Construct a Script that copies and wraps the programBytes array. The array is parsed and checked for syntactic
+     * validity.
+     * @param pattern The pattern used form methods like isSentToAddress
+     * @param programBytes Array of program bytes from a transaction.
+     */
+    public Script(ScriptPattern pattern, byte[] programBytes) throws ScriptException {
+        this(pattern, programBytes, 0);
+    }
+
+    public Script(ScriptPattern pattern, byte[] programBytes, long creationTimeSeconds) throws ScriptException {
```
schildbach: Maybe add a JavaDoc here, too? (Yes, it's very similar to the above.)

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -809,7 +799,7 @@ private static boolean castToBool(byte[] data) {
     /**
      * Cast a script chunk to a BigInteger.
      *
-     * @see #castToBigInteger(byte[], int) for values with different maximum
+     * @see #castToBigInteger(byte[], int, boolean) for values with different maximum
```
schildbach: These two changes seem to be unrelated? Maybe put them in a different commit, or even in a different PR?

core/src/main/java/org/bitcoinj/script/ScriptChunk.java
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright 2013 Google Inc.
  * Copyright 2014 Andreas Schildbach
+ * Copyright 2017 John L. Jegutanis
```
schildbach: Since you removed your change to ScriptChunk, can you remove this copyright addition too?

core/src/main/java/org/bitcoinj/script/Script.java
natzei: What's the reason you are storing the pattern? Am I missing some use-case?
I know anyone would define its - no bitcoin specific - patterns, but here `pattern` is used to answer methods like `isPayToScriptHash`, which are bitcoin specific.

erasmospunk: It is useful if you wanted to add a modified ruleset but in the case of the Bitcoin specific `isPayToScriptHash` you could return false.

core/src/main/java/org/bitcoinj/script/Script.java
erasmospunk: Added

core/src/main/java/org/bitcoinj/script/Script.java
erasmospunk: Removed

core/src/main/java/org/bitcoinj/script/ScriptChunk.java
erasmospunk: Removed

erasmospunk:
Amended code.

core/src/main/java/org/bitcoinj/script/Script.java
schildbach: @natzei Are you fine with this? I'd merge if you think your comment has been addressed.

natzei: Sorry for the delay.
@erasmospunk You didn't convinced me. After you added a modified ruleset within a Script, what's the purpose? The method `chunksMatch` you wrote is actually fine.
At the time, I don't see the necessity to store the pattern. Maybe I'm wrong, btw.

schildbach:
@erasmospunk I'm not sure about storing the ScriptPattern either. At least it increases the memory usage of power-user wallets. And we certainly would not mix Scripts of different currencies within one wallet, let alone Transaction.

I wonder if it should simply be a NetworkParameter field?

erasmospunk:
@schildbach @natzei Thanks for the feedback.

I removed the extra abstraction as the use case was pretty narrow.

erasmospunk:
As before I moved all the pattern matching code to the `ScriptPattern`, except for the `isPayToScriptHash`. Before, P2SH detection was done by checking the raw bytes of the script, now it checks the `ScriptChunk`s.

natzei:
@erasmospunk I really like your amend.

erasmospunk:
@natzei thanks

I'm not sure why travis is randomly failing on some unrelated code:
```
encryptDecrypt(org.bitcoinj.wallet.BasicKeyChainTest) Time elapsed: 0.098 sec <<< ERROR!
java.lang.IllegalArgumentException: private key exceeds 32 bytes: 376 bits
	at org.bitcoinj.wallet.BasicKeyChainTest.encryptDecrypt(BasicKeyChainTest.java:147)
```

[Before](https://travis-ci.org/bitcoinj/bitcoinj/builds/292222875) it was failing on `ECKeyTest::testEncryptionIsReversible` but with the same `private key exceeds 32 bytes: 376 bits` message.

natzei:
I think travis is failing randomly. 

The thrown exception is due to a check I introduced with PR #1373 (and #1404). Briefly, if a private key exceeds 32 bytes, it cannot be represented with the WIF format (before I did my PRs, the key was truncated).

I suppose some tests generate a key randomly via `new ECKey()`.

erasmospunk:
Rebased to the latest master.

natzei:
Up

schildbach:
Merged!

PR #1447, opened 2017-09-11
Change Wallet's saveNow() to be public instead of protected

ferostar:
Comment for the `public WalletFiles autosaveToFile(File f, long delayTime, TimeUnit timeUnit, @Nullable WalletFiles.Listener eventListener)` method reads: 

> You should still save the wallet manually when your program is about to shut down as the JVM will not wait for the background thread.

But being saveNow() `protected`, it's impossible to do it. By being marked as `public`, it's possible to manually save the wallet when the app terminates.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -1458,7 +1458,7 @@ public class Wallet extends BaseTaggableObject
     }
 
     /** If auto saving is enabled, do an immediate sync write to disk ignoring any delays. */
-    protected void saveNow() {
+    public void saveNow() {
         WalletFiles files = vFileManager;
         if (files != null) {
             try {


schildbach:
I'm not sure if that method is supposed to be public. Did you notice the `saveToFileStream()` and `saveToFile()` methods?

ferostar:
Yes, i saw them. That does it, in a way, but... kind of beats the point of using WalletAppKit i think. If you can make the class handle most parts of wallet saving automatic, shouldn't be force saving too? If the app is terminating shouldn't be possible to simply save? Might be confused because of the comment in Wallet.java i mentioned in my notes, though. If that's the case, i'm sorry.

schildbach:
I added a clarifying link to the JavaDoc of `autosaveToFile()`. I'm not sure what you mean by the WalletAppKit reference. It uses saveToFile() too when shutting down, so no data should be lost.

schildbach:
Closing this, feel free to comment if you still think you need access to the `saveNow()` method.

PR #1460, opened 2017-10-02
Set DNS seeds according to current values specified in chainparams.cpp

prusnak:
This fix updates the DNS seed values to match the ones specified in chainparams.cpp of current bitcoind.

### DIFF
--- a/core/src/main/java/org/bitcoinj/params/MainNetParams.java
+++ b/core/src/main/java/org/bitcoinj/params/MainNetParams.java
@@ -74,11 +74,9 @@ public class MainNetParams extends AbstractBitcoinNetParams {
                 "seed.bitcoin.sipa.be",         // Pieter Wuille
                 "dnsseed.bluematt.me",          // Matt Corallo
                 "dnsseed.bitcoin.dashjr.org",   // Luke Dashjr
-                "seed.bitcoinstats.com",        // Chris Decker
-                "seed.bitnodes.io",             // Addy Yeow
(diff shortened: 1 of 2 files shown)


schildbach:
Why do you remove some seeds? They seem to be doing fine.

prusnak:
In 30 days they will be announcing nodes that are incompatible with nodes announced by other dns seeds. See also https://github.com/bitcoinj/bitcoinj/pull/1461

schildbach:
That's not as bad as it seems. Bitcoinj uses the PoW to sort this out (plus implemented consensus rules in full verification mode). Given the centralized nature of DNS seeds, diversity is a good thing.

sturles:
@schildbach PoW is just proof of work, not proof of validation.  SPV clients should be on the minority chain, if the majority chain is invalid.  There is a very real and current threat from miners representing a majority of the hashrate to produce invalid blocks to steal BTC from SPV wallets.  It accomplishes this by replaying transactions from the invalid chain on the valid chain, and by presenting transactions with inputs from the invalid chain as valid to SPV clients.  SPV can not do full validation of blocks, and will be vulnerable to this attack regardless, but it can do basic countermeasures like avoid contacting nodes which are know to be bad.

I would prefer that the seed node related to a blockchain analysis company is removed, and this patch accomplishes this as well.

schildbach:
I'm closing this.

schildbach:
However I think we should go through the seeds and remove those not working.

PR #1461, opened 2017-10-02
Don't follow incompatible Segwit2x chain.

prusnak:
In 30 days there will be nodes in the network that will have different consensus rules. It's wise to disconnect from them, because otherwise user might get inconsistent/unexpected behaviour.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -564,8 +564,10 @@ public class Peer extends PeerSocketHandler {
             close();
             return;
         }
-        if ((vPeerVersionMessage.localServices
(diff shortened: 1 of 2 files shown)


schildbach:
This will most likely break SPV clients by forcing them on a minority chain.

Also, service bit 7 will not be very effective for what you want to archive, given the heterogenous usage and masquerading of this bit. Have you considered implementing a consensus rule for full verifying mode instead?

lichtamberg:
utAck - a first step has to be done.

nomnombtc:
As a user that connects his bitcoinj based app to a segwit2x node, I would prefer if it would not disconnect itself. Not sure if this is possible, but maybe this can be made optional for the no2x people.

prusnak:
@nomnombtc IMHO you should used a patched version which connects to Segwit2x nodes ONLY. I think that having one codebase which connects to both nodes is not a very good idea.

schildbach:
I think this is obsolete now.

PR #1469, opened 2017-10-23
Add a french translation for contexts.md documentation

cozouf:
Hello,

I would contribute to bitcoinj but I think my skills in Java programming language aren't high enough to contribute effectively.

Then, I want to contribute as a document writter. I have translated the Contexts.md file from English to French (my mother tongue). 
I guess some people would like to have more accurate information on that file. 
I noticed this document is an old document but it is always better to have a clear and well documented project to help other contributors
However, as you can notice in this file, some words are still in English because there are no equivalents of these words in French (ex: Churn).

Sincerely,
cozouf


schildbach:
Thanks a lot for your contribution. I admit I'm a bit hesitant to support localized versions of bitcoinj. Keeping localizations up-to-date places a huge burden on the project. We're even struggling to keep the code samples on bitcoinj.github.io up-to-date.

PR #1472, opened 2017-10-26
Translation of README.md to french language

LucieJ95:
The translation of the README.md file would aim at helping french native speaker to understand the main goal of bitcoinj repository, considering that the basic concepts could be misunderstood, based on the complexity of the project.  Indeed, many french native speakers are reluctants when it comes to understand important concepts using english language, that is why a translation would be a way of encouraging them to contribute. Even though most of them know english, when it comes to such a project, it might be good to have a clear idea of how it works.
Kinds regards.

### DIFF
--- a/README_FR.md
+++ b/README_FR.md
@@ -3,9 +3,9 @@ La bibliothèque bitcoinj est une implémentation du protocole Bitcoin en Java, 
 
 ## Technologies
 
-•	Java 6 pour les modules principaux, Java 8 pour le reste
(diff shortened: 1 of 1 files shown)


schildbach:
Thanks for your contribution. I admit I'm a bit hesitant to support localized versions of bitcoinj. Keeping localizations up-to-date places a huge burden on the project. We're even struggling to keep the code samples on bitcoinj.github.io up-to-date.

PR #1473, opened 2017-10-26
Fix minimal push data for -1

natzei:
Pushing -1 via method `ScriptBuilder#number(long n)` cause a not minimal representation.

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/ScriptBuilder.java
+++ b/core/src/main/java/org/bitcoinj/script/ScriptBuilder.java
@@ -112,11 +112,7 @@ public class ScriptBuilder {
      * shortest encoding possible.
      */
     public ScriptBuilder number(long num) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/script/ScriptBuilder.java
```diff
@@ -177,6 +177,8 @@ protected ScriptBuilder bigNum(int index, long num) {
 
         if (num == 0) {
             data = new byte[0];
+        } else if (num == -1) {
```
schildbach: Hmm, this change violates the contract from the method comment, which guarantees to always use pushdata (hence the "big"). I think the fix needs to go into the two number() methods.

natzei:
Thanks, I agree. I amended the commit.

schildbach:
Merged, with a minimal change (an assert message).

Thanks again!

PR #1479, opened 2017-11-03
Fix DumpedPrivateKey.toBase58() if the key is compressed

natzei:
DumpedPrivateKeyTest#roundtripBase58_compressed() failed before this fix.
I've also added a new test, just to cover another case.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/DumpedPrivateKey.java
+++ b/core/src/main/java/org/bitcoinj/core/DumpedPrivateKey.java
@@ -30,7 +30,6 @@ import javax.annotation.Nullable;
  * the last byte is a discriminator value for the compressed pubkey.
  */
 public class DumpedPrivateKey extends VersionedChecksummedBytes {
-    private boolean compressed;
 
     /**
      * Construct a private key from its Base58 representation.
@@ -50,7 +49,6 @@ public class DumpedPrivateKey extends VersionedChecksummedBytes {
     // Used by ECKey.getPrivateKeyEncoded()
     DumpedPrivateKey(NetworkParameters params, byte[] keyBytes, boolean compressed) {
         super(params.getDumpedPrivateKeyHeader(), encode(keyBytes, compressed));
-        this.compressed = compressed;
     }
 
     private static byte[] encode(byte[] keyBytes, boolean compressed) {
@@ -72,12 +70,7 @@ public class DumpedPrivateKey extends VersionedChecksummedBytes {
         super(encoded);
         if (params != null && version != params.getDumpedPrivateKeyHeader())
             throw new WrongNetworkException(version, new int[]{ params.getDumpedPrivateKeyHeader() });
-        if (bytes.length == 33 && bytes[32] == 1) {
-            compressed = true;
-            bytes = Arrays.copyOf(bytes, 32);  // Chop off the additional marker byte.
-        } else if (bytes.length == 32) {
-            compressed = false;
-        } else {
+        if (bytes.length != 32 && bytes.length != 33) {
             throw new AddressFormatException("Wrong number of bytes for a private key, not 32 or 33");
         }
     }
@@ -86,8 +79,11 @@ public class DumpedPrivateKey extends VersionedChecksummedBytes {
      * Returns an ECKey created from this encoded private key.
      */
     public ECKey getKey() {
-        final ECKey key = ECKey.fromPrivate(bytes);
-        return compressed ? key : key.decompress();
+        return ECKey.fromPrivate(Arrays.copyOf(bytes, 32), isCompressed());
+    }
+
+    public boolean isCompressed() {
+        return bytes.length==33 && bytes[32] == 1;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


schildbach:
Interesting, thanks! I wonder what's the consequence of this. We've been printing base58-encoded private keys for a long time, e.g. Wallet.toString() and in Bitcoin Wallet's old (pre-2014) backup format. Does this mean these were the wrong keys?

core/src/test/java/org/bitcoinj/core/DumpedPrivateKeyTest.java
```diff
@@ -66,7 +67,22 @@ public void cloning() throws Exception {
...
+    public void roundtripBase58_compressed() throws Exception {
+        String base58 = "cSthBXr8YQAexpKeh22LB9PdextVE1UJeahmyns5LzcmMDSy59L4";     // 33 bytes, compressed == true
+        assertEquals(base58, DumpedPrivateKey.fromBase58(null, base58).toBase58());
+    }
+
+    @Test(expected=AddressFormatException.class)
+    public void roundtripBase58_invalidCompressed() {
+        String base58 = "5Kg5shEQWrf1TojaHTdc2kLuz5Mfh4uvp3cYu8uJHaHgfTGUbTD";     // 33 bytes, compressed == false
+        byte[] bytes = Base58.decodeChecked(base58);
+        bytes = Arrays.concatenate(bytes, new byte[]{0});
```
schildbach: I wonder is there an alternative to using this spongycastle method?

schildbach: E.g. `com.google.common.primitives.Bytes.concat(byte[]...)`.

natzei: Oh, sorry. I thought it was on java.util.Arrays. I'll change it.

natzei:
It depends. Are private keys compressed by default?

schildbach:
Probably not, but I admit I'm confused. Previously we always talked about *public* keys being compressed or not, and indeed public keys are compressed by default (see d268270d7b864bf35234633a4ac8264d585882c6). *Private* keys being compressed is new to me.

schildbach:
FWIW, here is an old (worthless) Bitcoin Wallet backup:

``# KEEP YOUR PRIVATE KEYS SAFE! Anyone who can read this can spend your Bitcoins.
929tTVp2tq7MmDMnFk4j73TZMvsWe66DdseeTArvUw2KSrpcsVc 2011-12-05T11:59:29Z
939o9ZK9zTp2Fi7B1km8EZxQPFUU8EuVUZzfBmLWPyXpGFpSF3n 2012-05-04T16:19:02Z``

schildbach:
Ok, I now understand that private keys do not have that compressed/uncompressed distinction, but their base58-encoded form carries the information if their corresponding public key is compressed or not.

The private keys from the above examples correspond to uncompressed public keys. I'll investigate further.

schildbach:
I just noticed that Travis didn't pick up your PR. In fact it doesn't compile, there are the following errors:

```Tests in error: 
  ECKeyTest.base58Encoding_stress:206 » IllegalArgument Private keys must be 32 ...
  ECKeyTest.roundTripDumpedPrivKey:362 » IllegalArgument Private keys must be 32...
  BIP38PrivateKeyTest.bip38testvector_compression_noEcMultiply_test1:77 » IllegalArgument
  BIP38PrivateKeyTest.bip38testvector_compression_noEcMultiply_test2:86 » IllegalArgument
```

natzei:
Oh, sorry. I didn't see these errors, my bad. I saw these checks are from BIP32. I'm further investigating.

natzei:
Ok, I found a problem in my implementation of DumpedPrivateKey#toBase58(). 

I propose a little refactoring for the class.
The deprecated constructor chop off the extra byte for compression flag, removing it from the superclass attribute. I think it shouldn't and should avoid the `boolean compressed` attribute.

I'll amend in a while.

natzei:
Amend.

The remaining compilation errors are the well-known errors on the key length that exceeds 32 bytes (I'm looking for a workaround).

schildbach:
Your refactoring looks good to me. I think getKey() needs to chop off the marker byte if the key marked as compressed.

natzei:
Sorry, it was wrong. I've also added two new tests to cover that case.

core/src/main/java/org/bitcoinj/core/DumpedPrivateKey.java
```diff
@@ -86,20 +79,25 @@ public DumpedPrivateKey(@Nullable NetworkParameters params, String encoded) thro
      * Returns an ECKey created from this encoded private key.
      */
     public ECKey getKey() {
-        final ECKey key = ECKey.fromPrivate(bytes);
-        return compressed ? key : key.decompress();
+        final ECKey key = ECKey.fromPrivate(Arrays.copyOf(bytes, 32));
```
schildbach: Rather than creating a compressed key and then decompressing it, shall we use the `ECKey.fromPrivate(bytes, compressed)` constructor in the first place?

natzei: Sure, of course!

schildbach:
One last suggestion: shall we perhaps rename the new `isCompressed()` to `isPubKeyCompressed()`? If yes, you don't need to amend this, I can simply rename before merging.

natzei:
I agree, at least it's not misleading.

schildbach:
Ok, I renamed the method, added two asserts for it and merged. Thanks a lot for this PR!

PR #1490, opened 2017-12-26
Avoid NPE in Block#hasTransactions

dwhjames:
Field is nullable;

```java
@Nullable List<Transaction> transactions;
```


BigAdam2005:
Small change but usable.
@schildbach should merge it;

schildbach:
Merged.

PR #1491, opened 2017-12-29
Enable jitpack builds with jdk7 (android apps)

acrespo:
Currently if someones tries to add bitcoinj as a dependency via [Jitpack](https://jitpack.io/), it compiles using Jitpack's default, Java 8. This breaks Android apps builds. I think we should find a way to workaround this as it forbids people wanting to try out latest changes (e.g segwit) until there is a release published. Otherwise, I'm leaving my fork for people experiencing the same problem.


dbkingsb:
I'm surprised that it breaks when building with Java 8. I can build all the modules with 1.8.0_131. What build error are you seeing?

Does jitpack not honor the maven-compiler-plugin configuration in the core pom that specifies 1.7?

acrespo:
I'm sorry if I wasn't clear, it's not that the bitconj module fail to build with Java 8. It's just that Android doesn't support Java 8 so dependencies must be compiled with 1.7. 

And referring to your question, no, there's some kind of problem because jitpack build ALL modules with the same jdk, even if you only want to use just one module (and using Java 8 does not serve for Android apps, and using Java 7 breaks other modules like examples).

schildbach:
Isn't it possible to tell jitpack to cd into core and build there? It would achieve the same thing, without changing any pom.

Issue #1361, opened 2017-04-01
BIP39 passphrase not normalized


schildbach:
@erasmospunk notified me that `MnemonicCode.toSeed(words, passphrase)` doesn't normalize the passphrase as spec'ed in BIP39. John, do you have a testcase for us? The english tests all use "TREZOR" as a passphrase. I know there are Japanese test cases but they also require the Japanese wordlists which bitcoinj currently does not support.

Issue #10, opened 2014-02-07
Add NFKD normalization form to BIP39 code


slush0:
Hello,

during BIP39 acceptance discussion we've found that it would be helpful to better specify the way how to handle UTF-8 strings. Please check https://github.com/bitcoin/bips/pull/17 for more details and for relevant changes in the draft.

This link will probably help you with NFKD normalization: http://docs.oracle.com/javase/tutorial/i18n/text/normalizerapi.html

I think we're quite near to find a concensus across developers and BIP39 will turn into Accepted state quite soon. However I recommend to wait with the changes until it will be clear that everybody is happy with NFKD and there won't be any further changes in the paper.

PR #1493, opened 2018-01-01
Mnemonic code and passphrase normalization, expanded tests

dbkingsb:
Resolve issues #1361 and #10.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/MnemonicCode.java
+++ b/core/src/main/java/org/bitcoinj/crypto/MnemonicCode.java
@@ -17,12 +17,11 @@
 
 package org.bitcoinj.crypto;
 
+import com.google.common.base.Stopwatch;
 import org.bitcoinj.core.Sha256Hash;
 import org.bitcoinj.core.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import com.google.common.base.Stopwatch;
 
 import java.io.BufferedReader;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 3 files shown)


dbkingsb:
Close / Reopen to restart build.

erasmospunk:
@dbkingsb as this bug was present for many years, could we add a method for users to recover funds by not normalising? Also what about a detection when a password uses unsafe characters?

schildbach:
I just merged the passphrase null check (but used checkNotNull).

schildbach:
Perhaps recovery could happen behind the scenes, simply by checking all possible variants of the passphrase? But I doubt this affects many users.

dbkingsb:
> Perhaps recovery could happen behind the scenes, simply by checking all possible variants of the passphrase?

We could derive seeds for the 4 variants (passphrase as-is / normalized, salt as-is / normalized), but is there a way to determine which resulting seed is the intended seed?

erasmospunk:
@dbasch there isn't a way to know for sure. It would be nice to have a detection system if the provided password is different than the normalised form.

Maybe we could have a second ```toSeed(List<String> words, String passphrase, boolean normalise)``` and the normal method calls ```toSeed(words, passphrase, true)```?

Also we could have a utility method like ```MnemonicCode#needsNormalisation``` to check if the input data needs normalisation.

I forgot to add that [libwally](https://github.com/ElementsProject/libwally-core/blob/master/src/bip39.c#L202) does not perform any normalisation as well (it expects that the library user does it).

dbkingsb:
I'm going to close this pull request and split this branch so that we can get tests in place for the partial merge; see pull request #1562.

PR #1494, opened 2018-01-02
ScriptBuilder: add methods opTrue() and opFalse()

natzei:


schildbach:
Looks good to me. However do we want the "op" prefix on ScriptBuilder method names? I know we cannot use just "true" or "false".

natzei:
I was also considering `addTrue`/`addFalse`, but I ended up preferring the op prefix. However, any other choice is fine for me.

natzei:
I've seen that some ScriptBuilder tests are located in ScriptTest. Should testOpTrue and testOpFalse be placed there?

schildbach:
No actually I noticed the same thing and was considering to move those tests to ScriptBuilderTest after I merge your PR.

PR #1495, opened 2018-01-04
ScriptChunk: Add method toByteArray(), remove method write().

natzei:
Not really exciting PR: basically I did not like passing a stream as parameter.
I think the class can be easier to test, maybe :)

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/ScriptChunk.java
+++ b/core/src/main/java/org/bitcoinj/script/ScriptChunk.java
@@ -21,6 +21,8 @@ import org.bitcoinj.core.Utils;
 import com.google.common.base.Objects;
 
 import javax.annotation.Nullable;
+
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Arrays;
@@ -136,6 +138,17 @@ public class ScriptChunk {
         } else {
             stream.write(opcode); // smallNum
         }
+    }
+
+    public byte[] toByteArray() {
+        ByteArrayOutputStream stream = new ByteArrayOutputStream();
+        try {
+            write(stream);
+        } catch (IOException e) {
+            // Should not happen as ByteArrayOutputStream does not throw IOException on write
+            throw new RuntimeException(e);
+        }
+        return stream.toByteArray();
     }
 
     @Override
--- a/core/src/test/java/org/bitcoinj/script/ScriptChunkTest.java
+++ b/core/src/test/java/org/bitcoinj/script/ScriptChunkTest.java
@@ -16,15 +16,23 @@
 
 package org.bitcoinj.script;
 
+import static org.bitcoinj.script.ScriptOpCodes.OP_0;
+import static org.bitcoinj.script.ScriptOpCodes.OP_IF;
 import static org.bitcoinj.script.ScriptOpCodes.OP_PUSHDATA1;
 import static org.bitcoinj.script.ScriptOpCodes.OP_PUSHDATA2;
 import static org.bitcoinj.script.ScriptOpCodes.OP_PUSHDATA4;
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
+import java.util.Random;
+
 import org.junit.Test;
+import org.spongycastle.util.Arrays;
 
 public class ScriptChunkTest {
+
+    private static final Random RANDOM = new Random(42);
 
     @Test
     public void testShortestPossibleDataPush() {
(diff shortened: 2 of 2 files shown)


natzei:
@schildbach I figured out that other methods, also involving Script and ScriptBuilder, could me rearranged.
I'm currently working on a branch, trying to keep the commits self contained. 
I was wondering: do you prefer several small PRs (1 or 2 commits) or just one (7+ commits) in which discuss the overall changes?
Thank you!

schildbach:
If the change is expected to need broader discussion, maybe ask for feedback first on the mailing list.

Generally if you think your change can be separated into parts that make sense on its own then I prefer independant PRs.

As you say commits should be self-contained. Bugfixes to the (still unmerged) PR should be squashed into the respective commit rather than added as another commit on top.

natzei:
Then I'll do several PRs. Thank you

erasmospunk:
@natzei it is preferable to create a helper method that passes a ByteArrayOutputStream and returns an array. It is not recommended removing public APIs like this.

natzei:
I agree, I was not considering that it could break the API. 

Then, I suggest to deprecate the method write() and keep the method toByteArray().

core/src/main/java/org/bitcoinj/script/ScriptChunk.java
```diff
@@ -108,33 +108,37 @@ public boolean isShortestPossiblePushData() {
...
-                stream.write(0xFF & data.length);
-                stream.write(0xFF & (data.length >> 8));
+                res.add((byte) OP_PUSHDATA2);
+                res.add((byte) (0xFF & data.length));
+                res.add((byte) (0xFF & (data.length >> 8)));
             } else if (opcode == OP_PUSHDATA4) {
                 checkState(data.length <= Script.MAX_SCRIPT_ELEMENT_SIZE);
-                stream.write(OP_PUSHDATA4);
-                Utils.uint32ToByteStreamLE(data.length, stream);
+                res.add((byte) OP_PUSHDATA4);
```
schildbach: I think you should be using `uint32ToByteArrayLE()` here. Kind of inlining the logic from the method is a step back imho.

core/src/main/java/org/bitcoinj/script/ScriptChunk.java
```diff
@@ -108,33 +108,37 @@ public boolean isShortestPossiblePushData() {
...
+                res.add((byte) (0xFF & data.length));
+                res.add((byte) (0xFF & (data.length >> 8)));
+                res.add((byte) (0xFF & (data.length >> 16)));
+                res.add((byte) (0xFF & (data.length >> 24)));
             } else {
                 throw new RuntimeException("Unimplemented");
             }
-            stream.write(data);
+            return Bytes.concat(Bytes.toArray(res), data);
```
schildbach: Imho this has the potential of a performance regression, as it copies around res and data (multiple times).

schildbach:
I think this change should be seen in conjunction with `Script.getProgram()`. I doubt `ScriptChunk.write()` is used much externally, so maybe `Script` is the only user.

The main problem with using the IO framework is the checked IOException imho. Other than that, it's probably quite performant and does its job. `ArrayList` also feels a bit weird to me, how performant is it, also considering the `Bytes.concat()` and `Bytes.toArray()` you do on top of it? Maybe using NIO (ByteBuffer?) would be a modern alternative to IO?

natzei:
Thanks for your feedbacks. I changed the implementation in favour of `ByteBuffer`, which I agree is more "low-level" than using an `ArrayList`.

I've also added some tests for the method.

core/src/main/java/org/bitcoinj/script/ScriptChunk.java
natzei: The drawback of using `uint32ToByteArrayLE` is that a new array must be created and passed. I preferred to keep the code more readable inlining the logic, since the same logic is already done with the case of PUSHDATA2.

schildbach: I took the opportunity and migrated all remaining manual bitstuffing to Utils. See #1518.

If we stay with ByteBuffer (it was just a random idea out of my head), then probably we should add the appropriate utils methods too. Bitstuffing has been implemented incorrectly many times in the past, which is why I'm resitant to duplicate it in code.

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -152,7 +152,7 @@ public String toString() {
                 return Arrays.copyOf(program, program.length);
             ByteArrayOutputStream bos = new ByteArrayOutputStream();
             for (ScriptChunk chunk : chunks) {
-                chunk.write(bos);
+                bos.write(chunk.toByteArray());
```
erasmospunk: Here we will be making many allocations.

schildbach:
Today I spoke to @cketti about this PR. He's not convinced about `ByteBuffer` either and was recommending OkIo (source/sink).

erasmospunk:
@natzei the advantage of the OutputStream is that you can write to it without allocating and this can add up when you are processing a lot of data.

Consider adding a method like this if you need it:

```
public byte[] toByteArray() {
    ByteArrayOutputStream stream = new ByteArrayOutputStream();
    try {
        write(stream);
    } catch (IOException e) {
        // Should not happen as ByteArrayOutputStream does not throw IOException on write
        throw new RuntimeException(e);
    }
    return stream.toByteArray();
}
```

natzei:
@schildbach Thank you for your interest, I'd like to know something more about why he/she is not convinced about `ByteBuffer`. I've never used Okio, so I cannot judge it.

@erasmospunk Yes, of course it should be more efficient. However, scripts are never so long to be considered "a lot of data" and the overhead introduced could be negligible (my point is that premature optimization sometimes leads to wrong design choices).
I like your suggestion though, it could be a good compromise that make the class more testable (and `write` could became only visible within the same package).

schildbach:
@natzei ByteBuffer seems to be more designed for asynchronous I/O.

I also know OkIo only from OkHttp and Moshi. Both are using OkIo to minimize buffer copying and because it has a very [straightforward API](http://square.github.io/okio/1.x/okio/okio/Buffer.html). It even has support for big and little endian short/int/long (sadly, it doesn't support unsigned types though). It also has [ByteString](http://square.github.io/okio/1.x/okio/okio/ByteString.html), basically an immutable byte array!

Regarding this PR, I think for now we should just go with the solution @erasmospunk was suggesting. Sorry about the "ByteBuffer detour". If we want to go OkIo, I think we should plan this on a broader scale: there are many classes that are kind of similar to ScriptChunk.

natzei:
Totally agree. I changed my commit and left the addition test case :)

schildbach:
Merged!

PR #1497, opened 2018-01-08
ScriptChunk: Encapsulation of attributes 'opcode' and 'data'

natzei:


schildbach:
Is there any discussion somewhere around this change? The fields were final before, so encapsulation is not strictly needed I'd say.

natzei:
Hi Andreas. There is no discussion unfortunately. My point is that ScriptChunk should be an immutable class, like String, Integer and so on, and making public attributes `final` is not sufficient.

Advantages of encapsulation and immutability are well known in OOP and I'd just like to improve Bitcoinj in that sense.

schildbach:
Ok I see your point. I'm also a proponent of immutability.

For `opcode` I'd argue the current version is already immutable.
For `data` I'm a bit scared of performance loss due to array duplication. ScriptChunks are amongst the most used objects in bitcoinj.

natzei:
You're right. I moved the array copy to the constructor.
> For opcode I'd argue the current version is already immutable.

I agree, making it `final` makes it not updatable (same for `data` field, now copied).
Now it's just a matter of personal preference using encapsulation or not. If you prefer, I'll remove it.

schildbach:
I think for `opcode`, all encapsulation would add is API breakage and a little overhead. While I'm not afraid of breaking an API, it should be worth it. IMHO in this case it's not: I see no advantage.

For `data`, copying in the array in the constructor is unfortunately not enough to make it immutable. A user of `getData()` (or `.data`) could still change the array. I'm still ok with adding the `Array.copyOf()` as it's at least a little bit of progress at the expense of only a small overhead.

What do others think?

natzei:
> I think for opcode, all encapsulation would add is API breakage and a little overhead. While I'm not afraid of breaking an API, it should be worth it. IMHO in this case it's not: I see no advantage.

Your repo, your rules :)

> For data, copying in the array in the constructor is unfortunately not enough to make it immutable.

Yes, I've updated it adding `Array.copyOf()` both in the constructor and in the getter method.

core/src/test/java/org/bitcoinj/script/ScriptChunkTest.java
```diff
@@ -46,4 +50,26 @@ public void testShortestPossibleDataPush() {
         assertFalse("push of 255 bytes", new ScriptChunk(OP_PUSHDATA2, new byte[255]).isShortestPossiblePushData());
         assertFalse("push of 65535 bytes", new ScriptChunk(OP_PUSHDATA4, new byte[65535]).isShortestPossiblePushData());
     }
+
+    @Test
+    @Ignore
+    public void testGetData() {
```
natzei: I've evaluated the overhead of invoking `getData()` in this way. It seems that no overhead is introduced.

natzei:
Rebased

natzei:
I added the methods getData() and getDataValue():
     - if `isPushData()==true`, `getData()` returns the byte array that is pushed on the stack, null otherwise
     - if `isPushData()==true`, `getDataValue()` returns the decoded value of data, null otherwise


Finally, I've removed the test to evaluate how much overhead `getData()` introduces.
For the curious ones, this is the code to evaluate the overhead:
```java
    int data_length = 1000;
    int iterations = 100_000;
    
    Random rnd = new Random(42);
    final byte[] data = new byte[data_length];
    rnd.nextBytes(data);
    ScriptChunk sc = new ScriptChunk(OP_RESERVED, data);
    
    long t_field = 0;
    long t_method = 0;
    
    for (int i=0; i<iterations; i++) {

        long t0 = System.currentTimeMillis();
        byte[] dataFromField = sc.data;
        long t1 = System.currentTimeMillis();
        byte[] dataFromGetter = sc.getData();
        long t2 = System.currentTimeMillis();
        
        t_field += t1-t0;
        t_method += t2-t1;
    }
    System.out.println("data length   "+data_length);
    System.out.println("iterations    "+iterations);
    System.out.println("sc.data       "+t_field+"msec");
    System.out.println("sc.getData()  "+t_method+"mesc");
}
```
```
data length   1000
iterations    100000
sc.data       11msec
sc.getData()  46mesc
```

natzei:
Switch to `Optional` for return type of `getData()` and `getDataValue()`.
Don't know why CI is failing.

natzei:
Hi all. If nobody is interested in, I'll close the PR. It makes no sense to keep rebasing and stuff.

PR #1498, opened 2018-01-15
Infinite recursion, causing StackOverflowError in deprecated version of Wallet.toString(...) method.

tomasz-ludek:
Fixed infinite recursion, causing StackOverflowError in deprecated version of Wallet.toString(boolean, boolean, boolean, AbstractBlockChain) method.


schildbach:
Thanks for your PR, but can you show a sample (e.g. in JUnit test form) that shows the problem?

tomasz-ludek:
Sure thing, I can provide a unit test. However, you can easily reproduce this bug calling exactly this version of method: toString(boolean, boolean, boolean AbstractBlockChain) on any Wallet (AbstractBlockChain can be null).

schildbach:
Merged, thanks.

Issue #1501, opened 2018-01-24
Unnecessary accountPath included in Watch Key Methods


HashEngineering:
While working on a pull request that I plan to submit soon, changes were observed in this commit that are unnecessary and in one case change the path of a key: 
 https://github.com/bitcoinj/bitcoinj/commit/e036f8174aa5acb4fa8acf3b05f2c743f4b64353 (Allow to create wallets with an arbitrary account path.)

The main point of that commit was to allow for wallets to be created that used arbitrary paths such as BIP44.  This required a change to the protobuf format to store the path related to the seed, so the correct derivation path would be used in the DeterministicKeyChain.  Other methods in bitcoinj pertaining to creating wallets with a seed would also include the path.  Some changes were made to the methods relating to watched keys to also include an account path in the same way.  However those changes were not necessary because DeterministicKey already has its own path attribute.

Take for instance this constructor:
```
public DeterministicKeyChain(DeterministicKey watchingKey) {
       this(watchingKey, ACCOUNT_ZERO_PATH);
}
```
`watchingKey` has a path and it may be different from `ACCOUNT_ZERO_PATH`.

This is the constructor called from above and the accountPath parameter is unnecessary.
`public DeterministicKeyChain(DeterministicKey watchingKey, ImmutableList<ChildNumber> accountPath)`

If a key is ever created in bitcoinj, even if from deserializing an extended private key, then it will have a relative path. 

Several other changes were made to bitcoinj due to adding the accountPath as a parameter to watched keys, including several `from...` methods in Wallet to created watched wallets from DeterministicKeys or serialized Base58 strings.

I can submit a pull request for the changes reverting the watched key methods back to their original forms before the referenced commit.

schildbach:
Sounds good to me.

@gmagnotta @DaLN What do you think?

DaLN:
@schildbach Seems ok to me. It would need testing that everything is working properly.

HashEngineering:
This has been merged as https://github.com/bitcoinj/bitcoinj/commit/d3dca96a3cfdccdce65712eed0207d1c1c0e7e49 and https://github.com/bitcoinj/bitcoinj/commit/be382c68000590b9205c078e3cd8247c0ab2b905.

PR #1509, opened 2018-02-01
Removing unnecessary accountPath parameter on watch key methods

HashEngineering:
Thus pull request is related to #1501.

**Title:**  Removing unnecessary accountPath parameter on watch key methods preserving that the watched key account path can be arbitrary (does not have to be 0H).

A previous commit https://github.com/bitcoinj/bitcoinj/commit/e036f8174aa5acb4fa8acf3b05f2c743f4b64353 added arbitrary account support to Deterministic Seeds and Watched Keys.

During some of my own development for another pull request I plan to submit at a later time, it was discovered that the `accountPath` parameter that was added to methods relating to watched keys was unnecessary.  The reason being that the account path is already part of the DeterministicKey (even if it is created from serialized base58 format).  The account path is a single `ChildNumber`, which can be 0H.  It can also be any other arbitrary value such as 1 in the included test `watchingChainAccountOne`.

Before the above mentioned commit, bitcoinj required that a watched key be the 0H account.  After that commit, requiring 0H was no longer necessary.  `watchingChainAccountOne` was added to prove that a different account, such as 1, can be watched.  In this test, the extended public key was obtained from the external tool https://iancoleman.io/bip39/ using the `AccountOneChain`'s mnemonic seed and the path `m/1`

Further notes are given at particular lines of code.


core/src/main/java/org/bitcoinj/wallet/DefaultKeyChainFactory.java
```diff
@@ -48,11 +48,14 @@ public DeterministicKeyChain makeKeyChain(Protos.Key key, Protos.Key firstSubKey
     @Override
     public DeterministicKeyChain makeWatchingKeyChain(Protos.Key key, Protos.Key firstSubKey, DeterministicKey accountKey,
                                                       boolean isFollowingKey, boolean isMarried) throws UnreadableWalletException {
+        if (accountKey.getPath().size() != DeterministicKeyChain.ACCOUNT_ZERO_PATH.size())
```
HashEngineering: This check was added back, with the change that it checks for the size of the account path (being 1) rather than the value being 0H.

Otherwise, if path with longer length would cause different exception in `DeterministicKeyChain(DeterministicKey watchingKey)`

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -322,17 +322,9 @@ public DeterministicKeyChain(DeterministicSeed seed, ImmutableList<ChildNumber>
...
      */
     public DeterministicKeyChain(DeterministicKey watchingKey) {
-        this(watchingKey, ACCOUNT_ZERO_PATH);
-    }
-
-    /**
-     * Creates a deterministic key chain that watches the given (public only) root key. You can use this to calculate
-     * balances and generally follow along, but spending is not possible with such a chain.
-     */
-    public DeterministicKeyChain(DeterministicKey watchingKey, ImmutableList<ChildNumber> accountPath) {
```
HashEngineering: In many places, there were two methods.  One with the extra accountPath parameter and one without (that would then supply a default path of 0H).  In those cases, one of the methods was removed.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -287,42 +287,21 @@ public static Wallet fromSeed(NetworkParameters params, DeterministicSeed seed,
     }
 
     /**
-     * Creates a wallet that tracks payments to and from the HD key hierarchy rooted by the given watching key. A
-     * watching key corresponds to account zero in the recommended BIP32 key hierarchy.
```
HashEngineering: `watching key corresponds to account zero in the recommended BIP32 key hierarchy.` was removed from the comments, because the path of the doesn't have to be 0H.

schildbach:
Looks great so far. I only wonder didn't we say that accountPaths can now be of arbitrary length too?

HashEngineering:
That is a great question.  

Before the commit https://github.com/bitcoinj/bitcoinj/commit/e036f8174aa5acb4fa8acf3b05f2c743f4b64353, watch keys required a path equal to 0H.  

Now it can be any path with size of 1.  If a `DeterministicKey `was created with by deserializing an "xpub..." then it would have a path of size 1.  If the `DeterministicKey `was obtained with `DeterministicKeyChain.getWatchingKey`, then it would have the full path.

I will see how it works by removing restrictions on the size of the accountPath.  If all tests pass, I will add another commit with those changes.

HashEngineering:
I noticed something else while looking into these matters.  This is in `DefaultKeyChainFactory`, in the second `makeKeyChain` override that takes an an arbitrary accountPath:

```
    @Override
    public DeterministicKeyChain makeKeyChain(Protos.Key key, Protos.Key firstSubKey, DeterministicSeed seed,
                                              KeyCrypter crypter, boolean isMarried, ImmutableList<ChildNumber> accountPath) {
        DeterministicKeyChain chain;
        if (isMarried)
            chain = new MarriedKeyChain(seed, crypter);
        else
            chain = new DeterministicKeyChain(seed, crypter, accountPath);
        return chain;
    }
```

If the Seed is not married, then it will create a DeterministicKeyChain with the seed and the accountPath.  If the Seed is married, then it will create a MarriedKeyChain with the seed and the default BIP32 account path of m/0H.  In this case, the accountPath is ignored.  Further looking at MarriedKeyChain shows that it doesn't support arbitrary paths.  Perhaps this should be added?

Additionally, the DeterministicKeyChain.Builder class does not support arbitrary paths.  bitcoinj does not currently use DeterministicKeyChain.Builder from what I can tell.

HashEngineering:
In my second commit, the checks for `accountPath.size() == 1` were removed.  Another test was added:  `watchingChainArbitraryPath`.  This test will watch a key from the bip44chain that preserves the path with arbitrary length.  No exceptions are thrown and all tests pass.

schildbach:
@HashEngineering Thanks. If this is finished I'll merge it.

HashEngineering:
This is finished.

schildbach:
I just merged this!

PR #1511, opened 2018-02-10
Fix the print info in examples/SendRequest.java

harryge00:
I am a beginner using bitcoinj. I think the `to` is the actual address sending btc to.


schildbach:
Good catch! Merged.

Issue #1502, opened 2018-01-24
Add support for importing HD keys that can sign transactions


HashEngineering:
I am preparing a pull request to add support for creating wallets that can import an HD key, much like a watched key, but with the ability to sign transactions (spend).

While working on this, I was copying the schemes used by watched keys, but with some modifications to bitcoinj to ensure that it reads correctly from the protobuf format.

In my work, I found that changes made in a previous commit for arbitrary account paths were not necessary for the function of watched keys.  I created an issue for this here #1501.  I would rather not submit a pull request for my changes and include the fixes regarding simplifying the methods for watched keys.  It would be best that both watched and "spendable keys" use similar and consistently named methods with the same parameter lists.

My current branch for this is here:  
https://github.com/HashEngineering/bitcoinj/commits/importhdkey4

I will submit this as a pull request after hopefully getting some feedback regarding issue #1501.  The pull request will have much more documentation about changes made and reasoning for those changes.

PR #1514, opened 2018-02-14
Add ability to import an account key that allows for a wallet to spend

HashEngineering:
**Objective:**  Add support for importing HD account keys that allows the wallet to sign transactions.

Currently bitcoinj allows for setting up a DeterministicKeyChain to watch an account level HD Key.  It has restrictions that prohibit watching a key with private key information as it requires the use of dropPrivate() in such cases.

The purpose of this pull request is to allow for spending (signing transactions) from an account level HD Key that is supplied to create a Wallet, KeyChainGroup, DeterministicKeyChain or DeterministicKey.

The use case for this pull request is for an app that will manage its own DeterministicSeed information separate from bitcoinj and this app will manage multiple accounts where each account is handled by a different bitcoinj Wallet object.  The Wallet object that manages a single account will not have knowledge of the deterministic seed from which account key is derived, nor does it necessarily know the derivation scheme used (such as BIP44 or BIP32).

Given the limitation mentioned at the beginning with respect to watched keys (no private key information, some modifications had to be made to a few places in bitcoinj to allow for an HD key to sign transactions and be loaded from the protobuf format of the wallet.  These changes are as follows:

- A new constructor was added for `DeterministicKeyChain` to support creating a KeyChain with a spendable HD account key by setting the isWatching parameter to false.
- A static spend method was added to follow the convention of the watch method of `DeterministicKeyChain`.
- `DeterminisiticKeyChain.fromProtobuf` was modified to recognize an HD account key as a key that has no seed, but has secret bytes.  This distinguishes it from a watching key (which has no seed or secret bytes) and keys derived from a parent that has a seed stored in the protobuf file.
- No changes were made to how protobuf files are saved or its format.
- DefaultKeyChainFactory has a new method makeSpendingKeyChain to allow creating the DeterministicKeyChain from the protobuf format in DeterminisiticKeyChain.fromProtobuf.

The Wallet class has a few new methods for creating a new wallet
`fromSpendingKey` – similar to `fromWatchingKey`, but allows spending from the provided account key.
`fromSpendingKeyB58` – similar to `fromWatchingKeyB58`, but allows spending from the provided Base58 extended private key.
`fromMasterKey` – this is a new method that takes a master key and the account number, from which it derives the account key.  Master key could be the root key of BIP32 or any arbitrary key that is the parent of an account.  For BIP44, this key would be `m/44H/1H`

`WalletAppKit` was modified to allow restoring a wallet from an account key in a similar way that it restores from a seed.

A few tests are added to ensure that these changes can be verified to work in `DeterministicKeyChainTest`

- `spendingChain` - similar to watchingChain, but ensures that signing works.
- `spendingChainAccountTwo` - similar to watchingChain, but for a different account and ensures that signing works.
- `masterKeyAccount` - this test takes the bip44chain and simulates `fromSpendingKeyB58` and `fromMasterKey` and makes sure that the keys resulting from both are the same as bip44chain and that the serializations match the included text files.

Contributor: @DaLN (Nelson MELINA <nelson.melina@mycelium.com>)

#1502


core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -332,6 +332,31 @@ public DeterministicKeyChain(DeterministicKey watchingKey) {
...
+     * creates a deterministic key chain that allows spending. If <code>isFollowing</code> flag is set(only allowed
+     * if <code>isWatching</code> is set) then this keychain follows some other keychain.  In a married wallet following
+     * keychain represents "spouse's" keychain.
+     */
+    public DeterministicKeyChain(DeterministicKey key, boolean isFollowing, boolean isWatching) {
+        if(isWatching)
+            checkArgument(key.isPubKeyOnly(), "Private subtrees not currently supported for watching keys: if you got this key from DKC.getWatchingKey() then use .dropPrivate().dropParent() on it first.");
+        else
+            checkArgument(key.hasPrivKey(), "Private subtrees are required.");
+        checkArgument(isWatching ? true : !isFollowing, "Can only follow a key that is watched");
```
Giszmo: ¿Would this not be the same as `checkArgument(isWatching || !isFollowing, ...`?

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -882,7 +915,15 @@ public boolean isFollowing() {
                 if (chain == null) {
                     // If this is not a following chain and previous was, this must be married
                     boolean isMarried = !isFollowingKey && !chains.isEmpty() && chains.get(chains.size() - 1).isFollowing();
-                    if (seed == null) {
+                    // If this has a private key but no seed, then all we know is the spending key H
+                    if (seed == null & key.hasSecretBytes())
```
Giszmo: coding style ... I'm pretty sure Schildbach will ask for style consistent with the project, so `if (...) `~\n~`{`.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -898,7 +939,7 @@ public boolean isFollowing() {
                 }
                 // Find the parent key assuming this is not the root key, and not an account key for a watching chain.
                 DeterministicKey parent = null;
-                if (!path.isEmpty() && !isWatchingAccountKey) {
+                if (!path.isEmpty() && !isWatchingAccountKey & !isSpendingKey) {
```
Giszmo: Binary `&`? Unless `!isSpendingKey` has to get evaluated no matter what, use `&&` here and even if `isSpendingKey` had side effects, I would find it confusing to rely on [this particularity](https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.22.2) of the binary `&` in a logical expression.

core/src/test/java/org/bitcoinj/wallet/DeterministicKeyChainTest.java
```diff
@@ -470,6 +475,180 @@ public void watchingChainAccountOne() throws UnreadableWalletException {
         assertEquals(key4.getPubKeyPoint(), rekey4.getPubKeyPoint());
     }
 
+    @Test
+    public void spendingChain() throws UnreadableWalletException {
+        Utils.setMockClock();
+        DeterministicKey key1 = chain.getKey(KeyChain.KeyPurpose.RECEIVE_FUNDS);
```
Giszmo: avoid variables ending in numbers. If those are to be treated equally, instantiate a list/array. If not, name them by their purpose. Makes review easier.

In particular: Are key1 and key2 supposed to differ or is key2 a copy of key1?

(Granted, this is the same style as other tests in the file.)

core/src/test/java/org/bitcoinj/wallet/DeterministicKeyChainTest.java
```diff
@@ -470,6 +475,180 @@ public void watchingChainAccountOne() throws UnreadableWalletException {
...
+        assertEquals(100000, chain.getEarliestKeyCreationTime());
+        chain.setLookaheadSize(10);
+        chain.maybeLookAhead();
+
+        assertEquals(key1.getPubKeyPoint(), chain.getKey(KeyChain.KeyPurpose.RECEIVE_FUNDS).getPubKeyPoint());
+        assertEquals(key2.getPubKeyPoint(), chain.getKey(KeyChain.KeyPurpose.RECEIVE_FUNDS).getPubKeyPoint());
+        final DeterministicKey key = chain.getKey(KeyChain.KeyPurpose.CHANGE);
+        assertEquals(key3.getPubKeyPoint(), key.getPubKeyPoint());
+        try {
+            // We can sign with a key from a spending chain.
```
Giszmo: this comment could go into the `catch` logic:

`fail("Keys from spending chain should be able to sign");`

core/src/test/java/org/bitcoinj/wallet/DeterministicKeyChainTest.java
```diff
@@ -470,6 +475,180 @@ public void watchingChainAccountOne() throws UnreadableWalletException {
...
+        }
+
+        @Override
+        protected ImmutableList<ChildNumber> getAccountPath() {
+            return ImmutableList.of(new ChildNumber(2, true));
+        }
+    }
+
+    @Test
+    public void spendingChainAccountTwo() throws UnreadableWalletException {
```
Giszmo: Quite a lot of duplicate code here. Please factor it out.

core/src/test/java/org/bitcoinj/wallet/DeterministicKeyChainTest.java
```diff
@@ -470,6 +475,180 @@ public void watchingChainAccountOne() throws UnreadableWalletException {
...
+        // Test we can serialize and deserialize a spending chain OK.
+        List<Protos.Key> serialization = chain.serializeToProtobuf();
+        checkSerialization(serialization, "spending-wallet-account-two-serialization.txt");
+        chain = DeterministicKeyChain.fromProtobuf(serialization, null).get(0);
+        final DeterministicKey rekey4 = chain.getKey(KeyChain.KeyPurpose.CHANGE);
+        assertEquals(key4.getPubKeyPoint(), rekey4.getPubKeyPoint());
+    }
+
+    @Test
+    public void masterKeyAccount() throws UnreadableWalletException {
```
Giszmo: déjà vu?

core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
```diff
@@ -196,6 +198,19 @@ public WalletAppKit restoreWalletFromSeed(DeterministicSeed seed) {
         this.restoreFromSeed = seed;
         return this;
     }
+    /**
```
schildbach: Is this a method comment? To me it's not immediately clear, also because there is a newline after the comment.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
schildbach: Indeed :smile:

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
schildbach: Looks unintended to me.

HashEngineering:
@schildbach @Giszmo Yes, I will definitely address all of the above mentioned concerns.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
HashEngineering: Yes, that is a typo (`&` instead of `&&`).

core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
HashEngineering: It is supposed to be a method comment, yes.

core/src/test/java/org/bitcoinj/wallet/DeterministicKeyChainTest.java
HashEngineering: We probably don't need the previous test, spendingChainAccountTwo.  I will refactor this test, masterKeyAccount, as this one tests the main point of this pull request.  Another test already checks watched keys for other accounts or paths.

core/src/test/java/org/bitcoinj/wallet/DeterministicKeyChainTest.java
HashEngineering: I will rename the key1 as firstReceiveKey, key2 as secondReceiveKey, key3 as firstChangeKey and key4 as secondChangeKey.  These are all different keys for the first two receive addresses and first two change addresses in the key chain.

HashEngineering:
I am not sure why the travis check failed as all the tests passed on my system before pushing my fixes.  Nevertheless, I will run the tests again.

schildbach:
That's an unreliable test, don't worry :smile:

schildbach:
Thanks for your fixes to the PR. Merged!

PR #1515, opened 2018-02-17
Add AddressScript interface to ease handling of BECH32 addresses

erasmospunk:
DO NOT MERGE, feedback is needed.

This is a proof of concept of how to handle Bech32 addresses using a new `AddressScript` interface.

The `Address#getHash160()` method that is replaced by the `getValue()` as SegWit uses 160 or 256 bit hashes.

The new `getScriptType()` method is a replacement for the `Address#isP2SHAddress()` where we get the exact `Script.ScriptType` that the address represent. For example the existing addresses will return either `P2PKH` or `P2SH` and the future Bech32 addresses will return either `P2WPKH` or `P2WSH`. Notice that it is easier to add SegWit support to `ScriptBuilder#createOutputScript` when we switch the `Script.ScriptType` enum.

Finally the `getAddressFormat()` returns `AddressFormat.BASE58` for the existing `Address` and `AddressFormat.BECH32` for the new SegWit addresses.

Ideally all APIs should be changed to use the new interface and it is easy when we pass parameters to methods but return values are trickier. Unless we are OK with breaking the API.

It is missing the `Comparable` that is defined in `VersionedChecksummedBytes`, any ideas of how it could be handled?

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AddressScript.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressScript.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2018 John Jegutanis
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bitcoinj.core;
+
+import org.bitcoinj.script.Script.ScriptType;
+
+public interface AddressScript {
+    enum AddressFormat {
+        BASE58,
+        BECH32
+    }
+
+    /** Get the script type of this address */
+    ScriptType getScriptType();
+
+    /** Get the hash value of this address */
+    byte[] getValue();
+
+    /** Get the network parameters of this address */
+    NetworkParameters getParameters();
+
+    /** Get the destination type */
+    AddressFormat getAddressFormat();
+}
--- a/core/src/main/java/org/bitcoinj/core/Address.java
+++ b/core/src/main/java/org/bitcoinj/core/Address.java
@@ -2,6 +2,7 @@
  * Copyright 2011 Google Inc.
  * Copyright 2014 Giannis Dzegoutanis
  * Copyright 2015 Andreas Schildbach
+ * Copyright 2018 John Jegutanis
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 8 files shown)


core/src/main/java/org/bitcoinj/core/AddressScript.java
```diff
@@ -0,0 +1,42 @@
...
+public interface AddressScript extends Serializable, Cloneable {
+    enum AddressFormat {
+        BASE58,
+        BECH32
+    }
+
+    /** Get the script type of this address */
+    ScriptType getScriptType();
+
+    /** Get the version or header value of this address */
```
schildbach: The equivalent of the address header in Bech32 is the HRP (human readable part). Because of the different types (int vs. String) I think this method should not be part of this interface.

erasmospunk: Slight correction: Bech32 addresses have a [witness version](https://github.com/Coinomi/bech32/blob/java-ref/ref/java/src/main/java/org/bech32/SegwitAddress.java#L37) that is different from the HRP that is only present in the encoded string.

erasmospunk:
I'm looking your PR.

> Implementing Comparable just between SegwitAddresses should be trivial, I guess you're asking for how to compare to base58-based data? I think if we really want that, an external Comparator would be better suited than implementing the Comparable interface.

Yes, an external Comparator simplifies things.

core/src/main/java/org/bitcoinj/core/AddressScript.java
schildbach: Yes, but as far as I understand the witness version serves and entirely different purpose than the base58 address header (extension of the scripting language vs. distinguishing between networks+usecases). So I think it should not be using the same accessor.

erasmospunk: Good point, removing it...

erasmospunk:
@schildbach Implemented the changes you asked and added the Address wrapper proof of concept.

schildbach:
Hmm, I don't understand the problem you're solving with the wrapper. Why not stay with a direct implementation?

Actually what would help greatly is seeing how a SegwitAddress would be implemented.

erasmospunk:
The idea was to avoid `VersionedChecksummedBytes` that was in effect a Base58 implementation (`toBase58()`, `VersionedChecksummedBytes(String encoded)` parses as Base58).

So if `SegWitAddress` extends `VersionedChecksummedBytes` then we get the situation where a Bech32 address is also a Base58 object.

erasmospunk:
I'm removing the latest commit and keeping the AddressScript here as a reference until we finally merge the #1516

erasmospunk:
Another advantage of the wrapper class is that we make don't break the API, for example `Wallet#currentReceiveAddress()`. Instead the new `Address` class handles Bech32 addresses seamlessly.

schildbach:
I was under the impression that a `SegwitAddress` would not extend `VersionChecksummedBytes` but only implement the `AddressScript` interface?

erasmospunk:
@schildbach I was referring to the #1516 where `SegWitAddress` extends `AbstractAddress` that extends `VersionChecksummedBytes` so that `(segWitAddressObj instanceof VersionChecksummedBytes)  == true`.

schildbach:
Let's stay with your proposal in this PR for now, I want to understand it.

I take it the `VersionedChecksummedBytes` hierarchy takes only base58-data, as it does today. `SegwitAddress` (and it's containing hierarchy, if any) takes only bech32-data. And to be able to seamlessly use addresses of any kind we create a new interface, currently called `AddressScript` implemented by the current `Address` and the future `SegwitAddress` classes.

If that's right, what do we need the wrapper for?

If we manage to expose all necessary functionality via the `AddressScript` interface and refactor bitcoinj to use it then that would be a nice solution, no?

erasmospunk:
The `AddressScript` is to be able to easily interchange a base58 and bech32 addresses. It could be used everywhere instead of `Address` but it will break a lot of APIs and people's code.

The second commit is to show a way of how we could re-structure things without breaking the existing APIs.

Let's say you add a BIP84 keychain to `Wallet` and need to call `currentReceiveAddress()` as an existing BitcoinJ user:
- With the #1516 approach it would return an `AbstractAddress` so a refactor is needed after an update.
- The `AddressScript` everywhere will also change the API and a refactor will be necessary
- With the new `Address` that just wraps a `SegWitAddress`, no refactor is needed as it will be seamlessly be converted to a valid Bech32 address by calling `toString()`.

erasmospunk:
Closing this as #1516 seems to be the way to go.

PR #1516, opened 2018-02-18
Implement native segwit address (BIP173)

schildbach:
This is an alternative approach to #1515 by @erasmospunk. I'm putting it up for "pre-review" a bit earlier than expected; it's far from polished.

In contrast to the clean AddressScript interface approach by John, I'm trying to integrate SegwitAddress into the VersionChecksummedBytes hierarchy. This will also not be possible without a bit of API churn, but at least the preparational first commit hasn't broken any unit tests, examples or my own app.

Also see the commit messages.


erasmospunk:
If we are OK with some API breakage, may I suggest a slightly different approach?

- Rename `Address` to `Base58Address` or something similar
- Both `Base58Address` and `SegwitAddress` implement the `AddressScript` interface
- Create a wrapper `Address` class that contains a `protected final AddressScript addressScript;` field
- The new `Address` shall implement `AddressScript` (and `Comparable<Address>` for some compatibility?) and pass through the method calls to the internal `addressScript` object.
- Keep but deprecate the old methods like `fromBase58`, `fromP2SHHash`

I'm going to push a proof of concept of the idea to the #1515 PR.

core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
```diff
@@ -33,29 +36,20 @@
  * <p>and the result is then Base58 encoded. This format is used for addresses, and private keys exported using the
  * dumpprivkey command.</p>
  */
-public class VersionedChecksummedBytes implements Serializable, Cloneable, Comparable<VersionedChecksummedBytes> {
-    protected final int version;
```
erasmospunk: Why dropping the `version` field? It's a loss of information and cannot be recreated from the network parameters alone.

core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
```diff
@@ -33,29 +36,20 @@
...
+        this.params = checkNotNull(params);
+        this.bytes = checkNotNull(bytes);
     }
 
     /**
      * Returns the base-58 encoded String representation of this
      * object, including version and checksum bytes.
      */
-    public final String toBase58() {
+    protected static String toBase58(int version, byte[] bytes) {
```
erasmospunk: Maybe this could go to `Base58` as `encodeChecked`?

core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
erasmospunk: Another "minor" use case is with Litecoin, where they have P2SH addresses with header byte of `5` or `50`.

core/src/main/java/org/bitcoinj/core/AbstractAddress.java
```diff
@@ -0,0 +1,54 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bitcoinj.core;
+
+import javax.annotation.Nullable;
+
+public class AbstractAddress extends VersionedChecksummedBytes {
```
erasmospunk: Could we rename this to GenericAddress?

core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
```diff
@@ -33,29 +36,20 @@
  * <p>and the result is then Base58 encoded. This format is used for addresses, and private keys exported using the
  * dumpprivkey command.</p>
  */
-public class VersionedChecksummedBytes implements Serializable, Cloneable, Comparable<VersionedChecksummedBytes> {
```
erasmospunk: Could we add an abstract `getFormat()` and an enum like this:

```
enum Format {
    BASE58,
    BECH32
}
```

When we handle the abstract types, it's handy to know what is the encoding format.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
```diff
@@ -77,6 +77,7 @@
     protected int addressHeader;
     protected int p2shHeader;
     protected int dumpedPrivateKeyHeader;
+    protected String segwitAddressHrp;
```
erasmospunk: Could we rename this to `addressPrefix`?

core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
schildbach: Good idea, I implemented this via b87d22e2beb328b140be84b5b7e3eced4637b4b3 on master.

core/src/main/java/org/bitcoinj/core/AbstractAddress.java
schildbach: My idea was to rename it to Address, so it can be a plug-in replacement for the current Address (which in turn I would rename to LegacyAddress).

core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
schildbach: I can add it back of course, but for the current classes it's not needed. The only information it carries is the network (use `VersionChecksummedBytes.params`) and the usecase (inspect the type of your class if you don't already know it).

What's the difference between 5 and 50 on Litecoin?

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
schildbach: Hmm, is this "bc" string planned to be used for more than the HRP of segwit addresses?

core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
schildbach: In a way, I wanted to abstract away the encoding from the user. Just use toString() and you'll get the correct encoding.

But of course if it's important, we can add it. You could also inspect the type though.

core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
erasmospunk: There is no difference except do distinguish it visually

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
erasmospunk: Only for HRP. I think addressPrefix is more descriptive.

core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
erasmospunk: OK, in that case we can leave it.

core/src/main/java/org/bitcoinj/core/VersionedChecksummedBytes.java
erasmospunk: Another issue that I see is in `SegWitAddress` where we need to store a version but we do it in a "hacky" way by using the internal `bytes`.

core/src/main/java/org/bitcoinj/core/SegwitAddress.java
```diff
@@ -0,0 +1,133 @@
...
+            throw new AddressFormatException("Invalid length for address version 0");
+    }
+
+    private int witnessVersion() {
+        return bytes[0] & 0xff;
+    }
+
+    private byte[] witnessProgram() {
+        // skip version byte
+        return convertBits(bytes, 1, bytes.length - 1, 5, 8, false);
```
erasmospunk: We can avoid this conversion if we saved the witness version and program in `VersionedChecksummedBytes`.

core/src/main/java/org/bitcoinj/core/AbstractAddress.java
erasmospunk: Renaming to Address is a much cleaner approach.

I forgot to ask about adding a method from `AddressScript`:
`public abstract Script.ScriptType getScriptType();`

core/src/main/java/org/bitcoinj/core/SegwitAddress.java
schildbach: We could transiently cache them in `SegwitAddress` if needed.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
schildbach: BIP173 uses the term HRP or Human Readable Part. It doesn't use Prefix. Would you prefer `segwitAddressHumanReadablePart`?

erasmospunk: What about addressHrp, then?

core/src/main/java/org/bitcoinj/core/AbstractAddress.java
schildbach: Added.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
schildbach: Ah, so it's the "segwit" that you want to get rid of? But legacy addresses have no HRP.

schildbach:
I rebased on current master, added the `getScriptType()` and put a commit on top that makes BitcoinURIs aware of segwit addresses.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
erasmospunk: My comments are mainly to make the library abstract and alt-chain friendly. For example BitcoinCash refers this as "addressPrefix".

schildbach:
Again I rebased on current master and put a commit on top that allows sending to native segwit address. I tested this in an adapted Bitcoin Wallet 6.01 against a segwit wallet in Electrum 3.0.6.

erasmospunk:
@schildbach didn't you want to rename ```AbstractAddress``` to ```Address``` and extend from it a ```LegacyAddress``` as a way to preserve API compatibility?

> My idea was to rename it to Address, so it can be a plug-in replacement for the current Address (which in turn I would rename to LegacyAddress).

schildbach:
@erasmospunk I'm planning to try this rename really soon now. I would also like to rename `VersionedChecksummedBytes` to `PrefixedChecksummedBytes`, if that doesn't break too much.

schildbach:
I just did a quick test rename with bitcoinj and bitcoin-wallet in the same refactoring to get a rough feel of the impact.

The `VersionedChecksummedBytes`→`PrefixedChecksummedBytes` impact is minimal, especially on other projects. That's actually expected, as most projects probably don't use this class directly.

`Address`→`LegacyAddress` and `AbstractAddress`→`Address` has quite some impact, as follows:

- A project that doesn't care about segwit (yet) has the most changes because of the first rename. The bitcoinj subprojects (tools, examples, wallettemplate) are examples for this. However, in this case global find & replace should do the trick, so I'm not too concerned.
- A wallet that wants to be able to send to segwit addresses has about half of the impact. The sending part is actually nearly a drop-in replacement, very nice! The receiving part (display of own addresses, etc) however suffers from the renames again.

- A wallet that would go full segwit would be inpacted even less; it would be as close to a drop-in replacement as it can get. The problem is that option is theoretical, as bitcoinj is currently far from being able to _receive_ to native addresses.

In any case, there is a constant amount of API churn due to the already merged (on master) incompatible API changes.

erasmospunk:
In the #1514 PR I did something similar by replacing the `Address` and the effect was minimal if you add all the old methods to it and `@Deprecate` them. They deprecated stuff could stay in release-0.15 and completely removed in the next version.

I have some code that integrates segwit (native & p2sh) with the `LocalTransactionSigner` that I will push in the next days.

schildbach:
I take it you mean #1515. Yes, that seems quite similar.

schildbach:
Ok, after lots of polishing I propose these three commits for merging. The other commits you were seeing on this PR will re-appear in a later PR if this gets merged.

I went through all of the comments again:

- Added back LegacyAddress.getVersion() (idea by John)
- Added Address.getOutputScriptType() (idea by John)
- Kept NetworkParameters.segwitAddressHrp as is, I think this is the best name for it
- I added and completed lots of Javadoc
- The class renaming is done

@erasmospunk If I remember right, the Bech32 and Bech32Test classes are written by you (Coinomi). Would you consider relicensing it under Apache2 license? I know the current license is very permissive already, but it would make things easier if bitcoinj would not mix licenses.

erasmospunk:
@schildbach Bech32 and Bech32Test classes can be relicensed under Apache2.

schildbach:
Merged. Thanks John for your feedback!

PR #1520, opened 2018-02-23
Remove concept of 'from address'

schildbach:
Segwit is another nail on the coffin, because for each pubkey there is now also a segwit address.
A number of wallets will not recognize returned payments if sent to the wrong address type.


erasmospunk:
ACK

PR #1521, opened 2018-02-24
Move ECKey.toAddress() to Address.fromKey()

schildbach:
This is a preparation for supporting native segwit addresses. Keys can't know what
type of address is wanted. In future, there will also be a SegwitAddress.fromKey().


schildbach:
Note: I thought about just deprecating the old method. However, I think this can be dangerous as you might generate a wrong receiving address, even if it's the same private/public key.

erasmospunk:
You are right, it is dangerous if you are not explicit.

What about using the following ```Address.from(params, keyScheme, ecKey)```?

The keyScheme could be the following enum:
```
public enum KeyScheme {
    DEFAULT(44),
    SEGWIT_NATIVE(84),
    SEGWIT_COMPAT(49);

    public final int bip43Index;

    KeyScheme(int bip43Index) {
        this.bip43Index = bip43Index;
    }

    public static Optional<KeyScheme> fromBip43Index(int index) {
        for (KeyScheme scheme : values()) {
            if (scheme.bip43Index == index) return Optional.of(scheme);
        }
        return Optional.absent();
    }
}
```

schildbach:
Keep in mind that `Address` is (in the current naming scheme) a legacy address, not an abstract address. Maybe you can base your idea on the segwit-address branch so we can see how it plays out? I suspect the two methods `Address.fromKey()` and `SegwitAddress.fromKey()` would do the trick too?

PR #1525, opened 2018-03-02
 PeerGroup: Start using VersionMessage.receivingAddr as it was intended

schildbach:
I'm not sure why this field was never set. According to the operator of a full node, bitcoinj clients seem to be the only ones who don't use it.


schildbach:
Merged, and fixed VersionMessage in general.

PR #1526, opened 2018-03-04
Added a SIGINT hook, so a partial checkpoint file gets written

Giszmo:
When the checkpoints tool gets stuck, you loose the syncing you did, as SIGINT makes it skip writing the checkpoints file. This change allows you to rescue the scanned checkpoints.


schildbach:
I prefer this to stay an "atomic" operation I must admit. What's the use of a partial checkpoints file?

Issue #1029, opened 2015-07-13
Race condition between dependency retrieval and filtered transaction receipt


devrandom:
Symptom: transaction stays in pending even though it was in a received filtered block.

Dependency retrieval in `Peer` is asynchronous.  While Peer is retrieving dependencies for a broadcast transaction, the download peer can send us the transaction hash in a filtered block.  But since the `Wallet` hasn't seen the transaction yet, it will ignore the filtered transaction hash.

schildbach:
Related to https://github.com/bitcoinj/bitcoinj/issues/981?

devrandom:
It is similar, but doesn't require the peer to fail in order to occur.  Also, I'm not sure I completely understand @mikehearn analysis there, will follow up.

mikehearn:
Yes, makes sense. We should probably queue up processing of filtered blocks if a dependency download is currently in progress?

PR #1530, opened 2018-03-05
Change method BlockChain.rollbackBlockStore from protected to public.

kmels:
This method can be used by a wallet to replay from a given height. For example a wallet receives a transaction with OP_RETURN that incurs in new keys imported to the wallet. The wallet needs to process again the block if relevant transactions are found in it. 

I believe this approach can be used to solve bad wallet syncs as a product of race conditions between filtered blocks and transactions as mentioned in #1029 .


schildbach:
Can you base your PR on master? Normally we don't introduce such a change to the release version.

kmels:
@schildbach ok, done.

PR #1533, opened 2018-03-11
Migrate from Maven to Gradle for building.

schildbach:
This targets Gradle 3.2.1, as this is the version that will be in the upcoming Ubuntu 18.04 LTS.

You can use:
- `gradle clean`
- `gradle generateProto`
- `gradle build`
- `gradle install`

Currently, the protobuf classes are generated on each build. I hope this will not cause unwanted changes in these classes due to differences in protobuf libraries. If it turns out problematic, maybe in future we should not put generated classes into git.

Dependency pinning (to certain hashes) is gone for now. I doubt this will add much security if the feature isn't built into gradle itself. As long as it's a plugin, there is the trust bootstrapping problem.

Bundled jars are currently gone. If someone still needs them, please submit a PR for building them again. The Windows versions of the tool launcher scripts (e.g. wallet-tool.cmd) are gone, because I have no Windows to test with. If you have Windows, feel free to help here.

Because Gradle has no optional dependencies, there is now a dependency on `org.fusesource.leveldbjni:leveldbjni-all:1.8` that you likely want to manually exclude from your build.


yichen88:
README.md missing change?

schildbach:
Good catch! Amended, and rebased.

schildbach:
I'll skip the signing/uploading part. I think this in fact should be part of the build process.

schildbach:
Merged!

schildbach:
Maybe we should exclude the generated protobuf files from version control?

natzei:
I think yes, since the build generates fresh ones.

schildbach:
@natzei See 002fe8184cf6121856c966272d00f7a80c97b803

Issue #1532, opened 2018-03-10
Bug:  Wallet Encryption of DeterministicKeyChain with Arbitrary Path


HashEngineering:
This is a followup issue to this feature:  **Allow to create wallets with arbitrary path (e.g m/44'/0'/0/0)** 
https://github.com/bitcoinj/bitcoinj/commit/e036f8174aa5acb4fa8acf3b05f2c743f4b64353
#1341

When using the Wallet.encrypt feature, there is an IllegalArgumentException exception thrown if the wallet contains a DeterministicKeyChain with an arbitrary path other than (m/0H - the default for bitcoinj).

The cause is in the `DeterministicKeyChain(KeyCrypter crypter, KeyParameter aesKey, DeterministicKeyChain chain)` constructor where the account path is not set to that of `chain`.  This results in an `IllegalArgumentException` in `DeterministicHierarchy.get`.  For example in the case of BIP44, `get` will be looking for a path of `0H` but should be looking for `44H/0H`

I will further work to determine the fix and if other methods are affected by this bug.

schildbach:
Fixed via e0d9683bd54c4e60d71840ad1ab2409016a3e0f9

PR #1534, opened 2018-03-12
Wallet:  Allow encrypting DeterministicKeyChain with arbitrary path

HashEngineering:
This is in reference to issue #1532.

This pull request contains a test that will fail, if `setAccountPath(chain.getAccountPath());` is missing from the DeterministicKeyChain constructor used during the encryption process.  The account path needs to be set in the new DeterministicKeyChain so it will find the correct path instead of assuming the default `0H`.

The failure report would have been as follows if the account path is not set:
```
[INFO] Running org.bitcoinj.wallet.WalletTest
[ERROR] Tests run: 123, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 16.327 s <<< FAILURE! - in org.bitcoinj.wallet.WalletTest
[ERROR] encryptWalletWithArbitraryPath(org.bitcoinj.wallet.WalletTest)  Time elapsed: 0.328 s  <<< ERROR!
java.lang.IllegalArgumentException: No key found for absolute path M/0H.
        at org.bitcoinj.wallet.WalletTest.encryptWalletWithArbitraryPath(WalletTest.java:175)
```

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
@@ -433,6 +433,7 @@ public class DeterministicKeyChain implements EncryptableKeyChain {
         checkNotNull(chain.seed);
 
         checkArgument(!chain.rootKey.isEncrypted(), "Chain already encrypted");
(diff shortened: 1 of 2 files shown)


schildbach:
Merged!

PR #1541, opened 2018-03-19
Transaction: Limit initial size of inputs and outputs ArrayList.

oscarguindzberg:
Limits initial size of inputs and outputs ArrayLists when parsing a tx.

This prevents this DoS attack:

- Somehow the attacker needs to get a p2p connection to the bitcoinj node.
- The attacker sends a tx msg that says the tx contains a trillion inputs.
- bitcoinj tries to instantiate an ArrayList with a size of a trillion.
- OutOfMemoryError and the bitcoinj node is down.


schildbach:
Good catch! Merged.

PR #1542, opened 2018-03-21
Refactor some codes.

cryptcoin-junkey:
They are a kind of 'cosmetic changes'. But they will increase readability for 'BitcoinJ newbie' like me.


schildbach:
Sorry, I think "organize imports" isn't of much use – especially since we don't have a standard defined for import order etc.

Issue #877, opened 2015-06-11
Rethink WalletEventListener.onTransactionConfidenceChanged


mikehearn:
Originally reported on Google Code with ID 511

```
Profiling the replay process for Andreas' megawallet shows a surprising amount of time
spent simply iterating transactions in order to call them for this event listener.
The problem is confidence stores the depth and work done rather than recalculate it,
so every transaction is updated for every block. With a lot of transactions, that's
a lot of work per block.

Andreas uses this event for a bunch of things in the UI, for example, as a central
place to trigger recalculation of the list (he adds some throttling on top).

With some tweaks in both bitcoinj and the Android app we can be much smarter about
this and avoid doing lots of useless work.
```

Reported by `hearn@google.com` on 2014-01-15 16:22:53

Giszmo:
We are definitely hitting a brick wall with this one, too. Anybody working on a fix, so we don't have to stumble around, finding the same blockers?

schildbach:
I'm not aware of anyone working on this.

schildbach:
I just took a look at this and opened preview PR #1553.

PR #1553, opened 2018-04-07
WIP: Get rid of depth in TransactionConfidence.

schildbach:
This is work on #877. It's very preview, breaks tests and therefore not intended for merging soon.

Issues:
- Yes, it's an API breaking change. Mostly this means you have to replace `confidence.getDepthInBlocks()` by `confidence.getDepthInBlocks(wallet.getLastBlockSeenHeight());`.
- Height is removed from the wallet protobuf. The field number is blocked for future use. Luckily appearsAtBlockHeight was already present so no migration is needed.
- As intended, ConfidenceListener will not fire on depth change, since the depth is now not part of the confidence any more. This means wallets with responsive UI will miss blockchain confirmations – unless they subscribe to some of the block height listeners too. Perhaps we should add a wallet-based "lastBlockSeenListener", since I suspect the existing listeners will perhaps not fire at the right time.
- The TransactionConfidence.getDepthFuture() needs a replacement, based on the above listener.
- Theoretically, if a block arrives which is irrelevant to the wallet saving could be skipped. Unfortunately, we still need to persist the lastBlockSeenHeight field so for now Wallet.saveLater() still needs to be called on each new block.
- DefaultCoinSelector spends coins based on priority (depth * value). Since priority is no longer a thing with miners (they pick by fee rate) and depth is now not easily available to the DefaultCoinSelector, I tried changing the algorithm to "older first". Unfortunately this breaks many tests. I guess this decision/work should be broken out to a separate PR.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -2146,9 +2146,8 @@ public class Wallet extends BaseTaggableObject
                         // included once again. We could have a separate was-in-chain-and-now-isn't confidence type
                         // but this way is backwards compatible with existing software, and the new state probably
                         // wouldn't mean anything different to just remembering peers anyway.
-                        if (confidence.incrementDepthInBlocks() > context.getEventHorizon())
+                        if (confidence.getDepthInBlocks(getLastBlockSeenHeight()) > context.getEventHorizon())
                             confidence.clearBroadcastBy();
-                        confidenceChanged.put(tx, TransactionConfidence.Listener.ChangeReason.DEPTH);
                     }
                 }
             }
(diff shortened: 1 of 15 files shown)


micheal-swiggs:
I can see a possible solution to this. It would involve creating a new `TxConfidenceTable` maybe `ConfidenceManager` object that is used by `Wallet`. Methods in `Wallet` are the consumers of `depth`. The only thing I wonder about is if we want to implement the changes as breaking or non-breaking changes? To accomplish non-breaking changes it would require accessing `ConfidenceManager` via `Context` from within `TxConfidenceTable`. The `depth` related methods of `TxConfidenceTable` would start making calls to `ConfidenceManager` via `Context`. The `depth` methods within `TxConfidenceTable` would be deprecated and the new methods on `ConfidenceManager` should be used instead. It doesn't feel like the most pretty solution but is perhaps justified for one release as the dependency on `Context` could be removed once the deprecated methods on `TxConfidenceTable` are removed...

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -2146,9 +2146,8 @@ public void notifyNewBestBlock(StoredBlock block) throws VerificationException {
                         // included once again. We could have a separate was-in-chain-and-now-isn't confidence type
                         // but this way is backwards compatible with existing software, and the new state probably
                         // wouldn't mean anything different to just remembering peers anyway.
-                        if (confidence.incrementDepthInBlocks() > context.getEventHorizon())
+                        if (confidence.getDepthInBlocks(getLastBlockSeenHeight()) > context.getEventHorizon())
                             confidence.clearBroadcastBy();
-                        confidenceChanged.put(tx, TransactionConfidence.Listener.ChangeReason.DEPTH);
```
micheal-swiggs, line 2151: @schildbach I suppose this is the line that is causing the so called "useless work" referenced here: https://github.com/bitcoinj/bitcoinj/issues/877 ?

PR #1562, opened 2018-04-21
Refactor mnemonic code tests for maintainability and empty passphrases

dbkingsb:
The pull request consists of a split of the MnemonicCodeTest into two test classes, the addition of test vectors for an empty passphrase, and disabled test vectors for a normalized passphrase.

The test vectors portion of MnemonicCodeTest was split into another test class called MnemonicCodeVectorsTest. This new class uses the Parameterization feature of JUnit to execute multiple vectors through the same test. The MnemonicCodeTest class now contains tests predominantly for guard clauses (i.e. throw exception if input not valid).

There could be mistakes with handling an empty passphrase, (e.g. is the prefix of "mnemonic" still used) so test vectors for an empty passphrase were generated and added.

Test vectors for passphrase normalization are included, however they are disabled due to Issue #1361. These can be enabled for the testing and merging of that issue's resolution.

### DIFF
--- a/core/src/test/java/org/bitcoinj/crypto/MnemonicCodeVectorsTest.java
+++ b/core/src/test/java/org/bitcoinj/crypto/MnemonicCodeVectorsTest.java
@@ -1,3 +1,20 @@
+/*
+ * Copyright 2013 Ken Sedgwick
+ * Copyright 2014 Andreas Schildbach
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.bitcoinj.crypto;
 
 import org.bitcoinj.core.Utils;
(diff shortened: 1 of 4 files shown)


schildbach:
Thanks a lot, it's looking good. Just two nits:
- The new test file is missing a license header. Feel free to copy from the old file.
- I think the commented testcases belong into a separate PR, ideally together with the corresponding fix.

core/src/test/java/org/bitcoinj/crypto/MnemonicCodeVectorsTest.java
```diff
@@ -0,0 +1,270 @@
...
+    @Before
+    public void setup() throws IOException {
+        mc = new MnemonicCode();
+    }
+
+    @Test
+    public void testMnemonicCode() throws Exception {
+        final List<String> mnemonicCode = mc.toMnemonic(HEX.decode(entropy));
+        final byte[] seed = MnemonicCode.toSeed(mnemonicCode, passphrase);
+        final byte[] entropy = mc.toEntropy(split(this.mnemonicCode));
```
schildbach: Hmm, can't we also use Utils.SPACE_JOINER here?

dbkingsb: Sorry, I'm not seeing where it is applicable beyond line 66; can you specify?

dbkingsb: Ah, did you mean to move the "split" method to Utils, because it is now used in two different classes? Agreed; I'll do this along with the other two comments.

schildbach:
Ah sorry, I meant using a Splitter. Never mind, I amended it to your contribution and merged. Thanks again!

schildbach:
Do you wish to open a PR about the normalization issue?

dbkingsb:
Ah, Splitter -- I'm not too familiar with the Google Commons 😁.

I opened PR #1564 (WIP) for normalization of the passphrase.

PR #1563, opened 2018-04-21
Full support for native segwit.

schildbach:
This allows to create wallets with a DeterministicKeyChain that derives native segwit addresses and receives on them. By default, new wallets are still created with a legacy P2PKH key-chain.

- Hierarchical-deterministic derivation of native segwit addresses.
- Receive payments to native segwit addresses.
- Spend and sign payments from native segwit addresses.
- Watch-only wallets with native segwit addresses (zpub/vpub).

There is currently no automatic migration from P2PKH to P2WPKH wallets. Either you add a second DeterministicKeyChain via API or you create a new Wallet(NetworkParams, Script.ScriptType.P2WPKH).

Be aware this adds a new field in the wallet protobuf: output_script_type in Key, which keeps track of the script type of DeterministicKeyChains. Protobufs will be migrated; old DeterministicKeyChains are assumed to be of type P2PKH.

_Warning: As of today, I have only tested this excessively on testnet. So for now only use with testnet or really small amounts!_


lontivero:
@schildbach i would like to help you adding support for segwit tx and bech32 addresses support but I don't know where to start, there are a couple of PR trying to do the same it seems. How could I help?

schildbach:
I rebased this PR on current master, and implemented proper base58 de/serialization of P2WPKH account keys (zpub, zprv rather than xpub, xprv). We're getting closer (-:

schildbach:
Great progress on the segwit front: This PR is ready to merge! Please review and test, and most importantly, please report your findings. Also please report if you found no problems. I've updated the first post on this PR.

nopara73:
Tests pass, Windows 10, IntelliJ.  

image

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -106,20 +106,33 @@
...
     public static final ImmutableList<ChildNumber> ACCOUNT_ZERO_PATH = ImmutableList.of(ChildNumber.ZERO_HARDENED);
+    // m / 1'
+    public static final ImmutableList<ChildNumber> ACCOUNT_ONE_PATH = ImmutableList.of(ChildNumber.ONE_HARDENED);
     // m / 44' / 0' / 0'
     public static final ImmutableList<ChildNumber> BIP44_ACCOUNT_ZERO_PATH = ImmutableList.of(new ChildNumber(44, true),
             ChildNumber.ZERO_HARDENED, ChildNumber.ZERO_HARDENED);
     public static final ImmutableList<ChildNumber> EXTERNAL_SUBPATH = ImmutableList.of(ChildNumber.ZERO);
     public static final ImmutableList<ChildNumber> INTERNAL_SUBPATH = ImmutableList.of(ChildNumber.ONE);
 
+    private static ImmutableList<ChildNumber> defaultAccountPathFor(Script.ScriptType outputScriptType) {
```
lontivero: Why is this? Is it for bip49?

schildbach: No. BIP49 is out of scope, as it deals with P2SH-P2WPKH only.

schildbach: If you look at the implementation of the method, there is one path for P2PKH (the default old one) and another for the new P2WPKH output script type.

lontivero:
@schildbach Ive finished my review and it looks great. I also compared the test related to segwit (no in this PR) against the test vectors in the bips and nbitcoin and they are okay (in fact I had to update the test vectors for bip173 in nbitcoin as result of this work).

Bitcoinj is huge and contains many concepts that I was not familiar with so, there are part of the code that I just couldn't review properly because of my lack of context. There are also some test data files (the `deterministic-(sic)-serialization.txt` files) that I don't know how to validate, I guess that I could load it with nbitcoin and check they pass too...

Please let me know if there is something else I can do.

oscarguindzberg:
Would you confirm this is the case: MarriedKeyChain ability to work with P2WSH is still to be implemented. MarriedKeyChain still works fine with legacy P2SH.

oscarguindzberg:
> There is currently no automatic migration from P2PKH to P2WPKH wallets. Either you add a second DeterministicKeyChain via API or you create a new Wallet(NetworkParams, Script.ScriptType.P2WPKH).

Since bech32 addresses are not supported everywhere yet I assume many apps using bitcoinj would want to be able to expose both types of addresses (1xxx and bc1xxx) to users.

I see that could be problematic using any of the 2 approaches you suggest. Adding a second DeterministicKeyChain is problematic since KeyChainGroup.getActiveKeyChain() is used a lot in Wallet and KeyChainGroup. Having 2 wallets forces apps to store 2 files and it would be hard to combine utxos from the 2 wallets when spending.

Please, let me know if I am missing something here.

schildbach:
@lontivero Thanks for your review, it's much appreciated and useful.

@oscarguindzberg Yes, for P2WSH we only have some building blocks but no usecase. The tests for MarriedKeyChain still work, but I didn't try it manually.

Yes, the migration is still todo. I planned to provide a `KeyChainGroup.getActiveKeyChain(Script.ScriptType)` variant so that you can have two active keychains (one for each script type). And in Wallet maybe only provide similar `getCurrentAddress()` and `getFreshAddress()` variants?

For determinism purposes, I'd like to stick to this upgrade path:

Basic Keychain --> P2PKH keychain --> P2WPKH keychain

Then for old wallets that get migrated to P2WPKH a fallback P2PKH keychain is always available, but unfortunately for a newly created P2WPKH there would be no downgrade path to P2PKH. I currently don't know how to deal with this case or maybe just don't care.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
schildbach: @lontivero Also see https://github.com/bitcoinj/bitcoinj/pull/1563#issuecomment-462271008; I moved the method into a pluggable interface.

schildbach:
Merged – yipiieh!

githorray:
Congratulations!!! Lots of upstreams folks can't put off segwit support anymore! ;-)

PR #1565, opened 2018-04-25
Add bip47 implementation and tests

kmels:
**Overview**

BIP47 allows merchants to transact with bitcoin without the possibility of linking any two transactions together. It achieves this privacy by establishing a shared secret key when sending a "notification transaction" that is basically an OP_RETURN with the payer's payment code. After the notification transaction is broadcasted, the secret key is used in the BIP47AppKit with BIP32 to generate the payee's payment address on every payment transaction

Version 1 of Payment Codes support a notification transaction with MIN_NONDUST_OUTPUT (576 satoshis) that the payee will receive. 

**Considerations for integrating BIP47**

While the wallet's PeerGroup is connected to a peer, other bip47 wallets are expected to broadcast the notification before the payment and the receiver's wallet will process each transaction in a serial way. When the wallet's PeerGroup is disconnected and then needs to catchup with the blockchain, a race condition occurs that we solved by PR #1530. The problem seemed to be that a filteredBlock containing a notification *and* a payment transaction doesn't guarantee the order and thus failed to import the DeterministicKeys to watch.

**Goals**
- Send and receive BIP47 transactions using SPV wallets and V1 payment codes
- Provide unit tests for [BIP47 vector tests](https://gist.github.com/SamouraiDev/6aad669604c5930864bd)


kmels:
re: build.

To fix the "SC is not a valid provider error" we would need Oracle to sign the jce provider. So i removed the `prov` dependency in another [WIP branch](https://github.com/kmels/bitcoincashj/tree/bitcoinj/bip47-noprovider), the problem is now: `java.security.NoSuchAlgorithmException: ECDH AlgorithmParameters not available`
I put openjdk8 in .travis.yml but travis doesn't have it with JavaFX. Would you mind that we change to openjdk8? We can have 2 builds and maybe have travis be silent about the oracle build. :+1:

schildbach:
I'm under the impression that the concept of reusable payment codes didn't catch traction in the Bitcoin community. Also, the BIP hasn't been updated for SegWit, hasn't it?

Issue #1568, opened 2018-04-29
Factor out recId search logic


matthewleon:
`ECKey.signMessage` contains a block of logic for determining a key's `recId`, beginning here: https://github.com/bitcoinj/bitcoinj/blob/0b0ea0c7e5a875c1b5fabe64aad6a86a8d69b0fc/core/src/main/java/org/bitcoinj/core/ECKey.java#L869

This logic is itself potentially useful to library users, and could be exposed as a new method, perhaps `public byte getRecId(ECDSASignature sig, Sha256Hash hash)`.

I would be happy to provide a PR for this.

schildbach:
Sure, feel free to!

PR #1570, opened 2018-04-30
ECKey.findRecoveryId

matthewleon:
addresses https://github.com/bitcoinj/bitcoinj/issues/1568

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/ECKeyTest.java
+++ b/core/src/test/java/org/bitcoinj/core/ECKeyTest.java
@@ -240,6 +240,19 @@ public class ECKeyTest {
     }
 
     @Test
+    public void findRecoveryId() {
+        ECKey key = new ECKey();
+        String message = "Hello World!";
+        Sha256Hash hash = Sha256Hash.of(message.getBytes());
+        ECKey.ECDSASignature sig = key.sign(hash);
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/bitcoinj/core/ECKeyTest.java
```diff
@@ -239,6 +239,19 @@ public void verifyMessage() throws Exception {
         assertEquals(expectedAddress, gotAddress);
     }
 
+    @Test
```
schildbach: Thanks for the test. But shouldn't it test for a specific recovery id? Currently it just asserts that the recovery id is between 0 and 3.

Is there a test vector in a specification, or in a different implementation perhaps?

matthewleon: It would certainly be ideal to test for a specific recovery ID. I'll see what I can dig up.

matthewleon:
@schildbach I've added two new tests. One uses a vector I found on Stack Overflow, which, while not ideal, seems to be common practice in lieu of a set of "best practice" vector for sepc256k1. The second is a version of another test in the file that replaces a looped search for the recovery ID with one that uses the new method.

schildbach:
Thanks! I just squashed and merged.

PR #1594, opened 2018-06-18
null pointer exception happens when adding witness program

nayuta-ueno:
Using `addWatchedAddress()` to adding witness program happens NullPointerException.

```
SegwitAddress watchAddr = SegwitAddress.fromBech32(params, "tb1....");
wallet.addWatchedAddress(watchAddr);
```


schildbach:
Merged. Thanks!

PR #1596, opened 2018-06-21
input connected if out's parent transaction equals to outpoit.fromTx

igurev:
It seems that "out" is missing before getParentTransaction()


schildbach:
I agree. I have not merged because I could not find out why the tests are failing. I has probably to do with how we fake test data.

PR #1604, opened 2018-07-14
Connected exception

BigAdam2005:
Created `PeerConnectionException` which  replaces old `NotYetConnectedException`


schildbach:
I think we need to postpone changes to the networking layer to after 0.17. So I'm moving the milestone to 0.18.

PR #1615, opened 2018-07-25
Use gradle wrapper

Bo-Ye:
Using gradle wrapper means that not only don't you have to manually install Gradle yourself, but you are also sure to use the version of Gradle that the build is designed for. This makes your historical builds more reliable.

https://docs.gradle.org/current/userguide/gradle_wrapper.html


Bo-Ye:
@natzei thanks for the comment, the jar file shouldn't be ignored, it's meant to be uploaded once and shared by others.

> gradle-wrapper.jar

    The Wrapper JAR file containing code for downloading the Gradle distribution.
Please see https://docs.gradle.org/current/userguide/gradle_wrapper.html for the more details.

BigAdam2005:
I think it is great idea for projects which without gradle.
@natzei why?

natzei:
I think too that the adoption of gradle wrapper is a good idea.
I'm not very familiar with it, so I could be wrong, but the executable jar of gradle should be downloaded once **locally** and not pushed on the repository.
Analogously, It's like pushing maven dependencies on a repository instead of making the user download them from maven central.

BigAdam2005:
@natzei no we need to have jar in repository. Gradle wrapper allow you to build project without local gradle instance. Look at gradle wrapper in other java projects.

natzei:
Sorry, but I disagree.

Moreover, since the content of `gradle/wrapper` is "generated", I think that the whole directory should be ignored too.
Pros of pushing the gradle jar is that developers are not obligated to install gradle. Cons are that changes in gradle version result in a new jar to push which cause the repository to grow over the time.

My point reflects this answer https://stackoverflow.com/a/28473686, however is personal and probably @BigAdam2005 and @Bo-Ye agree on the previous answer instead (https://stackoverflow.com/a/20351077).

It's a matter of choice. The maintainer will choose the approach he prefer.

schildbach:
If it were possible, I'd satisfy all dependencies from the Ubuntu or Debian (or similar) repository. Lazily downloading deps is really not a good security model.

Is there a specific problem we need to solve? For example, does bitcoinj not build with a more recent Gradle version (I'm testing with 3.4.1, that's what comes with Ubuntu 18.04.1)?

Note everybody can create their own wrapper with `gradle wrapper`.

natzei:
I'm using gradle 4.9-1 and no problem here.

> Note everybody can create their own wrapper with `gradle wrapper`.

I agree.

Bo-Ye:
@BigAdam2005 Thanks for comments.
@natzei please see my replies.

> I'm not very familiar with it, so I could be wrong, but the executable jar of gradle should be downloaded once locally and not pushed on the repository.

the executable jar is not downloaded, instead, it's generated by the person who initialize the project and only generated once.

> Analogously, It's like pushing maven dependencies on a repository instead of making the user download them from maven central.

the gradle-wrapper.jar has noting to do with project dependencies, it's a tool to download gradle installation distribution.

> Moreover, since the content of gradle/wrapper is "generated", I think that the whole directory should be ignored too.

Unfortunately, if the content of gradle/wrapper were ignored, the whole gradle wrapper idea wouldn't work because the gradle wrapper contains a TOOL to download gradle installation distribution other than gradle installation itself. Regardless what the gradle installation version is, the gradle wrapper's jar file is always same, you can resemble it as curl on linux.

> Cons are that changes in gradle version result in a new jar to push which cause the repository to grow over the time.

This is not true, changes in gradle version would NOT result in a new jar to push, only the gradle version configuration file will be changed.

@schildbach thanks for the comments.

> If it were possible, I'd satisfy all dependencies from the Ubuntu or Debian (or similar) repository. Lazily downloading deps is really not a good security model.

I didn't interpret your words well, if you trust gradle in the first place, you would trust gradle wrapper as well. Therefor, using gradle wrapper automatically installs gradle is a safe way as well.

> Is there a specific problem we need to solve? For example, does bitcoinj not build with a more recent Gradle version (I'm testing with 3.4.1, that's what comes with Ubuntu 18.04.1)?

The goal using gradle wrapper is to improve the convenience. Only the project creator should manually install gradle once, for other contributors, gradle installation is automatic and later on, for whatever reasons, the gradle needs to be upgraded(new features coming up or security issues fixes) only the project creator/maintainer should change gradle version config file under gradle wrapper folder. Other followers/contributors will be agnostic to any gradle version changes and do the business as usual, which means they don't need to manually install/upgrade a new gradle again.

> Note everybody can create their own wrapper with gradle wrapper. We can git-ignore all the files that are created by this command.

Everybody creating their own wrapper is not necessary and is against the will using gradle wrapper because it means either everyone manually installs gradle or generates gradle wrapper files twice after using gradle wrapper.

Bo-Ye:
Let me clarify how the gradle wrapper works entirely.

1. The project creator manually installs gradle and run `gradle wrapper,` the following files will be created, remember to add these files and folders to your version control system (e.g GitHub or Bitbucket).

|-gradle
  |--- wrapper
      |--- gradle-wrapper.jar
      |--- gradle-wrapper.properties
|-gradlew
|-gradlew.bat 

gradle-wrapper.jar is a tool to download gradle installation, because other contributors don't install gradle manually, so they are not able to generate this file before gradle-wrapper.jar installs the gradle for you, so the file has to be uploaded to github. gradle-wrapper.properties controls the gradle version upgrading.

2. Other project contributors/followers git clones the project, they're not required to install gradle and can run any gradle commands by ./gradlew <task name> directly.

3. Any gradle version upgrading just requires to change version number in gradle-wrapper.properties and it only requires to be carried out by one person.

One of the most prominent Java project -- Spring Framework is also using gradle wrapper and make all files under gradle/wrapper uploaded to github. Please refer to the following link:

https://github.com/spring-projects/spring-framework/tree/master/gradle/wrapper

natzei:
> the gradle-wrapper.jar has noting to do with project dependencies, it's a tool to download gradle installation distribution.

Thanks for clarification.

The workflow of gradle wrapper is clear, I'm complaining about **pushing a tool on the repository**  just to avoid **developers** to have installed any version of gradle and execute `gradle wrapper`.

In my opinion, the benefit of using the wrapper is to uniform developers to build using the same version of gradle, not to avoid gradle installation.

> One of the most prominent Java projects -- Spring Framework is also using gradle wrapper and make all files under gradle/wrapper uploaded to github. 

Again, it is a matter of choice. No one can be considered wrong IMHO.

schildbach:
For now, I've added an ignore of the wrapper files via 90c3cccb6fab21e26db10e4649ca2d6af6f9bb9c. So everyone can create their own `gradle wrapper` if they like.

Installing gradle is just a matter of `sudo apt install gradle` and you can do that on the same command as installing your JDK, e.g. `sudo apt install openjdk-8-jdk openjfx gradle`.

msgilligan:
> Lazily downloading deps is really not a good security model.

Note that there is a setting in `gradle-wrapper.properties` that tells the Gradle wrapper to check a SHA 256 sum of the download and abort the build if it doesn't match.  Example from [gradle-wrapper.properties in ConsensusJ](https://github.com/ConsensusJ/consensusj/blob/master/gradle/wrapper/gradle-wrapper.properties#L4) :

```
distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.3-bin.zip
distributionSha256Sum=8626cbf206b4e201ade7b87779090690447054bc93f052954c78480fa6ed186e
```

We'd also need to be careful about the contents of the binary `gradle-wapper.jar`, of course.

There's a conflict here between Gradle best practices and secure/reproducible build best practices, but the benefits to using the wrapper are significant and we've already exerted quite a bit of effort to make the builds work everywhere without out it. My recommendation is that we use the wrapper, but with use of the SHA sum and extreme caution in committing the `gradle-wapper.jar` file.

@schildbach If you're willing to re-consider this, I can rebase this PR and add the SHA sum check.

schildbach:
I think it's fine as is. Everyone can use Gradle wrapper on their own if they like.

msgilligan:
@schildbach I think we can go ahead and close this one.

PR #1620, opened 2018-07-29
Removed unCacheHeader method

BigAdam2005:
Remove unnecessary method `unCacheHeader` from `Block`


schildbach:
Can you elaborate on why you think it's unnecessary?

BigAdam2005:
It is unneeded because we can just overwrite that value without setting to null.

schildbach:
I think it is an error to not set the `headerBytesValid` to `false` if one of the header fields changes.

Travis fails the build. Did you run the tests locally?

BigAdam2005:
Yes and it worked.

PR #1629, opened 2018-08-04
Removed memory mapping in SPVBlockStore

BigAdam2005:
Removed memory mapping from SPVBlockStore.


schildbach:
You removed the cache, not the memory mapped buffer. Hint: it's `MappedByteBuffer buffer`. However, it's not that easy. What the comment doesn't tell of course that we need to find a suitable replacement. And actually I personally think the memory mapping is not that bad.

BigAdam2005:
I am working on moving caching to blockchain. But removing memory mapping was in TODO.
About replacement, I think about removing normal ByteBuffer and replace it with DirectByteBuffer.

BigAdam2005:
I created BlockStore map instead of replacing memory mapping. This map contains information about block hash, index of first byte of block and last byte of block.  It has two times faster than yours.

schildbach:
Can you please not reformat comments and otherwise untouched code? It makes your PRs hard to review. Only change what's needed to implement your idea.

BigAdam2005:
Ok, fixed code formatting.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -175,89 +188,50 @@ private void initNewStore(NetworkParameters params) throws Exception {
...
-                    return storedBlock;
-                }
-            } while (cursor != startingPoint);
-            // Not found.
-            notFoundCache.put(hash, NOT_FOUND_MARKER);
-            return null;
-        } catch (ProtocolException e) {
-            throw new RuntimeException(e);  // Cannot happen.
-        } finally { lock.unlock(); }
+        ChainMap map = chainMaps.get(hash);
```
schildbach: The error handling isn't right here. You need to keep `lock.unlock();` in the finally blocks or otherwise you risk leaking the lock.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -325,4 +300,24 @@ private void setRingCursor(ByteBuffer buffer, int newCursor) {
         checkArgument(newCursor >= 0);
         buffer.putInt(4, newCursor);
     }
+
+    class ChainMap {
```
schildbach: Missing access modifiers in this class. Private?

schildbach:
You say your PR makes the store quicker. What devices did you test this on?

Is this PR changing the format of the file? It seems so, as you are inserting a new field at position 8. In this case, we need a migration plan and tests for going from the old format to the new. Or – more conservatively – keep the old SPVBlockStore class and offer your new one in parallel?

BigAdam2005:
I tested it on my computer(linux). We don't need to change tests because it is tested now in every get or put. I think we need to migrate, maybe change headers magic to `SPV2` or something else to recognize older versions.

BigAdam2005:
Added migration option for older stores.

BigAdam2005:
@schildbach what do you think about it

schildbach:
SPVBlockStore is mainly meant for and used on mobile devices (Android). We need testing there too.

You didn't answer the question: does this PR change the file format?

BigAdam2005:
Yes it changes file format. I will add test for android soon.

schildbach:
So then if it changes the file format, can you document the changes and add a rationale why you changed it?

BigAdam2005:
Only change is adding int at position 8. I added it because it was needed for creating index after loading block store

BigAdam2005:
Ok, checked for android and changed code to be less error prone.

BigAdam2005:
@schildbach something more to fix?

PR #1635, opened 2018-08-10
Added AddressMessage constructor

BigAdam2005:
I added constructor for `AddressMessage` which allow for creating message from `PeerAddress` list.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AddressMessage.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressMessage.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedList;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


schildbach:
Looks like you're replying to `getaddr` with just the own address – which is guaranteed to be known by the caller already. If I understand this right, it doesn't make much sense to me.

A minimal implementation would reply with multiple known good peers, perhaps those currently connected by PeerGroup (excluding the caller)? A more elaborate implementation would collect known good peers and keep them across restarts. That database could then be used for our own connections too so we don't have to query PeerDiscoveries all the time.

core/src/main/java/org/bitcoinj/core/AddressMessage.java
```diff
@@ -47,7 +48,16 @@
...
         super(params, payload, offset, serializer, length);
     }
-
+    public AddressMessage(NetworkParameters params, List<PeerAddress> addrs){
+        super(params);
+        addresses = addrs;
+    }
+    public AddressMessage(NetworkParameters params, final PeerAddress addr){
+        super(params);
+        addresses = new LinkedList<PeerAddress>(){{
```
schildbach: I guess you could use `Collections.singletonList()` here.

BigAdam2005:
I only sending addresses from `PeerGroup`. About saving it, where store it, in which format it should be stored and when save it.

BigAdam2005:
So it is ready? @schildbach

PR #1639, opened 2018-08-31
Fixed potential locking error in TxConfidenceTable

tau3:
There was an incorrect lock release in `org.bitcoinj.core.TxConfidenceTable#seen`, so i made an attempt to fix this and implemented a unit test


schildbach:
Thanks for spotting this! Merged.

PR #1662, opened 2018-11-02
Introduce SignatureDecodeException

schildbach:
ECKey: If DER-encoded signatures cannot be parsed, throw SignatureDecodeException rather than RuntimeException.


schildbach:
Merged.

PR #1669, opened 2018-11-27
Clarify that just bare (non-P2SH) multisig txs which require more tha…

oscarguindzberg:
…n 3 public keys are non-standard


schildbach:
Merged.

PR #1672, opened 2018-12-12
SegWit wallet features

Chm:
Hi,

We are building a BIP49 wallet app in RPC client mode, found some segwit features were missing.
I grabbed some code from segwit branch and added some override points.

Code passed unit tests in my app with rpc service, and the changes didn't intervene unrelated code.

TODO:
local witness check with executeScript in Script.checkWitness

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/core/ECKey.java
@@ -27,6 +27,8 @@ import com.google.common.primitives.UnsignedBytes;
 import org.bitcoin.NativeSecp256k1;
 import org.bitcoin.NativeSecp256k1Util;
 import org.bitcoin.Secp256k1Context;
+import org.bitcoinj.script.Script;
+import org.bitcoinj.script.ScriptBuilder;
 import org.bitcoinj.wallet.Protos;
 import org.bitcoinj.wallet.Wallet;
 import org.slf4j.Logger;
@@ -157,6 +159,9 @@ public class ECKey implements EncryptableItem {
     protected EncryptedData encryptedPrivateKey;
 
     private byte[] pubKeyHash;
+
+    private Script p2wpkhScript;
+    private byte[] p2wpkhHash;
 
     /**
      * Generates an entirely new keypair. Point compression is used so the resulting public key will be 33 bytes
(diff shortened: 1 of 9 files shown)


schildbach:
Thanks for picking this up again.

Please try to keep PRs small and concise. In particular:

- If you find bugs in current master, put it up as a separate PR. Commit 138d193923e5ff00869f27275b8cf272e556daf2 is such a case – it has got nothing to do with Segwit. I cherry-picked this particular commit already, so in this case just a rebase is needed.
- Other unrelated work should go on a different PR as well. The extension points you're proposing are such an example.
- If you fix bugs in your own PR (or decide to change the implementation mid-way), don't expose everyone to your development history. Perfect example: 13c009506fe148ae5509533995a46734a385c8e9 and its revert f736fa01402c6446cb968ae644c307fdc6bc9760.
- Try to not format existing code, especially if you didn't touch it. It will make PRs hard to read.
- Another way to make this PR smaller could be by deferring the fully validating wallet code to a later PR. We still need to decide if we want to keep full validation – currently it's in a horrible state.

Unit tests:

- For a new feature like this, extensive unit testing is mandatory.
- For simple bug fixes unit tests are optional, but it's still recommended to add them if it makes sense.

Also I'd like to make you aware of the (unmerged) PR #1563. I'm not sure, but it looks a bit as if your approach could conflict with this PR. Basically the idea is that `DeterministicKeyChain` has a new `outputScriptType` field that is responsible for what type of addresses are generated, type of signature used, etc.

Sorry if I'm a bit pedantic here but it's easy to introduce lots of bugs if changes get too broad. And proper receiving and being able to spend in all cases is one of the most delicate responsibilities of the framework.

Chm:
Thanks for the advice, I'll break the commits a bit for review.

I didn't realize the other approach in keychain, I would also vote for a dedicated type.
For me it's most important to have extension points in release build, as I'm using a signer with JNI ECKey and only public key in wallet.

And the tx witness calculation code in segwit branch got one more byte in hash pre image, I'll get some test code for that soon.

schildbach:
Could you rebase this to current master so we can see what's left?

Chm:
Remaining changes:

Calculate P2WPKHash for key, store, map somewhere(basic keychain for now)
Set tx protocol version to make serialization work
Add witness signature in signer
Segwit fee

TBD
check witness signature with executeScript()

jonathancross:
FYI: https://github.com/bitcoinj/bitcoinj/pull/1563  has been merged.

schildbach:
I think this is now all more or less implemented on master. Thanks!

PR #1674, opened 2018-12-18
Calc TxSig for witness

Chm:
Tx witness signing and test case separated from PR #1672

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -1185,6 +1185,154 @@ public class Transaction extends ChildMessage {
         }
     }
 
+    public TransactionSignature calculateWitnessSignature(
+            int inputIndex,
+            ECKey key,
+            byte[] redeemScript,
+            Coin value,
+            SigHash hashType,
+            boolean anyoneCanPay) {
+        Sha256Hash hash = hashForSignatureWitness(inputIndex, redeemScript, value, hashType, anyoneCanPay);
+        return new TransactionSignature(key.sign(hash), hashType, anyoneCanPay);
+    }
+
+    public TransactionSignature calculateWitnessSignature(
+            int inputIndex,
+            ECKey key,
+            Script redeemScript,
+            Coin value,
+            SigHash hashType,
+            boolean anyoneCanPay) {
+        return calculateWitnessSignature(inputIndex, key, redeemScript.getProgram(), value, hashType, anyoneCanPay);
+    }
+
+    public TransactionSignature calculateWitnessSignature(
+            int inputIndex,
+            ECKey key,
+            @Nullable KeyParameter aesKey,
+            byte[] redeemScript,
+            Coin value,
+            SigHash hashType,
+            boolean anyoneCanPay) {
+        Sha256Hash hash = hashForSignatureWitness(inputIndex, redeemScript, value, hashType, anyoneCanPay);
+        return new TransactionSignature(key.sign(hash, aesKey), hashType, anyoneCanPay);
+    }
+
+    public TransactionSignature calculateWitnessSignature(
+            int inputIndex,
+            ECKey key,
+            @Nullable KeyParameter aesKey,
+            Script redeemScript,
+            Coin value,
+            SigHash hashType,
+            boolean anyoneCanPay) {
+        return calculateWitnessSignature(inputIndex, key, aesKey, redeemScript.getProgram(), value, hashType, anyoneCanPay);
+    }
+
+    public synchronized Sha256Hash hashForSignatureWitness(
+            int inputIndex,
+            byte[] scriptCode,
+            Coin prevValue,
+            SigHash type,
+            boolean anyoneCanPay) {
+        int sigHash = TransactionSignature.calcSigHashValue(type, anyoneCanPay);
+        return hashForSignatureWitness(inputIndex, scriptCode, prevValue, (byte) sigHash);
+    }
+
+    /**
+     * <p>Calculates a signature hash, that is, a hash of a simplified form of the transaction. How exactly the transaction
+     * is simplified is specified by the type and anyoneCanPay parameters.</p>
+     *
+     * <p>This is a low level API and when using the regular {@link Wallet} class you don't have to call this yourself.
+     * When working with more complex transaction types and contracts, it can be necessary. When signing a Witness output
+     * the scriptCode should be the script encoded into the scriptSig field, for normal transactions, it's the
+     * scriptPubKey of the output you're signing for. (See BIP143: https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki)</p>
+     *
+     * @param inputIndex   input the signature is being calculated for. Tx signatures are always relative to an input.
+     * @param scriptCode   the script that should be in the given input during signing.
+     * @param prevValue    the value of the coin being spent
+     * @param type         Should be SigHash.ALL
+     * @param anyoneCanPay should be false.
+     */
+    public synchronized Sha256Hash hashForSignatureWitness(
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/bitcoinj/script/ScriptBuilder.java
```diff
@@ -438,6 +438,22 @@ public static Script updateScriptWithSignature(Script scriptSig, byte[] signatur
         return builder.build();
     }
 
+    /**
+     * Creates a segwit scriptPubKey that sends to the given public key hash.
+     */
+    public static Script createP2WPKHOutputScript(byte[] hash) {
+        checkArgument(hash.length == 20);
```
schildbach: I believe we could use the constant `SegwitAddress.WITNESS_PROGRAM_LENGTH_PKH` here.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1089,10 +1102,12 @@ public Sha256Hash hashForSignature(int inputIndex, byte[] redeemScript,
      * @param type Should be SigHash.ALL
      * @param anyoneCanPay should be false.
      */
-    public Sha256Hash hashForSignature(int inputIndex, Script redeemScript,
-                                                    SigHash type, boolean anyoneCanPay) {
-        int sigHash = TransactionSignature.calcSigHashValue(type, anyoneCanPay);
-        return hashForSignature(inputIndex, redeemScript.getProgram(), (byte) sigHash);
+    public Sha256Hash hashForSignature(
```
schildbach: Please don't reformat this method signature (and the 5 above).

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1185,6 +1200,145 @@ public Sha256Hash hashForSignature(int inputIndex, byte[] connectedScript, byte
...
+        byte[] connectedScript = scriptCode.getProgram();
+        return hashForSignatureWitness(inputIndex, connectedScript, prevValue, type, anyoneCanPay);
+    }
+
+    public synchronized Sha256Hash hashForSignatureWitness(
+            int inputIndex,
+            byte[] connectedScript,
+            Coin prevValue,
+            SigHash type,
+            boolean anyoneCanPay) {
```
schildbach: This parameter seems to be ignored, because it is overridden in line 1284. Is this perhaps a bug? Or am I wrong?

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -305,6 +305,73 @@ public void witnessTransaction() {
...
+        Script scriptCode = new ScriptBuilder()
+                .data(ScriptBuilder.createOutputScript(LegacyAddress.fromKey(netParams, key))
+                        .getProgram())
+                .build();
+        assertEquals("1976a91479091972186c449eb1ded22b78e40d009bdf008988ac",
+                HEX.encode(scriptCode.getProgram()));
+
+        TransactionSignature txSig = unsignedTx.calculateWitnessSignature(0, key,
+                scriptCode, Coin.COIN.multiply(10),
+                Transaction.SigHash.ALL, false);
```
schildbach: Thanks for the test. It would be nice if there were tests for other SigHash and anyoneCanPay values. Are there any in Bitcoin Core that could be adapted?

core/src/main/java/org/bitcoinj/script/ScriptBuilder.java
Chm: Right

core/src/main/java/org/bitcoinj/core/Transaction.java
Chm: Added the reformat makes it easy to compare to segwit branch, Reversing.

core/src/main/java/org/bitcoinj/core/Transaction.java
Chm: SigHash type was enough, should be removed along call stack.

core/src/test/java/org/bitcoinj/core/TransactionTest.java
Chm: Just got through wiki, only basic sample was provided. Bitcoin core should have test cases for other flags in the code.
Maybe in a separate test later?

core/src/main/java/org/bitcoinj/core/Transaction.java
Chm: ANYONECANPAY flag was covered by SigHash type constants. Confusing to have enum as input while other info added along the call stack, and have a byte when checking for behavior.

Chm: I see the reasoning for backward compatibility. Need to have it organized.

core/src/test/java/org/bitcoinj/core/TransactionTest.java
schildbach: Ok, fair enough. But we should really add more tests before release.

schildbach:
I just noticed your PR seems to cover the P2SH-P2WPKH case only. Am I right? If yes: I think this should be clearly mentioned in the Javadoc. Do you plan to submit code for P2WPKH (aka "native segwit") too?

Btw. there are more testcases on the segwit branch in commit 01044ea0209f7e514338549b720dc6f736253f9e.

Chm:
Native P2WPKH test passed, except for a typo in bip wiki: missing sigHashType byte 0x01 in a signature in decomposed data. Final tx matched.

Chm:
Hi, @schildbach , seems all known review advices have been resolved. Could you take a look if any change is needed for this PR?

schildbach:
It's all fine and I am ready to merge. Shall we just wait for a (serious) comment on https://github.com/bitcoin/bips/pull/748, it seems to me it's related right?

schildbach:
I merged this. Thanks again for your contribution!

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -305,6 +305,150 @@ public void witnessTransaction() {
...
+                key1.getPublicKeyAsHex());
+
+        TransactionSignature txSig0 = unsignedTx.calculateSignature(0, key0,
+                ScriptBuilder.createOutputScript(key0).getProgram(),
+                Transaction.SigHash.ALL, false);
+        assertEquals("30450221008b9d1dc26ba6a9cb62127b02742fa9d754cd3bebf337f7a55d114c8e5cdd30be022040529b194ba3f9281a99f2b1c0a19c0489bc22ede944ccf4ecbab4cc618ef3ed01",
+                HEX.encode(txSig0.encodeToBitcoin()));
+
+        Script scriptCode = new ScriptBuilder()
+                .data(ScriptBuilder.createOutputScript(LegacyAddress.fromKey(netParams, key1))
```
schildbach: @Chm While trying to integrate your new methods into LocalTransactionSigner, I stumbled upon this line. Can you help with these two questions?

1. Why is it that a P2PKH script (rather than P2WPKH) is created here, with `ScriptBuilder.createOutputScript(LegacyAddress.fromKey(netParams, key1))`? `key1` is used for receiving to a native segwit address (P2WPKH), so I'd expect `SegwitAddress.fromKey()`. But of course with this change the unit test fails, so your code must be right (on the assumption that the BIP example is correct too).

2. Why is the output script wrapped into a data push? This seems unusual, and it also differs from how it is done for the P2PKH spend in line 328. Again, the unit test proves correctness, but is there any mention in the BIP? I read everything carefully and I couldn't find this wrapped construct anywhere.

My ultimate problem is that though everything works in your two unit tests, I cannot get the signature right (likely the sig hash).

PR #1677, opened 2018-12-24
Subclass override point

Chm:
Subclass override point required for providing private key with a dedicated signer, while only public key were stored in wallet keychain. Thus no decryption is needed when a mobile app is starting, and private keys could be generated on the fly only when performing a real sign action.


Chm:
Can't get failure details from travis.

Local test failed at random: a random generated keychain group got a private key longer than 256 bits after AES decryption.

schildbach:
I restarted the build and now it succeeds. Yes, I'm aware of the randomly failing test. When I looked at it last time I could not find a better solution than just retry. I can't remember the details right now, but if it's important enough I'll open a ticket with the details early next year.

Regarding this PR: I wonder if you could add a unit test that serves two purposes, show how you plan to use the extension points and make sure they continue to exist because they're being used in the test.

I was planning to overhaul the TransactionSigner hierarchy. Right now, there is some duplicated code and maybe for segwit it would be useful to have something like a "Segwit Signer". But I'm not sure, just wanted to mention the plan.

Chm:
Sure, I can demonstrate how we're doing with a test case.

Within our approach, a signer, which is not a key bag, holds a private key to generate a whole hd hierarchy, that's still a bit strange. But it works with minimum change within this library, and our wallet has been released on play store.

While writing my own signer, I wondered:
a. should a key bag or wallet or signer answer to if it can sign for an input?
b. each signer, including the missing signature signer, must check correctlySpends() for each input, with evm evaluation. Could some of these checks be relieved for a mobile device?
Definitely need some detailed plan for a refactor.

PR #1678, opened 2018-12-24
Remove not precise comment

oscarguindzberg:
This contructor is in fact "Used for deserialization" but not only for deserialization.


schildbach:
Merged.

PR #1679, opened 2018-12-28
DRMWorkaround: JceSecurity.isRestricted remove final modifier

oscarguindzberg:
On Oracle's jdk8 (u192) JceSecurity.isRestricted is declared final so the code without this fix fails.

See https://stackoverflow.com/questions/1179672/how-to-avoid-installing-unlimited-strength-jce-policy-files-when-deploying-an


schildbach:
Merged.

PR #1680, opened 2019-01-02
DRMWorkaround: Skip workaround on OpenJDK.

schildbach:
According to https://stackoverflow.com/questions/1179672/how-to-avoid-installing-unlimited-strength-jce-policy-files-when-deploying-an the workaround is not needed on OpenJDK.

@oscarguindzberg What do you think?


schildbach:
That's good enough, thanks for the testing!

PR #1688, opened 2019-01-24
Use "best chain" instead of "main chain".

oscarguindzberg:
Some comments used "main chain", while some others used "best chain".
Use "best chain" everywhere for consistency with the code.


schildbach:
Agreed & merged.

PR #1689, opened 2019-01-27
DeterministicKeyChain, MarriedKeyChain: Remove most constructors and static constructors.

schildbach:
Migrate all use of constructors to use the builders instead.

While working on #1563 I noticed we will either break API or have an insane amount of constructors in DeterministicKeyChain and MarriedKeyChain. My proposal is to remove them all but the bare necessary and those that remain are considered unstable API in future. Consumers should use DeterministicKeyChain.builder() and MarriedKeyChain.builder().

Any comments on this?

@devrandom You added the builder pattern a good while ago.


schildbach:
Merged.

Issue #1477, opened 2017-11-03
Cannot find symbol 'sun.misc.Cleaner' on MacOS/Java 9


bendyarm:
I followed the instructions for a full build but I got a compilation error.
MacOS 10.13.1
Java version: 9.0.1
Apache Maven 3.5.2
git clone ...bitcoinj
mvn clean package

Here's the relevant output:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.2:compile (default-compile) on project bitcoinj-core: Compilation failure
[ERROR] [snipped]/bitcoinj/core/src/main/java/org/bitcoinj/store/WindowsMMapHack.java:[36,9] cannot find symbol
[ERROR]   symbol:   class Cleaner
[ERROR]   location: class org.bitcoinj.store.WindowsMMapHack
```

schildbach:
What branch/tag/commit hash are you on?

bendyarm:
master.

```
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working tree clean
$ git log | head -n 1
commit 9f5f973d8a41b3f97157939d38479d17b884ce21
$ 
```

schildbach:
Can you try Java 8? Seems like the `sun.misc.Cleaner` class was removed.

tompreston:
Hi,
I'm also getting stuck here. How do I tell maven to use Java 8? I've tried the sample here https://www.mkyong.com/maven/how-to-tell-maven-to-use-java-8/ 

like so:
```
diff --git a/pom.xml b/pom.xml
index 4a06e1eb..bd78d667 100644
--- a/pom.xml
+++ b/pom.xml
@@ -192,6 +192,8 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <generated.sourceDirectory>gen</generated.sourceDirectory>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
   </properties>

   <profiles>
```

but I'm still seeing the error:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.2:compile (default-compile) on project bitcoinj-core: Compilation failure
[ERROR] [also snip]/bitcoinj/core/src/main/java/org/bitcoinj/store/WindowsMMapHack.java:[34,9] cannot find symbol
[...]

PR #1695, opened 2019-02-07
[WIP] Remove mmap usage in SPVBlockStore

oscarguindzberg:
There has been a "TODO" for years to remove mmap usage in SPVBlockStore.

I replaced it by direct usage of RandomAccessFile

I am motivated to implement this because I found https://github.com/bitcoinj/bitcoinj/issues/1477 fix does not work on windows.

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
@@ -30,8 +30,6 @@ import java.util.concurrent.locks.*;
 
 import static com.google.common.base.Preconditions.*;
 
-// TODO: Lose the mmap in this class. There are too many platform bugs that require odd workarounds.
-
 /**
  * An SPVBlockStore holds a limited number of block headers in a memory mapped ring buffer. With such a store, you
  * may not be able to process very deep re-orgs and could be disconnected from the chain (requiring a replay),
@@ -44,18 +42,10 @@ public class SPVBlockStore implements BlockStore {
     public static final int DEFAULT_CAPACITY = 5000;
     public static final String HEADER_MAGIC = "SPVB";
 
-    protected volatile MappedByteBuffer buffer;
     protected final NetworkParameters params;
 
     protected ReentrantLock lock = Threading.lock("SPVBlockStore");
 
-    // The entire ring-buffer is mmapped and accessing it should be as fast as accessing regular memory once it's
-    // faulted in. Unfortunately, in theory practice and theory are the same. In practice they aren't.
-    //
-    // MMapping a file in Java does not give us a byte[] as you may expect but rather a ByteBuffer, and whilst on
-    // the OpenJDK/Oracle JVM calls into the get() methods are compiled down to inlined native code on Android each
-    // get() call is actually a full-blown JNI method under the hood, meaning it's unbelievably slow. The caches
-    // below let us stay in the JIT-compiled Java world without expensive JNI transitions and make a 10x difference!
     protected LinkedHashMap<Sha256Hash, StoredBlock> blockCache = new LinkedHashMap<Sha256Hash, StoredBlock>() {
         @Override
         protected boolean removeEldestEntry(Map.Entry<Sha256Hash, StoredBlock> entry) {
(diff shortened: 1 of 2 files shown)


schildbach:
Thanks for the effort! Do you have an idea if that change might affect performance?

schildbach:
Ah, and the format is stays the same, right?

core/src/main/java/org/bitcoinj/core/StoredBlock.java
```diff
@@ -142,6 +144,33 @@ public static StoredBlock deserializeCompact(NetworkParameters params, ByteBuffe
         return new StoredBlock(params.getDefaultSerializer().makeBlock(header), chainWork, height);
     }
 
+    /** Serializes the stored block to a custom packed format. Used by {@link org.bitcoinj.store.SPVBlockStore}. */
```
schildbach: I think it's a bit unfortunate that these two methods duplicate the above two methods. In theory, we could implement the new methods like this:

```
FileChannel channel = randomAccessFile.getChannel();
ByteBuffer buffer = ByteBuffer.allocate(CHAIN_WORK_BYTES);
deserializeCompact(params, buffer);
```

Of course I don't about the performance implications and wether we run into specific problems in Windows again. Have you considered this?

oscarguindzberg: I added a commit with your suggestion "Avoid duplicating StoredBlock's serializeCompact() and deserializeCompact".
I will squash once we finish the discussion and testing.
I don't think this deduplication commit brings up the windows problem again, since the windows problem is just with mmaped files not with the uses of Buffers or FileChannels.
About performance, this deduplication commit does not improve/deteriorate performance. I will address your performance question in a broader PR comment.

oscarguindzberg:
@schildbach 
The block store file format is not changed.

About performance: 
I guess this change will, in fact, deteriorate i/o performance (See https://en.wikipedia.org/wiki/Memory-mapped_file). I don't know how much worse it will be, we can test it but I guess it will depend on the use case.
Mike suggested to eventually remove the use of mmaped files back in 2015 https://github.com/bitcoinj/bitcoinj/commit/07d85f24ad124d61f061d65a2fdf42800241dd5f but I don't know what other problems they cause on top of this windows specific problem on releasing the file control.

Since this will require lots of test and might be controversial, I think it would be better as a first step to make the change optional/just for windows. A couple of options:
- Overload constructors to add a `boolean useMmapedFile` parameter.
- Add some `if (System.getProperty("os.name").toLowerCase().contains("win"))`
- Create a `public class NotMmappedFileSPVBlockstore extends SPVBlockstore`

schildbach:
For Android, Mike had this to say about mmap: "on Android each get() call is actually a full-blown JNI method under the hood, meaning it's unbelievably slow". So from my Android perspective, I'd probably be ok to just change the implementation to RandomAccessFile and be done with it. On Windows, RandomAccessFile seems to be the only available option anyway. So only Linux and MacOS are left to decide.

Yes, testing needs to be done. But we're entering a stabilization phase anyway.

oscarguindzberg:
@schildbach What would be the next step with this PR?

About unit tests: they pass on my computer. Travis unit tests execution fails but I can't find out why.

schildbach:
I'd prefer to keep only one implementation: either mmap or RandomAccessFile. I'm not afraid about testing, as we're about to enter the stabilization phase for 0.15. I leave it up to you wether you think we should try it for 0.15.

oscarguindzberg:
A similar bugfix is being tested on the Bisq project https://github.com/bisq-network/bisq/pull/2403 . Let's see what happens there before merging the PR.

schildbach:
If you rebase this to current master, it should be picked up by the new performance test.

ManfredKarrer:
I did some basic runtime testw with the Bisq app and it seems the performacne degradation at initial download is acceptable. I still would prefer to only apply it for Windows not for all OS. See https://github.com/bisq-network/bisq/pull/2403#issuecomment-463799943

schildbach:
Ok, then I suggest a more straightforward name… SPVRandomAccessFileBlockStore?

oscarguindzberg:
I rebased from master.

I ran SPVBlockStoreTest.performanceTest() on a mac
old mmapped-version takes 0.6 seconds
new non-mmapped-version takes 5.24 seconds.

This is around 10x performance loss on writing.

I will think more about this and post another comment.

oscarguindzberg:
So...
Travis build fails because SPVBlockStoreTest.performanceTest() takes 2.2 seconds.
About performance:
- We can assume 10x performance loss on writing.
- The question is whether that matters or not. The blockchain is written a lot during chain sync. Chain sync might be huge for:
  - new wallets.
  - wallets being re-opened after a while.
  - restore from seed.
- If the wallet is connecting to a remote peer, p2p delays might be orders of magnitude bigger than the performance loss of not using mmapped file. But syncing to a local bitcoind is probably a common use case on several projects.
- We still no did test on performance when reading yet. There is a software cache on SPVBlockStore so it should prevent going to the file in many cases.

On a side note...
All this started because restore from seed fails on windows.
I started working on a different approach to solve that problem.
https://github.com/bitcoinj/bitcoinj/pull/1702

oscarguindzberg:
I close this PR and suggest to merge #1702 instead.

PR #1696, opened 2019-02-10
KeyChainGroup: Migrate constructors to builder pattern.

schildbach:
KeyChainGroup: Remove most constructors. Those that remain are considered an unstable API.

Migrate all use of constructors to use a builder instead.


schildbach:
Merged.

PR #1697, opened 2019-02-11
[WIP] Segwit migration

schildbach:
The goal of this PR is to define a migration path: `Basic chain --> P2PKH chain --> P2WPKH chain`.

The first migration had already been implemented years ago with the introduction of BIP32 HD-wallets. The second migation happens now as P2WPKH/Bech32 payments are becoming more popular. We will need the P2PKH fallback chain for a while since there are still a number of senders which cannot deal with P2WPKH/Bech32.

We need to pay attention to the migration path (now, and in future): If a user restores from a years old wallet containing only basic keys, we cannot skip deriving the P2PKH chain because the user might have money sitting on it. Only for wallets which have been created as pure P2WPKH wallets we can safely skip the fallback(s).

If you use the `Wallet` class, all you can do with a fallback chain is request a fresh receiving address from it via `Wallet.freshReceiveAddress(Script.ScriptType scriptType)`.

TODO the actual migration `P2PKH --> P2WPKH`
TODO maybe we need the concept of a "current address" for fallbacks too

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
+++ b/core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
@@ -44,8 +44,13 @@ import static com.google.common.base.Preconditions.*;
 /**
  * <p>A KeyChainGroup is used by the {@link Wallet} and manages: a {@link BasicKeyChain} object
  * (which will normally be empty), and zero or more {@link DeterministicKeyChain}s. The last added
- * deterministic keychain is always the active keychain, that's the one we normally derive keys and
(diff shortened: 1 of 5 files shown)


schildbach:
Merged.

lacksfish:
Quick question, is this PR already included in one of the 0.14.x versions?

I'm basically just clicking around, and noticed 'Bitcoin Lightning Wallet' (lnwallet) already offered features from this PR way in advance, and was wondering if that was simply from early usage of this PR, or a direct predecessor of it.

PR #1700, opened 2019-02-15
WalletAppKit: Use always provideBlockStore() to simplify using a diff…

oscarguindzberg:
…erent block store when needed


schildbach:
Merged!

PR #1702, opened 2019-02-15
Clear the blockchain file (as opposed to delete it) in preparation for restore

oscarguindzberg:
This is an alternative solution to https://github.com/bitcoinj/bitcoinj/pull/1695

### DIFF
--- a/core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
+++ b/core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
@@ -69,7 +69,7 @@ public class WalletAppKit extends AbstractIdleService {
     protected final KeyChainGroupStructure structure;
     protected final String filePrefix;
     protected volatile BlockChain vChain;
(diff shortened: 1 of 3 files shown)


schildbach:
Sounds like a good idea to me!

schildbach:
If that solves your problem, I like it more than #1695.

oscarguindzberg:
I just:
- Removed WalletAppKit.provideBlockStore() and modified WalletAppKit's vStore instance variable type to stress it is of SPVBlockStore type: WalletAppKit won't work with another store type. The "overridable" solution was half-way done. Overriding provideBlockStore() to return any of the other BlockStore implementations included in bitcoinj would make the system fail on restore from seed/key. 
- Rebased from master.
- Removed "[WIP]" label from the title.

oscarguindzberg:
@schildbach don't forget about this PR :)

schildbach:
Thanks for the reminder! I just merged.

Issue #1707, opened 2019-02-22
Build issue on master and release-0.15


msgilligan:
Running `gradle clean build` on `master` (123f7be3efac2d1864cad2cd1ae868578ad2617c) or `release-0.15` fails with the following error:

```
/Users/sean/git/bitcoinj/core/src/main/java/org/bitcoinj/core/PeerGroup.java:1285: error: reference not found
     * {@link PeerConnectionEventListener} and use the onPeerConnected/onPeerDisconnected methods.
              ^
/Users/sean/git/bitcoinj/core/src/main/java/org/bitcoinj/core/TransactionConfidence.java:227: error: reference not found
     * {@link BlockChainListener}, attach it to a {@link BlockChain} and then use the getters on the
              ^
```

This is undetected by CI because it uses `gradle build check` which doesn't compile JavaDoc.

msgilligan:
Add PR #1708 and PR #1709 to fix JavaDoc and do a more complete CI build.

schildbach:
Thanks for spotting this. Both issues are fixed now.

PR #1709, opened 2019-02-22
More complete build for Travis CI

msgilligan:
... so more build errors are caught by CI.

Gradle `build` rather than `check` in Travis CI. Also remove the exclusion of `:wallettemplate:check`.

This also addresses Issue #1707

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ jdk:
 install:
   - sudo apt-get install -y protobuf-compiler
 script:
-  - gradle clean check -x :wallettemplate:check
+  - gradle clean build
 
 notifications:
   irc:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -8,7 +8,7 @@ jdk:
 install:
   - sudo apt-get install -y protobuf-compiler
 script:
-  - gradle clean check -x :wallettemplate:check
+  - gradle clean build
```
schildbach: This won't work -- we need to exclude wallettemplate from the build since there is no reliable way to declare the JavaFX dependency.

schildbach: I suggest #2760.

schildbach:
Fixed via 3974b38bad78c703a358650f78c9e0ba6ce6b82b

.travis.yml
msgilligan: Yeah sorry, my local build was using Oracle JDK8.

> there is no reliable way to declare the JavaFX dependency

I might be interested in trying to find a solution for that in the future. There's got to be a way.

> I suggest #2760.

What are you referring to by #2760?

schildbach: No idea where I got this number from. I actually meant  #1711.

Issue #1713, opened 2019-02-23
Upgrade JavaFX wallettemplate to require JDK 11+


msgilligan:
It would be nice to be able to build the JavaFX components of bitcoinj with CI. In [a comment](https://github.com/bitcoinj/bitcoinj/pull/1709#discussion_r259294706) @schildbach says "there is no reliable way to declare the JavaFX dependency."

One possibility: Require a minimum of Java 11 for the JavaFX module and use OpenJFX which can be downloaded from Maven Central and should work with other Oracle JDK or Open JDK. (See [Run HelloWorld using Gradle](https://openjfx.io/openjfx-docs/#gradle).)

As the JavaFX `wallettemplate` isn't widely used (I think) and (in my opinion) the best way to deliver JavaFX apps is with a bundled JVM, it may be acceptable to require (what is currently) the latest JDK.

If there's agreement on this approach, I can try to make a PR for `wallettemplate` that requires JDK 11 and pulls the OpenJFX from Maven Central.

schildbach:
So would that JDK 11 build on wallet-template then run in addition to what we have already? I'd certainly not want to drop JDK 8 for building in the near future.

Generally I think wallet-template in CI has not much benefit since the "wallet usecase" is already covered by wallet-tool (which _is_ under CI). This is why I excluded it quite lightheartedly when the OpenJFX dependency became a problem.

msgilligan:
> So would that JDK 11 build on wallet-template then run in addition to what we have already?

Yes, `wallet-template` would only build on JDK 11 and later -- or at least that's my idea. Here's my first attempt at doing that: https://github.com/msgilligan/bitcoinj/commit/58e83d4b57d1af2bf1bb421df5ecd85e0973dc68

>  I'd certainly not want to drop JDK 8 for building in the near future.

I totally agree. In fact,I'm thinking about opening an issue/PR to try to get the CI builds to use JDK 7 for part of the build so we can be sure to find any accidental usages of JDK 8 features.

msgilligan:
OK, PR #1715 is ready for your review.

msgilligan:
PR #1715 is now (indirectly) merged. Closing.

PR #1715, opened 2019-02-25
Build wallettemplate with JDK 11+ and OpenJFX downloaded from Maven Central

msgilligan:
Possible resolution to Issue #1713 

* Create new `settings-jdk11.gradle` Gradle settings file
* Update `.travis.yml` to use `-c settings-jdk11.gradle` when building with JDK11
* Update build instructions in README

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ jdk:
   - openjdk11
 install: true
 script:
-    # excluding wallet-template because of the unreliable JavaFX dependency
-  - gradle clean build -x :wallettemplate:compileJava
+  # wallettemplate subproject is conditionally built if JDK 11+ is installed
+  - gradle clean build
 
 notifications:
   irc:
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,4 +1,12 @@
+import org.gradle.util.GradleVersion
+
 include 'core'
 include 'tools'
 include 'examples'
-include 'wallettemplate'
+def minGradleVersion = GradleVersion.version("4.10")
+if (GradleVersion.current().compareTo(minGradleVersion) >= 0 &&  JavaVersion.current().isJava11Compatible()) {
+    println "Including wallettemplate because ${GradleVersion.current()} and Java ${JavaVersion.current()}"
+    include 'wallettemplate'
+} else {
+    println "Skipping wallettemplate: ${GradleVersion.current()} and Java ${JavaVersion.current()}"
+}
--- a/wallettemplate/build.gradle
+++ b/wallettemplate/build.gradle
@@ -1,5 +1,11 @@
+plugins {
+    id 'org.openjfx.javafxplugin' version '0.0.7' apply false
+}
+
 apply plugin: 'java'
 apply plugin: 'eclipse'
+apply plugin: 'application'
+apply plugin: 'org.openjfx.javafxplugin'
 
 eclipse.project.name = 'bitcoinj-wallettemplate'
 
@@ -12,11 +18,16 @@ dependencies {
     implementation 'org.slf4j:slf4j-jdk14:1.7.25'
 }
 
-sourceCompatibility = 1.8
+javafx {
+    modules = [ 'javafx.controls', 'javafx.fxml' ]
+}
+
+sourceCompatibility = 1.11
 compileJava.options.encoding = 'UTF-8'
 compileTestJava.options.encoding = 'UTF-8'
 
-task wallet_template(type: JavaExec) {
-    main = 'wallettemplate.Main'
-    classpath = sourceSets.main.runtimeClasspath
+mainClassName = 'wallettemplate.Main'
+
+// task wallet_template can be replaced with the 'run' task
+task wallet_template(dependsOn: 'run') {
 }
--- a/README.md
+++ b/README.md
@@ -8,8 +8,10 @@ The bitcoinj library is a Java implementation of the Bitcoin protocol, which all
 
 ### Technologies
 
-* Java 7 for the core modules, Java 8 for everything else
-* [Gradle 3.4+](https://gradle.org/) - for building the project
+* Java 7 for the `core` module
+* Java 8 for `tools` and `examples`
+* Java 11 for the JavaFX-based `wallettemplate`
+* [Gradle 3.4+](https://gradle.org/) - for building the project (Gradle 4.10.2 or later is recommended)
 * [Google Protocol Buffers](https://github.com/google/protobuf) - for use with serialization and hardware communications
 
 ### Getting started
@@ -18,17 +20,20 @@ To get started, it is best to have the latest JDK and Gradle installed. The HEAD
 
 #### Building from the command line
 
-To perform a full build use
+Official builds are currently using with JDK 8, even though the `core` module is compatible with JDK 7 and later.
+
+To perform a full build (*including* JavaDocs and unit/integration *tests*) use JDK 8+
 ```
 gradle clean build
 ```
-You can also run
-```
-gradle javadoc
-```
-to generate the JavaDocs.
+If you are running JDK 11 or later the build will automatically include the JavaFX-based `wallettemplate` module. The outputs are under the `build` directory.
 
-The outputs are under the `build` directory.
+To perform a full build *without* unit/integration *tests* use:
+```
+gradle clean assemble
+```
+
+
 
 #### Building from an IDE
 


msgilligan:
I'm almost certain the build failure is one of those intermittent integration test failures and has a good chance of running if restarted.

.travis.yml
```diff
@@ -2,13 +2,15 @@
...
 language: java
-jdk:
-  - openjdk8
-  - openjdk11
-install: true
-script:
-    # excluding wallet-template because of the unreliable JavaFX dependency
-  - gradle clean build -x :wallettemplate:compileJava
+install:
+  - sudo apt-get install -y protobuf-compiler
```
schildbach: Is this a merge error, or do you intentionally want to install the protobuf debian package again?

.travis.yml
```diff
@@ -2,13 +2,15 @@
...
+
+matrix:
+  include:
+    - jdk: openjdk8
+      script: gradle clean build
+    - jdk: openjdk11
+      script: gradle -c settings-jdk11.gradle clean build
```
schildbach: Rather than using a separate settings-jdk11.gradle, did you try using the `-x` (exclude) switch?

.travis.yml
msgilligan: Shit. It's a merge error. Sorry.

.travis.yml
msgilligan: I did try that, the problem is the `wallettemplate` module is using a JDK11-requiring plugin and even with the `-x` switch the build still fails.

.travis.yml
msgilligan: Fixed.

.travis.yml
```diff
@@ -2,13 +2,13 @@
 sudo: false
 dist: xenial
 language: java
-jdk:
-  - openjdk8
-  - openjdk11
-install: true
```
schildbach: According to [Travis](https://docs.travis-ci.com/user/job-lifecycle/#skipping-the-installation-phase), the `install: true` needs to stay in if you don't have anything to do in the install phase. This is because the install phase isn't optional.

msgilligan: It seemed to work OK without out, but I'll put it back.

.travis.yml
msgilligan: Actually, I have an even better solution --- stay tuned.

schildbach:
Can also rebase and squash everything?

msgilligan:
> Can also rebase and squash everything?

I can, but Github will do that for you automatically (if you choose the option) when you merge. Will that work for you?

.travis.yml
schildbach: In my experience, if you leave it out the `script` phase gets replaced with the default `gradle assemble`.

schildbach:
> > Can also rebase and squash everything?
> 
> I can, but Github will do that for you automatically (if you choose the option) when you merge. Will that work for you?

I generally don't use the Github UI for merging. But even if I would, I couldn't due to the merge conflict.

msgilligan:
By the way, I'm enjoying using [sdkman](https://sdkman.io) to switch between JVMs:

```
$ sdk use java 8.0.201-oracle
$ gradle clean build
$ sdk use java 11.0.2-open
$ gradle clean build
```

msgilligan:
> I generally don't use the Github UI for merging. But even if I would, I couldn't due to the merge conflict.

OK, I'll fix it.

msgilligan:
I'm not use to doing rebase/squash from the command-line (but can learn quickly I hope) -- is this good enough?

msgilligan:
Actually wait for my next commit...

msgilligan:
@schildbach Sorry for the hassle -- I think everything is ready now.

settings.gradle
```diff
@@ -1,4 +1,6 @@
 include 'core'
 include 'tools'
 include 'examples'
-include 'wallettemplate'
+if (JavaVersion.current().isJava11Compatible()) {
```
schildbach: Unfortunately Gradle 3.4.1 (which is the version that is in Ubuntu 18.04) does not have `.isJava11Compatible()`. It would have `.isJava10Compatible()` though, in case the javafxplugin is compatible with Java 10, too.

settings.gradle
```diff
@@ -1,4 +1,6 @@
...
+    include 'wallettemplate'
+}
```
schildbach: EOL missing

wallettemplate/build.gradle
```diff
@@ -12,11 +20,16 @@ dependencies {
...
+sourceCompatibility = 1.11
 compileJava.options.encoding = 'UTF-8'
 compileTestJava.options.encoding = 'UTF-8'
 
-task wallet_template(type: JavaExec) {
-    main = 'wallettemplate.Main'
-    classpath = sourceSets.main.runtimeClasspath
+mainClassName = 'wallettemplate.Main'
+
+// task wallet_template and can be replaced with the 'run' task
```
schildbach: This migration could actually be its own PR, since it's an independent idea that can be discussed and merged independently.

settings.gradle
schildbach: Even Gradle 4.4.1 (the current version on Debian testing and Ubuntu 19.04) doesn't have `.isJava11Compatible()`.

schildbach:
Perhaps we should go back to the drawing board with this. What advantages does the `org.openjfx.javafxplugin` actually have? The Gradle project page is very vague about this: 

> Plugin that makes it easy to work with JavaFX

Did you try simply adding a dependency to OpenJFX? I believe the OpenJDK dependencies are all listed here: https://mvnrepository.com/artifact/org.openjfx

settings.gradle
msgilligan: > in case the javafxplugin is compatible with Java 10, too.

Java 11 is the first version that doesn't include JavaFX by default. It's also the long-term release so it makes some sense that the plugin requires 11+.

wallettemplate/build.gradle
msgilligan: Actually, I think it's a byproduct of the OpenJFX plugin which uses/requires the Gradle 'application' plugin.

schildbach: Ah ok, didn't know. Thanks.

settings.gradle
schildbach: I agree. I'm just looking for workarounds. Another workaround may be using `getMajorVersion()` and do some string foo.

schildbach: Or maybe a construct like `isAtLeastGradle4.7() && isJava11Compatible()` would guard us from the exception? This would of course make the wallet-template build require Gradle 4.7.

msgilligan:
> Perhaps we should go back to the drawing board with this.

We've got (at least) three options that I can think of:

1. Try adding OpenJFX dependencies directly (as you suggest)
2. Use an alternative (non-Gradle, perhaps) mechanism to test for JDK 11+
3. Reconsider using the Gradle wrapper

> What advantages does the `org.openjfx.javafxplugin` actually have?

I'm not really sure, this is the first time I've used it and I just followed the [instructions](https://openjfx.io/openjfx-docs/#gradle). I am eager to update another JavaFX project I have to be JDK 11+, so I will be learning more about OpenJFX over the next month or two.

The documentation isn't very good about "Why", but I think the Plugin does some/all of the following:

1. Figures out which native JARs to depend on based upon the platform
2. Supports building fat jars and using the `jlink` tool to build a stripped, bundled JVM.
3. Works wit both module-based (Java 9+ module system) and classpath-based builds

I also noticed on the instructions page it says: 

> Note: Make sure to use gradle version 4.10+ and set the JAVA_HOME environment variable to JDK 11.

So this would require either using `gradlew` or checking the Gradle version before trying to include the `wallettemplate` module.

Personally, I'm a big fan of using the Gradle Wrapper. I know it breaks the rule of not checking in binaries, but it also helps significantly with ease-of-use and creating a consistent, reproducible build.

settings.gradle
msgilligan: > workaround may be using `.getMajorVersion()` and do some string foo.

> a construct like `isAtLeastGradle4.7() && isJava11Compatible()`

This corresponds to "option 2" I suggested here: https://github.com/bitcoinj/bitcoinj/pull/1715#issuecomment-467086167

> This would of course make the wallet-template build require Gradle 4.7.

It looks like the OpenJFX plugin requires Gradle 4.10 or later.

settings.gradle
msgilligan: Fixed and a rebased branch pushed.

msgilligan:
> Don't worry about the Travis errors, I'm currently always restarting the failing jobs. Most of the time, the reason is multiple concurrently execution tests that bind to the same port.

I would really like to try to fix this. Can we turn off concurrently executing tests?

schildbach:
The Travis errors should be almost be fixed via 0be1008339ea0297221d378877a9c5ec45845869. I recommend rebasing again.

schildbach:
I'm against using Gradle wrapper. Travis uses Gradle 4.10.2 so it should be fine with the conditional. I investigated briefly but didn't find a method to programmatically check the Gradle version though.

msgilligan:
OK, I added a programmatic check of the Gradle version, testing with various combinations of JDK 8, 11 and Gradle 3.5.1, 4.10.2, 4.10.3, and 5.2.1 (all on macOS) and this seems to be working reasonably well.

I also added "Gradle 4.10.2 or later is recommended" to the README.

schildbach:
I reworded the README a bit and merged! Thanks a lot for working on this.

PR #1720, opened 2019-02-28
 Remove useless brackets

show-rosarugosa:
In the description of the parameter eventHorizon of the method context(), a useless  brackets appears.


schildbach:
Merged.

PR #1726, opened 2019-03-08
prevent possible memory leak in Peer

### LINKED ISSUE #1719: memory leak if remote peer doesn't reply to a ping
When we are sending pings to a Peer that never sends `Pong` back for any reason, the `org.bitcoinj.core.Peer#pendingPings` is getting filled with `PendingPing` instances that are never cleared. (Normally they are cleared only on Pong receival)

Can we throw old `PendingPings` away (add timeout)? Disconnect from such Peer (how)? 

Similar problem seems to be with `org.bitcoinj.core.Peer#pendingTxDownloads` that gets filled if some transactions never arrive(?).

pvyhnal-generalbytes:
- when a peer does not respond to Ping,
- when a peer does not send announced transactions 

fixes #1719 

Please review if this way to disconnect is OK.

I was not sure about the number of pending objects before disconnect. Using too low number may lead to reconnecting too often. I saw a peer with around 15000 pending pings when running for around a day but was not able to reproduce it recently. Another peer had around 2000 pending transactions when running over one night.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -1213,6 +1213,11 @@ public class Peer extends PeerSocketHandler {
             } else {
                 log.debug("{}: getdata on tx {}", getAddress(), item.hash);
                 getdata.addTransaction(item.hash, vPeerVersionMessage.isWitnessSupported());
+                if (pendingTxDownloads.size() > 100) {
+                    log.info("{}: Too many pending transactions, disconnecting", this);
+                    close();
+                    return;
+                }
                 // Register with the garbage collector that we care about the confidence data for a while.
                 pendingTxDownloads.add(conf);
             }
@@ -1552,6 +1557,10 @@ public class Peer extends PeerSocketHandler {
         final VersionMessage ver = vPeerVersionMessage;
         if (!ver.isPingPongSupported())
             throw new ProtocolException("Peer version is too low for measurable pings: " + ver);
+        if (pendingPings.size() > 50) {
+            log.info("{}: Too many pending pings, disconnecting", this);
+            close();
+        }
         PendingPing pendingPing = new PendingPing(nonce);
         pendingPings.add(pendingPing);
         sendMessage(new Ping(pendingPing.nonce));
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -1213,6 +1213,11 @@ protected void processInv(InventoryMessage inv) {
             } else {
                 log.debug("{}: getdata on tx {}", getAddress(), item.hash);
                 getdata.addTransaction(item.hash, vPeerVersionMessage.isWitnessSupported());
+                if (pendingTxDownloads.size() > 100) {
```
schildbach: I would put these constants into a constant field.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -1552,6 +1557,10 @@ private void addPingTimeData(long sample) {
         final VersionMessage ver = vPeerVersionMessage;
         if (!ver.isPingPongSupported())
             throw new ProtocolException("Peer version is too low for measurable pings: " + ver);
+        if (pendingPings.size() > 50) {
```
schildbach: You're fiixing two different cases. I'd prefer to have them in two commits, one for each.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -150,6 +150,8 @@ public GetDataRequest(Sha256Hash hash, SettableFuture future) {
     private final ReentrantLock lastPingTimesLock = new ReentrantLock();
     @GuardedBy("lastPingTimesLock") private long[] lastPingTimes = null;
     private final CopyOnWriteArrayList<PendingPing> pendingPings;
+    // Disconnect form a peer that is not responding to Pings
```
schildbach: typo: form

pvyhnal-generalbytes: Fixed, thanks

schildbach:
I just merged the 'Ping' aspect of your PR. Can you rebase this PR on current master and squash everything into one commit? I need to evaluate the possible side effects of the 'pendingTxDownloads' aspect some more.

pvyhnal-generalbytes:
I made a separate PR for the `pendingTxDownloads` https://github.com/bitcoinj/bitcoinj/pull/1734

schildbach:
Btw. you don't need the create a new PR for rebases. Just force push (git push -f) your rebased brach over your old remote branch – github will pick it up automatically.

pvyhnal-generalbytes:
Sorry, I wasn't sure about changing the commits that were already pushed, so I think we can close this PR and continue in #1734 
Just future reference the merged part of this PR is 29f450b8609fd17c33a2d6da5613fe63cb003601

schildbach:
Closing this in favour of #1734.

PR #1733, opened 2019-03-14
Remove useless null check

show-rosarugosa:
Delete the useless null check, I also saw that you said this place is not suitable for the transaction check, I am currently working hard to read the source code, maybe I can help you write this check in the future.


schildbach:
Merged, thanks!

PR #1735, opened 2019-03-15
Add some transactions null check

show-rosarugosa:
In some locations, some null checks for transactions have been added. Currently, this check can only be done. In the future, I would like to modify this part of the code to do a more efficient check.


show-rosarugosa:
@schildbach 
Is there any problem with my modification? Please reply me when you have time.😄

schildbach:
Can you explain why are these checks needed? Blocks normally have at least one txn: the coinbase.

show-rosarugosa:
I know that the block should contain at least one coinbase transaction, but I saw the code in both places, and I am confused, so I think it would be better to add null check code everywhere.
image
image

Issue #997, opened 2015-06-18
FullPrunedBlockChain leaks memory


amichair:
I have a server using bitcoinj, connected to testnet3. It is idle (no user activity), other than bitcoin network traffic. It seems to be gaining 10MB-20MB of memory per day, until an OutOfMemoryError is thrown. I started analysing heap dumps, and it looks like FullPrunedBlockChain is the culprit, and specifically its orphanBlocks map. The docs say this is used when downloading the chain, but it appears to be growing all the time and is never cleared.

I'm not sure if/what scenarios the orphans are actually used for, but they should probably be evicted from memory when a chain download completes, or after a time limit, or just stored in the store rather than in memory, or maybe it's just a bug somewhere else that's causing them not to be found and used properly...

btw this may also be a security vulnerability - maybe a malicious peer can send a bunch of orphan blocks and cause a denial of service via OOME crashing the server.

There may be other leaks, but this appears to be taking up 50% of all heap usage after a few days of uptime - after it is fixed it will be easier to find the smaller leaks, if they exist.

mikehearn:
BitcoinJ doesn't try and be DoS resistant against peers in any part of its code.

However, on testnet especially orphan blocks happen very frequently. If the store is keeping them around forever then yeah, there should be a mechanism to clear them.

Patches welcome! The full verification code is largely community maintained at the moment.

amichair:
I'm surprised to hear that security isn't a priority, but it is what it is. I thought it was worth mentioning.

To be more precise than I was above, although the heap dump shows the FullPrunedBlockChain instance hogging the memory, the actual orphanBlocks field and its handling is in AbstractBlockChain, so this may be an issue in the regular BlockChain case as well.

I don't feel comfortable enough with the inner workings of bitcoinj to make non-trivial changes, otherwise I would have made the fix myself :-)

mikehearn:
DoS is a slightly different issue to 'security'. To me security means something like, can hackers steal your bitcoins. The worst case scenario for a peer DoS on bitcoinj is the process restarts and picks a new random set of peers.

That rationale holds very true for the mobile/desktop wallet use case bitcoinj is designed for. DoSing a mobile wallet by trying to stuff it full of data doesn't have a whole lot of benefit to the attacker. It will just bloat up and be restarted by the OS. On Android the screen might flash for a second and then everything will be back to normal.

[...]

PR #1738, opened 2019-03-18
Try connecting orphans when previous block not found

pvyhnal-generalbytes:
I'm not sure if this is the right way to fix this but I have let it run for about a week and it got into a state when last block connected into blockchain was ~5 days old and all the following ones were in orphanBlocks. I'm not sure how this happened but then all the new blocks were added to orphans and orphans were never connected (tryConnectingOrphans was never called). I tried to call tryConnectingOrphans() in debug which resolved the issue and it recovered without a need to restart and all orphans were connected to the chain. 

I found this because it causes a big memory consumption/leak as orphanBlocks takes much more memory and were never cleared, see #997

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -461,18 +461,20 @@ public abstract class AbstractBlockChain {
                 checkState(tryConnecting, "bug in tryConnectingOrphans");
                 log.warn("Block does not connect: {} prev {}", block.getHashAsString(), block.getPrevBlockHash());
                 orphanBlocks.put(block.getHash(), new OrphanBlock(block, filteredTxHashList, filteredTxn));
+                if (tryConnecting) {
+                    tryConnectingOrphans();
+                }
                 return false;
             } else {
                 checkState(lock.isHeldByCurrentThread());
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
```diff
@@ -461,18 +461,20 @@ private boolean add(Block block, boolean tryConnecting,
                 checkState(tryConnecting, "bug in tryConnectingOrphans");
                 log.warn("Block does not connect: {} prev {}", block.getHashAsString(), block.getPrevBlockHash());
                 orphanBlocks.put(block.getHash(), new OrphanBlock(block, filteredTxHashList, filteredTxn));
+                if (tryConnecting) {
```
schildbach: If the checkState() three lines above is "correct", this if() shouldn't be needed. But of course it doesn't hurt either.

schildbach:
Btw. analyzing the log file from during that week should allow to find the reason for how your node got into that state. Just out of my head: maybe it was a re-org?

pvyhnal-generalbytes:
I can confirm I didn't observe this memory leak after this fix. I did let it run for about two weeks on bitcoinj and also dashj. 
The last memory leak (but a minor one) I can observe is #1842

schildbach:
Thanks for your thorough testing – it's much appreciated! I just merged.

PR #1740, opened 2019-03-19
refactored test to add parametrized data

juanse254:


schildbach:
I'm not sure what's the advantage here? The concrete passwords are not that important for the test, it's just two arbitrary values.

Also, note that parametrized tests are more difficult to migrate to JUnit 5. So I'd say we really need an advantage if we use parametrized tests.

PR #1741, opened 2019-03-20
PeerGroup.addAddress(): don't increase max connections if address alr…

oscarguindzberg:
…eady exist


schildbach:
I'm a bit scared about the now lock.lock() inside lock.lock(), but it seems to work. Merged!

PR #1743, opened 2019-03-23
Modify inappropriate message length check

show-rosarugosa:
I think it is more appropriate to do this length check, because the above comment explains that the maximum message length refers to the length without the prefix, so here is only the verification of the true length of the message, I think it should not involve the length of the prefix. 
This code can be more easily understood, and the comments above this line of code also write to verify that the length is out of range or negative, so I think this is more readable.


schildbach:
Thanks! I just merged.

PR #1744, opened 2019-03-23
Exclude SeedPeersTest.java from normal test build

MethDamon:
SeedPeersTest.java seems to also use the network, so it should probably be exluded from the test build as well


schildbach:
Why do you think it uses the network? I don't think it does, and also I can run the test with network switched off.

MethDamon:
Hm, you're right, seems like I am mistaken.

schildbach:
Thanks anyway for looking over it.

PR #1745, opened 2019-03-23
replace AESFastEngine with AESEngine

matthewleon:
AESFastEngine raises a deprecation warning.

see https://nvd.nist.gov/vuln/detail/CVE-2016-1000339

Also see the bouncycastle 2.6.4 release notes regarding `AESFastEngine` deprecation.

First time contribution here: Not sure how worthwhile the perf vs. security tradeoff here is. My naive impression is that AES operations aren't generally on a hot path. Is this correct?

I searched through discussions here and didn't see this particular vuln come up.


schildbach:
Yeah, I guess we should merge it.

Wallet encryption usually only runs once, when the user decides to encrypt her/his wallet. From then on, typically only selected keys are decrypted for the address(es) the user wants to spend from.

schildbach:
Merged!

PR #1748, opened 2019-03-25
Refactor/kill Throwables.propagate

matthewleon:
This eliminates all three usages of `Throwables.propagate`, which is giving deprecation warnings. Two of the usages can simply be replaced by a re-`throw`, while one seems to require inlining.

See the Guava project's elaborate wiki page on the subject here: https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate


schildbach:
Thanks a lot! Merged.

PR #1750, opened 2019-03-26
Bugfix/monetary format equality

matthewleon:
`equals` and `hashCode` for `MonetaryFormat` were broken. This fixes them.


schildbach:
Good catch! Squashed & merged.

PR #1751, opened 2019-03-26
NetworkParameters: make double-checked locking safe with volatile

matthewleon:
In `NetworkParameters`, double-checked locking is performed during lazy initialization of `defaultSerializer` in `getDefaultSerializer`: https://github.com/matthewleon/bitcoinj/blob/ee8f86aaaa00efbf580eeb20e45df55f228e1ead/core/src/main/java/org/bitcoinj/core/NetworkParameters.java#L398

For this to actually be safe, `defaultSerializer` must be marked `volatile`.


schildbach:
Merged, thanks!

PR #1752, opened 2019-03-27
fix a possible NPE in peer discovery

matthewleon:
Attempting to `add` `null` to a `PriorityQueue` will cause an NPE. It seems possible for `addrToTry` to be null at this point in the code, or at least this possibility is quite hard to reason about. The log line above guards against it.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerGroup.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerGroup.java
@@ -495,7 +495,9 @@ public class PeerGroup implements TransactionBroadcaster {
                 if (retryTime > now) {
                     long delay = retryTime - now;
                     log.info("Waiting {} ms before next connect attempt {}", delay, addrToTry == null ? "" : "to " + addrToTry);
-                    inactives.add(addrToTry);
+                    if (addrToTry != null) {
+                        inactives.add(addrToTry);
+                    }
                     executor.schedule(this, delay, TimeUnit.MILLISECONDS);
                     return;
                 }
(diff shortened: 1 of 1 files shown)


matthewleon:
Reasoning through this code, I think you'd get an NPE here if `inactives` is full of _only_ IPv6 addresses (imagine it only has one address in it, and that address is IPv6... seems realistic enough) and `ipv6Unreachable` is `true`. You'll then exhaust through the queue looking for an address to try, leaving `addrToTry == null`, hit this rescheduling block, and throw an NPE on trying to push the `addrToTry` back onto the queue.

matthewleon:
Yes, you're absolutely right. I updated the PR accordingly.

schildbach:
Hm, I had I slightly different fix in mind. Why not the entire lines 497-505 in an if block? After all, we possible don't want to do anything if we don't have an address.

Btw. if you rebase your PR to master then it becomes a bit clearer.

matthewleon:
I see your point. I've rebased and rewritten this as a null guard rather than a giant if. I believe that's the most robust, and readable, solution here: the common case is to fall through.

From what I can tell, this loop is the _only_ place where anything gets popped from the `inactive` queue, so this should be safe. Multithreading here is a bit tough for me to reason about, though, so please take extra care in checking my thinking.

matthewleon:
> Multithreading here is a bit tough for me to reason about, though, so please take extra care in checking my thinking.

(looking back, I see that all this code takes is wrapped by the `lock`. Hopefully that keeps us safe here.)

schildbach:
Yeah, that looks great!

A moment I pondered we should re-schedule the job in any case, but then I think if someone adds an inactive address it's their duty to schedule the job.

schildbach:
Merged!

PR #1754, opened 2019-03-28
fix NPEs on wallet encryption

matthewleon:
The `checkState` logic here was somewhat backward, allowing NPEs to be thrown in the code below. Further, two operations on `chains` were not protected by a nullity check.


matthewleon:
Sorry for the quick force-push on this. Had to update some documentation.

matthewleon:
FWIW a lot of this logic could probably be simplified by guaranteeing non-nullity of `chains`. Is there actually any reason to allow it to be `null`?

schildbach:
I actually introduced nullablity of chains in 0.15 to express that such KeyChainGroups don't support deterministic chains. After all these years, it became clear that some users of bitcoinj simply don't want deterministic keys. (Including myself, in one specific usecase.)

schildbach:
Don't worry about force pushed on your own PR. I personally prefer not reading through history anyway when reviewing PRs.

schildbach:
Merged!

PR #1755, opened 2019-03-29
fix an npe in decryption of non-HD KeyChainGroups

matthewleon:
similar to the encryption fix in https://github.com/bitcoinj/bitcoinj/pull/1754

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
+++ b/core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
@@ -693,16 +693,17 @@ public class KeyChainGroup implements KeyBag {
     public void decrypt(KeyParameter aesKey) {
         // This code must be exception safe.
         checkNotNull(aesKey);
+
         BasicKeyChain newBasic = basic.toDecrypted(aesKey);
-        List<DeterministicKeyChain> newChains = new ArrayList<>(chains.size());
-        if (chains != null)
+        basic = newBasic;
+        if (chains != null) {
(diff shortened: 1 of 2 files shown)


matthewleon:
Sorry, looks like this one needs more work.

matthewleon:
Okay, this is ready now. It turns out that the ordering of operations here is very important if we want to ensure that exceptions don't screw up our keychain state.

schildbach:
Hmm, you've hidden a critical (if what you say is right) bugfix in this PR! I've extracted the fix from your PR and merged it separately.

Can you rebase this PR on current master?

matthewleon:
You are right about the bugfix. My apologies, I had no intention to obscure it. I'm going to rebase and take extra care not to introduce that kind of issue.

matthewleon:
It also seems to me that as long as we _are_ careful about only mutating state at the end of these methods, then the comment `// This code must be exception safe.` is not technically correct: we can throw during attempted encryption/decryption. Only the assignments at the end can't throw. But we know that assignment is safe.

matthewleon:
This is ready for review again. I've eliminated the comments regarding exception safety: I believe that the encryption/decryption methods called here can safely throw, since mutation only occurs after they are called.

schildbach:
You're right about the comment at the top, but shouldn't we move the comment(s) down to the beginning of the block of code that should not throw?

matthewleon:
I've restore comments to the sections of code that are sensitive.

schildbach:
Perfect, thanks! I merged.

PR #1756, opened 2019-03-29
eliminate a redundant null check

matthewleon:
`req.feePerKb` was already dereferenced at line 4113.

It it can be null, it should be marked as `@Nullable`.


schildbach:
You're right! Merged.

PR #1757, opened 2019-03-30
Replace Guava primitive compare() with JDK7+ compare()

msgilligan:
In Java 7, Integer.compare() and Long.compare() were added. Use those
methods rather than the Guava equivalents which are “deprecated” for
JDK7+.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Sha256Hash.java
+++ b/core/src/main/java/org/bitcoinj/core/Sha256Hash.java
@@ -18,7 +18,7 @@
 package org.bitcoinj.core;
 
 import com.google.common.io.ByteStreams;
-import com.google.common.primitives.*;
+import com.google.common.primitives.Ints;
 
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/bitcoinj/core/Sha256Hash.java
```diff
@@ -18,7 +18,7 @@
 package org.bitcoinj.core;
 
 import com.google.common.io.ByteStreams;
-import com.google.common.primitives.*;
```
schildbach: This is an unrelated change.

msgilligan: It's somewhat related, since I'm trying to identify places in the code that are using Guava functions that are now provided by the JDK. When I looked at this file, I saw it was using `Ints` (only) but not for the methods I was replacing. Do you want me to remove it and send it separately?

schildbach:
I merged it without the change in Sha256Hash. I wouldn't worry about the import unless you want to change the usage of that import too.

At one point we could "organize imports" over the whole project, but I never bothered enough.

PR #1758, opened 2019-03-30
Use JDK7+ Objects.hash() and equals() rather than Guava

msgilligan:
(Guava says their methods are “deprecated” for JDK7+)

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/core/ECKey.java
@@ -23,7 +23,6 @@ import org.bitcoinj.script.Script;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
-import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import com.google.common.primitives.UnsignedBytes;
 import org.bitcoin.NativeSecp256k1;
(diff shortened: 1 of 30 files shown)


msgilligan:
Not sure why that test failed. It passes on my Mac with OpenJDK 11.0.2.

schildbach:
Don't worry about the test. I re-ran it and it passed.

core/src/main/java/org/bitcoinj/core/ECKey.java
```diff
@@ -23,7 +23,7 @@
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
-import com.google.common.base.Objects;
+import java.util.Objects;
```
schildbach: Import order isn't right here (multiple occurences).

core/src/main/java/org/bitcoinj/core/FilteredBlock.java
```diff
@@ -17,7 +17,7 @@
 
 package org.bitcoinj.core;
 
-import com.google.common.base.Objects;
+import java.util.Objects;
```
schildbach: Duplicate import.

msgilligan: Good catch. I'll fix it

core/src/main/java/org/bitcoinj/core/ECKey.java
msgilligan: Good catch. I'll fix it

msgilligan:
OK, I force-pushed an update with correctly sorted `import`s

schildbach:
Thanks! Merged.

PR #1759, opened 2019-03-30
Replace deprecated Wallet constructor with createDeterministic()

msgilligan:


schildbach:
Yes, I think we can remove constructor in the same commit.

And maybe do the same for the `Context` variant of the constructor?

msgilligan:
Sure. I'll add those changes.

msgilligan:
Both of those constructors are used to create anonymous objects in test classes. Any suggestions on how best to preserve that test functionality after removing the constructors?

schildbach:
Reduce their visibility?

msgilligan:
> Reduce their visibility?

`protected` would allow them to be used by subclasses including anonymous subclasses in tests. We can't use package visibility because the tests are in multiple packages currently. So `protected` would work, but the question is do we want the ability to create subclasses using this constructor to be part of the API?

Two other alternatives that I can think of are:
1. Create some kind of explicit mechanism (subclasses or a mechanism using composition rather than inheritance)
2. Use a mocking framework to get the test behavior we need

I see we're already using *EasyMock* -- I haven't used that one before, so let me see if it will do what we need if we remove those constructors.

Related question: Are we running unit tests in the Android simulator? (That isn't happening on GitLab CI, but someone might be doing it locally/manually.)

msgilligan:
I decided that I didn't want to mess with subclasses or mocks as part of this PR and did a force-push using your idea of making the constructor `protected`. I also removed the constructor that took a `NetworkParameters` and replaced its usage with the constructor that takes `Context`. Although I changed the constructor to `protected`, I left it as `@Deprecated` -- I'm not sure if you want to remove that now.

msgilligan:
Oops, accidentally clicked the wrong button and closed and re-opened. Sorry.

schildbach:
I just got rid of one of the two Wallet subclasses via 8718a51c5555dd3764c11b13087ec45e431ce5bc.

The other one is more tough though.

schildbach:
Since the remaining constructur is only there for a unit test we could use the `@VisibleForTesting` annotation?

msgilligan:
The `@VisibleForTesting` annotation is only informative, right?

msgilligan:
@schildbach I added `@VisibleForTesting` and rebased to `master`.

This should be ready for merging _unless_ you want to remove the `@Deprecation` now that we have made visibility `protected` and added `@VisibleForTesting`.

schildbach:
Merged, thanks!

PR #1760, opened 2019-03-31
Wallet: Introduce BadWalletEncryptionKeyException…

schildbach:
…which is thrown if the private keys and seed of this wallet cannot be decrypted due to the supplied encryption key or password being wrong.


schildbach:
Merged.

Issue #1761, opened 2019-03-31
PeerGroup peer lists: replace with Maps?


matthewleon:
The declarations of `PeerGroup`'s peer lists has the following comment: https://github.com/bitcoinj/bitcoinj/blob/06d9b6489060584fa4c9b455b0a5856f3348953f/core/src/main/java/org/bitcoinj/core/PeerGroup.java#L105

It should be possible to get the best of both worlds using `LinkedHashMap<PeerAddress, Peer>`. `Collections.synchronizedMap` should win us the thread-safety provided by the current use of `CopyOnWriteArrayList`.

This would provide nice protection against duplication of peers. While I haven't personally seen this happen, it's a scenario that is quite difficult to test against. A type-level guarantee seem like a very nice thing to have.

Thoughts?

PR #1762, opened 2019-03-31
replace inactive peer PriorityQueue with TreeSet

matthewleon:
In the spirit of https://github.com/bitcoinj/bitcoinj/issues/1761 this PR replaces the inactive peer `PriorityQueue` with a `TreeSet`.

There is also a one-line code change here to ensure fully deterministic ordering of the inactive peer queue by tie-breaking on hash rather than just port number. The tests that would have been impacted by the change are already disabled.


matthewleon:
(I've avoided discussing perf here because I imagine the effects in that regard to be a bit difficult to determine but also probably not very important)

PR #1764, opened 2019-03-31
Simplify BaseTaggableObject

msgilligan:
* Make the tags member non-nullable and final by initializing it
* Remove comment about being lazily-created (it doesn’t need to be)
* Remove now unnecessary null-handling code
* Eliminate dependency on Guava (`Maps.newHashMap()`)

### DIFF
--- a/core/src/main/java/org/bitcoinj/utils/BaseTaggableObject.java
+++ b/core/src/main/java/org/bitcoinj/utils/BaseTaggableObject.java
@@ -16,7 +16,6 @@
 
 package org.bitcoinj.utils;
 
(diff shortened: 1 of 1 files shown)


schildbach:
Hmmm, does anyone actually use these tags? Anyway, I merged. Thanks!

msgilligan:
> Hmmm, does anyone actually use these tags?

I was wondering the same thing. It is used in the `wallettemplate` JavaFX app, but that could be changed. Do you want to deprecate it?

schildbach:
No, it's fine for now. It doesn't really hurt.

PR #1766, opened 2019-04-04
Simplify wallettemplate/build.gradle

msgilligan:
* Put all Gradle plugins in plugins block
* Use “11” for sourceCompatibility (not “1.11”)

### DIFF
--- a/wallettemplate/build.gradle
+++ b/wallettemplate/build.gradle
@@ -18,7 +18,7 @@ javafx {
     modules = [ 'javafx.controls', 'javafx.fxml' ]
 }
 
-sourceCompatibility = 11
+sourceCompatibility = 1.11
 compileJava.options.encoding = 'UTF-8'
 compileTestJava.options.encoding = 'UTF-8'
 
--- a/wallettemplate/build.gradle
+++ b/wallettemplate/build.gradle
@@ -1,11 +1,9 @@
 plugins {
-    id 'org.openjfx.javafxplugin' version '0.0.7' apply false
+    id 'java'
+    id 'eclipse'
+    id 'application'
+    id 'org.openjfx.javafxplugin' version '0.0.7'
 }
-
-apply plugin: 'java'
-apply plugin: 'eclipse'
-apply plugin: 'application'
-apply plugin: 'org.openjfx.javafxplugin'
 
 dependencies {
     implementation project(':bitcoinj-core')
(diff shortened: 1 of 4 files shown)


wallettemplate/build.gradle
```diff
@@ -1,12 +1,10 @@
 plugins {
-    id 'org.openjfx.javafxplugin' version '0.0.7' apply false
```
schildbach: This change looks good to me. To be consistent across all subprojects, could you change it for the other subprojects too?

wallettemplate/build.gradle
```diff
@@ -20,7 +18,7 @@ javafx {
     modules = [ 'javafx.controls', 'javafx.fxml' ]
 }
 
-sourceCompatibility = 1.11
```
schildbach: I'm not so sure about this change. I know OpenJDK 11 supports this syntax, but I believe OpenJDK 8 does not. And I'd like to keep the type of syntax used consistent with the other subprojects.

schildbach:
Ah, drats! This syntax seems to require a newer Gradle version that 3.4.1.

schildbach:
I propose to merge this after we've upgraded to Gradle 4.4.1. Which we should be able to do pretty soon – Gradle 4.4.1 is about to enter Ubuntu 18.04.

msgilligan:
Sorry, I've been testing with Gradle 4.3.10 (the final version of 4.x) but that isn't old enough. It would be nice if CI could somehow test with 3.4.1. (This is where the wrapper comes in handy, but we might be able to do it with a custom .yml file)

schildbach:
On F-Droid, we modify the local copy of build.gradle like this:

> `echo 'task wrapper(type: Wrapper) { gradleVersion = "3.4.1" }' >> ../build.gradle`

And then I think gradlew is executed by the build daemon. Maybe we can adapt that pattern to Travis?

msgilligan:
> Maybe we can adapt that pattern to Travis?

We can give it a try.

I created PR #1781 so that our minimum Gradle requirements are better documented and enforced.

BTW, I use the [sdkman](https://sdkman.io) for my testing so I can do things like:
```
$ sdk use gradle 3.3
$ gradle clean build
$ sdk use java 8.0.201-oracle
$ gradle clean build
$ sdk use gradle 3.4
$ gradle clean build
$ sdk use gradle 4.10.3
$ gradle clean build
$ sdk use java 12.0.0-open
$ gradle clean build
```

schildbach:
I plan to merge this as soon as https://github.com/bitcoinj/bitcoinj/pull/1789 is agreed and merged.

schildbach:
Merged.

PR #1767, opened 2019-04-04
Upgrade QRGen to v2.5.0 in wallettemplate

msgilligan:
* QRGen 2.5.0 (eliminates split packages - also requires use of jitpack.io to get latest build)

### DIFF
--- a/wallettemplate/src/main/java/wallettemplate/controls/ClickableBitcoinAddress.java
+++ b/wallettemplate/src/main/java/wallettemplate/controls/ClickableBitcoinAddress.java
@@ -38,8 +38,8 @@ import javafx.scene.input.MouseButton;
 import javafx.scene.input.MouseEvent;
 import javafx.scene.layout.AnchorPane;
 import javafx.scene.layout.Pane;
-import net.glxn.qrgen.QRCode;
-import net.glxn.qrgen.image.ImageType;
+import net.glxn.qrgen.javase.QRCode;
+import net.glxn.qrgen.core.image.ImageType;
 import wallettemplate.Main;
 import wallettemplate.utils.GuiUtils;
 
--- a/wallettemplate/src/main/java/wallettemplate/controls/ClickableBitcoinAddress.java
+++ b/wallettemplate/src/main/java/wallettemplate/controls/ClickableBitcoinAddress.java
@@ -16,6 +16,7 @@
 
 package wallettemplate.controls;
 
+import com.google.zxing.BarcodeFormat;
 import com.google.zxing.Writer;
 import com.google.zxing.WriterException;
 import com.google.zxing.client.j2se.MatrixToImageWriter;
@@ -50,7 +51,6 @@ import java.awt.*;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.OutputStream;
 import java.net.URI;
 
 import static javafx.beans.binding.Bindings.convert;
@@ -172,7 +172,7 @@ public class ClickableBitcoinAddress extends AnchorPane {
         Writer qrWriter = new QRCodeWriter();
         BitMatrix matrix;
         try {
-            matrix = qrWriter.encode(uri, com.google.zxing.BarcodeFormat.QR_CODE, 320, 240);
+            matrix = qrWriter.encode(uri, BarcodeFormat.QR_CODE, 320, 240);
         } catch (WriterException e) {
             throw new RuntimeException(e);
         }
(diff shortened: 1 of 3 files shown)


schildbach:
I'm skeptical that adding another repo is worth the new QRGen version.

If you separate the commits I'd merge just the EasyBind update for now.

msgilligan:
It's the upgraded QRGen version that I really want. It fixes a problem with split packages that breaks Java Modules -- which is something I'm working on for a future pull request. Notice that I added the `jitpack.io` repo only to the build for `wallettemplate` which will only run with Java 11 is installed.

msgilligan:
I'll go ahead and create a separate PR for the **EasyBind** update and make this one for **QRGen** only.

msgilligan:
OK, I force-pushed a change to this PR so it has the **QRGen** v2.5.0 upgrade only. I also submitted PR #1770 to contain just the **EasyBind** update.

schildbach:
What version of QRGen fixes the issue with split packages? I noticed there _are_ more recent QRGen releases on JCenter, just not the most recent ones.

Also I wonder if we actually need QRGen. Creating QR-codes with ZXing is a one-liner, and it seems as QRGen is using ZXing behind the scenes anyway.

schildbach:
Just for clarification: I'm not against Jitpack, but I see the disadvantages of using multiple repos at the same time. I would suggest if we switch, we switch with the entire project. And since this has implications on the consumers of bitcoinj too I think such a change should be openly discussed on the mailing list first.

msgilligan:
Yeah, I'm not a fan of using jitpack for what should be released JARs. I definitely don't want to switch the entire project.

I like your idea of using ZXing directly and moving the QRGen dependency. I'll make that change in the next few days and force-push.

msgilligan:
> I like your idea of using ZXing directly and moving the QRGen dependency. I'll make that change in the next few days

Done!

> and force-push

Not done (because a local pop/merge interfered.)

If the cumulative diff looks acceptable to you, I'll do a squash and force-push.

wallettemplate/src/main/java/wallettemplate/controls/ClickableBitcoinAddress.java
```diff
@@ -153,4 +150,38 @@ protected void showQRCode(MouseEvent event) {
...
+    /**
+     * Create a PNG-format image byte string from a Bitcoin URI
+     * @param uri Bitcoin URI
+     * @return PNG-format bytes
+     */
+    private static byte[] qrCodePNGBytes(String uri) {
+        Writer qrWriter = new QRCodeWriter();
+        BitMatrix matrix;
+        try {
+            matrix = qrWriter.encode(uri, com.google.zxing.BarcodeFormat.QR_CODE, 320, 240);
```
schildbach: Maybe import BarcodeFormat?

msgilligan: Good idea, will do.

msgilligan:
> I wonder do I understand right that the QR code will be PNG-encoded and decoded immediately after that? 

Yeah, see Mike's comment. I decided not to try to fix that in this commit.

> I don't know about the capabilities of JavaSE/JFX, but can't the QR code be rendered directly into a black/white bitmap that is used for displaying?

There must be a way. And we may want to switch libraries anyway because ZXing says it is in "maintenance mode" and has a problem with [Split Packages](https://github.com/zxing/zxing/issues/1153) that is making a modular implementation of `wallettemplate` more difficult.

Maybe we should look at [nayuki/QR-Code-generator](https://github.com/nayuki/QR-Code-generator) and the code in (rejected) [Issue 33: Pure JavaFX WriteImage code](https://github.com/nayuki/QR-Code-generator/issues/33).

nayuki:
I would be happy to see my QR Code generator library be used in bitcoinj. Is there anything I can help you with?

schildbach:
Merged.

Regarding switching out ZXing for QR Code generator: Encoding is one part, decoding would be the other. Apparently wallet-template doesn't have a QR code scanner, but a fully fledged wallet would need to have that. So I'm not so much in favour of switching, especially because ZXing works very well.

Maintenance mode is also not a problem for me, as long as it's still maintained.

The problem with split packages could be worked around by only using the ZXing core package. It has a generic API that returns a "BitMatrix" that you can easily copy into your own bitmap. See https://github.com/bitcoin-wallet/bitcoin-wallet/blob/master/wallet/src/de/schildbach/wallet/util/Qr.java#L54 for example.

schildbach:
I just saw this PR uses BitMatrix already. But why do we need the ZXing-JavaSE package then?

msgilligan:
> But why do we need the ZXing-JavaSE package then?

You're right. The only thing we're using from the ZXing-JavaSE package is the PNG conversion. The solution is to remove ZXing-JavaSE and convert directly to a JavaFX `Image`. This will kill two birds with one stone.

I'll submit a new PR with this change.

msgilligan:
Thanks @nayuki your QR Code generator library looks awesome, but we're sticking with ZXing for the time being.

PR #1771, opened 2019-04-04
core jar: Automatic module name org.bitcoinj.core

msgilligan:
See: http://branchandbound.net/blog/java/2017/12/automatic-module-name/

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -45,6 +45,15 @@ test {
     }
 }
 
+ext.moduleName = 'org.bitcoinj.core'
+
+jar {
+    inputs.property("moduleName", moduleName)
+    manifest {
+        attributes  'Automatic-Module-Name': moduleName
+    }
+}
+
 task javadocJar(type: Jar, dependsOn: javadoc) {
     classifier = 'javadoc'
     from javadoc.destinationDir


core/build.gradle
```diff
@@ -45,6 +45,15 @@ test {
     }
 }
 
+ext.moduleName = 'org.bitcoinj.core'
+
+jar {
+    inputs.property("moduleName", moduleName)
```
schildbach: Why not use `instruction "Automatic-Module-Name", "org.bitcoinj.core"` like suggested in the article you linked? What's the difference?

msgilligan: `inputs.property("moduleName", moduleName)` tells Gradle to ru-run the `jar` task if `ext.moduleName` changed since the last run.

I'm pretty sure `instruction` in the article is an error (looks like they copied code from a build file using the 'osgi' plugin) -- `attributes` is what is used for plain-old-everyday jars.

schildbach: I see. Also I just saw it's recommended in the Gradle docs: https://guides.gradle.org/building-java-9-modules/#optional_add_code_automatic_module_name_code_manifest_entries_for_all_other_projects

schildbach:
Thanks for investigating into Java9 modules! It's much appreciated. I just merged.

core/build.gradle
msgilligan: FYI @schildbach I got the article corrected: https://disqus.com/home/discussion/branchandbound/automatic_module_name_calling_all_java_library_maintainers_48/#comment-4414097676

PR #1772, opened 2019-04-05
missing an @Nullable annotation on ECKey.toString

matthewleon:
Not only is `NetworkParameters params` `@Nullable` in theory, `toString()` passes it as `null`.


schildbach:
Merged.

PR #1775, opened 2019-04-05
Fix spelling errors

show-rosarugosa:
The TXOs should be UTXOs.


schildbach:
TXO stands for "transaction output", so it's not necessarily a spelling error.

show-rosarugosa:
But in the code, I see a code that says UTXO is getting here, so I think it's more appropriate to change the comment to UTXO.
`throw new ProtocolException("Unknown tx version in getutxo output: " + version);`

schildbach:
I decided the change the comment to the wording used in the JavaDoc of `getHitMap()` via 40eff8cba23a3a1b57039bd0bb555be9d267942b.

Thank you for pointing out the unusually worded comment.

PR #1776, opened 2019-04-06
Remove dependency on zxing-javase, unnecessary step of PNG conversion

msgilligan:
Rather than using zxing-javase to create a PNG, convert
the BitMatrix directly to a JavaFX Image

Also re-organize import statements

This is a follow-up (and hopefully improvement) to PR #1767.

### DIFF
--- a/wallettemplate/build.gradle
+++ b/wallettemplate/build.gradle
@@ -12,7 +12,7 @@ dependencies {
     implementation 'com.google.guava:guava:27.1-android'
     implementation 'org.fxmisc.easybind:easybind:1.0.3'
     implementation 'de.jensd:fontawesomefx:8.0.0'
-    implementation 'com.google.zxing:javase:3.3.3'
+    implementation 'com.google.zxing:core:3.3.3'
     implementation 'org.slf4j:slf4j-jdk14:1.7.25'
 }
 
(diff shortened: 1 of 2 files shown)


msgilligan:
Good news: with these dependencies I've been able to build and run a *modular* `wallettemplate`! It even works with `jlink` and `jpackage`!

wallettemplate/build.gradle
```diff
@@ -12,7 +12,7 @@ dependencies {
     implementation 'com.google.guava:guava:27.1-android'
     implementation 'org.fxmisc.easybind:easybind:1.0.3'
     implementation 'de.jensd:fontawesomefx:8.0.0'
-    implementation 'com.google.zxing:javase:3.3.3'
+    implementation 'com.google.zxing:core:3.3.3'
```
msgilligan: I believe this change removes a total of 4 dependencies from the project.

msgilligan: Can I resolve my own conversation?

schildbach:
Merged. Great news about the modularization!

msgilligan:
Thanks for all your help on this one!

> Great news about the modularization

Yeah, that has been one of my goals with these recent PRs. I hope to make a PR in the next few days for the modular build. I just need to do a little bit of cleanup. One issue is that to get `jpackage` to work from Gradle you need to "manually" install  a prerelease version from JDK 13 EA (and not the main build of JDK 13 EA, either)  So I'm thinking we just check if a(some) variable(s) are set and if so run `jpackage`. `jlink` works with JDK 11 only.

PR #1777, opened 2019-04-06
Remove wallet_template Gradle task (in favor of run task)

msgilligan:
The `run` task is standard (`application` plugin) and does what we need.
I don’t think anything else depends on the `wallet_template` task.


schildbach:
Merged.

PR #1779, opened 2019-04-08
Make wallettemplate use Java Module System, native jlink and jpackage builds.

msgilligan:
* Upgrade dependencies
* Use [BadAss JLink Plugin](https://badass-jlink-plugin.beryx.org/releases/latest/)
* `jlink` and `jpackage` tasks are now both working

`jpackage` requires an early-access build of the [jpackage Packaging Tool](https://jdk.java.net/jpackage/) from OpenJDK 13 to be installed (and typically located via the `BADASS_JLINK_JPACKAGE_HOME` environment variable.)

You can run the `jlink` generated image with: `./wallettemplate/build/image/bin/WalletTemplate` or `./wallettemplate/build/image/bin/WalletTemplate.bat` for Windows (Windows not tested)

The **macOS** app is at `wallettemplate/build/jpackage/WalletTemplate.app` and is double-clickable. Windows and Linux apps should be in the same directory when built under those OSes.

Tested on **macOS**.  Output sizes are:
```
$ du -s -m wallettemplate/build/image wallettemplate/build/jpackage/WalletTemplate*
53	wallettemplate/build/image
57	wallettemplate/build/jpackage/WalletTemplate-1.0.dmg
111	wallettemplate/build/jpackage/WalletTemplate-1.0.pkg
71	wallettemplate/build/jpackage/WalletTemplate.app
```

If I recall correctly, under Java 8 these files would be around 200 MB each.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,1 @@
+systemProp.org.gradle.java.compile-classpath-packaging=true
--- a/wallettemplate/build.gradle
+++ b/wallettemplate/build.gradle
@@ -2,13 +2,15 @@ plugins {
     id 'java'
     id 'eclipse'
     id 'application'
-    id 'org.openjfx.javafxplugin' version '0.0.8'
+    id 'org.openjfx.javafxplugin'           version '0.0.8'
+    id 'org.beryx.jlink'                    version '2.17.7'
 }
 
 dependencies {
     implementation project(':bitcoinj-core')
     implementation 'de.jensd:fontawesomefx:8.0.0'
     implementation 'com.google.zxing:core:3.4.0'
+    implementation 'org.slf4j:slf4j-api:1.7.30'
     implementation 'org.slf4j:slf4j-jdk14:1.7.30'
 }
 
@@ -17,9 +19,30 @@ javafx {
     modules = [ 'javafx.controls', 'javafx.fxml' ]
 }
 
-sourceCompatibility = 1.11
+sourceCompatibility = 11
 compileJava.options.encoding = 'UTF-8'
 compileTestJava.options.encoding = 'UTF-8'
 javadoc.options.encoding = 'UTF-8'
 
-mainClassName = 'wallettemplate.Main'
+mainClassName = "$moduleName/wallettemplate.Main"
+
+jlink {
+    addExtraDependencies("javafx")
+    options = ['--strip-debug', '--compress', '2', '--no-header-files', '--no-man-pages']
+    launcher {
+        name = 'WalletTemplate'
+        jvmArgs = []
+    }
+    mergedModule {
+        requires 'java.logging'
+        requires 'java.desktop'
+
+        requires 'javafx.graphics'
+        requires 'javafx.controls'
+    }
+    jpackage {
+        // See https://badass-jlink-plugin.beryx.org/releases/latest/#_jpackage for
+        // where the plugin's jpackage task finds the path to the jpackage tool by default
+        skipInstaller = false
+    }
+}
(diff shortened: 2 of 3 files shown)


msgilligan:
p.s. I"m not sure why `WalletTemplate.app` is ~20 MB bigger than `wallettemplate/build/image`

schildbach:
Interesting view of the future (-:  Thanks for that!

At the current time, I'd rather not hike the OpenJDK requirement even further and depend on beta versions (slf4j). I guess the jlink part of your PR will have to wait. Some of the other parts are contained in some of your other PRs and I will merge them in the next weeks.

msgilligan:
>  I'd rather not hike the OpenJDK requirement even further

This PR doesn't change the OpenJDK requirement -- it still builds with JDK 11 (or later, I've tested it with 12) However, the `jpackage` command-line tool was removed from the JDK (in version 11, I believe) and is being reintroduced in JDK 13. So (for now) if you want to run the optional `jpackage` task, you will need to have the `jpackage` tool installed and set one of the configuration variables used by the BadAss Jlink Plugin to tell it where to find it.

So, `jpackage` will have to remain an optional part of the build. However, you can still build and run the modular app and you can still build and run the custom `jlink` image with JDK 11.

msgilligan:
> I'd rather not … depend on beta versions (slf4j).

I just pushed 5a5c517a7c3216ab57ea0524ad1c47504080c352 which reverts to slf4j 1.7.25. (Note that slf4j does not have an "Automatic Module Name" entry in its manifest so the automatic module name reverts to the filename with the `-` replaced with a `.`)

msgilligan:
@schildbach Have you had a chance to look at my two comments above and 5a5c517a7c3216ab57ea0524ad1c47504080c352?

schildbach:
Your change looks good to me, thanks. As soon as https://github.com/bitcoinj/bitcoinj/pull/1789 is agreed and merged, I'll try to work into jlink and jpackage so I can start using/testing it myself.

msgilligan:
> I'll try to work into jlink and jpackage so I can start using/testing it myself.

Yes, that is a good idea because I've only done limited testing and only on macOS. I assume anyone with an app based on `wallettemplate` does independent testing of their app.

I personally feel that JavaFX is becoming a much more viable desktop app solution in recent versions because of `jlink` and `jpackage`.  I don't think apps that require separate installation of the JVM will ever become widely used. So that's why I'm pushing so hard to move bitcoinj into the modular world. If developers are building/shipping native packages then they really should always be on the latest version of the JDK.  (So I'm actually using JDK 12 to build my JavaFX apps at the moment.)

msgilligan:
@schildbach I squashed and force-pushed.

msgilligan:
I rebased and updated this PR and for some unexplained reason the `wallettemplate` build is failing with a "module not found" warning for `org.bitcoinj.core` even though we are building it as an automatic module in `bitcoinj-core`. I think the problem might be a bug in Gradle or one of the plugins. I downgraded this PR to "Draft" until this issue can be resolved.

msgilligan:
@schildbach I figured out what the issue is and have reported it to the Java Modularity Gradle Modules Plugin project: https://github.com/java9-modularity/gradle-modules-plugin/issues/143

msgilligan:
@schildbach I found a workaround and tried to add it via https://github.com/bitcoinj/bitcoinj/pull/1779/commits/7522a54d9f9d4f04ef7f94c8e7ce91a59038be7a but this doesn't work with Gradle 4.10.3. It does work with Gradle 5.6+, however. I don't think this is the best long-term fix for this issue, but it is an illustration of the costs of trying to support such an old (and such a wide range of) Gradle releases. I also noticed right away how slow the old versions are to compile.

schildbach:
It might be worth rebasing this to current master.

msgilligan:
> Interesting view of the future (-: 

Looks like the future is almost here!

msgilligan:
Closed in favor of PR #3566. Because 3566 only added a `jlink` build, Issue #3572 was created to remind us to implement a `jpackage` wrapper.

PR #1781, opened 2019-04-09
Better document and enforce min Gradle and JDK versions for build

msgilligan:
* Create variables for min Gradle version for basic build, Java FX build
* Throw exception if Gradle less than minimum version
* Change existing println statements to print to stderr

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,6 +1,16 @@
 import org.gradle.util.GradleVersion
+import org.gradle.api.GradleScriptException
+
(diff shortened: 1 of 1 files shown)


msgilligan:
The TravisCI build failure looks like one of those random unit test failures...

schildbach:
Fixed the test failure via 8a19bc14e6537fd7cfff85f8f0aef82935c51c03.

schildbach:
Merged!

PR #1784, opened 2019-04-14
fix handling of non-canonical DER signatures

matthewleon:
In Bitcoin Core, when the relevant flags are set, trying to decode a non-canonical signature results in immediate script failure.

In bitcoinj, the behavior is incorrect: the interpreter simply adds `0` to the stack.

The additional tests are ported directly from Bitcoin Core's `script_tests.json`. As you can see, they test that failure is immediate by running a `NOT` instruction after `CHECKSIG`.


schildbach:
Thanks, great catch! I compared the tests with bitcoin core 0.17.1 and 0.18.rc and found that there are subtle differences. E.g.

```
["0x4a 0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000", "0 CHECKSIG NOT", "DERSIG", "SIG_DER", "Overly long signature is incorrectly encoded for DERSIG"]
```

versus 

```
["0x4a 0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000", "0 CHECKSIG NOT", "", "OK", "Overly long signature is correctly encoded"],
```

Could we perhaps match it?

matthewleon:
@schildbach good idea. Will give it a shot shortly.

matthewleon:
@schildbach From what I can see, _both_ those tests are present on core's current master. They are different tests: the first has the `DERSIG` flag enabled, the second does not.

Both tests are present on my branch:

The first (already was there): https://github.com/bitcoinj/bitcoinj/blob/f26abf03fc915594af54e3ebc506ce4f0e01d58a/core/src/test/resources/org/bitcoinj/script/script_tests.json#L695

The second (added by my PR):
https://github.com/bitcoinj/bitcoinj/blob/f26abf03fc915594af54e3ebc506ce4f0e01d58a/core/src/test/resources/org/bitcoinj/script/script_tests.json#L1230

schildbach:
You're right, I missed the second set of tests.

I just merged.

PR #1789, opened 2019-04-17
Require Gradle 4.4 (or above) for building.

schildbach:
Debian Buster and Ubuntu Bionic, Cosmic and Disco now come with Gradle 4.4.1. Thus, I propose we raise the build requirement to Gradle 4.4.

This will allow us to use Gradle syntax and features that have become available between versions 3.4 and 4.4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ The bitcoinj library is a Java implementation of the Bitcoin protocol, which all
 
 ### Technologies
 
-* Java 7+ and Gradle 3.4+ for the `core` module
-* Java 8+ and Gradle 3.4+ for `tools` and `examples`
+* Java 7+ and Gradle 4.4+ for the `core` module
+* Java 8+ and Gradle 4.4+ for `tools` and `examples`
 * Java 11+ and Gradle 4.10+ for the JavaFX-based `wallettemplate`
 * [Gradle](https://gradle.org/) - for building the project
 * [Google Protocol Buffers](https://github.com/google/protobuf) - for use with serialization and hardware communications
--- a/settings.gradle
+++ b/settings.gradle
@@ -2,7 +2,7 @@ import org.gradle.util.GradleVersion
 import org.gradle.api.GradleScriptException
 
 // Minimum Gradle version for main build
-def minGradleVersion = GradleVersion.version("3.4")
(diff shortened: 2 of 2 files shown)


msgilligan:
Looks good to me @schildbach ! I'll run some tests on the Mac to make sure things work as expected and share the results here.

schildbach:
Great, thanks! I'll wait with merging for your findings.

msgilligan:
| JDK | Gradle |Result|
| --- | --- | --- |
| 7.0.222-zulu | 4.4 |✅|
| 8.0.201-oracle | 4.3.1 |✅|
| 8.0.201-oracle | 4.4 |✅|
| 8.0.201-oracle | 5.3 |✅|
| 8.0.201-oracle | 5.4 |✅|
| 11.0.2-open | 5.4 |✅|
| 12.0.1-open | 5.4 |✅ Note 1|

Note 1: The following warning was displayed

```
> Task :bitcoinj-core:compileJava
warning: [options] source value 7 is obsolete and will be removed in a future release
warning: [options] target value 7 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
```

schildbach:
Cool! Just to be clear, the test represented by the second line should yield an error "bitcoinj build requires Gradle 4.4 or later". Did this happen?

msgilligan:
Notice that the JDK 7 build failed. That was the final test I ran and I hit enter on the comment a little too soon, so had to edit the comment to add the error.

> Just to be clear, the test represented by the second line should yield an error "bitcoinj build requires Gradle 4.4 or later". Did this happen?

Yes. That is correct ✅ operation for that combination.

schildbach:
We could, but anyway it is not related to updating the Gradle version. I'll merge.

msgilligan:
@schildbach Thinking about it a little more, we need to distinguish between the build JDK and the target JDK. We should require a minimum of JDK 8 to build and `core` will just target JDK 7. There is no need to support building with JDK 7.

schildbach:
Yes, I was thinking around these lines too. I don't even have the possibility to run a build wth JDK 7 any more.

Then again, as soon as I've got the desugaring working in my Android builds I wouldn't mind targetting JDK 8 (but still require to only use Java 7 API).

PR #1790, opened 2019-04-17
Use ScriptsChange instead of ScriptChange.

oscarguindzberg:
There was a mix of ScriptsChange (with s) and ScriptChange (without s).
Use ScriptsChange everywhere.


schildbach:
Merged!

PR #1791, opened 2019-04-17
.travis.yml: Specify exact Gradle version to use.

schildbach:
- Specify exact Gradle to use by running the Gradle wrapper in before_script.
- Pick several combinations of JDK version and Gradle version by building a matrix. See this example build: https://travis-ci.org/bitcoinj/bitcoinj/builds/521380652

@msgilligan What do you think of this? Would you pick other combinations?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,10 @@ jdk:
   - openjdk8
   - openjdk11
 install: true
+before_script:
+  - gradle wrapper --gradle-version=4.10.3
 script:
-  - gradle clean build
+  - ./gradlew clean build
 
 notifications:
   irc:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,17 @@
 sudo: false
 dist: xenial
 language: java
-jdk:
-  - openjdk8
-  - openjdk11
+matrix:
+  include:
+  - jdk: openjdk8
+    env: GRADLE_VERSION=4.4.1
+  - jdk: openjdk8
+    env: GRADLE_VERSION=4.10.3
+  - jdk: openjdk11
+    env: GRADLE_VERSION=4.10.3
 install: true
 before_script:
-  - gradle wrapper --gradle-version=4.10.3
+  - gradle wrapper --gradle-version=$GRADLE_VERSION
 script:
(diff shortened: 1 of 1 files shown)


msgilligan:
> What do you think of this? 

I like it. (Assuming it works, I've never used this approach before because I always use a fixed version via a wrapper.)

> Would you pick other combinations?

I definitely think we should keep the latest (or at least a recent 5.x) version of Gradle in the mix.

Less important, but it might make sense to add JDK 12 in there to get early warnings of any breakages with the latest builds.

There is also a mechanism in JDK 9+ that allows builds for earlier versions. One problem with using JDK 8 to build for JDK 7 is that it doesn't detect usage of library methods that don't exist in JDK 7. I actually got burned by this in one of my early (2015?) PRs to bitcoinj. So we might want to look into the feature in JDK 9+ that lets us build for earlier JDKs **and** automatically uses the correct library versions. If you're interested I could make sure it would really work for us. (I also have used builds that have both JDK 7 and JDK 8 so that I can detect accidental usage of JDK 8+ methods in a JDK 7 build)

schildbach:
In this PR, I tried to cover what we already have/use, which is why I didn't add other Gradle or OpenJDK versions except from maintenance updates. I'm open for extending on that, so no objection against adding a build for Gradle 5.x. Currently on Travis we build with the default, which is 4.10.2.

At the same time I want to stay responsible with the VM time Travis gives us for free. Just this PR increases build time to about 9 minutes by just adding 1 combination.

Hmm, since the `--release` flag according to the SO you linked is just an abbreviation for `--source --target --bootclasspath`, why not use the unabbreviated version so that it works for all our current builds? Thinking it further, we probably don't need `--target` because our build artifacts are thrown away on Travis anyway.

msgilligan:
> Hmm, since the --release flag according to the SO you linked is just an abbreviation for --source --target --bootclasspath, why not use the unabbreviated version so that it works for all our current builds?

My understanding is that prior to JDK 9 you have to manually install the files for the `bootclasspath` option, where as in JDK 9+ they are included in the JDK. I was equating that with the `release` option, but I guess it should also work with the unabbreviated version, as well.

PR #1792, opened 2019-04-17
simplify TX verification error checking

matthewleon:
This simplifies some of the error checking in `Transaction.verify()` and `TransactionOutput.getValue()`. It seems that some of the calls in these methods used to throw exceptions, but no longer do so, leaving some stale code for checking errors that cannot be thrown.


matthewleon:
Tests on JDK 8 appear to be failing due to some kind of fluke. I suspect they'll pass if simply rerun.

schildbach:
Looks good, thanks! I just merged this.

PR #1793, opened 2019-04-17
kill an unnecessary instance variable in TransactionOutput

matthewleon:
This seems to have been left over from earlier code changes and now serves no purpose.


schildbach:
Merged. I love such small simplifications (-:

PR #1795, opened 2019-04-19
Access block transactions with getter method rather than directly thr…

jakobjanecek:
…ough the field.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -507,7 +507,7 @@ public abstract class AbstractBlockChain {
         if (!params.passesCheckpoint(storedPrev.getHeight() + 1, block.getHash()))
             throw new VerificationException("Block failed checkpoint lockin at " + (storedPrev.getHeight() + 1));
         if (shouldVerifyTransactions()) {
-            for (Transaction tx : requireNonNull(block.getTransactions()))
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
```diff
@@ -506,8 +507,7 @@ private void connectBlock(final Block block, StoredBlock storedPrev, boolean exp
         if (!params.passesCheckpoint(storedPrev.getHeight() + 1, block.getHash()))
             throw new VerificationException("Block failed checkpoint lockin at " + (storedPrev.getHeight() + 1));
         if (shouldVerifyTransactions()) {
-            checkNotNull(block.transactions);
-            for (Transaction tx : block.transactions)
+            for (Transaction tx : requireNonNull(block.getTransactions()))
```
schildbach: I believe this will fail on lots of Android devices, since it is a JDK 8 method. Perhaps use the Guava equivalent? 

Also if course you just as well omit it because the `for ()` construct will throw a NPE.

jakobjanecek: This method ia already available in JDK 7 - https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#requireNonNull(T)

But you are correct that in this case it's not useful.  I've removed the explicit check.

schildbach:
Thanks for the update! I just merged this.

PR #1796, opened 2019-04-20
kill deprecated methods in FilteredBlockAndPartialMerkleTreeTests

matthewleon:
The compiler complains about these. And they're not the methods being tested. I also sneaked in a tiny clean up of the use of large numbers in one of the method invocations that I touched.


schildbach:
Merged.

PR #1797, opened 2019-04-20
suppress deprecation warning in ChildKeyDerivationTest

matthewleon:
This applies a warning suppression annotation to the lines in ChildKeyDerivationTest that use deprecated serialization methods.

### DIFF
--- a/core/src/test/java/org/bitcoinj/crypto/ChildKeyDerivationTest.java
+++ b/core/src/test/java/org/bitcoinj/crypto/ChildKeyDerivationTest.java
@@ -215,7 +215,9 @@ public class ChildKeyDerivationTest {
         {
             final String pub58 = key1.serializePubB58(params);
             final String priv58 = key1.serializePrivB58(params);
+            @SuppressWarnings("deprecation")
             final byte[] pub = key1.serializePublic(params);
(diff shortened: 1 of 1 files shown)


schildbach:
I think I would like to keep the warnings as a reminder that the refactoring still needs to be done.

matthewleon:
My reading of these methods is that they are purposely testing deprecated methods. Refactoring them to use the non-deprecated methods would, in effect, be creating _new_, different tests. Does that make sense?

matthewleon:
(If you disagree, I'm happy to close this in favor of a new PR that actually refactors these to use non-deprecated methods. Please let me know.)

schildbach:
Yes, I guess they are testing the deprecated methods because the methods were the main public API before 0.15.

I think we should first become clear about how an API should look like and if we need it at all. E.g. should we make `byte[] serialize(NetworkParameters params, boolean pub, Script.ScriptType outputScriptType)` public? Or should it be split into serializePublic/serializePrivate again?

Until that is resolved, I think we can just live with the warnings.

matthewleon:
> Until that is resolved, I think we can just live with the warnings.

Makes sense. Closing this PR.

PR #1798, opened 2019-04-20
fix deprecation warnings in WalletTest

matthewleon:
Replace usage of two deprecated invocations, and add warning suppression to another.

The `fromKeys()` method here is testing a deprecated method, which is why I'm suppressing the warning rather than calling a more modern method.


matthewleon:
In light of our conversation in https://github.com/bitcoinj/bitcoinj/pull/1797 I've altered this PR to not use `SuppressWarnings`. It now simply consists of the two updated method invocations that avoid the deprecated methods.

schildbach:
Merged.

PR #1799, opened 2019-04-20
writeBinaryCheckpoints method

tnitan:
- wrapped fileOutputStream in try/finally block and moved .close() calls  to avoid possible resource leaks

### DIFF
--- a/tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
+++ b/tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
@@ -172,27 +172,26 @@ public class BuildCheckpoints {
     }
 
     private static void writeBinaryCheckpoints(TreeMap<Integer, StoredBlock> checkpoints, File file) throws Exception {
-        try (FileOutputStream fileOutputStream = new FileOutputStream(file, false)) {
-            MessageDigest digest = Sha256Hash.newDigest();
-            try (
-                    DigestOutputStream digestOutputStream = new DigestOutputStream(fileOutputStream, digest);
-                    DataOutputStream dataOutputStream = new DataOutputStream(digestOutputStream)
-            ) {
-                digestOutputStream.on(false);
-                dataOutputStream.writeBytes("CHECKPOINTS 1");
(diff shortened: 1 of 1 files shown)


schildbach:
Thanks for looking into this.

May I suggest to use the try-with-resources construct here? Then you could do without the manual closing at all. And while you're at it, you could do the same for the writeTextualCheckpoints() method.

tnitan:
Sure, I've updated the pull-request. :)

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
```diff
@@ -172,42 +172,43 @@ public void notifyNewBestBlock(StoredBlock block) throws VerificationException {
...
-        dataOutputStream.writeInt(0);  // Number of signatures to read. Do this later.
-        digestOutputStream.on(true);
-        dataOutputStream.writeInt(checkpoints.size());
-        ByteBuffer buffer = ByteBuffer.allocate(StoredBlock.COMPACT_SERIALIZED_SIZE);
-        for (StoredBlock block : checkpoints.values()) {
-            block.serializeCompact(buffer);
-            dataOutputStream.write(buffer.array());
-            buffer.position(0);
+        try (FileOutputStream fileOutputStream = new FileOutputStream(file, false)) {
+            MessageDigest digest = Sha256Hash.newDigest();
```
schildbach: I think if you move this line up above the try block, the two try blocks could be merged into one. Or am I overlooking something?

tnitan: try-with-resources blocks can only use final variables that imlpement AutoCloseable interface, and the lines in second try are using: "MessageDigest digest = Sha256Hash.newDigest();" which doesn't meet the requirements.

tnitan: You're right, I've moved them in a single try. Don't know why I didn't merge them in the first place. Thanks!

schildbach:
Perfect, thanks! I just merged this.

PR #1800, opened 2019-04-20
updated outdated documentation in LegacyAddress

matthewleon:
Some of the documentation on `LegacyAddress` is false, or suggests using a deprecated method.


matthewleon:
Closing this in favor of a more systematic approach.

PR #1801, opened 2019-04-20
breaking: Type-safe legacy addresses

matthewleon:
My apologies for the large size of this PR: I couldn't see any way to break it down further without making it less coherent. If you like the general idea here, but would like it to be broken into chunks for easier review, I'd be happy to follow any suggested strategy for that.

This refactors `LegacyAddress` to be an `abstract` class with two concrete subclasses: `LegacyP2PKHAddress` and `LegacyP2SHAddress`. This eliminates the now-redundant `p2sh` field (and some of the awkwardness it imposed), and eliminates the need for some calls to methods such as `isP2SHAddress` or `getOutputScriptType`.

There are two breaking changes here that come to mind, one obvious and one more subtle:

1. The previously-deprecated, `public` constructor on `LegacyAddress` is necessarily made `protected`. This was not used elsewhere in the codebase, and should cause minimal disruption.
2. Because we have eliminated `p2sh` from calculation of `hashCode()`, P2SH and P2PKH addresses that are derived from the exact same hash (in the exceedingly unlikely scenario that the public key hash of the P2PKH address is identical to the script hash of the P2SH address) will now have the same `hashCode()`. This doesn't violate the contract of `hashCode()`, which naturally allows for hash collisions. It seems more like a theoretical curiosity than anything, but it's worth mentioning.

Tests for `LegacyAddress` are maintained and augmented. They are also copied/ported to the concrete subclasses.

Next steps would follow along two paths:

1. There are many places in the code base that could switch from using `LegacyAddress` to using the concrete subclasses. These will naturally lead to the development of type-safe convenience methods dealing with the concrete subclasses (accepting them as parameters or returning them).
2. A similar refactor can be done to `SegwitAddress`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
@@ -29,8 +29,8 @@ import org.bitcoinj.script.ScriptPattern;
 
 /**
  * A legacy (base58) Bitcoin address looks like 1MsScoe2fTJoq4ZPdQgqyhgWeoNamYPevy and is divided into two
- * concrete types: {@link LegacyP2PKHAddress}, which is derived from the RIPE-MD160 hash of public key bytes,
- * and {@link LegacyP2SHAddress}, which is derived from the hash of a {@link Script}.
+ * concrete types: {@link P2PKHAddress}, which is derived from the RIPE-MD160 hash of public key bytes,
+ * and {@link P2SHAddress}, which is derived from the hash of a {@link Script}.
  */
 public abstract class LegacyAddress extends Address {
     /**
@@ -56,7 +56,7 @@ public abstract class LegacyAddress extends Address {
     }
 
     /**
-     * Construct a {@link LegacyP2PKHAddress} that represents the given pubkey hash. The resulting address will be a P2PKH type of
+     * Construct a {@link P2PKHAddress} that represents the given pubkey hash. The resulting address will be a P2PKH type of
      * address.
      * 
      * @param params
@@ -65,12 +65,12 @@ public abstract class LegacyAddress extends Address {
      *            20-byte pubkey hash
      * @return constructed address
      */
-    public static LegacyP2PKHAddress fromPubKeyHash(NetworkParameters params, byte[] hash160) throws AddressFormatException {
-        return LegacyP2PKHAddress.fromPubKeyHash(params, hash160);
+    public static P2PKHAddress fromPubKeyHash(NetworkParameters params, byte[] hash160) throws AddressFormatException {
+        return P2PKHAddress.fromPubKeyHash(params, hash160);
     }
 
     /**
-     * Construct a {@link LegacyP2PKHAddress} that represents the public part of the given {@link ECKey}. Note that an address is
+     * Construct a {@link P2PKHAddress} that represents the public part of the given {@link ECKey}. Note that an address is
      * derived from a hash of the public key and is not the public key itself.
      * 
      * @param params
@@ -79,12 +79,12 @@ public abstract class LegacyAddress extends Address {
      *            only the public part is used
      * @return constructed address
      */
-    public static LegacyP2PKHAddress fromKey(NetworkParameters params, ECKey key) {
+    public static P2PKHAddress fromKey(NetworkParameters params, ECKey key) {
         return fromPubKeyHash(params, key.getPubKeyHash());
     }
 
     /**
-     * Construct a {@link LegacyP2SHAddress} that represents the given P2SH script hash.
+     * Construct a {@link P2SHAddress} that represents the given P2SH script hash.
      * 
      * @param params
      *            network this address is valid for
@@ -92,13 +92,13 @@ public abstract class LegacyAddress extends Address {
      *            P2SH script hash
      * @return constructed address
      */
-    public static LegacyP2SHAddress fromScriptHash(NetworkParameters params, byte[] hash160) throws AddressFormatException {
-        return LegacyP2SHAddress.fromScriptHash(params, hash160);
+    public static P2SHAddress fromScriptHash(NetworkParameters params, byte[] hash160) throws AddressFormatException {
+        return P2SHAddress.fromScriptHash(params, hash160);
     }
 
     /** @deprecated use {@link #fromScriptHash(NetworkParameters, byte[])} */
     @Deprecated
-    public static LegacyP2SHAddress fromP2SHHash(NetworkParameters params, byte[] hash160) {
+    public static P2SHAddress fromP2SHHash(NetworkParameters params, byte[] hash160) {
         return fromScriptHash(params, hash160);
     }
 
@@ -107,7 +107,7 @@ public abstract class LegacyAddress extends Address {
      *             {@link ScriptPattern#extractHashFromP2SH(Script)}
      */
     @Deprecated
-    public static LegacyP2SHAddress fromP2SHScript(NetworkParameters params, Script scriptPubKey) {
+    public static P2SHAddress fromP2SHScript(NetworkParameters params, Script scriptPubKey) {
         checkArgument(ScriptPattern.isP2SH(scriptPubKey), "Not a P2SH script");
         return fromScriptHash(params, ScriptPattern.extractHashFromP2SH(scriptPubKey));
     }
@@ -128,10 +128,10 @@ public abstract class LegacyAddress extends Address {
     public static LegacyAddress fromBase58(@Nullable NetworkParameters params, String base58)
             throws AddressFormatException, AddressFormatException.WrongNetwork {
         try {
-            return LegacyP2PKHAddress.fromBase58(params, base58);
+            return P2PKHAddress.fromBase58(params, base58);
         } catch (AddressFormatException.WrongAddressType ignore) {
             try {
-                return LegacyP2SHAddress.fromBase58(params, base58);
+                return P2SHAddress.fromBase58(params, base58);
             } catch (AddressFormatException.WrongAddressType e) {
                 throw new IllegalStateException("Address type is somehow neither P2SH or P2PKH", e);
             }
--- a/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
@@ -20,36 +20,23 @@ package org.bitcoinj.core;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
-import java.util.Arrays;
-import java.util.Objects;
-
 import javax.annotation.Nullable;
 
-import org.bitcoinj.params.Networks;
 import org.bitcoinj.script.Script;
 import org.bitcoinj.script.Script.ScriptType;
 import org.bitcoinj.script.ScriptPattern;
 
 
 /**
- * <p>A Bitcoin address looks like 1MsScoe2fTJoq4ZPdQgqyhgWeoNamYPevy and is derived from an elliptic curve public key
- * plus a set of network parameters. Not to be confused with a {@link PeerAddress} or {@link AddressMessage}
- * which are about network (TCP) addresses.</p>
- *
- * <p>A standard address is built by taking the RIPE-MD160 hash of the public key bytes, with a version prefix and a
- * checksum suffix, then encoding it textually as base58. The version prefix is used to both denote the network for
- * which the address is valid (see {@link NetworkParameters}, and also to indicate how the bytes inside the address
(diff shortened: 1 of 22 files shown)


Polve:
As a user of this library, I would love to see this refactor.

schildbach:
Interesting idea; thanks a lot for this contribution! I think I considered it when I did the SegwitAddress refactoring, but I couldn't clearly answer the question if it's useful for developers so I postponed. Maybe @Polve can you elaborate on why you would love to see this?

The split between `SegwitAddress` and `LegacyAddress` was introduced because devs might want to base their application exclusively on modern (Segwit) addresses. Also, because of base58 vs bech32 it seemed to make sense from the implementation perspective. For the further split(s) you're proposing I'd actually expect developers to almost never use these classes directly. The splits can still make sense! I imagine it's a bit like the relatiionship between `InetAddress` vs `Inet4Address` and `Inet6Address`. You'd always use `InetAddress` in your code, but the two subclasses still make sense.

I agree that if we do this refactoring, we should separate P2WPKH and P2WSH is a similar way (likely in a separate PR).

I wonder assuming we implement the class splits as suggested, if we hypothetically add P2SH-P2WPKH to bitcoinj would that mean we add a subclass of `LegacyP2SHAddress`? Similar question: If something like Segwit v1 arrives, say with Schnorr crypto, would we somehow subclass the SegwitAddress hierarchy even further? I'm asking these questions because I'd like to understand where we're heading to.

I'm fine with the two listed breaking changes. The constructor is scheduled for removal/being made unaccessible anyway. And hash collisions can theoretically happen all the time – also between addresses of the same type of course.

If you add new source files, can you add the license header to each of them?

If you're worried about PR size, I suggest reviewing your changes and for each change ask the question if it would make sense to put it into a different PR. For example, the removals of `throws Exception` in LegacyAddressTest look like they could be done right away, without much discussion. Also, there are some formatting-only changes that add to the length of this PR.

Polve:
I extend this library to cope with other addresses format (for example Bitcoin Cash bech32 ones) and I already do something similar, I think it would become more clean with the proposed change.

matthewleon:
Thanks for the preliminary review. Will work on getting this into better/cleaner shape. I'd like to answer all your questions bit by bit. Will start with one that seems a bit shorter to answer, before delving into others:

> I wonder assuming we implement the class splits as suggested, if we hypothetically add P2SH-P2WPKH to bitcoinj would that mean we add a subclass of `LegacyP2SHAddress`? Similar question: If something like Segwit v1 arrives, say with Schnorr crypto, would we somehow subclass the SegwitAddress hierarchy even further? I'm asking these questions because I'd like to understand where we're heading to.

The issue here is that an address can't be determined to be P2SH-P2WPKH simply by looking at its serialized form. There is potentially the same issue with further `SegwitAddress` subdivisions. You'd have to ask yourself if you are comfortable with the idea that an address in the subclass `P2SHP2WPKHLegacyAddress` (yikes!) is _definitely_ P2SHP2WPKH, while an address outside of that subclass _might_ still be, but you can't tell just by looking at it. That's a much tougher philosophical question, and I don't have a ready, simple answer to it. I would lean toward saying "no" though.

With the simple P2SH vs. P2PKH division, you don't have that issue because of the different header byte. We can _always_ know whether an address is one or the other. The taxonomy is unambiguous.

schildbach:
Related: https://github.com/bitcoinj/bitcoinj/pull/1810

matthewleon:
> For the further split(s) you're proposing I'd actually expect developers to almost never use these classes directly. The splits can still make sense! I imagine it's a bit like the relationship between `InetAddress` vs `Inet4Address` and `Inet6Address`. You'd always use `InetAddress` in your code, but the two subclasses still make sense.

I think the analogy here makes a lot of sense. That said, there will be some cases where you really do care that something is an `Inet4Address`, and there are places that you do care that something is a `P2PKHAddress`.

Example: A wallet that does not support multisig will probably only ever _create_ `P2PKHAddress`es. Being able to guarantee that in the wallet's method signatures provides, at a minimum, a small ergonomic benefit to the developer, and possibly the reassurance that it will never be necessary to `assert` that the address has `getOutputScriptType == ScriptType.P2SH`.

matthewleon:
> If you add new source files, can you add the license header to each of them?

Happy to do so. Can you suggest what to do with the copyright lines? There's very little _new_ code here, it's mostly just moved around from other files.

schildbach:
> Happy to do so. Can you suggest what to do with the copyright lines?

You could use the phrase "Copyright by the original author or authors.". Then it's up to the potential license takers to dig up the git history.

I think this PR could use a hearty rebase to master.

matthewleon:
> I think this PR could use a hearty rebase to master.

done

core/src/main/java/org/bitcoinj/core/LegacyAddress.java
```diff
@@ -20,59 +20,43 @@
...
      * 
      * @param params
      *            network this address is valid for
-     * @param p2sh
-     *            true if hash160 is hash of a script, false if it is hash of a pubkey
      * @param hash160
      *            20-byte hash of pubkey or script
      */
-    private LegacyAddress(NetworkParameters params, boolean p2sh, byte[] hash160) throws AddressFormatException {
+     protected LegacyAddress(NetworkParameters params, byte[] hash160) throws AddressFormatException {
```
schildbach: There is a stray leading space here :grin:

core/src/test/java/org/bitcoinj/core/LegacyAddressTest.java
```diff
@@ -125,7 +126,7 @@ public void getNetwork() {
     public void getAltNetwork() {
         // An alternative network
         class AltNetwork extends MainNetParams {
-            AltNetwork() {
+            private AltNetwork() {
```
schildbach: This is an unrelated change I guess.

core/src/main/java/org/bitcoinj/core/LegacyP2PKHAddress.java
```diff
@@ -0,0 +1,140 @@
...
+            throws AddressFormatException, AddressFormatException.WrongAddressType, AddressFormatException.WrongNetwork {
+        byte[] versionAndDataBytes = Base58.decodeChecked(base58);
+        int version = versionAndDataBytes[0] & 0xFF;
+        byte[] bytes = Arrays.copyOfRange(versionAndDataBytes, 1, versionAndDataBytes.length);
+        if (params == null) {
+            for (NetworkParameters p : Networks.get()) {
+                if (version == p.getAddressHeader())
+                    return fromPubKeyHash(p, bytes);
+                if (version == p.getP2SHHeader())
+                    throw new AddressFormatException.WrongAddressType("P2PKH");
```
schildbach: Shouldn't this say "P2SH"? Same 7 lines below.

core/src/main/java/org/bitcoinj/core/LegacyP2SHAddress.java
```diff
@@ -0,0 +1,128 @@
...
+            throws AddressFormatException, AddressFormatException.WrongNetwork {
+        byte[] versionAndDataBytes = Base58.decodeChecked(base58);
+        int version = versionAndDataBytes[0] & 0xFF;
+        byte[] bytes = Arrays.copyOfRange(versionAndDataBytes, 1, versionAndDataBytes.length);
+        if (params == null) {
+            for (NetworkParameters p : Networks.get()) {
+                if (version == p.getP2SHHeader())
+                    return fromScriptHash(p, bytes);
+                if (version == p.getAddressHeader())
+                    throw new AddressFormatException.WrongAddressType("P2SH");
```
schildbach: Shouldn't this say "P2PKH"? Same 7 lines below.

schildbach:
@matthewleon Are you going to do the same refactoring for SegwitAddress? (in a separate PR)

schildbach:
Btw. thanks for your explanation about P2SH-P2WPKH – I'm on the same side (-:

matthewleon:
> Are you going to do the same refactoring for SegwitAddress? (in a separate PR)

yes

core/src/main/java/org/bitcoinj/core/LegacyP2PKHAddress.java
matthewleon: I've changed the error message to try to be clearer.

core/src/test/java/org/bitcoinj/core/LegacyAddressTest.java
matthewleon: removed

core/src/main/java/org/bitcoinj/core/LegacyAddress.java
matthewleon: thanks!

matthewleon:
@schildbach Thanks for the review. This should be ready for another look.

core/src/main/java/org/bitcoinj/core/P2PKHAddress.java
```diff
@@ -0,0 +1,144 @@
...
+import org.bitcoinj.script.Script;
+
+import javax.annotation.Nullable;
+import java.util.Arrays;
+
+/**
+ * A legacy P2PKH (Pay 2 Public Key Hash) address is built by taking the RIPE-MD160 hash of the public key bytes,
+ * with a version prefix and a checksum suffix, then encoding it textually as base58. The version prefix is used to
+ * both denote the network for which the address is valid (see {@link NetworkParameters}, and also to indicate how the
+ * bytes inside the address * should be interpreted (in this case, as P2PKH rather than P2SH).
```
schildbach: There is a stray asterisk in the JavaDoc here.

core/src/main/java/org/bitcoinj/core/P2PKHAddress.java
```diff
@@ -0,0 +1,144 @@
...
+package org.bitcoinj.core;
+
+import org.bitcoinj.params.Networks;
+import org.bitcoinj.script.Script;
+
+import javax.annotation.Nullable;
+import java.util.Arrays;
+
+/**
+ * A legacy P2PKH (Pay 2 Public Key Hash) address is built by taking the RIPE-MD160 hash of the public key bytes,
```
schildbach: Maybe use the word 'to' rather than '2' here…

core/src/main/java/org/bitcoinj/core/P2SHAddress.java
```diff
@@ -0,0 +1,132 @@
...
+package org.bitcoinj.core;
+
+import org.bitcoinj.params.Networks;
+import org.bitcoinj.script.Script;
+
+import javax.annotation.Nullable;
+import java.util.Arrays;
+
+/**
+ * A legacy P2SH (Pay 2 Script Hash) address is built by taking the RIPE-MD160 hash of the redemption {@link Script}
```
schildbach: …and here.

core/src/main/java/org/bitcoinj/core/P2SHAddress.java
```diff
@@ -0,0 +1,132 @@
...
+ * bytes, with a version prefix and a checksum suffix, then encoding it textually as base58. The version prefix is used
+ * to both denote the network for which the address is valid (see {@link NetworkParameters}, and also to indicate how
+ * the bytes inside the address * should be interpreted (in this case, as P2SH rather than P2PKH).
```
schildbach: Another stray asterisk.

schildbach:
Just to let you know, I've not forgotten about this PR. I thought it would make sense to wait for the accompanying SegwitAddress refactoring to have a look at the final picture.

msgilligan:
I'm not sure I agree with this change. I don't think I have any practical objections, just theoretical and philosophical. Here are my philosophical objections:

1. Single responsibility principle: the Address format should be a representation of an Address format not a redeem script format. I realize that P2SH addresses are a unique set of strings, but their format (base58) is the same.
2. Favor composition over inheritance: We already have a boolean field `p2sh` that indicates the difference between the two and that is probably sufficient, so why not keep things simple?

matthewleon:
Given the points raised above, I'm not sure I believe that this PR adds enough value to honestly be worth a breaking change.

schildbach:
@msgilligan has some good points. Adding to point 1, I think for privacy reasons Bitcoin will be moving to addresses that don't reveal the type of redeem script anyway.

msgilligan:
@schildbach Are we going to close this and remove it from the 0.17 milestone?

schildbach:
Hmmm, I think @msgilligan is right about what an address should be. It breaks my heart to dismiss contributions like this, especially since I encouraged working on it in the beginning. But I now think it's better to not merge this. So I'm closing this.

PR #1802, opened 2019-04-21
Transaction: Add methods for calculating transaction weight…

schildbach:
…and virtual transaction size.


schildbach:
This is a prerequesite for implementing the current "fee consensus".

PR #1803, opened 2019-04-21
Feature/bip67 compliance

matthewleon:
This adds the BIP67-compliant `LegacyAddress.multisigFromKeys` factory method. It also adds a (breaking) check to `Script.createRedeemScript` to ensure that all passed-in keys are compressed.

The actual sorting mechanism is unchanged, so this is only breaking in cases where users were passing in uncompressed keys.

The bulk of the new code is the test suite in `core/src/test/java/org/bitcoinj/script/Bip67Test.java`, which uses the vectors from BIP 67: https://github.com/bitcoin/bips/blob/master/bip-0067.mediawiki

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
@@ -19,9 +19,11 @@
 package org.bitcoinj.core;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static org.bitcoinj.script.ScriptBuilder.createP2SHOutputScript;
 
 import java.util.Arrays;
 import java.util.Objects;
+import java.util.Set;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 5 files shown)


matthewleon:
This failed on a bloom filter test _only_ on JDK 11. This seems quite unrelated to this PR, and hopefully won't happen if the tests are rerun. The test in question seems somewhat flawed, as it will produce false positives on rare occasion. Will try to address this with a different PR.

schildbach:
I just triggered the unit test again.

Interestingly, BIP67 states that there is already an implementation in bitcoinj and links to https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/script/ScriptBuilder.java#L331

matthewleon:
> Interestingly, BIP67 states that there is already an implementation in bitcoinj and links to

Yes, I noticed. I should have noted that here. The only thing missing from the implementation is the check that the keys are compressed. I've added that check in this PR. Otherwise, the method I've added here is just a convenience.

PR #1805, opened 2019-04-21
add a gradle task for running examples

matthewleon:
I couldn't see any other way of getting these running after the move away from Maven.


schildbach:
Well, you can always run them in your IDE?

matthewleon:
> Well, you can always run them in your IDE?

Fair point. I sought a way of invoking them more easily from the CLI.

PR #1806, opened 2019-04-21
Remove dependency on EasyBind by adding a few lines of code

msgilligan:
* Remove dependency on EasyBind
* Replace EasyBind with two static methods

Hopefully the code is more readable than before.

### DIFF
--- a/wallettemplate/src/main/java/wallettemplate/MainController.java
+++ b/wallettemplate/src/main/java/wallettemplate/MainController.java
@@ -16,6 +16,9 @@
 
 package wallettemplate;
 
+import javafx.beans.binding.Binding;
+import javafx.beans.binding.Bindings;
+import javafx.beans.value.ObservableValue;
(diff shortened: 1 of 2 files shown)


wallettemplate/src/main/java/wallettemplate/MainController.java
```diff
@@ -48,6 +50,7 @@
 
     private BitcoinUIModel model = new BitcoinUIModel();
     private NotificationBarPane.Item syncItem;
+    private static MonetaryFormat monetaryFormat = MonetaryFormat.BTC.noCode();
```
schildbach: Perhaps make it a true constant by adding the final keyword and uppercase the name?

msgilligan: Will do.

msgilligan: Done!

msgilligan:
Merged by @schildbach

PR #1809, opened 2019-04-23
Create utility class for QR Code FX image creation

msgilligan:
Move the code to generate a JavaFX image for a QR code to a reusable
utility method.

### DIFF
--- a/wallettemplate/src/main/java/wallettemplate/utils/QRCodeImages.java
+++ b/wallettemplate/src/main/java/wallettemplate/utils/QRCodeImages.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright by the original author or authors.
+ *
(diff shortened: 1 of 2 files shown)


wallettemplate/src/main/java/wallettemplate/utils/QRCodeImages.java
```diff
@@ -0,0 +1,60 @@
+package wallettemplate.utils;
```
schildbach: Can you add the license header?

msgilligan: Added!

msgilligan:
@schildbach The build failure looks like it will clear if the job is resubmitted.

schildbach:
Merged!

PR #1811, opened 2019-04-23
Use TestNet3 in WalletTemplate

msgilligan:
WalletTemplate is not a production-ready wallet and (for now, at least)
should default (currently hard-coded) to TestNet3.


schildbach:
Merged.

PR #1813, opened 2019-04-23
Refactor AlertMessageTest.java

MethDamon:
Hey, I've been going through some tests in order to identify test smells and tests that can be refactored. I figured I'd start somewhere and refactor this. Splitting many asserts in one test method into several test methods has some advantages, it increases readability imo and the test itself is easier to use as documentation. Also, when some test fails you'll immediately see what's wrong instead of having to find the line number in the one test method with many asserts. I just wanted to mention that I am doing these pull requests based on an assignment at my university in a lecture about Software Testing. I am sorry for not mentioning this in the previous pull requests.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/AlertMessageTest.java
+++ b/core/src/test/java/org/bitcoinj/core/AlertMessageTest.java
@@ -1,6 +1,7 @@
 /*
  * Copyright 2011 Google Inc.
  * Copyright 2014 Andreas Schildbach
+ * Copyright 2019 Tim Strasser
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


schildbach:
This separation into many tiny tests seems excessive to me. I think it's not sustainable to continue this through all our tests.

That said, I agree that here or there a test could be split into two. It's just not something I'd like to unconditionally apply.

MethDamon:
I don't think any rule in software engineering unconditionally applies ;) Agreed, and I think that there are other test classes that would highly benefit from such refactoring. If you want to see any test classes refactored, please let me know, I will happily work on them.

MethDamon:
@schildbach Also bump :D

PR #1814, opened 2019-04-23
Put wallet data files in app data directory

msgilligan:
Previously **WalletTemplate** would store the `.spvchain` and `.wallet` files
in the app’s current working directory which could vary depending upon
how the app was launched.

This PR stores the app’s data files in the “app data directory” which
is always named **WalletTemplate** (`APP_NAME`) and located in the standard
location for application data directories on the current OS platform.

For macOS: `~/Library/Application Support/WalletTemplate`
For Linux/unix: `~/.WalletTemplate`
For Windows: `${APPDATA}/WalletTemplate`  (where `APPDATA` is a standard
Windows environment variable)

Changes:

* Added `OSDetector` class to detect Mac, *Nix, Windows
* Added `AppDataDirectory` class to return current dir Path
* Modified `Main` to use `appDataDirectory` rather than `“.”`

### DIFF
--- a/wallettemplate/src/main/java/wallettemplate/Main.java
+++ b/wallettemplate/src/main/java/wallettemplate/Main.java
@@ -35,6 +35,7 @@ import javafx.scene.layout.Pane;
 import javafx.scene.layout.StackPane;
 import javafx.stage.Stage;
 import wallettemplate.controls.NotificationBarPane;
+import wallettemplate.utils.AppDataDirectory;
 import wallettemplate.utils.GuiUtils;
 import wallettemplate.utils.TextFieldValidator;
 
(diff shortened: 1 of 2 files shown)


wallettemplate/src/main/java/wallettemplate/utils/OSDetector.java
```diff
@@ -0,0 +1,36 @@
...
+package wallettemplate.utils;
+
+/**
+ * Utility class to detect host operating system
+ */
+public class OSDetector {
+    private static final String osName = System.getProperty("os.name").toLowerCase();
+
+    static public boolean isWindows() {
+        return osName.contains("windows");
```
schildbach: This class is redundant to Utils.isWindows() / .isMac() / .isLinux().

msgilligan: I force-pushed an update that uses the existing `Utils` class.

msgilligan:
@schildbach It's using `o.b.c.Utils` now!

schildbach:
Merged, thanks!

PR #1817, opened 2019-04-24
Refactor ScriptPatternTest.java

MethDamon:
Again, the benefit of splitting multiple assertions in one test method into multiple test methods is increased readability, the test functions as documentation and if a test fails it's immediately visible what function of the class under test has caused it.


schildbach:
Thanks! Makes sense, I merged this.

PR #1818, opened 2019-04-24
Refactor UTXOTest.java and add two new test methods to it

MethDamon:
Changes:
- Split multiple asserts into test methods
- Added test method for ```getIndex()```
- Added test method for ```getHash()```

Also, this test uses the normal java serialization rather than ```writeObject()``` and ```readObject()```


MethDamon:
@schildbach Can you take another look at this?

schildbach:
I think this is very similar to #1813. Splitting the test into many "micro tests" doesn't scale.

I see however a couple of new asserts "hidden" in your refactoring. If you think they're worth to have, we could add them.

PR #1823, opened 2019-04-28
Remove Transaction.MIN_NONDUST_OUTPUT and NetworkParameters.getMinNonDustOutput().

schildbach:
The threshold has always been dependent on the actual output script type, even more so with the
SegWit discount. So use TransactionOutput.getMinNonDustValue() or .isDust().


schildbach:
I merged a version of this which just deprecated the field and method. The rest of it (complete removal) is scheduled for after 0.16.

schildbach:
Was removed via 6a7ac8e2c7f96ae5743dfa2035a00a265fa37924 and 2b5c50db259211b1dfe66dc5f6d26fdefcba2cd2.

PR #1827, opened 2019-04-29
refactor_test_smell

mansohal:
refactor test smells in the test methods.


schildbach:
This PR adds a text file with two lines. Seems like a mistake to me.

PR #1830, opened 2019-05-01
Use Bouncy Castle's Scrypt implementation rather than the standalone library.

schildbach:
Amazingly, on a Pixel 2 this implementation is a bit quicker than Will Glozer/Colin Percival's native code.


schildbach:
Merged.

msgilligan:
@schildbach and one less dependency! 👍

PR #1832, opened 2019-05-04
don't encode witness version to OP_N when constructing SegwitAddress

matthewleon:
This bug was never exposed outside the `SegwitAddress` class (and thus can't be tested) because the `encode` method is only ever called by `fromHash`, where the script version is always set to `0`.

Nevertheless, this bug could result in faults in future code, as other witness program versions begin to be used: The other construction methods, such as `fromBech32`, leave the version byte as a simple representation of the version number. This means that, were the` SegwitAddress` constructor ever called with a non-`0` version number, `getWitnessVersion` would return an incorrect result, and ScriptBuilder.createOutputScript would fail, re-running `encodeToOpN()` on the now greater-than-16 version number.


schildbach:
Great catch!

schildbach:
Merged!

PR #1833, opened 2019-05-05
simplify SegwitAddress.encode by using Arrays.prepend

matthewleon:
This code documents its intentions somewhat more clearly by moving array prepending into a safe utility method.


schildbach:
Can we perhaps use Guava or Java7 for this? Not a long while ago I replaced all BouncyCastle "non-cryptography-uses" by Guava/Java7.

PR #1838, opened 2019-05-05
deprecate LazyECPoint.getEncoded and .isCompressed

matthewleon:
Copies the exact deprecation messages from `ECPoint`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
+++ b/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
@@ -65,6 +65,10 @@ public class LazyECPoint {
         return get().getDetachedPoint();
     }
 
+    /**
+     * @deprecated per-point compression property will be removed, refer {@link #getEncoded(boolean)}
+     * @return a byte encoding.
+     */
     public byte[] getEncoded() {
         if (bits != null)
             return Arrays.copyOf(bits, bits.length);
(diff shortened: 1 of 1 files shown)


matthewleon:
merged in https://github.com/bitcoinj/bitcoinj/commit/09c8a8e8550fd001554228cfe832304fd0ad1fa3

schildbach:
I wonder what do we do if these methods are actually removed?

matthewleon:
> I wonder what do we do if these methods are actually removed?

They claim they will do this.

I have been reflecting on this issue. I have some ideas I'll be exploring on a few branches. I will try to come up with something concrete.

Regardless, I do think it makes sense to deprecate these independently of whether or not they actually go ahead and get rid of them upstream.

matthewleon:
To be a bit more specific, the `ECKey` class has a TODO somewhere in it seeking to begin tracking compression state within `ECKey` itself, rather than delegating to `ECPoint`. That should be a perfectly feasible way to do things with minimal breakage.

schildbach:
I think we should add `@deprecated` only if we plan to remove these methods. But as far as I understand we will not remove them anytime soon because compressed state will be important for Bitcoin basically forever.

Just because the implementation used deprecated API now, that doesn't mean we need to remove the method. As you pointed out, there are alternative ways to implement the desired behaviour.

PR #1839, opened 2019-05-05
Update a bit of outdated documentation in the ECKey code

matthewleon:
From what I can tell, `pub` is _always_ set. It just uses a lazy class.


schildbach:
Indeed. Merged!

PR #1840, opened 2019-05-05
SegwitAddress simplification: store witness program bytes

matthewleon:
Modifies the `SegwitAddress` class to directly store the witness program and witness version, rather than their combined, Base32 encoded version.

The two previous private constructors are combined into one simplified constructor.

Getters for `witnessVersion` and `bytes` are simplified.

All encoding and decoding is shifted to serialization to and from Bech32 `String`s.

This can logically be followed by moving Base32 en/decoding into the `Bech32` class.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/SegwitAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/SegwitAddress.java
@@ -20,6 +20,8 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 
 import java.io.ByteArrayOutputStream;
+import java.util.Arrays;
+import java.util.Objects;
 
 import javax.annotation.Nullable;
 
@@ -212,6 +214,21 @@ public class SegwitAddress extends Address {
         bech32Bytes[0] = (byte) (witnessVersion & 0xff);
         System.arraycopy(convertedProgram, 0, bech32Bytes, 1, convertedProgram.length);
         return Bech32.encode(params.getSegwitAddressHrp(), bech32Bytes);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), witnessVersion);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        SegwitAddress other = (SegwitAddress) o;
+        return this.params.equals(other.params)
+                && Arrays.equals(this.bytes, other.bytes)
+                && (this.witnessVersion == other.witnessVersion);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


matthewleon:
This might not be ideal as-is: I can see that it probably requires overriding `compareTo`, at least. Thoughts here?

matthewleon:
(my own inclination is to say that despite requiring `equals` and `compareTo` overrides, this is worth it for the added simplicity/clarity)

matthewleon:
Have overridden `equals` and `hashCode`, but will leave `compareTo`. Making a sensible override of the `compareTo` inherited from `PrefixedChecksummedBytes` that properly keeps the contract of `compareTo` when compared to other members of the class hierarchy is seemingly impossible. I'd hope that any sensible user of the class would use a custom `Comparator` with whatever ad-hoc logic they need.

For what it's worth, I'd vote to actually make a breaking change that kills the `Comparable` implementation on `PrefixedChecksummedBytes`, but that 's a separate issue.

matthewleon:
`compareTo` issue addressed in a separate PR: https://github.com/bitcoinj/bitcoinj/pull/1844

schildbach:
Looks good to me. Is this PR dependent on #1844 being merged?

matthewleon:
> Looks good to me. Is this PR dependent on #1844 being merged?

Not dependent on #1844 . The `Comparable` implementation here is already inconsistent with `equals` when comparing between different subclasses of `PrefixedChecksummedBytes`. This PR will make `SegwitAddress.compareTo` also inconsistent with `equals` when comparing between two `SegwitAddress` instances. I'm not sure how much of a problem this is: `LegacyAddress` already shows exactly this problem because of its additional field.

I'd say we accept that the `Comparable` implementation is broken, and accept that even if this PR makes it "more broken" in a certain sense, the real solution is to kill `Comparable`.

msgilligan:
This should be rebased to `master` since PR #2018 was merged. This PR can now be considered completely separately from the `Comparable` issues.

@schildbach @matthewleon What should we do?

msgilligan:
> Modifies the `SegwitAddress` class to directly store the witness program and witness version, rather than their combined, Base32 encoded version.

Technically, the combined version is only partially Base32-encoded, with 8-bits/byte to 5-bytes/byte `convertBits` having been applied, but not the conversion to the corresponding ascii codes.

msgilligan:
I think this PR contains at least one good idea. But I propose we first move `byte[] bytes` to a format that maintains the `witnessVersion` prefix followed by the standard (8-bits/byte) `witnessProgram`. We can then separately consider removing the `witnessVersion` prefix from `bytes` and moving it to a separate variable in the `SegwitAddress` subclass.

See Issue #2697.

msgilligan:
This PR has been effectively implemented by two other PRs in a series of related refactorings, see:

* https://github.com/bitcoinj/bitcoinj/pull/2744
* https://github.com/bitcoinj/bitcoinj/pull/2745

Thanks @matthewleon for this suggestion!

msgilligan:
@schildbach This can be closed.  (See my previous comment.)

schildbach:
Thanks to both of your for improving SegwitAddress!

PR #1844, opened 2019-05-09
breaking: kill PrefixedChecksummedBytes Comparable implementation

matthewleon:
Sadly, this implementation falls prey to the problem explained here: http://www.javapractices.com/topic/TopicAction.do?Id=10

> When a class extends a concrete Comparable class and adds a significant field, a correct implementation of compareTo cannot be constructed. The only alternative is to use composition instead of inheritance.

Currently, `LegacyAddress` behaves this way. My proposed changes to `SegwitAddress` add significant fields. `Bip38PrivateKey` has a number of such fields.

Besides this, the implementation of `equals` used throughout these classes checks for class equality, making it inconsistent with `compareTo` _even_ in cases where significant fields are not added. This is nasty for reasons explained well in the `Comparable` javadoc:

> It is strongly recommended (though not required) that natural orderings be consistent with equals. This is so because sorted sets (and sorted maps) without explicit comparators behave "strangely" when they are used with elements (or keys) whose natural ordering is inconsistent with equals. In particular, such a sorted set (or sorted map) violates the general contract for set (or map), which is defined in terms of the equals method.

This PR kills the `Comparable` implementation and replaces it with public static `Comparator`s that are, in the case of `LegacyAddress`, covariant. I've added a test case that demonstrates a bit of the issue.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
@@ -21,10 +21,13 @@ package org.bitcoinj.core;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.Objects;
 
 import javax.annotation.Nullable;
 
+import com.google.common.collect.ComparisonChain;
+import com.google.common.primitives.UnsignedBytes;
 import org.bitcoinj.params.Networks;
 import org.bitcoinj.script.Script;
 import org.bitcoinj.script.Script.ScriptType;
@@ -50,6 +53,17 @@ public class LegacyAddress extends Address {
 
     /** True if P2SH, false if P2PKH. */
     public final boolean p2sh;
+
+    public static final Comparator<LegacyAddress> byteComparator = new Comparator<LegacyAddress>() {
+        @Override
+        public int compare(LegacyAddress o1, LegacyAddress o2) {
+            return ComparisonChain.start()
+                    .compare(o1.params.getId(), o2.params.getId())
+                    .compareFalseFirst(o1.p2sh, o2.p2sh)
+                    .compare(o1.bytes, o2.bytes, UnsignedBytes.lexicographicalComparator())
+                    .result();
+        }
+    };
 
     /**
      * Private constructor. Use {@link #fromBase58(NetworkParameters, String)},
--- a/core/src/main/java/org/bitcoinj/core/PrefixedChecksummedBytes.java
+++ b/core/src/main/java/org/bitcoinj/core/PrefixedChecksummedBytes.java
@@ -23,6 +23,7 @@ import java.io.ObjectOutputStream;
 import java.io.Serializable;
 import java.lang.reflect.Field;
 import java.util.Arrays;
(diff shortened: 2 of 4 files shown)


schildbach:
I'm adding @msgilligan to this PR because he lobbied for comparable addresses in the first place (via c72c48cd586c510befb5262b39648a85fc3ead71).

matthewleon:
Curiously, if you examine the source code at https://github.com/bitcoinj/bitcoinj/commit/c72c48cd586c510befb5262b39648a85fc3ead71 you can see that the `Comparable` implementation has been broken since inception: it is inconsistent with `equals` both between `VersionedChecksummedBytes` and `Address` objects (which will _always_ be unequal according to `equals`, but can still be equal by the logic of `compareTo`) and among two `Address` objects, where comparison will ignore `params`.

matthewleon:
I should add that the `Comparable` interface doesn't technically _require_ consistency with `equals`. I can see there being a "solution" where you just loudly advertise that it is inconsistent with `equals` and leave it at that, rather than breaking anyone who has chosen to depend on this interface.

The unfortunate thing is that the existence of `Comparable` on `PrefixedChecksummedBytes` makes certain refactors a bit scary, because the exact places where `compareTo` and `equals` behave consistently or inconsistently shifts when you add/remove fields to/from subclasses.

msgilligan:
Yeah, I really like (and need) comparable addresses and would like to figure out a way to keep that functionality. At the very least for `LegacyAddress`. Unfortunately I currently don't have a copy of *Effective Java, Third Edition* (which is referenced @matthewleon's link above) so I just ordered one for next-day delivery.

Is there a way we can keep `Comparable` addresses? I'll look into it further over the next few days and see if I can come up with a proposal.

schildbach:
Can't we simply base `PrefixedChecksummedBytes.compareTo()` on the `toString()` of the two objects to compare?

matthewleon:
> Can't we simply base `PrefixedChecksummedBytes.compareTo()` on the `toString()` of the two objects to compare?

Likely, yes. That's actually probably the most realistic way to have a `compareTo` that remains consistent with `equals`. It also is probably the technique that has the largest performance penalty. That said, if you are comparing enough of these objects for perf to matter, you maybe are willing to use a custom `Comparator` anyway.

It's a real improvement over the current situation, and is also far less breaking than getting rid of `Comparable`. Good compromise.

schildbach:
If performance really matters, we could start caching the `toString()` outputs. The `PrefixedChecksummedBytes` hierarchy is meant to be immutable, so it would be easy to do.

msgilligan:
> When a class extends a concrete Comparable class and adds a significant field, a correct implementation of compareTo cannot be constructed. 

Minor detail: Actually in this case we are not extending a _concrete_ `Comparable` class but an abstract class.

msgilligan:
I'm working on a PR to improve `compareTo` for addresses and reviewing this thread.

The PR I'm working on will remove compareTo from `PrefixedChecksummedBytes` and add it to the abstract `Address` class.

msgilligan:
FYI: PR #2018 was merged which addressed the same issues as this PR.

PR #1852, opened 2019-05-14
KeyChainGroup: New CurrentKeyChangeEventListener…

schildbach:
…that fires when the current key and/or address changes.


schildbach:
This also changes wallet-template to make use of the new listener for updating its current address in the UI.

@msgilligan Are you ok with this change to wallet-template?

msgilligan:
The change to wallet-template looks good to me!

PR #1854, opened 2019-05-17
switch PrefixedChecksummerBytes.compareTo to use toString()

matthewleon:
See discussion at https://github.com/bitcoinj/bitcoinj/pull/1844


schildbach:
The change looks good to me. @msgilligan Are you happy with this, too?

The failing unit test is "for real" this time. Guess that's the downside of strict mocking – it somewhat hardcodes implementation details into the test.

matthewleon:
@schildbach thanks for the feedback. Happy to fix failing tests, but I would like to wait for feedback from @msgilligan before proceeding.

schildbach:
@msgilligan Do you want to comment on this PR?

msgilligan:
Yes, I do. Sorry for taking so long. I've been focused on the AirGap signing stuff, which I'm hoping to finish today.

I'm hoping I/we can find a way to do it without String conversion and will look into it this weekend and then comment.

msgilligan:
See [AddressSpec."Create valid MainNet Address from private key"()](https://github.com/ConsensusJ/consensusj/blob/master/bitcoinj-spock/src/test/groovy/com/msgilligan/bitcoinj/spock/AddressSpec.groovy#L65-L77) for a Spock test that illustrates a case where string comparison has a different result than field comparison.

msgilligan:
I did a bit of reading on the concerns(s) discussed in Issue #1844 and came to the conclusion that `compareTo` should only need to compare Address classes -- or perhaps only `Address` classes of the same subtype (`LegacyAddress` and `SegwitAddress`) so I would support removing `compareTo` from the base class `PrefixedChecksummedBytes` and implementing it simply for `Address` (and other classes on an as-needed basis)

msgilligan:
My application requires a high-performance implementation that is memory efficient as well. I am creating sorted lists and maps, with using address comparison to sort and filter very large datasets. In some cases the datasets can actually be _every_ address on the blockchain.

matthewleon:
> I would support removing `compareTo` from the base class `PrefixedChecksummedBytes` and implementing it simply for `Address` (and other classes on an as-needed basis)

This might be something of an improvement, but the issue doesn't really seem to be truly solved unless `Comparable` is only allowed on classes marked `final`. Granted, maybe it's okay to only "sort of" solve the issue by saying "hey, if you choose to extend these classes, the burden is on you to ensure that you don't break `Comparable<OpenClass>`... Seems a bit fishy though.

msgilligan:
Can we mark the classes final? Are users of *bitcoinj* extending these classes?

Polve:
Please, please, don't declare any classes or method final: there are users extending and/or subclassing classes :-)

schildbach:
> RegTest and TestNet addresses can have the same string but be different objects.

Fair point. We could compare the network too though. It's part of `PrefixedChecksummedBytes` as well, so it would be easy and cheap.

> My application requires a high-performance implementation that is memory efficient as well. I am creating sorted lists and maps, with using address comparison to sort and filter very large datasets. In some cases the datasets can actually be _every_ address on the blockchain.

I know we had this discussion already months ago when we introduced the natural ordering of `PrefixedChecksummedBytes`, but can't you use your own custom `Comparator`?

Would caching the textual representation be ok with your memory efficiency requirements?

schildbach:
btw. caching could perhaps also mean to use a weak reference, so that the cached texts would be garbage collected under memory pressure.

msgilligan:
> I know we had this discussion already months ago when we introduced the natural ordering of PrefixedChecksummedBytes, but can't you use your own custom Comparator?

I could, but that's a little awkward. It sure is nice to just be able to use `Address` a a key in a (sorted) `Map`. I'd really like to preserve this simple, intuitive behavior.

msgilligan:
> I […] came to the conclusion that `compareTo` should only need to compare Address classes […] so I would support removing `compareTo()` from the base class `PrefixedChecksummedBytes` and implementing it simply for `Address` [and subclasses]

My `TreeMap` based sorting of addresses just broke on a large dataset (all ~347745 addresses currently holding USDT)  The current implementation of `compareTo()` is not accounting for the `boolean p2sh` field (it was previously represented in the version byte.)

Omni is also working towards adding official Segwit support, so soon I'll be wanting to create `Map` keys for datasets that include both `SegwitAddress` and `LegacyAddress` addresses. Addresses are unique and should be able to be used as keys, whether they're legacy or segwit.

So I'm working on a PR that:

* removes `Comparable` from `PrefixedChecksummedBytes`
* adds `implements Comparable<Address>` to `Address` with an `abstract` method declaration
* provides  `compareTo()` implementations for  `LegacyAddress` and `SegwitAddress`.

This would allow @Polve and others to subclass, but they would be forced to implement `compareTo` and think about its ramifications.  

> saying "hey, if you choose to extend these classes, the burden is on you to ensure that you don't break `Comparable<Address>`

With this PR, we will be saying the above in the way _Gosling_ (or _Bloch_) intended by requiring subclasses to implement `compareTo()` 😁.

This seems like the correct (i.e. consistent with _Effective Java_, etc.) way to do it and also seems to address everyone's concerns on this thread.

msgilligan:
Currently I'm implementing sort order as follows:

1.  netParams
1.  Legacy vs. Segwit
1.  (Legacy only) Version byte
1.  The `bytes` array

This means that addresses on the same network would be grouped together, within networks legacy and segwit would be grouped, and then finally they'd be sorted based on the "bytes". That seems to be what most people would want.

Obviously this is not the same order as the human readable address-strings. If someone wants that they can create a custom (and very slow) Comparator, or simply convert to String and use the String as the key. But for datasets with 100's of thousands to 100's of millions of addresses the memory savings of `byte[]` versus `String` is significant.

msgilligan:
My work-in-progress is in PR #2018.

msgilligan:
@schildbach Now that PR #2018 is merged, I think we can close this one.

PR #1857, opened 2019-05-18
kill ScriptChunk.startLocationInProgram

matthewleon:
The `startLocationInProgram` field of `ScriptChunk` was used as a cacheing mechanism for storing the location of each `ScriptChunk` in a `Script`. This was memory-inefficient, as this integer was only actually used during interpretation of `OP_CODESEPARATOR`. Furthermore, it was dangerous and inconsistent: the `ScriptBuilder` methods, for example, do not properly set `startLocationInProgram`, leading to a potential runtime failure when interpreting `ScriptBuilder`-constructed Scripts.

Rather than cacheing each chunk's location, we add a method called `size()` that efficiently returns the size of a serialized `ScriptChunk`. During `Script` interpretation, we use this method to track the program counter.

Additionally, we add an exhaustive test to ensure that `size()` properly returns the size of a serialized `ScriptChunk`.

This will break any code that relied on `ScriptChunk.getStartLocationInProgram()`.

### DIFF
--- a/core/src/test/java/org/bitcoinj/script/ScriptChunkSizeTest.java
+++ b/core/src/test/java/org/bitcoinj/script/ScriptChunkSizeTest.java
@@ -27,20 +27,20 @@ public class ScriptChunkSizeTest {
     @Parameterized.Parameters
     public static Collection<ScriptChunk> data() {
         ArrayList<ScriptChunk> opcodes = new ArrayList<>(0xff);
-        for (int op = OP_NOP; op < 0xff; op++) opcodes.add(new ScriptChunk(op, null));
+        for (int op = OP_NOP; op < 0xff + 1; op++) opcodes.add(new ScriptChunk(op, null));
 
         ArrayList<ScriptChunk> smallData = new ArrayList<>(OP_PUSHDATA1);
         for (int op = 1; op < OP_PUSHDATA1; op++) smallData.add(new ScriptChunk(op, randomBytes(op)));
 
         ArrayList<ScriptChunk> pushData1 = new ArrayList<>(0xff);
-        for (int i = 0; i < 0xff; i++) pushData1.add(new ScriptChunk(OP_PUSHDATA1, randomBytes(i)));
+        for (int i = 0; i < 0xff + 1; i++) pushData1.add(new ScriptChunk(OP_PUSHDATA1, randomBytes(i)));
 
-        ArrayList<ScriptChunk> pushData2 = new ArrayList<>((int)Script.MAX_SCRIPT_ELEMENT_SIZE);
-        for (int i = 0; i < Script.MAX_SCRIPT_ELEMENT_SIZE; i++)
+        ArrayList<ScriptChunk> pushData2 = new ArrayList<>((int)Script.MAX_SCRIPT_ELEMENT_SIZE + 1);
+        for (int i = 0; i < Script.MAX_SCRIPT_ELEMENT_SIZE + 1; i++)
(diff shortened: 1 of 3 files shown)


schildbach:
Looks great! I'm glad you've tackled this. I tried years ago and failed (-:

My hope is it will save a bit of ScriptChunk instances even when used in SPV mode. ScriptChunk instances are the #1 bottleneck when it comes to large wallets on memory constrained devices like mobile.

matthewleon:
> ScriptChunk instances are the #1 bottleneck when it comes to large wallets on memory constrained devices like mobile.

Do you have any benchmarks, or even informal processes, that you're using to measure this? There are a number of changes that I think can be made that decrease memory consumption in exchange for a negligible, or even nonexistent, processing penalty.

matthewleon:
> My hope is it will save a bit of ScriptChunk instances even when used in SPV mode.

I'm curious to see. Naively, the saved integer space on each ScriptChunk instance probably doesn't amount to very much, especially since frequently-used operations were being cached in `STANDARD_TRANSACTION_SCRIPT_CHUNKS`.

I'm not sure, however, if the locations being used in `STANDARD_TRANSACTION_SCRIPT_CHUNKS` were correct for SegWit `Script`s, so this PR might inadvertently save you a bit of extra memory by using cached `ScriptChunk`s for SegWit as well. It would be a nice little win.

Now, if we can hide the `ScriptChunk` constructor and build the chunks from a factory, thus avoiding the entire allocation in the first place, we might also save on some unnecessary GC hits... But one step at a time :)

schildbach:
Well, a transaction can have many (several thousands of) outputs. Each output has, in the P2PKH case 5 ScriptChunks (P2WPKH: 2). This means that wallets receiving regular transactions of type "pay-to-many" (e.g. mining or microjob payouts) run out of memory sooner or later. I saw this in the Android memory profiler too.

I think other wallets generally don't keep transactions complete like we do. For calculating the balance of a wallet, only the relevant UTXOs are needed. And the transaction history could be kept in a list with a custom structure or even in a database. On the other hand, unconfirmed txns and txns at risk of being reorganized need to stay complete so that we can re-upload them to the network if needed. I think Mike's idea was to have this "event horizon" where past that point in time transactions could be disassembled at will to save resources. All this is a pretty disruptive change unfortunately.

Sometime I think about this other idea, which would be less disruptive (but could still be combined with the above): Transactions are kept in their original network serialized form. Pubkeys, hashes and signatures could be expressed by a sub-array of that one array (so no copying occurs) and optimized getters exist for inputs and outputs. Parsing into chunks would only happen if full validation is enabled. SPV mode would perhaps use chunks only for constructing when sending, which is a rare operation. For incoming transactions, it should not be needed.

P2WPKH has fewer of these standard ops (`OP_DUP` etc.), and none are new, so I didn't identify any new elements for `STANDARD_TRANSACTION_SCRIPT_CHUNKS`.

core/src/test/java/org/bitcoinj/script/ScriptChunkSizeTest.java
```diff
@@ -0,0 +1,65 @@
+package org.bitcoinj.script;
```
schildbach: This new class is missing the license. Also, while you're at it, can you make it that the last line is terminated with an EOL like all the others?

matthewleon: done

matthewleon:
merged in https://github.com/bitcoinj/bitcoinj/commit/5647f48b0d1796aca314fe159953ee13be5a0194

PR #1858, opened 2019-05-18
make ScriptTest constants private

matthewleon:
These constants haven't been touched in a number of years. Since then, commits have added several `private static` constants to this test. This confuses the reader, who naturally becomes curious as to why some constants are `private` while others are not.

This commit renders all of the constants `private`.


schildbach:
Merged.

PR #1859, opened 2019-05-19
Opcode names: accept OP_FALSE and OP_TRUE

matthewleon:
Previously, only `OP_0` was accepted as the name for the 0x00 opcode.


schildbach:
Hmm, is it possible `OP_TRUE` is missing too?

matthewleon:
> is it possible `OP_TRUE` is missing too?

Good catch. Fixed.

schildbach:
Thanks! Merged.

PR #1862, opened 2019-05-19
simplify the Script.correctlySpends() p2sh pushdata check

matthewleon:
Simplify the check that all p2sh script content is comprised of pushdata operations, and give a clearer error message when this is not the case.


schildbach:
Merged.

PR #1863, opened 2019-05-19
remove redundant opcode check in ScriptPattern.isSentToMultisig

matthewleon:
The following line has a check that is strictly stronger.


schildbach:
Merged this one too.

PR #1864, opened 2019-05-19
bugfix: ScriptPattern.isSentToMultisig throws

matthewleon:
`ScriptPattern.isSentToMultisig()` is meant to be a safe function returning a `boolean`. It was catching the wrong exception, causing it to throw when given crafted scripts.

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/ScriptPattern.java
+++ b/core/src/main/java/org/bitcoinj/script/ScriptPattern.java
@@ -210,7 +210,6 @@ public class ScriptPattern {
         try {
             // Second to last chunk must be an OP_N opcode and there should be that many data chunks (keys).
             ScriptChunk m = chunks.get(chunks.size() - 2);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/script/ScriptPattern.java
```diff
@@ -210,15 +210,14 @@ public static boolean isSentToMultisig(Script script) {
         try {
             // Second to last chunk must be an OP_N opcode and there should be that many data chunks (keys).
             ScriptChunk m = chunks.get(chunks.size() - 2);
-            if (!m.isOpCode()) return false;
```
schildbach: Hmmm, rather than continuing to rely on an exception flow why not add an `isOpN()` and use that? And then remove the try/catch.

matthewleon: Good idea. Will do this.

PR #1865, opened 2019-05-21
ScriptChunk.decodeOpN: remove checkState

matthewleon:
The `checkState` on `ScriptChunk.decodeOpN()` was excessively strict: it excluded `OP_0`. As the method calls `Script.decodeFromOpN()`, which itself has a check on the passed-in opcode, we simply remove the check entirely.

No code in bitcoinj was calling `ScriptChunk.decodeOpN()`.


schildbach:
Merged!

PR #1867, opened 2019-05-23
simplify isSentToMultisig

matthewleon:
Successor of https://github.com/bitcoinj/bitcoinj/pull/1864

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/ScriptPattern.java
+++ b/core/src/main/java/org/bitcoinj/script/ScriptPattern.java
@@ -198,7 +198,7 @@ public class ScriptPattern {
 
     /**
      * Returns whether this script matches the format used for multisig outputs:
-     * {@code [n] [keys...] [m] CHECKMULTISIG}
+     * {@code [m] [keys...] [n] CHECKMULTISIG}
      */
     public static boolean isSentToMultisig(Script script) {
         List<ScriptChunk> chunks = script.chunks;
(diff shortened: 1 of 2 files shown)


matthewleon:
Actually still not 100% sure this is documented and working correctly. Different sources online switch the `m` and the `n`. I'll have to compare to the core source.

matthewleon:
Have amended the javadoc. I believe the `m` and the `n` were switched in the doc, but functioning correctly in the code. Please check my thinking here carefully. Thanks!

schildbach:
I agree the JavaDoc has it the wrong way round. See https://github.com/bitcoin/bips/blob/master/bip-0011.mediawiki

core/src/main/java/org/bitcoinj/script/ScriptPattern.java
```diff
@@ -198,7 +198,7 @@ public static boolean isP2WSH(Script script) {
 
     /**
      * Returns whether this script matches the format used for multisig outputs:
```
schildbach: Perhaps we should use the tern "M-of-N multisig" here? This would define what m and n actually means.

And the Javadoc change could go on its own commit, as it's not related to the exception change as far as I understand.

matthewleon: https://github.com/bitcoinj/bitcoinj/pull/1867

matthewleon:
It might also make sense to merge https://github.com/bitcoinj/bitcoinj/pull/1863 first.

schildbach:
Thanks a lot! I just merged this.

PR #1875, opened 2019-05-31
Replace Guava newArrayList()/newLinkedList()

msgilligan:
... with direct call to the JDK 7 constructor (as recommended by deprecation
comment in Guava)


schildbach:
Looks like there are a couple of missing imports. But the concept looks fine to me.

msgilligan:
Sorry that was sloppy. It should be fixed now.

schildbach:
Merged.

PR #1877, opened 2019-06-06
Breaking: New HDPath type to replace ImmutableList<ChildNumber>

msgilligan:
`HDPath` replaces `ImmutableList<ChildNumber>` as the preferred representation for a hierarchical deterministic path.

Some methods take `List<ChildNumber>` as a parameter which will accept `ImmutableList<ChildNumber>`. Since `HDPath` implements `List<ChildNumber>`, if you're calling a method that now returns `HDPath`, you can change the receiving variable from `ImmutableList<ChildNumber>` to `List<ChildNumber>` or even call `ImmutableList.copyOf()` if necessary.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -21,7 +21,6 @@ import org.bitcoinj.core.*;
 import org.bitcoinj.script.Script;
 
 import com.google.common.base.MoreObjects;
-import com.google.common.collect.ImmutableList;
 import org.bouncycastle.crypto.params.KeyParameter;
 import org.bouncycastle.math.ec.ECPoint;
 
@@ -29,7 +28,9 @@ import javax.annotation.Nullable;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
+import java.util.List;
 import java.util.Objects;
 
 import static org.bitcoinj.core.Utils.HEX;
@@ -54,7 +55,7 @@ public class DeterministicKey extends ECKey {
     };
 
     private final DeterministicKey parent;
-    private final ImmutableList<ChildNumber> childNumberPath;
+    private final HDPath childNumberPath;
     private final int depth;
     private int parentFingerprint; // 0 if this key is root node of key hierarchy
 
@@ -62,7 +63,7 @@ public class DeterministicKey extends ECKey {
     private final byte[] chainCode;
 
     /** Constructs a key from its components. This is not normally something you should use. */
-    public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
+    public DeterministicKey(List<ChildNumber> childNumberPath,
                             byte[] chainCode,
                             LazyECPoint publicAsPoint,
                             @Nullable BigInteger priv,
@@ -70,13 +71,13 @@ public class DeterministicKey extends ECKey {
         super(priv, compressPoint(checkNotNull(publicAsPoint)));
         checkArgument(chainCode.length == 32);
         this.parent = parent;
-        this.childNumberPath = checkNotNull(childNumberPath);
+        this.childNumberPath = HDPath.of(checkNotNull(childNumberPath));
         this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
         this.depth = parent == null ? 0 : parent.depth + 1;
         this.parentFingerprint = (parent != null) ? parent.getFingerprint() : 0;
     }
 
-    public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
+    public DeterministicKey(List<ChildNumber> childNumberPath,
                             byte[] chainCode,
                             ECPoint publicAsPoint,
                             @Nullable BigInteger priv,
@@ -85,21 +86,21 @@ public class DeterministicKey extends ECKey {
     }
 
     /** Constructs a key from its components. This is not normally something you should use. */
-    public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
+    public DeterministicKey(List<ChildNumber> childNumberPath,
                             byte[] chainCode,
                             BigInteger priv,
                             @Nullable DeterministicKey parent) {
         super(priv, compressPoint(ECKey.publicPointFromPrivate(priv)));
         checkArgument(chainCode.length == 32);
         this.parent = parent;
-        this.childNumberPath = checkNotNull(childNumberPath);
+        this.childNumberPath = HDPath.of(checkNotNull(childNumberPath));
         this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
         this.depth = parent == null ? 0 : parent.depth + 1;
         this.parentFingerprint = (parent != null) ? parent.getFingerprint() : 0;
     }
 
     /** Constructs a key from its components. This is not normally something you should use. */
-    public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
+    public DeterministicKey(List<ChildNumber> childNumberPath,
                             byte[] chainCode,
                             KeyCrypter crypter,
                             LazyECPoint pub,
@@ -131,7 +132,7 @@ public class DeterministicKey extends ECKey {
      * information about its parent key.  Invoked when deserializing, but otherwise not something that
      * you normally should use.
      */
-    public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
+    public DeterministicKey(List<ChildNumber> childNumberPath,
                             byte[] chainCode,
                             LazyECPoint publicAsPoint,
                             @Nullable DeterministicKey parent,
@@ -140,7 +141,7 @@ public class DeterministicKey extends ECKey {
         super(null, compressPoint(checkNotNull(publicAsPoint)));
         checkArgument(chainCode.length == 32);
         this.parent = parent;
-        this.childNumberPath = checkNotNull(childNumberPath);
+        this.childNumberPath = HDPath.of(checkNotNull(childNumberPath));
         this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
         this.depth = depth;
         this.parentFingerprint = ascertainParentFingerprint(parent, parentFingerprint);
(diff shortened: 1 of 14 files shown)


core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -54,29 +55,29 @@ public int compare(ECKey k1, ECKey k2) {
...
+    private final HDPath childNumberPath;
     private final int depth;
     private int parentFingerprint; // 0 if this key is root node of key hierarchy
 
     /** 32 bytes */
     private final byte[] chainCode;
 
     /** Constructs a key from its components. This is not normally something you should use. */
-    public DeterministicKey(ImmutableList<ChildNumber> childNumberPath,
+    public DeterministicKey(List<ChildNumber> childNumberPath,
```
schildbach: Why not migrate these constructors to HDPath too?

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -0,0 +1,120 @@
+package org.bitcoinj.crypto;
```
schildbach: Can you add the license?

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -0,0 +1,120 @@
...
+
+import java.util.AbstractList;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Representation of HD Key derivation path
```
schildbach: You might want to use some paragraph markup in your JavaDoc.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -0,0 +1,120 @@
...
+
+
+    /**
+     * Extend the path by appending two ChildNumber objects
+     *
+     * @param child1 the first child to append
+     * @param child2 the second child to append
+     * @return A new immutable path
+     */
+    public HDPath extend(ChildNumber child1, ChildNumber child2) {
```
sparecycles: nitpick: (style) maybe this should directly follow the `extend(ChildNumber child)` method.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Extend the path by appending a relative path
+     *
+     * @param path2 the relative path to append
+     * @return A new immutable path
+     */
+    public HDPath extend(List<ChildNumber> path2) {
+        List<ChildNumber> mutable = new ArrayList<>(this.unmodifiableList); // Mutable copy
+        mutable.addAll(path2);
+        return new HDPath(mutable);
```
sparecycles: question: the `hasPrivateKey` field ends up being false in HDPaths produced by `extend()` operations, should it be preserved or should there be a way to introduce true values?

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
msgilligan: Because that would be (perhaps unnecessarily) breaking. The idea is that `List<ChildNumber>` will accept both `ImmutableList<ChildNumber> ` and `HDPath`. If you'd prefer migrating those to `HDPath`, I'd be happy to change it.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: Your'e right -- it should be preserved!

Fixed in force push: https://github.com/bitcoinj/bitcoinj/compare/7e832e4fd77a5cabd1df09caf3da946c4a140167..df2d543d4204193be42424bbaa6c15687cd82910

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: Fixed in force push: https://github.com/bitcoinj/bitcoinj/compare/7e832e4fd77a5cabd1df09caf3da946c4a140167..df2d543d4204193be42424bbaa6c15687cd82910

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: Fixed in force push: https://github.com/bitcoinj/bitcoinj/compare/7e832e4fd77a5cabd1df09caf3da946c4a140167..df2d543d4204193be42424bbaa6c15687cd82910

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: Fixed in force push: https://github.com/bitcoinj/bitcoinj/compare/7e832e4fd77a5cabd1df09caf3da946c4a140167..df2d543d4204193be42424bbaa6c15687cd82910

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
msgilligan: ... alternatively we could use overloading to accept either with the `List` variant being deprecated. We also have to avoid the _Catch-22_ situation where we need an `HDPath` to create an `HDPath`...

schildbach: I see. Guess it's ok for now, thanks for the clarification.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * @param child the child to append
+     * @return A new immutable path
+     */
+    public HDPath extend(ChildNumber child) {
+        List<ChildNumber> mutable = new ArrayList<>(this.unmodifiableList); // Mutable copy
+        mutable.add(child);
+        return new HDPath(this.hasPrivateKey, mutable);
+    }
+
```
schildbach: Nitpick: One newline too many.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -0,0 +1,135 @@
...
+ * <p>
+ * {@code HDPath} is immutable and uses the {@code Collections.UnmodifiableList} type internally.
+ * <p>
+ * It implements {@code java.util.List<ChildNumber>} to ease migration
+ * from the previous Guava {@code ImmutableList<ChildNumber>}. It should be a minor breaking change
+ * to replace {@code ImmutableList<ChildNumber>} with {@code List<ChildNumber>} where necessary in your code. Although
+ * it is recommended to use the {@code HDPath} type for clarity and for access to {@code HDPath}-specific functionality.
+ */
+public class HDPath extends AbstractList<ChildNumber> {
+    static final char PREFIX_PRIVATE = 'm';
```
schildbach: Can these constants be private?

schildbach:
Side note: I'm a bit worried to put all that HD-related stuff into the `.crypto` package. Somehow I think we could move that to `.wallet` or maybe a new `.hd`. However we can (and probably should) do that after merging this PR in a separate commit.

On a related side note, do we want to keep the `HD*` naming? Possible alternatives would be `HierarchicalDeterministicPath` or maybe `HierarchicalPath`.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: They can, but it seems like they should be made available as public constants _somewhere_.

msgilligan:
> On a related side note, do we want to keep the `HD*` naming? Possible alternatives would be `HierarchicalDeterministicPath` or maybe `HierarchicalPath`.

I prefer the brevity because `HDPath` is already used in many places. In the Bitcoin/Cryptocurrency context it is a well-known abbreviation.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: Will fix.

schildbach:
> I prefer the brevity because `HDPath` is already used in many places. In the Bitcoin/Cryptocurrency context it is a well-known abbreviation.

Ok, agreed.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
schildbach: Somehow I can't think of a reason to have them public. The textual representation of a path is owned by HDPath, no?

msgilligan: I can think of a couple of at least one use case: parsing a path string. I don't think we have code to do that. And if even if we did have that code, we're using `44H` and someone might want to implement the `44'` format.

msgilligan: That said, I'm happy to make them private if you prefer.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: Fixed in a force-push.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
schildbach: I think yes, for now I'd prefer to have them private.

msgilligan: Fixed in force-push.

msgilligan:
> Somehow I think we could move that to .wallet or maybe a new .hd. However we can (and probably should) do that after merging this PR in a separate commit.

I agree. I'm also continue to think about how we could crate a `base` library with fewer dependencies `.hd` could be a part of that.

Polve:
I strongly agree on a base library, what I would love in it is just crypto stuff and parsing classes for base bitcoin objects: no network code or logic to handle the blockchain and wallets.

msgilligan:
> just crypto stuff and parsing classes for base bitcoin objects: no network code or logic to handle the blockchain and wallets.

That's my idea, too. But it's not a trivial change (e.g. just splitting out certain packages into a new jar)

schildbach:
Looks like this is ready to merge. Agreed?

msgilligan:
Agreed.

schildbach:
Merged!

PR #1878, opened 2019-06-11
Move AppDataDirectory.java to o.b.utils in core

msgilligan:
This will allow usage in both `wallettemplate` and `tools` and to replace
similar code in `BlockFileLoader.java` in `core`.
Required minor back-port of `Path.of()` functionality from JDK11.

### DIFF
--- a/core/src/main/java/org/bitcoinj/utils/AppDataDirectory.java
+++ b/core/src/main/java/org/bitcoinj/utils/AppDataDirectory.java
@@ -14,16 +14,25 @@
  * limitations under the License.
  */
 
-package wallettemplate.utils;
+package org.bitcoinj.utils;
 
 import org.bitcoinj.core.Utils;
 
 import java.io.IOException;
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
 /**
  * Find/create App Data Directory in correct platform-specific location.
+ * This class is based on the conventions used in Bitcoin Core which
+ * uses the following locations:
+ * <dl>
+ *     <dt>Windows</dt><dd>{@code ${APPDATA}/.bitcoin}</dd>
+ *     <dt>macOS</dt><dd>{@code ${HOME}/Library/Application Support/Bitcoin}</dd>
+ *     <dt>Linux</dt><dd>{@code ${HOME}/.bitcoin}</dd>
+ * </dl>
+ * Note that {@code appName} is converted to lower-case on Windows and Linux/Unix.
  */
 public class AppDataDirectory {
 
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/bitcoinj/utils/AppDataDirectoryTest.java
```diff
@@ -0,0 +1,57 @@
+package org.bitcoinj.utils;
```
schildbach: Can you add the license?

core/src/test/java/org/bitcoinj/utils/AppDataDirectoryTest.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private static String macPath(String appName) {
+        return HOMEDIR + "/Library/Application Support/" + appName;
+    }
+
+    private static String unixPath(String appName) {
+        return HOMEDIR + "/." + appName.toLowerCase();
+    }
+
```
schildbach: Surplus newline.

core/src/main/java/org/bitcoinj/utils/AppDataDirectory.java
```diff
@@ -45,19 +54,36 @@ public static Path get(String appName) {
...
+     *
+     * @param appName The name of the current application
+     * @return Path to the application data directory
+     */
+    public static Path getPath(String appName) {
         final Path applicationDataDirectory;
 
         if (Utils.isWindows()) {
-            applicationDataDirectory = Path.of(System.getenv("APPDATA"), appName);
+            applicationDataDirectory = pathOf(System.getenv("APPDATA"), appName.toLowerCase());
```
schildbach: Are these `toLowerCase()` backward compatible? Or could it perhaps happen that wallettemplate doesn't find its wallets created by an earlier version any more on case-sensitive filesystems?

msgilligan: > Are these toLowerCase() backward compatible?

No, but I'm pretty sure the AppDataDirectory stuff hasn't been released yet. And WalletTemplate isn't (yet) something we have to worry about that kind of backward-compatibility for (IMO.)

schildbach: Ok, you're right, I forgot. Still I think the `.toLowerCase()` change should be put into a separate commit because it seems unrelated to the refactoring described in this PR.

msgilligan: I'm happy to put it in a separate commit, but I do consider it's absence a bug and was shocked that I had missed it. The whole idea was to replicate the behavior of Bitcoin Core.

core/src/test/java/org/bitcoinj/utils/AppDataDirectoryTest.java
msgilligan: Yes, force push coming up....

core/src/main/java/org/bitcoinj/utils/AppDataDirectory.java
schildbach: Even more so it could be a separate bugfix that can be merged much quicker than the proposed refactoring.

msgilligan: OK, will do. The bug doesn't exist in the App Dir code in `core` because that code hardcodes `bitcoin` or `Bitcoin` as appropriate, but it does exist in the `wallettemplate` `AppDataDirectory` class...

msgilligan: Fixed in PR #1882

schildbach:
This PR could now use a rebase.

msgilligan:
@schildbach rebased and ready to go!

schildbach:
I just merged this.

PR #1881, opened 2019-06-11
Make StoredBlock member fields final

msgilligan:
They are private and are never updated, so this should be a non-breaking change.


schildbach:
Merged.

PR #1882, opened 2019-06-19
Use lower-case appName for data dir on Windows/Linux/Unix

msgilligan:
This is a bug fix, since `AppDataDirectory` was intended to match the directory names used by Bitcoin Core. Thus we need to convert `appName` to lowercase on Windows, Linux, and Unix.

See PR #1878 for previous discussion.


schildbach:
Thanks! Merged.

PR #1891, opened 2019-06-25
Implement a test for AppDataDirectory for case of invalid directory name

MethDamon:
Hey, I realized that there is no test method that tests that exceptions caused by invalid folder names are correctly catched.

Also, why is the IOException from the Files API catched and then rethrown as a RunTimeException? IMO it would be nicer to keep the checked IOException and deal with it at top level.

### DIFF
--- a/core/src/test/java/org/bitcoinj/utils/AppDataDirectoryTest.java
+++ b/core/src/test/java/org/bitcoinj/utils/AppDataDirectoryTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Michael Sean Gilligan
+ * Copyright 2019 Michael Sean Gilligan, Tim Strasser
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


msgilligan:
The PR looks fine to me.

> Also, why is the IOException from the Files API catched and then rethrown as a RunTimeException?

Because the idea is that the folder name does not come from user input or change at runtime and is chosen by the app developer to be a valid folder name. So there is no need for the developer to handle the exception.

MethDamon:
> The PR looks fine to me.
> 
> > Also, why is the IOException from the Files API catched and then rethrown as a RunTimeException?
> 
> Because the idea is that the folder name does not come from user input or change at runtime and is chosen by the app developer to be a valid folder name. So there is no need for the developer to handle the exception.

Thanks for the general approval. I think it would make sense if someone would check this out and run the tests with different environment although I did research about invalid characters in folder names on different platforms.

The exception thing makes sense to me now, thanks for elaborating!

schildbach:
Tim, thanks for your contribution! I changed your copyright to be on its own line.

Merged.

msgilligan:
> I think it would make sense if someone would check this out and run the tests with different environment although I did research about invalid characters in folder names on different platforms.

@MethDamon @schildbach I just ran this test on macOS and the test fails because macOS _does not_ throw an exception for `":"`.

schildbach:
According to https://kb.acronis.com/content/39790:

> Since Mac OS X is build on top of UNIX there are a few inherent conventions that OS 9 users may not expect. Because of this, migrating certain files and folders from OS 9 to OS X may cause unexpected behavior.
>
> The only illegal character for file and folder names in Mac OS X is the colon ":"
>
> File and folder names are not permitted to begin with a dot "."
>
> File and folder names may be up to 255 characters in length

@msgilligan Could you try if ".a" throws on MacOS?

msgilligan:
That doesn't sound right to me -- I have tons of dotfiles and dotdirs on macOS. I'll try it (and a few other options) and see if I can find something that throws on the Mac.

I don't feel this is the greatest test, because it's really only testing Java and the underlying OS, not our code. But I'll try to fix it on macOS anyway.

msgilligan:
OK, see PR #1911 for a fix.

PR #1892, opened 2019-06-26
enhanced test coverage

juanse254:
complete branch test coverage for Fiat Test

### DIFF
--- a/core/src/test/java/org/bitcoinj/utils/FiatTest.java
+++ b/core/src/test/java/org/bitcoinj/utils/FiatTest.java
@@ -17,8 +17,7 @@
 package org.bitcoinj.utils;
 
 import static org.bitcoinj.utils.Fiat.parseFiat;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/bitcoinj/utils/FiatTest.java
```diff
@@ -105,6 +105,7 @@ public void testValueFetching() {
         Fiat fiat = parseFiat("USD", "666");
         assertEquals(6660000, fiat.longValue());
         assertEquals("6660000", fiat.toString());
+        assertEquals(206545287, fiat.hashCode());
```
schildbach: This seems like overtesting to me. It's not important that `.hashCode()` yields a specific value – only that it stays at the same value if the object stays "equal".

At the very least, such a test should go to its own test method. It has got nothing to do with "value fetching".

core/src/test/java/org/bitcoinj/utils/FiatTest.java
```diff
@@ -84,6 +83,7 @@ public void testToPlainString() {
     public void testComparing() {
         assertTrue(parseFiat("EUR", "1.11").isLessThan(parseFiat("EUR", "6.66")));
         assertTrue(parseFiat("EUR", "6.66").isGreaterThan(parseFiat("EUR", "2.56")));
+        assertEquals(parseFiat("EUR", "0.80").compareTo(parseFiat("USD","0.80")), -16);
```
schildbach: With `assertEquals()`, the expected value comes first. Assuming you exchanged the who arguments and `-16` is the expected value: why do you expect -16?

schildbach:
@juanse254 Do you want to comment on my review?

PR #1893, opened 2019-06-26
Implement a (partial) test for BlockFileLoader

MethDamon:


core/src/test/java/org/bitcoinj/utils/BlockFileLoaderTest.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright 2019 Tim Straser.
```
schildbach: Is your last name spelled correctly? (-:

core/src/test/java/org/bitcoinj/utils/BlockFileLoaderTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Test
+    public void testGetReferenceClientBlockFileListOneFile() {
+        try {
+            FileWriter fw = new FileWriter(TEST_FILE);
+            fw.write("test");
+            fw.close();
+        } catch (Exception e) {
+            System.out.println(e);
```
schildbach: Why catch this exception? If it throws, it should fail the test, no?

core/src/test/java/org/bitcoinj/utils/BlockFileLoaderTest.java
```diff
@@ -0,0 +1,83 @@
...
+    @Test
+    public void testGetReferenceClientBlockFileListMultipleFiles() {
+        try {
+            FileWriter fw = new FileWriter(TEST_FILE);
+            FileWriter fw2 = new FileWriter(TEST_FILE2);
+            fw.write("test");
+            fw2.write("test2");
+            fw.close();
+            fw2.close();
+        } catch (Exception e) {
```
schildbach: Same here.

core/src/test/java/org/bitcoinj/utils/BlockFileLoaderTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.util.List;
+
+public class BlockFileLoaderTest {
+
+    private final File TEST_DIRECTORY = new File("./test");
```
schildbach: I think we should create temp directories in `/tmp` (or similar on Windows/MacOS). Or alternatively in the `build` directory.

In any case, the directory name should be randomized, as `test` is much too likely to clash with existing directories. Something similar to what `File.createTempFile()` does.

schildbach:
@MethDamon Do you want to comment on my review?

PR #1894, opened 2019-06-27
Implement test for DownloadProgressTracker

MethDamon:


core/src/test/java/org/bitcoinj/core/listeners/DownloadProgressTrackerTest.java
```diff
@@ -0,0 +1,78 @@
...
+public class DownloadProgressTrackerTest {
+
+    @Mock
+    private Peer peer;
+
+    @Mock
+    private Block block;
+
+    @Mock
+    FilteredBlock filteredBlock;
```
schildbach: Shouldn't this be `final` too?

PR #1898, opened 2019-06-30
Move HDPath methods from HDUtils to HDPath

msgilligan:
All HDPath-related methods in HDUtils are now deprecated and delegate
to HDPath. HDPath now contains parsePath code.


msgilligan:
@schildbach I would look at this one next as it's a fairly obvious win.

schildbach:
Thanks for your guidance... merged!

PR #1901, opened 2019-07-03
Fix NPE in o.b.c.Utils runtime and os detection

msgilligan:
If `getProperty()` returns `null`, calling `.toLowerCase()` results in a `NullPointerException`. Instead
we’ll provide a default fallback value and check for `.equals(“”`) in the following `if` statement.

Note: I found the NPE when compiling/running with native-image/SubstrateVM.


schildbach:
Merged.

Issue #1899, opened 2019-07-01
DeterministicKey::toString skips creationTimeSeconds when parent key has creationTimeSeconds > 0


msgilligan:
`.toString()` in `DeterministicKey` uses the following code to skip printing `creationTimeSeconds` when it is zero:

```
        if (creationTimeSeconds > 0)
            helper.add("creationTimeSeconds", creationTimeSeconds);
```

However, the getter will return the creationTimeSeconds from the `parent` key:

```
    public long getCreationTimeSeconds() {
        if (parent != null)
            return parent.getCreationTimeSeconds();
        else
            return super.getCreationTimeSeconds();
```

`toString` should probably call the getter instead of accessing the field.

msgilligan:
@schildbach If you agree, assign this to me and I'll make a PR.

schildbach:
> toString should probably call the getter instead of accessing the field.

We could do this, but then I think we should add an info wether the value is inherited or not.

msgilligan:
I think we should do it. I was confused when debugging because the key didn't have a creation time when I expected it to have one. I'll make a PR.

msgilligan:
See PR #1903

schildbach:
Fixed via b59add47a7aba1f3bec8356b567bf0df4b036f81.

PR #1903, opened 2019-07-05
DeterministicKey#toString creationTimeSeconds fix for Issue #1899

msgilligan:
See Issue #1899

toString() output of creation time now corresponds with
getCreationTimeSeconds(), but we indicate if the creation time came
from the parent.


schildbach:
I changed it a bit and merged.

PR #1904, opened 2019-07-06
[0.15] Use Bouncy Castle's Scrypt implementation rather than the standalone library.

schildbach:
This is a backport of #1830 to the 0.15 release branch.

I'm proposing this as a PR because strictly speaking this is not a bugfix. The reason I think we should backport this is that starting on Aug 1, Android apps that are using bitcoinj 0.15 and have added the native scrypt code, [cannot publish updates to Google Play any more](https://android-developers.googleblog.com/2017/12/improving-app-security-and-performance.html).

Apps could add a 64-bit build of their native code. I've tried this myself, and while it worked recently it yielded slower execution times than the 32-bit version. And more importantly, 3 years ago I [managed to produce bad keys from passwords](https://github.com/wg/scrypt/issues/36) – not something we would want to happen in production code.

Lamdaworks Scrypt library appears to be unmaintained for years, so that's another reason to get rid of it.

Any objections?

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'com.google.protobuf'
 apply plugin: 'maven'
 apply plugin: 'eclipse'
 
-version = '0.15.3-SNAPSHOT'
+version = '0.15.3'
 archivesBaseName = 'bitcoinj-core'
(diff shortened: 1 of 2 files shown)


schildbach:
Merged.

PR #1906, opened 2019-07-09
Add basic .gitlab-ci.yml file

msgilligan:
Add a GitLab CI build using Ubuntu Bionic and OpenJDK 11

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,15 @@
+image: ubuntu:bionic
+
+variables:
+  GRADLE_VERSION: "4.10.3"
+
+before_script:
+  - apt-get update
+  - apt-get -y upgrade
+  - apt-get -y install openjdk-11-jdk gradle
+  - gradle wrapper --gradle-version=$GRADLE_VERSION --stacktrace
+
+build:
+  script:
+    - java -version
+    - ./gradlew build --stacktrace


schildbach:
What's the purpose of this? As long as PRs are opened on GitHub (and not GitLab), this CI will not run.

msgilligan:
If you mirror the repository to Gitlab it will run.

.gitlab-ci.yml
```diff
@@ -0,0 +1,21 @@
+image: ubuntu:bionic
+
+variables:
+  GRADLE_VERSION: "5.4.1"
```
schildbach: I think for now we should stay with the same Gradle version as in .travis.yml: 4.10.3

.gitlab-ci.yml
```diff
@@ -0,0 +1,21 @@
...
+before_script:
+  - apt-get update && apt-get -y upgrade
+  - apt-get -y install openjdk-11-jdk gradle
+  - gradle wrapper --gradle-version=$GRADLE_VERSION
+  - chmod +x ./gradlew
+
+build:
+  script:
+    - java -version
+    - LC_ALL=C.UTF-8 ./gradlew build
```
schildbach: Is the environment variable necessary? If yes, why? If not, can we remove the assignment?

.gitlab-ci.yml
```diff
@@ -0,0 +1,21 @@
...
+  artifacts:
+    paths:
+      - core/build
```
schildbach: What does Gitlab do with these artifacts?

.gitlab-ci.yml
msgilligan: Sure, I'll change it. (At some point we might want to do builds with multiple Gradle versions as we are doing on Travis)

.gitlab-ci.yml
msgilligan: It's needed to prevent fatal errors generating JavaDoc due to character set. There may be another, better way of resolving the issue, but I'm pretty sure if I remove this it will be broken. I'll verify.

.gitlab-ci.yml
msgilligan: Makes the available for download. (On commercial projects with private gitlab, I've used the CI-built artifacts as the official build binaries)

.gitlab-ci.yml
schildbach: Hmm, why is this not an issue locally or on Travis?

Maybe adding

```
javadoc {
    options.encoding = 'UTF-8'
}
```

to build.gradle could also help?

.gitlab-ci.yml
schildbach: Ok. I'm not sure if we want binary downloads on GitLib (or GitHub) however. Our website downloads currently point at maven.org.

.gitlab-ci.yml
schildbach: Yes, is there something like the `matrix` availble?

msgilligan:
One cool thing about GitLab CI is that you can run it locally with a command like:
```
 gitlab-runner exec docker --docker-privileged build
```

msgilligan:
@schildbach I started using GitLab on my wallet-framework branch because I wanted to run CI with Java 12 (although I've since dialed it back to 11.) You asked me to provide GitLab support as a PR against `master` -- did you have anything in particular in mind?

.gitlab-ci.yml
msgilligan: Yeah, we can turn them off if we want. I also sometimes configure it so that other output files such as logs or debug/error information is preserved. They're not super-easy to find on the GitLab site, so I'm not too worried that people will get confused by them and use them instead of the official Maven binaries.

.gitlab-ci.yml
msgilligan: There is, so I can try to add that in a subsequent PR. I just wanted to get a basic build working first.

.gitlab-ci.yml
msgilligan: Trying that now...

msgilligan:
> If you mirror the repository to Gitlab it will run.

To clarify: Not for PRs, but for any branches that are part of the mirror.

.gitlab-ci.yml
schildbach: Did it work?

.gitlab-ci.yml
schildbach: For a first step, I think I'd like to keep things simple and omit this part.

schildbach:
> You asked me to provide GitLab support as a PR against `master` -- did you have anything in particular in mind?

Nothing particular, other than I thought running the unit tests on two different CIs is desirable. And maybe in future we can use those two CIs to verify reproducible builds.

msgilligan:
I agree with you that two different CIs is desirable and I have done that on other projects. I also like the flexibility of GitLab CI's integration with Docker.

.gitlab-ci.yml
msgilligan: It didn't work. (and I got distracted by something else, sorry.)

Can we go ahead and merge this as is, and fix it in a separate PR when we find a better way?

.gitlab-ci.yml
msgilligan: Fixed in force-push.

msgilligan:
@schildbach If you're ok with the `LC_ALL=C.UTF-8` (at least for the time-being) then this PR is ready to merge.

empty-settings.gradle
```diff
@@ -0,0 +1,2 @@
+// Empty settings file to ignore subprojects
```
schildbach: This file seems to be unrelated to the PR. Can you remove it for now?

.gitlab-ci.yml
schildbach: Well, I'm not too happy with "shell foo" in build scripts but ok, for now I'm fine with it.

empty-settings.gradle
msgilligan: Actually, see line 9 of `.gitlab-ci.yml ` -- it solves the problem of the built-in Gradle not being able to process all the subproject of our `build.gradle`.

.gitlab-ci.yml
schildbach: Weird, I just reproduced your issue by removing the environment varable assignment. Then I fixed the issue by adding `javadoc.options.encoding = 'UTF-8'` to the core build.gradle, which should be identical to my above suggestion.

msgilligan: OK, I'll rebase and see what happens.

empty-settings.gradle
msgilligan: @schildbach Do yo see a better way of solving this issue?

.gitlab-ci.yml
```diff
@@ -0,0 +1,15 @@
+image: ubuntu:bionic
+
+variables:
+  GRADLE_VERSION: "4.10.3"
+
+before_script:
+  - apt-get update && apt-get -y upgrade
```
schildbach: This line could be split into two lines for easier reading.

.gitlab-ci.yml
```diff
@@ -0,0 +1,15 @@
...
+  - apt-get -y install openjdk-11-jdk gradle
+  - gradle -c empty-settings.gradle wrapper --gradle-version=$GRADLE_VERSION --info --stacktrace
+  - chmod +x ./gradlew
```
schildbach: This line isn't necessary; `gradlew` runs without.

.gitlab-ci.yml
```diff
@@ -0,0 +1,15 @@
...
+
+build:
+  script:
+    - java -version
+    - LC_ALL=C.UTF-8 ./gradlew build --info --stacktrace
```
schildbach: I think we should remove the `--info` from here and above; it creates very long output that GitLab can only hardly handle.

empty-settings.gradle
schildbach: Somehow I don't see the problem. For example see https://gitlab.com/bitcoinj/bitcoinj/-/jobs/260248387, it has run all the subprojects despite me having removed the empty-settings.gradle. Or am I missing something?

.gitlab-ci.yml
msgilligan: Yeah, `javadoc.options.encoding = 'UTF-8'` fixes the build failure (don't recall exactly why my previous attempt at setting javadoc options failed.

However, the test output still has character set issues, so maybe we need to add one more setting in `build.gradle`. Also, we probably want to remove/replace the `--info` parameter to the two Gradle commands.

empty-settings.gradle
msgilligan: OK, well that's cool. I probably changed something else that made the issue go away. It's likely that I had some other changes to the build files that weren't compatible with the gradle installed with apt-get. We may run into that in the future. The problem is that running `gradle wrapper` on a project requires the current version of gradle to be able to successfully process the root `build.gradle` and all the subprojects.

I'll go head and remove it.

.gitlab-ci.yml
msgilligan: OK.

.gitlab-ci.yml
msgilligan: OK.

.gitlab-ci.yml
msgilligan: Agree, once everything else is debugged we should definitely remove it.

.gitlab-ci.yml
msgilligan: Removed.

.gitlab-ci.yml
msgilligan: OK, everything seems to be working -- both locally with Docker/gitlab-runner on macOS and on hosted Gitlab. So I've removed `--info`.

.gitlab-ci.yml
msgilligan: > However, the test output still has character set issues

I was seeing `???` in the logging timestamps from the tests when running with Docker and `gitlab-runner` under macOS, but this is no longer happening.

We still might want to figure out how to set UTF-8 as the default character set at the system or Gradle level rather than task by task, but we can worry about that later if we need to.

msgilligan:
@schildbach OK, all issues resolved. Should be ready to merge.

msgilligan:
Travis failure looks like one of our normal intermittent failures...

schildbach:
Merged! Thanks a lot for your patience with me.

msgilligan:
> Thanks a lot for your patience with me.

It's a much better PR because of your work and feedback!

PR #1907, opened 2019-07-10
Added SPVBlockStore indexing

fireuse:
I have added indexing to SPVBlockStore so it will be faster when getting blocks.

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
@@ -43,7 +43,7 @@ public class SPVBlockStore implements BlockStore {
     /** The default number of headers that will be stored in the ring buffer. */
     public static final int DEFAULT_CAPACITY = 5000;
     public static final String HEADER_MAGIC = "SPVB";
-
+    private static final Map<Sha256Hash, Integer> blockIndex = new LinkedHashMap<>();
     protected volatile MappedByteBuffer buffer;
     protected final NetworkParameters params;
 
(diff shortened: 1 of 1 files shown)


schildbach:
Do you have a specific usecase in mind for which this is important?

I'm a bit reluctant about changes to SPVBlockStore since so many walles depend on it.

fireuse:
Yes, when you need to access block that isn't cached. In my solution i don't need to iterate over all block store to find block, so it is faster.

schildbach:
Yes, sure, but when is this necessary?

fireuse:
For example in blockchains explorers, or when looking at old transactions in wallet.

schildbach:
SPVBlockStore is not suitable for these usecases! It only covers the last weeks worth of blocks, never an entire block chain. For block explorer type scenarios, I suggest using a real database (e.g. PostgresFullPrunedBlockStore).

fireuse:
But even if you access block from yesterday, it will be faster because that block isn't cached. For example when you are getting list of yesterday's transactions it will be faster.

schildbach:
SPVBlockStore is simply not designed for historic queries, and certainly not on a transaction basis. SPVBlockStore is designed for the needs of SPV (simply payment verification) implementations. And SPV only goes back in history if a rollback occurs. For that event, the current caching is fine.

fireuse:
But assume we are opening wallet and printing last 20 transactions, transactions will be in uncached blocks(because we just opened wallet). In old version this app would load a few seconds longer.
And i am not removing current caching but i am extending it.

schildbach:
What wallet implementation are you talking about? org.bitcoinj.wallet.Wallet?

Why should a wallet access the block store? Wallets usually store their relevant transactions and/or their relevant UTXOs themselves. If they would rely on SPVBlockStore, they would not be able to access transactions older than a few weeks.

fireuse:
Ok so I am closing this pull request.

PR #1910, opened 2019-07-20
Improve JavaDoc for Wallet#commitTx, Wallet#maybeCommitTx

msgilligan:
It looks like comments were written when `maybeCommitTx` called  `commitTx`, not the other way around.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -2613,9 +2613,17 @@ public class Wallet extends BaseTaggableObject
     }
 
     /**
-     * Calls {@link Wallet#commitTx} if tx is not already in the pending pool
+     * Updates the wallet with the given transaction: puts it into the pending pool, sets the spent flags and runs
+     * the onCoinsSent/onCoinsReceived event listener.
+     * <p>
+     * Triggers an auto save (if enabled.)
+     * <p>
+     * Unlike {@link Wallet#commitTx} this method does not throw an exception if the transaction
+     * was already added to the wallet, instead it will return {@code false}
      *
+     * @param tx transaction to commit
      * @return true if the tx was added to the wallet, or false if it was already in the pending pool
+     * @throws VerificationException If transaction fails to verify
      */
     public boolean maybeCommitTx(Transaction tx) throws VerificationException {
         tx.verify();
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -2613,9 +2613,17 @@ private void maybeMovePool(Transaction tx, String context) {
     }
 
     /**
-     * Calls {@link Wallet#commitTx} if tx is not already in the pending pool
+     * Updates the wallet with the given transaction: puts it into the pending pool, sets the spent flags and runs
+     * the onCoinsSent/onCoinsReceived event listener.
```
schildbach: Are those listeners really called if the tx was already received?

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -2613,9 +2613,17 @@ private void maybeMovePool(Transaction tx, String context) {
...
+     * <p>
+     * Triggers an auto save.
```
schildbach: Note: auto-saving needs to be enabled.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -2703,15 +2711,22 @@ public boolean maybeCommitTx(Transaction tx) throws VerificationException {
...
+     * the onCoinsSent/onCoinsReceived event listener. Used in two situations:
+     * <p>
      * <ol>
      *     <li>When we have just successfully transmitted the tx we created to the network.</li>
      *     <li>When we receive a pending transaction that didn't appear in the chain yet, and we did not create it.</li>
      * </ol>
+     * <p>
+     * Triggers an auto save.
+     * <p>
+     *
```
schildbach: Surplus empty line.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: I added an "(if enabled)" to make this more clear.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: Fixed in force push.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: > Are those listeners really called if the tx was already received?

No, but don't the comments about throwing an exception or returning false cover that case. What do you suggest we say?

msgilligan: @schildbach Does my comment above answer your question? What are you looking for here?

schildbach: Somehow I don't see this. If I read from top to bottom, I'm under the impression that the actions described in this sentence happen in any case. But tthe method is a maybe*() method, so there is the possibility of a no-op (including the side effects like firing listeners).

I'm not sure if it's worth mentioning the listeners at all here. After all, they can be called in various situations.

msgilligan: I'll look at this over the weekend. Need to get back to other tasks for today...

msgilligan: Long weekend 😏 I'll get back on it soon.

schildbach:
I just merged this. Thanks!

PR #1914, opened 2019-07-24
Put shared dependency versions in gradle.properties

msgilligan:
Use common setting in `gradle.properties` for `slf4jVersion` and `guavaVersion`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,3 @@
+guavaVersion = 27.1-android
+slf4jVersion = 1.7.25
+
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -9,16 +9,16 @@ version = '0.16-SNAPSHOT'
 
 dependencies {
     compile 'org.bouncycastle:bcprov-jdk15on:1.60'
-    implementation 'com.google.guava:guava:27.1-android'
+    implementation "com.google.guava:guava:${guavaVersion}"
     compile 'com.google.protobuf:protobuf-java:3.7.1'
     implementation 'com.squareup.okhttp3:okhttp:3.12.3'
-    implementation 'org.slf4j:slf4j-api:1.7.26'
+    implementation "org.slf4j:slf4j-api:${slf4jVersion}"
     implementation 'net.jcip:jcip-annotations:1.0'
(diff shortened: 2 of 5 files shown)


msgilligan:
@schildbach I just rebased to `master` and resolved conflicts with PR #1913. I know you were reluctant to merge PR #1780 in part because of the 1-line `gradle.properties`, but putting project and dependency versions in `gradle.properties` is a standard (arguably best) practice rather than in gradle (groovy language) files as they are declarative and easier to manage/update. They are separate PRs, but please consider the use of `gradle.properties` for these two purposes (and other similar purposes in the future.)

msgilligan:
@schildbach What are your thoughts on this PR and related #1780 ?

schildbach:
Generally, I don't like another indirection. I prefer keeping things as simple as possible.

The one shared gradle.properties you're suggesting is tying the projects together. But they are kind of separate, even though for convenience they live in the same repo right now.

gradle.properties
```diff
@@ -0,0 +1,3 @@
+guavaVersion = 27.1-android
```
schildbach: Could the "dot notation" be used here? So `guava.version`?

msgilligan: I believe it could, but I've never seen it done for "custom" variables that are referenced in `build.gradle`. I believe it would then have to be accessed as `project.properties['guava.version']`

msgilligan:
> I prefer keeping things as simple as possible.

I think it's simpler to have one setting for the current version of dependencies, it makes it easier to update and make sure all subprojects get the same version (which is almost always what we want and is over-ridable if it isn't what we want) and it also cuts down on merge conflicts between branches, but I see your point.

>  But they are kind of separate, even though for convenience they live in the same repo right now.

I actually like having multiple subprojects in the same repo as I find it's easier to refactor, run unit tests etc between the projects. I know this can be done with Maven repos and CI, but that's harder to set up and maintain.

In any event, it's your project and I'm ok with whatever you decide.

msgilligan:
Closing to clear out backlog. If we want to do this in the future, we can open a new PR.

msgilligan:
I opened Issue #4128 suggesting that we reconsider using a `gradle.properties` file.

PR #1919, opened 2019-08-06
Bump SLF4J to 1.7.27 (adds Automatic-Module-Name in MANIFEST.MF)

msgilligan:
1.7.27 adds Automatic-Module-Name in MANIFEST.MF,
so for JPMS we have:

```
JAR name          Module name
slf4j-api	  org.slf4j
slf4j-jdk14	  org.slf4j.jul
```
See: https://jira.qos.ch/browse/SLF4J-420


schildbach:
Merged.

Is the info about the module names important to bitcoinj developers somehow? If yes I'd suggest to add some README that lists all those module names of deps.

msgilligan:
> Is the info about the module names important to bitcoinj developers somehow? If yes I'd suggest to add some README that lists all those module names of deps.

They'll need the module names if they are writing JPMS modules. Rather than add documentation about the modules/names of our dependencies the best way to provide that information would be in a `module-info.java` file.  A `module-info.java` would explain which dependencies are used, what is exported, etc. and would be verified by the build process. Of course, we can't put the compiled `module-info.class` file in a Java 7 byte code format JAR, but we could use a multi-release jar to do that.

schildbach:
Unfortunately, we cannot use multi-release jars in the JDK8 build, because the Android toolchain chokes on it.

So the `module-info.class` would for now be restricted to the JDK11 build, and it would not appear in the release JAR. Do you think it could still be useful, e.g. for people compiling from source?

PR #1922, opened 2019-08-09
Remove bitcoinj logos (and the misc folder).

schildbach:
We have copies in the webpage repository where they belong.


schildbach:
Merged.

PR #1926, opened 2019-08-26
Add application plugin to tools subproject

msgilligan:
* Enable and configure Gradle application plugin
* Add wallet_tool instructions in project README

Note: I tested this on macOS with Java 8/Gradle 4.4 and with Java 11/Gradle 5.6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,38 @@ gradle clean assemble
 
 Alternatively, just import the project using your IDE. [IntelliJ](http://www.jetbrains.com/idea/download/) has Gradle integration built-in and has a free Community Edition. Simply use `File | New | Project from Existing Sources` and locate the `build.gradle` in the root of the cloned project source tree.
 
+### Building and Using the Wallet Tool
+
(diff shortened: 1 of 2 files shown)


msgilligan:
@schildbach Are you around? Can you take a quick look at this?

msgilligan:
@schildbach Are you ok with merging this? The main point of this change is to build a `wallet_tool` shell script that can be easily run directly from the command line.

msgilligan:
@schildbach This one is pretty straightforward -- I'm hoping you can find time to review and hopefully merge it.

schildbach:
You're right. I renamed the command to `wallet-tool` (with a dash) and also removed the old artifacts for using the command line.

Since the application plugin seems to require one subproject per application, I think we should separate wallet-tool from build-checkpoints, and maybe move the tool-specific instructions into READMEs at subproject level.

Merged. And sorry for the delay.

msgilligan:
> Since the application plugin seems to require one subproject per application, I think we should separate wallet-tool from build-checkpoints, and maybe move the tool-specific instructions into READMEs at subproject level.

That all makes sense. My goad is to get a command-line wallet tool built to a native executable with the Graal native-image command. Graal has come a long way since my last attempt, so I'm looking forward (to finding the time) to try again.

Issue #800, opened 2015-06-11
bitcoinj-examples: maven-enforcer-plugin failure


mikehearn:
Originally reported on Google Code with ID 434

```
This is on the proposed release of bitcoinj 0.10, commit 4a480829244eafd0749fc03206d6b6bf66b08afd.

> mvn clean install

[...]
[INFO] --- maven-enforcer-plugin:1.2:enforce (enforce) @ bitcoinj-examples ---
[INFO] Applying DigestRule
[INFO] Building snapshot whitelist of all current artifacts
[WARNING] Artifact com.google:bitcoinj:0.10:jar:null:compile UNVERIFIED SHA1 (missing
in repo).
[INFO] List of verified artifacts. If you are confident in the integrity of your repository
you can use the list below:
[INFO] <urns>
[INFO]   <urn>com.google.code.findbugs:jsr305:1.3.9:jar:null:compile:40719ea6961c0cb6afaeb6a921eaa1f6afd4cfdf</urn>
[INFO]   <urn>com.google.guava:guava:13.0.1:jar:null:compile:0d6f22b1e60a2f1ef99e22c9f5fde270b2088365</urn>
[INFO]   <urn>com.google.protobuf:protobuf-java:2.4.1:jar:null:compile:0c589509ec6fd86d5d2fda37e07c08538235d3b9</urn>
[INFO]   <urn>com.h2database:h2:1.3.167:jar:null:compile:d3867d586f087e53eb12fc65e5693d8ee9a5da17</urn>
[INFO]   <urn>com.lambdaworks:scrypt:1.3.3:jar:null:compile:06d6813de41e177189e1722717979b4fb5454b1d</urn>
[INFO]   <urn>com.madgag:sc-light-jdk15on:1.47.0.2:jar:null:compile:d5c98671cc97fa0d928be1c7eb5edd3fb95d3234</urn>
[INFO]   <urn>io.netty:netty:3.6.3.Final:jar:null:compile:1eebfd2f79dd72c44d09d9917c549c60322462b8</urn>
[INFO]   <urn>net.jcip:jcip-annotations:1.0:jar:null:compile:afba4942caaeaf46aab0b976afd57cc7c181467e</urn>
[INFO]   <urn>org.apache.maven.plugins:maven-clean-plugin:2.3:maven-plugin:null:runtime:5cd0c5013640d238a941465520712e2a7610a320</urn>
[...]

PR #1931, opened 2019-09-09
Add checksum-dependency-plugin for verification of plugin/dependency checksums

vlsi:
`checksum-dependency-plugin` is a superset of `gradle-witness`, and it enables to increase the level of security.

See https://github.com/vlsi/vlsi-release-plugins#checksum-dependency-plugin

Key features:

* Gradle plugins can be verified (grade-witness doesn't track plugins)
* All Gradle configurations are supported (e.g. `java-library` plugin is supported). `checksum-dependency-plugin` intercepts detached configurations as well (e.g. the ones that are created on demand)
* PGP can be used for verification. PGP can be used with or without checksum. PGP enables to detect and prevent issues like https://blog.autsoft.hu/a-confusing-dependency/

Note: current PR adds uses just PGP verification if library has PGP signature.

`trust-requirement pgp='GROUP' checksum='NONE'` can be adjusted to match your expectations. For instance, `checksum='MODULE'` would require SHA-512 to be declared for all the dependencies. It would increase security, however it would make it more complicated to update dependencies.

See https://github.com/bitcoinj/bitcoinj/issues/800#issuecomment-480195290


msgilligan:
Thanks for this @vlsi ! Did you see the Travis [failure](https://travis-ci.org/bitcoinj/bitcoinj/jobs/582573250) in the Gradle 4.4.1 build?

vlsi:
@msgilligan , well.
I've seen that, however I was not sure if you are willing to merge this kind of PR.

I've created a blogpost (see https://medium.com/@vladimirsitniko/dependency-verification-checksum-vs-pgp-582e76207019?sk=7485298b76eaf9f935b899b002f4c3b5 ), and I've filed a bunch of PRs, however some of the maintainers decline the verification (e.g. https://github.com/mvmike/min-cal-widget/pull/44#issuecomment-529904381 )

Technically speaking, I was not targeting Gradle 4.4.1, and I could probably make it Gradle 4.4.1 compatible. Alternative option is to just disable the plugin for "old Gradle versions" by wrapping `apply plugin` with `if` in `settings.gradle`

vlsi:
Travis test passes.

msgilligan:
I'm not a maintainer, the decision will be made by @schildbach, but I'm generally supportive of this approach.

vlsi:
Hi, I see the PR has not been reviewed much yet :)

You might be interested that [Gradle 6.2 introduces in-core dependency verification](https://github.com/gradle/gradle/issues/10443#issuecomment-568741472)

The documentation can be reviewed here: https://github.com/gradle/gradle/pull/11755

So, for now, the options are like "upgrade to Gradle 6.2 and use in-core verification" or "use checksum-dependency-plugin.


From what I know Gradle would cover more cases when compared with `checksum-dependency-plugin`. For instance, it will be able to verify `pom.xml` which are implicitly fetched by Gradle when resolving transitive dependencies and probably other cases.

Some bits can be previewed in the current release candidates, release nightly builds and master nightly builds (see https://gradle.org/releases/ )

It would be nice if you could preview the feature and provide your feedback.

PR #1934, opened 2019-10-01
Fix javadoc for JDK 13

msgilligan:
Fix "heading used out of sequence" error in BtcFormat.java

This was the only thing preventing a build under JDK 13 from completing successfully.


schildbach:
That's great to hear!

Merged.

msgilligan:
> This was the only thing preventing a build under JDK 13 from completing successfully.

Well there seems to be some weird/intermittent problem running Gradle on JDK 13 that I hadn't run in to when I said the above.

PR #1938, opened 2019-10-08
Migrate from protobuf-java to protobuf-javalite

schildbach:
This is a preview of how a migration to the Protobuf Java Lite Runtime could look like.

There are some failing (now disabled) tests because of the textual protobuf format has changed. Those need to be fixed and enabled again.

The Java Lite Runtime is now recommended by Google for usage on Android.


HashEngineering:
While investigating a similar change to a `bitcoinj` fork (`dashj`), we have used this PR as a guide and found many tests failing.

While the full version of protobuf would yield this:
```
type: DETERMINISTIC_KEY
public_key: "\0029\373\030u\305\214S\345/\373y%\t\252~\267\f\016t|\354\020\356\306\313\317\027\325\376\232kh"
deterministic_key {
  chain_code: "\260\203\277\231\352\265y\020\356r_bO\374l\347\002\032i\216Ct\260\221-\207\200\243\364;\247I"
  path: 2147483692
  path: 2147483649
  path: 2147483648
  path: 0
  path: 8
}
```
The lite version yields this:
```
type: DETERMINISTIC_KEY
# org.bitcoinj.wallet.Protos$Key@d996fb3e
public_key: "\0029\373\030u\305\214S\345/\373y%\t\252~\267\f\016t|\354\020\356\306\313\317\027\325\376\232kh"
```

One person on Stack Exchange was asking about this issue:
https://stackoverflow.com/questions/40182623/android-protobuf-lite-tostring-not-working

> protobuf-lite does not support toString(). It can't, because TextFormat relies on protobuf reflection and descriptors, which are exactly the features that are omitted in lite mode. You will need to build against regular (non-lite) protobufs for this to work.

And from https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Message:

> See also MessageLite, which defines most of the methods that typical users care about. Message adds to it methods that are not available in the "lite" runtime. The biggest added features are introspection and reflection -- i.e., getting descriptors for the message type and accessing the field values dynamically.

Message has a toString method, while MessageLite does not.  This probably explains the above example from a test failure.

schildbach:
In my description, I should have added: We use this toString() output only for debug/testing purposes, so actual format is not that important. But our tests depend on that format currently, so we need to somehow convert the test. We could either

1) trust that the new protoful library does the right thing and just copy its output in place of the old test data, or
2) write some script to convert the old test data into the new format, or
3) compare everything manually (unfeasible), or
4) a combination of the above.

But in the end, I'd like to see the migration to protobuf-lite happening.

HashEngineering:
The current format seems to be as follows:

1. items are printed in order they are declared in wallet.proto.

```
type: DETERMINISTIC_KEY
public_key: "\002\357\\\252\376]\023\315\'\316`\317\362\032@\232\"\360\331\335\221] `\016,\351<\b\300\225\032m"
creation_timestamp: 100000000
deterministic_key {
  chain_code: "\370\017\223\021O?.@gZ|\233j\3437\317q-\241!\177J \323\'\264s\203\314\321\v\346"
  path: 2147483648
}
output_script_type: P2PKH
```

The new format is as follows:

1. Items are printed in alphabetical order
2. `# org.bitcoinj.wallet.Protos$Key@41bf1166` is part of the result and the value after the @ will be different every time.

```
# org.bitcoinj.wallet.Protos$Key@41bf1166
creation_timestamp: 100000000
deterministic_key {
  chain_code: "\370\017\223\021O?.@gZ|\233j\3437\317q-\241!\177J \323\'\264s\203\314\321\v\346"
  path: -2147483648
}
output_script_type: P2PKH
public_key: "\002\357\\\252\376]\023\315\'\316`\317\362\032@\232\"\360\331\335\221] `\016,\351<\b\300\225\032m"
type: DETERMINISTIC_KEY
```

The code related to creating the string appears to be here:
https://github.com/protocolbuffers/protobuf/blob/9b0fa3e7fbaf3463c3afa1c9ab111682108eeadb/java/core/src/main/java/com/google/protobuf/MessageLiteToString.java#L58

```java
  /**
   * Returns a {@link String} representation of the {@link MessageLite} object. The first line of
   * the {@code String} representation representation includes a comment string to uniquely identify
   * the object instance. This acts as an indicator that this should not be relied on for
   * comparisons.
   *
   * <p>For use by generated code only.
   */
  static String toString(MessageLite messageLite, String commentString) {
    StringBuilder buffer = new StringBuilder();
    buffer.append("# ").append(commentString);
    reflectivePrintWithIndent(messageLite, buffer, 0);
    return buffer.toString();
  }
```

Perhaps a solution is to use #2 (write some script to convert the old test data into the new format) where existing data files would be changed to follow the new sort order (alphabetical) and the line starting with `#` from the new format should be removed.  Changing the sort order of the new format looks to be very difficult.

HashEngineering:
Another difference was found in "deterministic-wallet-bip44-serialization.txt" but may be in other test files also.

In the "deterministic-wallet-bip44-serialization.txt" file, `account_path` ChildNumbers are seen as positive numbers. This file goes with the `serializeUnencryptedBIP44` test.

```
type: DETERMINISTIC_MNEMONIC
secret_bytes: "aerobic toe save section draw warm cute upon raccoon mother priority pilot taste sweet next traffic fatal sword dentist original crisp team caution rebel"
creation_timestamp: 1389353062000
deterministic_seed: "E\032\356\206\230,\275\263\364=\334^f\307\037\350\321X7R\262z\205\3564\371tp\2639R\342\027 J\266\253\250\320\022\031\233\271~O$\330\260\214\fz\231tI\353\215*\037\355\205\213.\224?"
account_path: 2147483692
account_path: 2147483649
account_path: 2147483648
```
However, with the javalite output, the `account_path` ChildNumbers are negative.

```
# org.bitcoinj.wallet.Protos$Key@999ded1c
account_path: -2147483604
account_path: -2147483647
account_path: -2147483648
creation_timestamp: 1389353062000
deterministic_seed: "E\032\356\206\230,\275\263\364=\334^f\307\037\350\321X7R\262z\205\3564\371tp\2639R\342\027 J\266\253\250\320\022\031\233\271~O$\330\260\214\fz\231tI\353\215*\037\355\205\213.\224?"
secret_bytes: "aerobic toe save section draw warm cute upon raccoon mother priority pilot taste sweet next traffic fatal sword dentist original crisp team caution rebel"
type: DETERMINISTIC_MNEMONIC
```

It seems as if the full `protobuf-java` format renders `account_path` as `long` values, where `protobuf-javalite` renders these numbers as `int` since the hardened bit becomes the sign bit.

HashEngineering:
I was able to make a script that converts the current format to the lite format as well as modify the lite format to remove the unstable part of the string.

```python
# convert protobuf-java text format to protobuf-javalite
# syntax:
# reformat.py file
#
# This command will reformat and write the output to file.lite
import sys

file = open(sys.argv[1])
outfile = open(sys.argv[1] + ".lite", "w")
line = file.readline()
lines = []


# Sorts a group alphabetically then prints to the outfile
def sortAndPrint(group, removeLastLB=False, removeFirstLB=False):
    group.sort()
    if removeFirstLB and group[0] == "\n":
        group.remove("\n")
    for i in range(len(group)):
        if i == len(group) - 1 and removeLastLB:
            group[i] = group[i].rstrip()
        print(group[i], end="", file=outfile)


# converts integers to protobuf-javalite format
def sign32neg(value):
    if 0x80000000 <= value <= 0xFFFFFFFF:
        value &= 0x7FFFFFFF
        value = int(value)
        value = ~value
        value ^= 0x7FFFFFFF
    return value


# converts a path to protobuf-javalite format 
def processPath(path):
    parts = path.rstrip().split(":")
    number = int(parts[1].lstrip())
    if number >= 0x80000000:
        number = sign32neg(number)
    return parts[0] + ": " + str(number) + "\n"


groupcount = 0
while len(line) > 0:
    if line.startswith("type:"):
        if len(lines):
            sortAndPrint(lines, removeFirstLB=groupcount==0)
            groupcount += 1
        lines.clear()
        lines.append(line)
    elif line.find("deterministic_key {") != -1:
        structure = line
        innerlines = []
        line = file.readline()
        while not line.startswith('}'):
            if line.find("path") != -1:  # preserve order of paths
                pathlines = processPath(line)
                line = file.readline()
                if line.startswith('}'):
                    innerlines.append(pathlines)
                    continue
                while line.find("path") != -1:
                    pathlines += processPath(line)
                    line = file.readline()
                innerlines.append(pathlines)
                if line.startswith('}'):
                    continue
            if line.startswith("  sigsRequiredToSpend"):
                line = line.replace("sigsRequiredToSpend", "sigs_required_to_spend")
            innerlines.append(line)
            line = file.readline()
        innerlines.sort()
        for l in innerlines:
            structure += l
        structure += line.rstrip() + "\n"
        lines.append(structure)
    elif line.find("path") != -1:  # preserve order of paths
        accountpathlines = processPath(line)
        line = file.readline()
        while line.find("path") != -1:
            accountpathlines += processPath(line)
            line = file.readline()
        lines.append(accountpathlines)
        lines.append(line)
    else:
        lines.append(line)
    line = file.readline()
# print the last group
print(file=outfile)
sortAndPrint(lines, removeLastLB=True)
# close open files
file.close()
outfile.close()
```

The `protoToString` method in DeterministicKeyChainTest must also be modified to replace lines that start with `#` with a line break to produce a stable string.
```java
    private String protoToString(List<Protos.Key> keys) {
        StringBuilder sb = new StringBuilder();
        for (Protos.Key key : keys) {
            sb.append(key.toString());
            sb.append("\n");
        }
        int hashTag = sb.indexOf("# org");
        while(hashTag != -1) {
            int endOfLine = sb.indexOf("\n", hashTag);
            sb.replace(hashTag, endOfLine+1, "\n");
            hashTag = sb.indexOf("# org");
        }
        return sb.toString().trim();
    }
```

schildbach:
Wow, thanks for your effort! I wonder if we have a test for the problematic index values at API level (rather than the `toString()`). Because if these values were deserialized incorrectly, that would surely be desastrous.

Your new `protoToString()` could probably be more elegantly implemented using `BufferedReader.readLine()`.

schildbach:
Btw. when I opened this PR I smoke-tested my app with the new library and I didn't notice any problems with serializing and deserializing wallets.

HashEngineering:
I think internally `protobuf-java` and `protobuf-javalite` are using the same format for the path `ChildNumbers` which is `int`.  What differs is how these are rendered with `toString()`

The code for rendering an integer in TextFormat (protobuf-java):
```
public static String unsignedToString(int value) {
    return value >= 0 ? Integer.toString(value) : Long.toString((long)value & 4294967295L);
}
```

The code that renders an integer in `protobuf-javalite`. So `account_path` or `path` is effectively `Integer.toString(value)`
```
buffer.append(": ").append(object.toString());
```

`0H` =  `0x80000000` (`int`) is rendered as ‭`2147483648‬` (`0x0000000080000000` as `long`) by `protobuf-java`
`0H` =  `0x80000000` (`int`) is rendered as ‭`-2147483648‬` (`0x80000000` as `int`) by `protobuf-javalite`

In both cases `0H` is not rendered in a human readable format.

maxxx:
any news for this? Android Firebase ('com.google.firebase:firebase-inappmessaging-display-ktx' and maybe more) depends on the lite version and conflicts with non-lite.

mpetlyuk:
Basically a lot of firebase dependencies are rely on protobuf-javalite. For now I don't see how to combine all android staff with bitcoinj. Any ideas? I am pretty sure that excluding **protobuf-java** from bitcoinj or excluding **protobuf-javalite** from firebase dependencies is wrong way. Anyone found working solution?

schildbach:
I just rebased this on current master.

@HashEngineering Are you still working on this, or do you have any not-yet published progress? If not, I guess I'll try to reproduce your work so we can adapt the unit tests with confidence.

HashEngineering:
@schildbach In my fork of this project for Dash, we migrated to `protobuf-javalite` about 1 year ago and didn't have any problems with the mobile apps that use it.  Also this change elimated issues with using firebase components.  Some tests that were related to PaymentChannels are failing in `dashj`, but I think you removed this from `bitcoinj` some time ago.

Here is a commit that has my changes, which includes modified test files. 
https://github.com/dashevo/dashj/commit/c709ff0611815439487d72aa9ce356e5de06135b

It seems that in the test files that are serialization of wallets, the only differences were how integers are rendered, which is discussed above, as well the rendering of ByteStrings.

mpetlyuk:
@schildbach does it mean that we can expect this PR in the master branch? Just wondering about terms of resolving this query, and what will be the better way for my team - try to move with fork or wait for changes here.

schildbach:
I would like to have it in master soon, so that it can get into the 0.16 release.

However, up to now I wasn't aware there is a conflict between protobuf-java and protobuf-javalite. It could be a showstopper for projects using protobuf-java other than via bitcoinj. Not sure how to deal with this other than trying it out.

msgilligan:
I have two possible (but non-trivial) solutions to this problem:

1. Use a "shadowing" tool to "roll up" the dependency to be internal to **bitcoinj**. This could also be useful for Guava and perhaps some of the other dependencies. I'm not sure what this would mean for use of protobuf types in the API.
2. Create two "variants" of **bitcoinj** using the Gradle variants feature -- one using protobuf and the other protobuf-lite.

schildbach:
I've added a second commit with all @HashEngineering 's changes I could find. However, there are still two segwit test datasets (.txt files) which are not contained in the dashj repo.

mpetlyuk:
> 2\. Create two "variants" of **bitcoinj** using the Gradle variants feature -- one using protobuf and the other protobuf-lite.

it sounds like the most right way to have compatibility for all target users of bitcoinj.

schildbach:
I added another commit with the two remaining test datasets. However those two are simply re-generated, without audit like the others.

schildbach:
> I have two possible (but non-trivial) solutions to this problem:

Somehow I think both of these solutions introduce too much complexity. At least I'd like to know if there actually *is* someone with this problem.

schildbach:
I can assert that the Python3 script posted in https://github.com/bitcoinj/bitcoinj/pull/1938#issuecomment-624823070 produces the same output as what is expected by the tests, and what was contributed via https://github.com/dashevo/dashj/commit/c709ff0611815439487d72aa9ce356e5de06135b.

If someone wants to review the Python script that would be great. In any case, I guess I will squash everything, add the script into the commit description and merge. Any objections?

schildbach:
This is now ready for final review.

HashEngineering (approved):

schildbach:
Merged!

mpetlyuk:
Any information when do you plan to prepare new release? @schildbach @msgilligan 
P.S. Thank you guys for resolving really important query from the side of all android developers!

schildbach:
The plan is to release before Taproot activates (~Nov).

PR #1940, opened 2019-10-21
ECKey: Deprecate method decompress().

schildbach:
The idea is to slowly remove API that allows to deal with uncompressed keys (except for the absolute minimum that is necessary for legacy keys).


schildbach:
I folded this in #1943.

PR #1943, opened 2019-10-23
ECKey: Clean up compress/decompress helpers

schildbach:
These commits aim at deprecating and eventually removing ECKey helpers that deal with public key compression. For normal users of this library, there should not be a reason to need this.


kasperwiese8:
write acces

schildbach:
Merged 98bc701 and 4dc4cf7, decided to not merge 18d7453.

PR #1959, opened 2020-01-08
Add lightning-wallet.com http seeds

btcontract:


schildbach:
Looks good to me, and I'm happy to see more HTTP seeds. Are you sure you want to continue to expose this via port 8081?

I'll do some functional testing on the weekend and would then merge this. I would also cherry-pick this to the 0.15 branch.

btcontract:
> Are you sure you want to continue to expose this via port 8081

Yes unless it's some kind of bad practice, I'd rather save 8080 for other possible future uses.

schildbach:
Merged!

PR #1960, opened 2020-01-17
Fix failing tests on certain systems and intoduce Utils::resetMocking()

erasmospunk:
It appears that the tests do not run in the same order or the static variables
are not reset between runs on all systemsi or gradle versions.
So it happened that the test UtilsTest::testRollMockClock was messing up
the time for the H2FullPrunedBlockChainTest (and the LevelDB one).

This fixes the issue by resetting the mock time for the
AbstractFullPrunedBlockChainTest.


schildbach:
Good catch! Merged.

PR #1961, opened 2020-01-17
Use protocolVersion in MessageSerializer

erasmospunk:
This change moves protocolVersion from Message to MessageSerializer.
The main reason is to support parsing transactions as non-segwit.
This is impossible to do otherwise, as MessageSerializer#makeTransaction
does not accept a protocolVersion parameter.
Additionally this change makes the MessageSerializer similar to the
Stream in BitcoinCore, where the protocol version can be modified.

I marked the API as beta so we can discuss a potentially better implementation.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
+++ b/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
@@ -17,6 +17,7 @@
 
 package org.bitcoinj.core;
 
+import com.google.common.base.Objects;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -28,6 +29,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
 
+import static org.bitcoinj.core.NetworkParameters.ProtocolVersion.CURRENT;
 import static org.bitcoinj.core.Utils.*;
 
 /**
@@ -47,6 +49,7 @@ public class BitcoinSerializer extends MessageSerializer {
     private static final int COMMAND_LEN = 12;
 
     private final NetworkParameters params;
+    private final int protocolVersion;
     private final boolean parseRetain;
 
     private static final Map<Class<? extends Message>, String> names = new HashMap<>();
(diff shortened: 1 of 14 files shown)


core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
```diff
@@ -378,4 +404,19 @@ public BitcoinPacketHeader(ByteBuffer in) throws ProtocolException, BufferUnderf
             cursor += 4;
         }
     }
+
```
schildbach: I would say adding equals/hashCode to BitcoinSerializer could go into a separate commit?

core/src/main/java/org/bitcoinj/core/Message.java
```diff
@@ -342,7 +346,18 @@ protected long readVarInt(int offset) throws ProtocolException {
...
         }
     }
-    
+
+    protected byte readByte() throws ProtocolException {
+        checkReadLength(1);
+        return payload[cursor++];
+    }
+
+    private void checkReadLength(int length) throws ProtocolException {
```
schildbach: This "extract method" refactoring could go into a separate commit, couldn't it?

schildbach:
> I marked the API as beta so we can discuss a potentially better implementation.

Is there something you expect to change specifically?

erasmospunk:
I think no, but I wanted to hear some feedback about the API.

(irrelevant but @schildbach how to contact you personally? I sent you an email but I'm not sure you are using that one)

core/src/main/java/org/bitcoinj/core/ChildMessage.java
```diff
@@ -43,7 +43,7 @@ public ChildMessage(NetworkParameters params, byte[] payload, int offset, int pr
         super(params, payload, offset, protocolVersion);
     }
 
-    public ChildMessage(NetworkParameters params, byte[] payload, int offset, int protocolVersion, Message parent, MessageSerializer setSerializer, int length) throws ProtocolException {
+    public ChildMessage(NetworkParameters params, byte[] payload, int offset, int protocolVersion, @Nullable Message parent, MessageSerializer setSerializer, int length) throws ProtocolException {
```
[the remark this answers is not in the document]
schildbach: I went ahead and fixed this via 73583776a6ef86e3d624047199747d43d38994f1.

core/src/main/java/org/bitcoinj/core/Message.java
schildbach: I just noticed you extracted the `checkReadLength()` method in order to add a  `readByte()` method. But what are you using this new method for? I don't see any references.

core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
schildbach: I added a variant of this via 3cbb10334edae29b86dbc08686bf61c487851061.

schildbach:
I merged the rest of this PR, minus the unused methods. Thanks for your contribution, I think this was a much needed refactoring!

PR #1965, opened 2020-02-28
Make Guava an `api` dependency using java-library

msgilligan:
Use the Gradle java-library plugin to make Guava an `api`
dependency since Guava types are used in the bitcoinj public API.

This is supported in Gradle 4.4.x: https://docs.gradle.org/4.4.1/userguide/java_library_plugin.html

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -9,7 +9,7 @@ version = '0.16-SNAPSHOT'
 
 dependencies {
     compile 'org.bouncycastle:bcprov-jdk15to18:1.64'
-    implementation 'com.google.guava:guava:28.2-android'
+    api 'com.google.guava:guava:28.2-android'
     compile 'com.google.protobuf:protobuf-java:3.9.2'
     implementation 'com.squareup.okhttp3:okhttp:3.12.8'
     implementation 'org.slf4j:slf4j-api:1.7.30'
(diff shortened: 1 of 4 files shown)


msgilligan:
Note that this changes the POM for `bitcoinj-core` as follows:

```
diff ~/.m2/repository/org/bitcoinj/bitcoinj-core/0.16-SNAPSHOT/bitcoinj-core-0.16-SNAPSHOT.pom bitcoinj-core-0.16-SNAPSHOT.pom 
13c13
<       <scope>runtime</scope>
---
>       <scope>compile</scope>
```

(Actually it changed the order of the `<dependency>` within the `<dependencies>` section, but other than that it only changes the scope from `runtime` to `compile` -- which is what we want.

schildbach:
I just merged the 'java-library' change. Regarding the rest, I think we should be able to purge the Guava dependency from all the other submodules, right? It should go into the same commit.

Can you rebase on master?

msgilligan:
>I think we should be able to purge the Guava dependency from all the other submodules, right?

I agree.

(However, I think in some cases it would make sense to leave an explicit dependency on an inherited dependency if that dependency is also being directly used in the current module. Since all these modules are in the same repo and we're not removing all Guava types from the bitcoinj API anytime soon I think removing it from these modules is the right thing to do.)
 
> Can you rebase on master?

Done.

schildbach:
Merged!

PR #1966, opened 2020-03-01
UTF-8 options, java & eclipse plugins in root build.gradle

msgilligan:
Simplify subproject `build.gradle` files by putting common
settings in root `build.gradle`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,9 +9,16 @@ buildscript {
 }
 
 allprojects {
+    apply plugin: 'java'
+    apply plugin: 'eclipse'
+
     repositories {
         jcenter()
     }
 
     group = 'org.bitcoinj'
+
+    compileJava.options.encoding = 'UTF-8'
+    compileTestJava.options.encoding = 'UTF-8'
+    javadoc.options.encoding = 'UTF-8'
 }
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -2,7 +2,6 @@ plugins {
     id 'java-library'
     id 'com.google.protobuf'
     id 'maven'
-    id 'eclipse'
 }
 
 version = '0.16-SNAPSHOT'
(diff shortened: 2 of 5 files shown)


build.gradle
```diff
@@ -9,9 +9,16 @@ buildscript {
...
 allprojects {
+    apply plugin: 'java'
+    apply plugin: 'eclipse'
+
     repositories {
         jcenter()
     }
 
     group = 'org.bitcoinj'
+
```
schildbach: I'm a bit turn. I like this change, because using UTF-8 is simply standard.

build.gradle
```diff
@@ -9,9 +9,16 @@ buildscript {
 }
 
 allprojects {
+    apply plugin: 'java'
```
schildbach: …however, I don't like this change because it makes the information about plugins more indirect.

The core subproject would now have both `java` and `java-library`. Seems unintended.

msgilligan: Let me think about this a little. There might be a better way for you.

schildbach:
@msgilligan I would merge the "encoding" part of this PR if you're ok with that.

msgilligan:
@schildbach I don't think the encoding settings will work unless the plugins are also applied in the root. 

> Let me think about this a little. There might be a better way for you.

I think the settings need to be applied conditionally (i.e. test to make sure the plugin is there before applying them.)  That's what I was going to try to do...

schildbach:
I'm not sure, but I think this is obsolete by now. Can it perhaps be closed? (Or rebased, it seems we have implemented at least part of it.)

msgilligan:
> I'm not sure, but I think this is obsolete by now. Can it perhaps be closed? (Or rebased, it seems we have implemented at least part of it.)

Yeah, I'll close it for now.

If you're willing to consider

```
    allprojects {
       apply plugin: 'java'
   etc...
```

In the root `build.gradle` I can create an updated version of it...

PR #1930, opened 2019-09-05
PSBT implementation (BIP174)

erasmospunk:
This initial PSBT implementation supports serializing/deserializing, creation,
merging and extracting operations.
It does not support signing as the current BitcoinJ lacks the appropriate
abstractions (although it can be done manually).

Additionally there are three separate commits with changes that are needed
for the PSBT implementation.

PR #1967, opened 2020-03-04
Make zero inputs/outputs transactions parsable.

erasmospunk:
Switch to the bitcoin core reference segwit parsing algorithm because
some no-segwit transactions with no inputs are incorrectly parsed as
segwit.

This change is to be 100% compatible with the reference implementation
and to be compatible with the PSBT implementation.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -42,6 +42,7 @@ import javax.annotation.Nullable;
 import java.io.*;
 import java.util.*;
 
+import static org.bitcoinj.core.NetworkParameters.ProtocolVersion.WITNESS_VERSION;
 import static org.bitcoinj.core.Utils.*;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
@@ -94,6 +95,12 @@ public class Transaction extends ChildMessage {
         }
     };
     private static final Logger log = LoggerFactory.getLogger(Transaction.class);
+
+    /**
+     * When this bit is set in protocolVersion, do not include witness. The actual value is the same as in Bitcoin Core
+     * for consistency.
+     */
+    public static final int SERIALIZE_TRANSACTION_NO_WITNESS = 0x40000000;
 
     /** Threshold for lockTime: below this value it is interpreted as block number, otherwise as timestamp. **/
     public static final int LOCKTIME_THRESHOLD = 500000000; // Tue Nov  5 00:53:20 1985 UTC
@@ -281,6 +288,15 @@ public class Transaction extends ChildMessage {
             }
         }
         return cachedTxId;
+    }
+
+    /**
+     * Returns if tx witnesses are allowed based on the protocol version
+     */
+    private boolean allowWitness() {
+        int protocolVersion = serializer.getProtocolVersion();
+        return (protocolVersion & SERIALIZE_TRANSACTION_NO_WITNESS) == 0
+                && protocolVersion >= WITNESS_VERSION.getBitcoinProtocolVersion();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -641,4 +641,19 @@ public void getWeightAndVsize() {
...
     }
+
+    @Test
+    public void edgeCases() {
+        MessageSerializer serializer = UNITTEST.getDefaultSerializer();
+        // Non SegWit tx with zero input and outputs
+        String txHex = "01000000000000000000";
+        Transaction tx = serializer.makeTransaction(HEX.decode(txHex));
+        assertEquals(txHex, HEX.encode(tx.bitcoinSerialize()));
+
```
schildbach: Can you split this into two test cases which can fail individually? And maybe a bit more speaking names?

core/src/test/java/org/bitcoinj/script/ScriptTest.java
```diff
@@ -394,7 +395,17 @@ public void dataDrivenInvalidTransactions() throws Exception {
             if (test.isArray() && test.size() == 1 && test.get(0).isTextual())
                 continue; // This is a comment.
             Map<TransactionOutPoint, Script> scriptPubKeys = parseScriptPubKeys(test.get(0));
-            Transaction transaction = TESTNET.getDefaultSerializer().makeTransaction(HEX.decode(test.get(1).asText().toLowerCase()));
+            byte[] txBytes = HEX.decode(test.get(1).asText().toLowerCase());
```
schildbach: I wonder if perhaps some added testcases to `tx_invalid.json` are missing here? This testcase succeeds with or without any of the core logic changes.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -95,6 +96,9 @@ public int compare(final Transaction tx1, final Transaction tx2) {
     };
     private static final Logger log = LoggerFactory.getLogger(Transaction.class);
 
+    /** When this bit is set in protocolVersion, do not include witness */
```
schildbach: I'm trying to find this (and the described logic) in the specs, but did not find. Can you point me to it?

erasmospunk:
> Just for my understanding: does this affect any valid transactions?
Not at all. If there is an upgrade in the witness format, the extra flags will make the current parser fail in a predictable manner (something that we want).

> Zero outputs are invalid and zero inputs can only happen for the coinbase right?
In the Bitcoin protocol, zero inputs and/or zero outputs transactions are invalid but parsable. In the coinbase we have an input but the prevout is empty (hash=0x0..00, index=0xf..ff)

core/src/main/java/org/bitcoinj/core/Transaction.java
erasmospunk: This is an internal serialisation flag that bitcoin core uses, I don't think it is in the specs. I just used the same for consistency.

https://github.com/bitcoin/bitcoin/blob/0.19/src/primitives/transaction.h#L15

core/src/test/java/org/bitcoinj/script/ScriptTest.java
erasmospunk: Without this modification, the test fails because the new parser cannot handle this zero-input, one-output transaction `0100000000010000000000000000015100000000`. It tries to parse it as segwit transaction and fails because it is incorrectly formed.

The old Transaction code was parsing it as a segwit tx without failing but the result was invalid: it was reading it as a zero-input and zero-output tx (the raw tx has one output). The actual test was passing because the point was to detect invalid transactions, not incorrectly parsed ones.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: Ok, then I think it would be a good idea to explain in the comment that this quirk was directly ported from Bitcoin Core.

core/src/test/java/org/bitcoinj/script/ScriptTest.java
schildbach: Ok thanks for the explanation.

schildbach:
> In the Bitcoin protocol, zero inputs and/or zero outputs transactions are invalid but parsable.

That would perhaps make a good summary for your commit: `Make zero inputs/outputs transactions parsable.`

core/src/main/java/org/bitcoinj/core/Transaction.java
erasmospunk: Done

erasmospunk:
This is the last dependency for the pull request #1930

schildbach:
Thanks! Merged.

PR #1968, opened 2020-03-06
Use VarInt not opcode for script length in tx.hashForWitnessSignature

craigraw:
The current implementation of tx.hashForWitnessSignature does not add the script length before the script as required by BIP143. 

The tests (and Script.correctlySpends) currently work because the script length is prepended before passing the script bytes into the method, by wrapping the witness script using `new ScriptBuilder.data(witnessScript.getProgram()).build()` which prepends the length as a script opcode. However, not only is it unclear that this treatment of the witness script is necessary before calling the method, but it will fail if the witness script is [longer than 75 bytes](https://github.com/bitcoinj/bitcoinj/blob/d67f060e27f38e4117ba20f38d6ae8bd1130ab34/core/src/main/java/org/bitcoinj/script/ScriptBuilder.java#L105), as may well be the case for P2WSH. 

I have added a test which uses an example from BIP143 with a witness script of 207 bytes to verify.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -1405,6 +1405,8 @@ public class Transaction extends ChildMessage {
             bos.write(hashSequence);
             bos.write(inputs.get(inputIndex).getOutpoint().getHash().getReversedBytes());
             uint32ToByteStreamLE(inputs.get(inputIndex).getOutpoint().getIndex(), bos);
+            VarInt scriptLength = new VarInt(scriptCode.length);
+            bos.write(scriptLength.encode());
             bos.write(scriptCode);
             uint64ToByteStreamLE(BigInteger.valueOf(prevValue.getValue()), bos);
             uint32ToByteStreamLE(inputs.get(inputIndex).getSequenceNumber(), bos);
--- a/core/src/main/java/org/bitcoinj/script/Script.java
+++ b/core/src/main/java/org/bitcoinj/script/Script.java
@@ -1559,8 +1559,7 @@ public class Script {
                 throw new ScriptException(ScriptError.SCRIPT_ERR_SIG_DER, "Cannot decode", x);
             }
             ECKey pubkey = ECKey.fromPublicOnly(witness.getPush(1));
-            Script scriptCode = new ScriptBuilder().data(ScriptBuilder.createP2PKHOutputScript(pubkey).getProgram())
-                    .build();
+            Script scriptCode = ScriptBuilder.createP2PKHOutputScript(pubkey);
             Sha256Hash sigHash = txContainingThis.hashForWitnessSignature(scriptSigIndex, scriptCode, value,
                     signature.sigHashMode(), false);
             boolean validSig = pubkey.verify(sigHash, signature);
--- a/core/src/test/java/org/bitcoinj/core/TransactionTest.java
+++ b/core/src/test/java/org/bitcoinj/core/TransactionTest.java
@@ -355,8 +355,16 @@ public class TransactionTest {
     }
     
     @Test
-    public void testWitnessSignatureP2SH_P2WPKH_Single() throws SignatureDecodeException {
-      String hex = "010000000136641869ca081e70f394c6948e8af409e18b619df2ed74aa106c1ca29787b96e0100000000ffffffff0200e9a435000000001976a914389ffce9cd9ae88dcc0631e88a821ffdbe9bfe2688acc0832f05000000001976a9147480a33f950689af511e6e84c138dbbd3c3ee41588ac00000000";
+    public void testWitnessSignatureP2SH_P2WSHSingleAnyoneCanPay() throws SignatureDecodeException {
+      // test vector P2SH-P2WSH from the final example at:
+      // https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#p2sh-p2wsh
+      String hex = "01000000" // version
+                + "01" // num txIn
+                + "36641869ca081e70f394c6948e8af409e18b619df2ed74aa106c1ca29787b96e" + "01000000" + "00" + "ffffffff" // txIn
+                + "02" // num txOut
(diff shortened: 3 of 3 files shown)


schildbach:
Hm, I somehow remember this was brought up a couple of weeks/months ago. Has that been you?

In any case, thanks for pointing this out!

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1405,6 +1405,8 @@ public synchronized Sha256Hash hashForWitnessSignature(
             bos.write(hashSequence);
             bos.write(inputs.get(inputIndex).getOutpoint().getHash().getReversedBytes());
             uint32ToByteStreamLE(inputs.get(inputIndex).getOutpoint().getIndex(), bos);
+            VarInt scriptLength = new VarInt(scriptCode.length);
```
schildbach: The last time I looked at this, I didn't see the error because BIP143 just states "serialized as scripts inside CTxOuts".

Now I agree. There is even a [TODO in `TransactionOutput.java`](https://github.com/bitcoinj/bitcoinj/blob/f94f8cd692144c5423950b9063db7e7ecd8a89c4/core/src/main/java/org/bitcoinj/core/TransactionOutput.java#L151) to move the script serialization to `Script.java`. Which would perhaps helped prevent the mistake.

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -1559,8 +1559,7 @@ public void correctlySpends(Transaction txContainingThis, int scriptSigIndex, @N
                 throw new ScriptException(ScriptError.SCRIPT_ERR_SIG_DER, "Cannot decode", x);
             }
             ECKey pubkey = ECKey.fromPublicOnly(witness.getPush(1));
-            Script scriptCode = new ScriptBuilder().data(ScriptBuilder.createP2PKHOutputScript(pubkey).getProgram())
-                    .build();
+            Script scriptCode = ScriptBuilder.createP2PKHOutputScript(pubkey);
```
schildbach: Looks reasonable.

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -240,14 +240,14 @@ public void testWitnessSignatureP2WPKH() {
         assertEquals("30450221008b9d1dc26ba6a9cb62127b02742fa9d754cd3bebf337f7a55d114c8e5cdd30be022040529b194ba3f9281a99f2b1c0a19c0489bc22ede944ccf4ecbab4cc618ef3ed01",
                 HEX.encode(txSig0.encodeToBitcoin()));
 
-        Script scriptCode = new ScriptBuilder().data(ScriptBuilder.createP2PKHOutputScript(key1).getProgram()).build();
```
schildbach: I will closely review these test changes asap.

schildbach: I agree with this change.

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -316,15 +316,15 @@ public void testWitnessSignatureP2SH_P2WPKH() {
         assertEquals("a9144733f37cf4db86fbc2efed2500b4f4e49f31202387",
                 HEX.encode(scriptPubKey.getProgram()));
 
-        Script scriptCode = new ScriptBuilder().data(ScriptBuilder.createP2PKHOutputScript(key).getProgram()).build();
```
schildbach: I agree with this change.

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -353,6 +353,18 @@ public void testWitnessSignatureP2SH_P2WPKH() {
         assertEquals(signedTxHex, HEX.encode(tx.bitcoinSerialize()));
         assertEquals(signedTxHex.length() / 2, tx.getMessageSize());
     }
+    
+    @Test
+    public void testWitnessSignatureP2SH_P2WPKH_Single() throws SignatureDecodeException {
```
schildbach: It looks like you took the example from https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#p2sh-p2wsh

If I am right, can you correct the title of this test and maybe add the URL in a comment like in the tests above.

Ideally we would implement the entire test vector(s), but it's not required for merging this.

schildbach: The hex could be separated a bit, like in the previous tests. But don't worry, I can do that too.

craigraw:
No problem, I've renamed the test and separated out the transaction hex.

schildbach:
Thanks! I just merged.

schildbach:
#2006 claims the changes from this PR cause verification errors from Bitcoin Core when creating and broadcasting a transaction.

craigraw:
I think we would need more details on the failed transaction to verify. Some [reports](https://bitcoin.stackexchange.com/questions/61030/got-64-non-mandatory-script-verify-flag-signature-must-be-zero-for-failed-chec) claim this error occurs "if the value you are signing is not the exact amount of the entire unspent output value"

java4tech:
you can reproduce this issue, if you run [wallettemplate](https://github.com/bitcoinj/bitcoinj/tree/master/wallettemplate) to send P2WPKH transaction. I couldn't find much from the bitcoincore logs either. The issue might be with the [SendRequest](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/wallet/SendRequest.java). 

here is my code that i tried to use with [WalletAppKit](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/kits/WalletAppKit.java) in my own project

        Wallet wallet = DashboardController.getKit().wallet();
        NetworkParameters params = TestNet3Params.get();
        Address destination = Address.fromString(params, strDest);
        Coin value = Coin.parseCoin(strValue);
        SendRequest request = SendRequest.to(destination, value);

schildbach:
I've put up PR #2014 which fixes what I think are two missed cases of scriptCode building. @craigraw do you agree?

PR #1972, opened 2020-03-11
Rename .github/workflow to .github/workflows

msgilligan:
This should fix Github Actions


msgilligan:
Note: The current `master` branch will fail on Github actions because version 0.8.6 of the Gradle Protobuf Plugin is incompatible with Gradle 6.2.1 which is the default on Github Actions.

https://github.com/msgilligan/bitcoinj/actions/runs/53842463

msgilligan:
I merged PR #1973 into this PR so that the builds will run better, although at present they're only completing successfully on **macOS** and failing under **Windows** and **Linux**:

https://github.com/msgilligan/bitcoinj/actions/runs/53939582

schildbach:
I merged the rename, and now Actions picks it up (but fails).

schildbach:
Hmm, the Ubuntu and Windows tests fail with

> VerificationException: Block too far in future: 2020-03-12T00:12:04Z (1583971924) vs allowed 1970-01-01T09:00:08Z (32408)

I wonder if GitHub Actions has a clock problem.

msgilligan:
Yeah, I was researching that issue. Weird.

msgilligan:
> I wonder if GitHub Actions has a clock problem.

It looks like the system time is OK. I added (see https://github.com/msgilligan/bitcoinj/commit/c079d92bf8f4e972b3e41ebe51efdebb10c25a19) a step to print the time, but the Linux and Windows builds [still fail](https://github.com/msgilligan/bitcoinj/actions/runs/54126506) even though the time is printed correctly.

msgilligan:
Also, the Github Actions Linux builds are working on the `consensusj-master` branch. (I just pushed a trivial change to force a build: https://github.com/msgilligan/bitcoinj/actions/runs/54130261)

msgilligan:
@schildbach It looks like Linux builds under Github Actions have (mysteriously?) started working.

msgilligan:
Also, we could close this issue and open a new one for the remaining **Windows** build problem. What do you think?

schildbach:
I'm not sure if this is just coincidence. If you look through [the past dozens of builds](https://github.com/bitcoinj/bitcoinj/actions) there is quite some on-off action going on (even including MacOS at some point).

msgilligan:
> I'm not sure if this is just coincidence. If you look through [the past dozens of builds](https://github.com/bitcoinj/bitcoinj/actions) there is quite some on-off action going on (even including MacOS at some point). I think it's just unstable.

Yes, the tests are unstable (and part of the problem is that some of the tests are integration tests not true unit tests.) I was hoping that by adding Github Actions with the non-Linux platforms that it would help us find and remove the instability.

However, I don't recall seeing `org.bitcoinj.core.VerificationException: Block too far in future` except under Github Actions.

msgilligan:
Here's PR https://github.com/bitcoinj/bitcoinj/pull/1982 that adds JDK 14 to the tests. Let's see if Windows tests work a little better under JDK 14...

msgilligan:
@schildbach I'm going to close this PR. I think we have existing issues about test reliability. If you want we can open a new issue for the failure of the Windows tests, but given the visible red X in Github Actions CI, we know we'll be constantly reminded about it.

schildbach:
Yes, but what are we going to do about Windows? I suspect currently everyone who submits a PR will receive several emails about tests failing. And those devs will be confused, because they didn't do anything wrong. If there is no idea, shall we remove Windows for now?

PR #1973, opened 2020-03-11
Upgrade Gradle Protobuf Plugin to 0.8.10

msgilligan:
The current latest version is 0.8.12, but version 0.8.11
and later require Gradle 5.6 or later which will change our
build requirements and documentation.

See:
https://github.com/google/protobuf-gradle-plugin/releases/tag/v0.8.11

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
     }
 
     dependencies {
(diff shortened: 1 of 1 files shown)


schildbach:
Merged.

PR #1977, opened 2020-04-16
enquote index column name in PostgresFullPrunedBlockStore.java 

mancini0:
enquote index column name in PostgresFullPrunedBlockStore.java to make compatible with Cockroach-DB

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/PostgresFullPrunedBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/PostgresFullPrunedBlockStore.java
@@ -77,7 +77,7 @@ public class PostgresFullPrunedBlockStore extends DatabaseFullPrunedBlockStore {
             "    toaddress character varying(74),\n" +
             "    addresstargetable smallint,\n" +
             "    coinbase boolean,\n" +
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/store/PostgresFullPrunedBlockStore.java
```diff
@@ -81,7 +81,7 @@
             ")\n";
```
schildbach, line 81: There is one more occurence in the line above. Does that need to be escaped, too?

schildbach:
My understanding of this change is that it's 100% compatible for Postgres users who use the all-lowercase spelling of `index`. Am I right?

core/src/main/java/org/bitcoinj/store/PostgresFullPrunedBlockStore.java
mancini0: yes, thanks

mancini0:
> My understanding of this change is that it's 100% compatible for Postgres users who use the all-lowercase spelling of `index`. Am I right?

yes that is correct

Issue #1873, opened 2019-05-30
Public methods whose parameters and return types are not public


lingocoder:
## Context

I'm developing [*an ABI inspection tool*](https://bit.ly/AbiImpl). I call it *The ABI*. It scans the project classes and dependencies of Java libraries. From the scan, it reports on classes whose public API exposes types from the library's dependencies.

I did a scan of the bitcoinj-core library. The abi reported several classes
 that it determined exports types from different dependencies through the library's public API. ([*attached*](https://github.com/bitcoinj/bitcoinj/files/3238989/bitcoinj.abi.report.0.b.txt))

The ABI inspector's results jibe with [*Gradle's guidelines for separating `api` and `implementation` dependencies*](http://bit.ly/ImplApiSep) into appropriate configurations. However the results differ from the transitivity currently defined for bitcoinj-core's dependencies.

The reason I'm engaging is because I'm looking to compare the results arrived at by the tool with bitcoinj authors' actual expectations for the transitivity of the library's dependencies. I'm hoping your feedback will help improve the accuracy of the abi's results.

The following table lists what the configurations are currently and what the inspection suggests that they could be changed to:

|      |   Currently    |   Analyzed as   |
| :--- | :--------------: | :-------------: |
| *`org.bouncycastle:bcprov-jdk15on:1.60`* | compile | api |
| *`com.squareup.okhttp3:okhttp:3.12.3`* | implementation | api |
| *`com.google.protobuf:protobuf-java:3.7.1`* | compile | api |
| *`com.google.guava:guava:27.1-android`* | implementation | api |
| *`net.jcip:jcip-annotations:1.0`* | implementation | api |  

The analysis the tool applies is [*based on Gradles recommended best practices*](http://bit.ly/iDimpLApi) for assigning appropriate configurations to dependencies.

## How the ABI Inspector's analysis was verified

1. Download bitcoinj-master.zip from Code page, unzip and cd to install dir
2. `gradle -x javadoc clean assemble`
3. Install the jar that is built to your local Maven repository (*assumes Maven is installed*)

        mvn org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install-file \
        -Dfile=${install_dir}\bitcoinj-master\core\build\libs\bitcoinj-core-0.16-SNAPSHOT.jar \
        -DgroupId=org.bitcoinj -DartifactId=bitcoin-core \
        -Dversion=0.16-SNAPSHOT -Dpackaging=jar \
[...]

PR #1979, opened 2020-04-22
Properly declare API dependencies in bitcoinj-core

msgilligan:
Properly declare Bouncy Castle, Protobuf, and OkHttp as Gradle `api`
dependencies in bitcoinj-core build.gradle.

This addresses Issue #1873.


schildbach:
Merged.

PR #1987, opened 2020-04-30
Added missing ScriptBuilder.createP2SHOutputScript for ECKey

milczarekIT:


schildbach:
I must admit I'm not really fluent in the various usecase of P2SH. Have you seen

```
public static Script createP2SHOutputScript(int threshold, List<ECKey> pubkeys)
```

which seems to serve a similar purpose?

milczarekIT:
Yes, mentioned method is for multisig: https://wiki.trezor.io/P2SH-multisig
Added method is for: https://wiki.trezor.io/P2WPKH-in-P2SH

milczarekIT:
Up

msgilligan:
This looks like effectively a part of PR #1995. I'm not sure what @schildbach would prefer, but I think this could/should be a single PR with a single commit that includes all the changes of PR #1995.

milczarekIT:
As I understand  @schildbach prefer 1 PR per 1 feature, so this is the reason why I created 2 PRs

schildbach:
If this utility method is deemed useful to more than use usecase, then yes I think this could be seen as a feature on itself and thus be merged as one commit. I really have to wrap around my head on the usecases now…

On another note, I think we should try to develop P2SH and P2WSH in parallel (aiming at API parity between the two). Do you think you could implement a `createP2WSHOutputScript()` too?

milczarekIT:
Yes, this method can be treated as a utility method and used in other places.

To be honest I worry I will not find time to implement  `createP2WSHOutputScript()`. @schildbach, @msgilligan if you have time feel free to implement it.

As @msgilligan noticed https://github.com/bitcoinj/bitcoinj/pull/1995 is enough to retrieve `LegacyAddress` from `ECKey` for P2SH

msgilligan:
> I must admit I'm not really fluent in the various usecase of P2SH. Have you seen
> 
> ```
> public static Script createP2SHOutputScript(int threshold, List<ECKey> pubkeys)
> ```
> 
> which seems to serve a similar purpose?

I created PR #2037 to improve the JavaDoc for the existing P2SH methods and to help better understand their use cases.

msgilligan:
> If this utility method is deemed useful to more than use usecase, then yes I think this could be seen as a feature on itself and thus be merged as one commit. I really have to wrap around my head on the usecases now…
> 
> On another note, I think we should try to develop P2SH and P2WSH in parallel (aiming at API parity between the two). Do you think you could implement a `createP2WSHOutputScript()` too?

There already are two `createP2WSHOutputScript()` methods (one that takes `byte[]` and one that takes `Script`, the same as for `createP2SHOutputScript()`.  I started writing 3 more methods to add (explicit, convenience) support for P2WSH multisig and also for P2SH-wrapped-P2WSH. I should have a pull request ready in the next day or so.

I'm also thinking it might be good to write up some use cases -- maybe in an Issue? (eventually I'd like to see the use cases in Kotlin/JUnit 5 test cases)

milczarekIT:
Up

PR #1988, opened 2020-05-02
DRMWorkaround cleanup

milczarekIT:
Cleaned up DRMWorkaround class

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DRMWorkaround.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DRMWorkaround.java
@@ -39,14 +39,11 @@ public final class DRMWorkaround {
         // even though that shipped in 2014! That's dumb. So we disable the ridiculous US government mandated DRM
         // for AES-256 here, as Tor/BIP38 requires it.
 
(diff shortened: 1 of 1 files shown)


schildbach:
Thanks! Can you separate your changes? I see two aspects:

- Make the class non-instantiatable
- Improve the parameter of a log message

Ideally each of those is in a separate commit. The "trigger CI build" is superfluous however – I can trigger rebuilds for you if you want.

milczarekIT:
I added "trigger CI build", because build randomly failed.
There are really small changes in DRMWorkaround, do we really need to split it into 2 commits?

schildbach:
I've pushed dbc4cb1d4cf47140c792fb141eb960666522d622 as an example of how "code smells" can by handled. All occasions of the same smell (originally discovered by you) are fixed at the same time, and there is a short but meaningful commit message.

milczarekIT:
Ok, thanks, but still some improvements can be introduced in `DRMWorkaround ` proposed in this PR.

schildbach:
Can you rebase this PR on current master so we can see what's left?

Issue #1992, opened 2020-05-03
HDPath does not support parsing paths for private keys


milczarekIT:
Following code throws exception:
`HDPath.parse("m/7H/3/3/1");`

... only paths for public keys are supported. Following code is working:
`HDPath.parse("M/7H/3/3/1");`

PR #1993, opened 2020-05-03
HDPath parsePath - fixed issue with parsing HDPaths for private keys

milczarekIT:
Fixed: https://github.com/bitcoinj/bitcoinj/issues/1992

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/HDPath.java
+++ b/core/src/main/java/org/bitcoinj/crypto/HDPath.java
@@ -39,9 +39,9 @@ import java.util.List;
  * concisely create HDPath objects (especially when statically imported.)
  */
 public class HDPath extends AbstractList<ChildNumber> {
-    private static final char PREFIX_PRIVATE = 'm';
-    private static final char PREFIX_PUBLIC = 'M';
-    private static final char SEPARATOR = '/';
+    private static final String PREFIX_PRIVATE = "m";
+    private static final String PREFIX_PUBLIC = "M";
+    private static final String SEPARATOR = "/";
     protected final boolean hasPrivateKey;
     protected final List<ChildNumber> unmodifiableList;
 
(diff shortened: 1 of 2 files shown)


milczarekIT:
Up

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -151,19 +151,22 @@ public static HDPath m(ChildNumber... children) {
      * Where a letter "H" means hardened key. Spaces are ignored.
      */
     public static HDPath parsePath(@Nonnull String path) {
-        String[] parsedNodes = path.replace("M", "").split("/");
+        boolean hasPrivateKey = path.startsWith(PREFIX_PRIVATE);
+        String[] parsedNodes = path.replace(PREFIX_PUBLIC, "")
+                .replace(PREFIX_PRIVATE, "")
+                .split("/");
```
schildbach, line 157: Maybe use the SEPARATOR constant at this point?

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -151,19 +151,22 @@ public static HDPath m(ChildNumber... children) {
      * Where a letter "H" means hardened key. Spaces are ignored.
      */
     public static HDPath parsePath(@Nonnull String path) {
-        String[] parsedNodes = path.replace("M", "").split("/");
+        boolean hasPrivateKey = path.startsWith(PREFIX_PRIVATE);
+        String[] parsedNodes = path.replace(PREFIX_PUBLIC, "")
```
schildbach, line 155: Actually I think the whole .replace() is wrong. Only a leading `M` or `m` should be removed. Later occurences would be illegal and should lead to some form of error. (I'm aware that the previous implementation had this problem too.)

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -151,19 +151,22 @@ public static HDPath m(ChildNumber... children) {
...
+                .replace(PREFIX_PRIVATE, "")
+                .split("/");
         List<ChildNumber> nodes = new ArrayList<>();
 
         for (String n : parsedNodes) {
-            n = n.replaceAll(" ", "");
+            n = n.replace(" ", "");
```
schildbach, line 161: This change seems unrelated to me.

core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
```diff
@@ -31,7 +33,7 @@
     public void testPrimaryConstructor() throws Exception {
         HDPath path = new HDPath(true, Collections.<ChildNumber>emptyList());
         Assert.assertTrue("Has private key returns false incorrectly", path.hasPrivateKey);
-        Assert.assertEquals("Path not empty", path.size(), 0);
```
schildbach, line 34: Fixing the use of `assertEquals()` in the entire project would be a nice separate PR.

schildbach:
I fixed this via ed845fd22b58ffb7de0b50a1f2e22f34b5d73277

Issue #1994, opened 2020-05-11
Creating P2SH address from ECKey is not supported


milczarekIT:
Creating P2SH address (Nested SegWit) from ECKey is not supported.

In the current version it is not possible to create P2SH address (Nested SegWit) from ECKey:
`Address.fromKey(params, key, ScriptType.P2SH)`

PR #1995, opened 2020-05-11
Added support for P2SH (Nested SegWit) Address.fromKey(params, key, P2SH)

milczarekIT:
Fixed https://github.com/bitcoinj/bitcoinj/issues/1994


milczarekIT:
This task depends on: https://github.com/bitcoinj/bitcoinj/pull/1987

It contains also changes from https://github.com/bitcoinj/bitcoinj/pull/1993, because https://github.com/milczarekIT/bitcoinj is few commits ahead from https://github.com/bitcoinj/bitcoinj

core/src/main/java/org/bitcoinj/core/LegacyAddress.java
```diff
@@ -91,7 +94,30 @@ public static LegacyAddress fromPubKeyHash(NetworkParameters params, byte[] hash
...
+     * @param key
+     *            only the public part is used
+     * @param scriptType
+     *            script type. P2PKH and P2SH supported.
+     * @return constructed address
+     */
+    public static LegacyAddress fromKey(NetworkParameters params, ECKey key, Script.ScriptType scriptType) {
+        if (ScriptType.P2PKH == scriptType) {
+            return fromPubKeyHash(params, key.getPubKeyHash());
+        } else if(ScriptType.P2SH == scriptType) {
```
msgilligan, line 115: From the context you can deduce that this is P2SH-P2WPKH, but maybe this should be in the comments.

I'm not sure if creating an explicit `Script.ScriptType` for `P2SH_P2WPKH` would be a good idea, but thought I'd at least throw the idea out. @schildbach ?

core/src/test/java/org/bitcoinj/core/LegacyAddressTest.java
```diff
@@ -188,6 +188,21 @@ public void p2shAddress() {
...
+    @Test
+    public void p2shAddressFromPrivateKey() {
+        final NetworkParameters networkParameters = MainNetParams.get();
+        final String privateKeyWIF = "L3E6dAmEwcRNbeUnzjd6YuhMm9fDxgzVmJHXAV1gtXm5e49Dctiv";
+
+        final ECKey key = DumpedPrivateKey.fromBase58(networkParameters, privateKeyWIF).getKey();
+        final Script script = ScriptBuilder.createP2SHOutputScript(key);
+
+        final LegacyAddress addressFromScriptHash = LegacyAddress.fromScriptHash(networkParameters, ScriptPattern.extractHashFromP2SH(script));
+        final LegacyAddress addressFromKey = LegacyAddress.fromKey(networkParameters, key, ScriptType.P2SH);
```
msgilligan, line 200: Minor point, but `final` is probably overkill here. Everything is one test/method and it's clear that no variable is or will be set twice. (I think this also varies from the other code in bitcoinj)

msgilligan:
@schildbach When you have a chance can you take a look at this?

core/src/test/java/org/bitcoinj/core/LegacyAddressTest.java
milczarekIT: `final` is just a hit for JVM that this variable is final and it will not be reassigned, so it is not an overkill, but good practice.
Also, this provides better readability for developers (you do not need to search in code if a variable has been reassigned).

core/src/main/java/org/bitcoinj/script/ScriptBuilder.java
```diff
@@ -478,6 +478,16 @@ public static Script createP2WPKHOutputScript(ECKey key) {
         return createP2WPKHOutputScript(key.getPubKeyHash());
     }
 
+    /**
+     * Creates a scriptPubKey that sends to script calculated by the given key. Read
+     * <a href="https://github.com/bitcoin/bips/blob/master/bip-0016.mediawiki">BIP 16</a> to learn more about this
+     * kind of script.
+     */
```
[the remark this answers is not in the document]
milczarekIT: The method name `createP2SHOutputScript`  has been used to adjust to the current convention name.
There are 8 methods in this class with this name. So I will leave this method name.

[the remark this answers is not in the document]
msgilligan: There are 3 methods in this class with the existing name. 2 of those create output scripts for any (i.e. non-segwit multisig, arbitrary P2SH, or P2SH-wrapped segwit) P2SH. The third is non-segwit multisig specific (and perhaps should be renamed.)

I think this new method needs to be clearly marked as for P2SH-P2WPKH as this is a special case. At the very least the comments need to explain this, and I think it would be a good idea for the method name to be something like: `createP2SHWrappedSegwitOutputScript`,

core/src/main/java/org/bitcoinj/script/ScriptBuilder.java
```diff
@@ -478,6 +478,16 @@ public static Script createP2WPKHOutputScript(ECKey key) {
...
+    public static Script createP2SHOutputScript(ECKey key) {
+        final Script redeemScript = new ScriptBuilder().smallNum(0).data(key.getPubKeyHash()).build();
```
msgilligan, line 487: The line creating the redeemScript should probably be in a separate method, similar to the existing `createRedeemScript` (for multisig) because we'll need that code for redeeming these P2SH-P2WPKH outputs.

alnagar:
When is this PR expected to be merged ?

milczarekIT:
@schildbach up, this PR is waiting for over 8 months. The community needs this code to be merged.

dolb:
@milczarekIT can I ask what does your pull request actually bring into the table?

I am asking, as this makes a little sense to me what is your code adding? I am just curious, as I am actually not using bitcoinj for spending UTXOs, but what will this line give to the average bitcoinj dev:

```
LegacyAddress.fromKey(params, key, Script.ScriptType.P2SH)
```

I was reading this pull request and I tried to figure out what is that script address actually for - can you explain this to me? I was currently working on spending p2sh-p2wpkh output, and this does not touch the surface of my needs.

alnagar:
@dolb 
This PR is for generating P2SH addresses that starts with 3***********

dolb:
> @dolb
> This PR is for generating P2SH addresses that starts with 3***********

I know what p2sh is - you can generate those now as well by creating a script with ScriptBuilder in current code and using fromScript method from LegacyAddress class. What I mean is how are you going to spend UTXO after sending value to this address, as this is a mistery for me ;)

PR #1996, opened 2020-05-20
BlockImporter: use switch statement instead of if/else if for parsing block store argument

msgilligan:
All changes are logically equivalent to old code.

* Use lambdas in a few places
* <>
* Combine catches
* etc.

### DIFF
--- a/tools/src/main/java/org/bitcoinj/tools/BlockImporter.java
+++ b/tools/src/main/java/org/bitcoinj/tools/BlockImporter.java
@@ -40,18 +40,22 @@ public class BlockImporter {
             params = MainNetParams.get();
 
         BlockStore store;
-        if (args[1].equals("MemFull")) {
-            Preconditions.checkArgument(args.length == 2);
-            store = new MemoryFullPrunedBlockStore(params, 100);
-        } else if (args[1].equals("Mem")) {
-            Preconditions.checkArgument(args.length == 2);
-            store = new MemoryBlockStore(params);
-        } else if (args[1].equals("SPV")) {
-            Preconditions.checkArgument(args.length == 3);
-            store = new SPVBlockStore(params, new File(args[2]));
-        } else {
-            System.err.println("Unknown store " + args[1]);
-            return;
+        switch (args[1]) {
+            case "MemFull":
+                Preconditions.checkArgument(args.length == 2);
+                store = new MemoryFullPrunedBlockStore(params, 100);
+                break;
+            case "Mem":
+                Preconditions.checkArgument(args.length == 2);
+                store = new MemoryBlockStore(params);
+                break;
+            case "SPV":
+                Preconditions.checkArgument(args.length == 3);
+                store = new SPVBlockStore(params, new File(args[2]));
+                break;
+            default:
+                System.err.println("Unknown store " + args[1]);
+                return;
         }
         
         AbstractBlockChain chain = null;


schildbach:
I generally agree with these Java7→8 migrations. However I would much prefer to bundle separate commits per IDE suggestion. So e.g. if the suggestion is to use the multi-catch, refactor all such occurrences (entire framework, where allowed) and then put that into one commit with nothing else.

msgilligan:
> I would much prefer to bundle separate commits per IDE suggestion. So e.g. if the suggestion is to use the multi-catch, refactor all such occurrences (entire framework, where allowed) and then put that into one commit with nothing else.

That would turn this PR into 8 separate PRs and for some of those 8, it might require me to go through all of the other modules to look for possible additional instances of the same refactorings. (obviously no lambdas in core, etc.) That significantly increases the amount of effort to contribute. In this case the work started as I was reviewing `BlockImporter` and `WatchMemPool` to see if they could be used for an application I was considering. I started fixing recommendations from the IDE as I saw them and decided to implement most of those changes across the entire `tools` subproject (since it had recently been updated to JDK 8.)

I understand your preference and it certainly makes sense and will try my best to comply with it in the future. What would you like me to do for this specific PR to help get it merged?

schildbach:
See #2043. In IntelliJ's inspection tool, it's just one click per type of refactoring. Though I admit I review all the changes and correct some formatting issues here and there, so yes you're right it's more work.

tools/src/main/java/org/bitcoinj/tools/BlockImporter.java
```diff
@@ -42,21 +42,26 @@ public static void main(String[] args) throws BlockStoreException, VerificationE
         new Context(params);
         
         BlockStore store;
-        if (args[1].equals("H2")) {
```
schildbach, line 45: This particular refactoring I'm not very fond of, because of the ugly switch syntax. Of course the repeated use of `args[1]` is ugly, but this could also be assigned to a temporary value. What do you think? In general, do you prefer `switch` to `if`?

schildbach: Btw. just a nitpick, but it's also not a JDK7 or JDK8 upgrade. Switch has been available from Java day one afair.

msgilligan: I think the switch syntax (despite the ugly way `break` is required) is much more readable. It's clear that only one of the `case` s ever fires and it's really easy to scan and find the `case` that you're looking for.

msgilligan: Also note that further improvements to `switch` are coming in newer JDKs.

msgilligan:
> separate commits per IDE suggestion. 

I had misunderstood this as one **PR** per IDE suggestion.

msgilligan:
> See #2043.

Approved. When you merge it to `master`, I'll rebase this PR and we'll see what's left.


> In IntelliJ's inspection tool, it's just one click per type of refactoring. 

I use IntelliJ, but I didn't know you could do that. Cool!

schildbach:
I've just rebased this and I think the only change that's left is the switch migration.

tools/src/main/java/org/bitcoinj/tools/WatchMempool.java
```diff
@@ -33,12 +33,10 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.collect.ImmutableList;
-
 public class WatchMempool {
     private static final Logger log = LoggerFactory.getLogger(WatchMempool.class);
     private static final NetworkParameters PARAMS = MainNetParams.get();
-    private static final ImmutableList<Transaction> NO_DEPS = ImmutableList.of();
```
micheal-swiggs, line 41: @msgilligan why replace `ImmutableList` with `emptyList()`? Just curious.

msgilligan: We've been migrating away from the Guava `ImmutableList` and related classes.  `Collections.emptyList()` is the best available replacement in JDK 8 for `ImmutableList.of()`.

msgilligan:
> I've just rebased this and I think the only change that's left is the switch migration.

I rebased and force pushed. Yes only the switch migration is left, so I renamed the PR as well.

schildbach:
Merged.

PR #1998, opened 2020-05-26
Improvements to Networks class

msgilligan:
* Replace Guava with JDK classes/methods
* Remove unnecessary wildcard types
* Improve comments
* Note that in Bech32 RegTest & TestNet have different HRPs.

### DIFF
--- a/core/src/main/java/org/bitcoinj/params/Networks.java
+++ b/core/src/main/java/org/bitcoinj/params/Networks.java
@@ -17,10 +17,11 @@
 package org.bitcoinj.params;
 
 import org.bitcoinj.core.NetworkParameters;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Lists;
 
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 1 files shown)


schildbach:
Thanks! I think the commits which are aiming to replace Guava should be separated into a different PR. Imho this needs discussion.

While the comment improvements could be merged right away.

Are the "unnecessary wildcard types" suggested by some code inspection? If yes, are there more such cases throughout the framework?

msgilligan:
> Thanks! I think the commits which are aiming to replace Guava should be separated into a different PR.

Will do.

> Imho this needs discussion.

Of course.

> Are the "unnecessary wildcard types" suggested by some code inspection?

No, it's just one of the guidelines in Effective Java, 3rd Edition. In general if you have a `List` of some type (e.g. `NetworkParameters`) it perfectly legal to put subclasses in the `List`. The `? extends` is used when you want to refer to an arbitrary generic list that _may_ have been declared with a subtype of `NetworkParameters` (e.g. a `List<AbstractBitcoinNetParams>` for which client code may expect all objects in the list to support the methods of the subclass.)

In the case of `Networks` there is only one _type of List_: a list of objects that implement `NetworkParameters`. So we can simplify the internal code and API by dropping the `? extends`.

> are there more such cases throughout the framework?

Possibly, but I don't know of an automated way to find them. It's possible there are code quality tools that might find them.

schildbach:
Could you rebase this PR so we can see what's left?

msgilligan:
> Could you rebase this PR so we can see what's left?

Done: https://github.com/bitcoinj/bitcoinj/pull/1998/commits/5ed4157297bd8330eb3dcc301a4cb35736705ef8

Note that Java 9 add `Set.of()` and Java 10 adds `Set.copyOf()` which could further simplify this class. At some point Android de-sugaring should allow us to use these methods.

schildbach:
Merged.

PR #1999, opened 2020-05-27
Replace Guava Stopwatch with custom utility class

msgilligan:
We were only using a subset of the Guava Stopwatch functionality so this
simple utility class removes yet another dependency on Guava.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerGroup.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerGroup.java
@@ -18,7 +18,7 @@
 package org.bitcoinj.core;
 
 import com.google.common.annotations.*;
-import com.google.common.base.*;
+import com.google.common.base.Throwables;
 import com.google.common.collect.*;
 import com.google.common.util.concurrent.*;
 import net.jcip.annotations.*;
--- a/core/src/main/java/org/bitcoinj/crypto/KeyCrypterScrypt.java
+++ b/core/src/main/java/org/bitcoinj/crypto/KeyCrypterScrypt.java
@@ -17,7 +17,7 @@
 
 package org.bitcoinj.crypto;
 
-import com.google.common.base.Stopwatch;
+import org.bitcoinj.utils.Stopwatch;
 import com.google.protobuf.ByteString;
 import org.bitcoinj.core.Utils;
 import org.bitcoinj.wallet.Protos;
--- a/core/src/main/java/org/bitcoinj/crypto/MnemonicCode.java
+++ b/core/src/main/java/org/bitcoinj/crypto/MnemonicCode.java
@@ -22,7 +22,7 @@ import org.bitcoinj.core.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.base.Stopwatch;
+import org.bitcoinj.utils.Stopwatch;
 
 import java.io.BufferedReader;
 import java.io.FileNotFoundException;
--- a/core/src/main/java/org/bitcoinj/params/AbstractBitcoinNetParams.java
+++ b/core/src/main/java/org/bitcoinj/params/AbstractBitcoinNetParams.java
@@ -20,7 +20,6 @@ package org.bitcoinj.params;
 import static com.google.common.base.Preconditions.checkState;
 
 import java.math.BigInteger;
-import java.util.concurrent.TimeUnit;
 
 import org.bitcoinj.core.Block;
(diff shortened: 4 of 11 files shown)


schildbach:
I must say I'm not a fan of re-inventing Guava classes like this. If your aim is to get rid of Guava completely, can you say how much work will be needed and how much code duplicated? I might be wrong, but it is my impression we're using Guava a lot (e.g. the Futures support) and might use it in future. In the past we got rid of usages of Guava when it was suggested to do so by the Guava documentation (usually migrations to JDK).

msgilligan:
> I must say I'm not a fan of re-inventing Guava classes like this.

I thought you might say that. My goal yesterday was to eliminate all usage of Guava in `org.bitcoinj.params` except `Preconditions.checkState`. So at first, I just replaced the usage of the Guava Stopwatch in `AbstractBitcoinNetParams`. But then I imagined you might ask that we do the same thing across the codebase, so I looked for other instances and looked to see what could be done.

(My initial goal yesterday was just to understand the `Networks` class a little better, and I saw opportunity for improvement.  We can discuss that in PR #1998 though.)

> If your aim is to get rid of Guava completely, can you say how much work will be needed and how much code duplicated?

Eventually removing Guava completely (at least from a subset of bitcoinj) is a personal Goal, but I realize it is a goal you may not (yet) share. Completely removing Guava is a big project and I don't know exactly how much work would be needed.

> I might be wrong, but it is my impression we're using Guava a lot (e.g. the Futures support)

I'm hoping we can migrate from Java Futures to `CompletableFuture` in the future when we migrate to a minimum of JDK 8.

> In the past we got rid of usages of Guava when it was suggested to do so by the Guava documentation (usually migrations to JDK).

Yes, I think we should absolutely continue to migrate to the JDK where possible, easy, and/or safe. The cases recommend by the Guava JavaDoc are cases we will agree on. We can decide on other possibilities on a case-by-case basis.

I knew this PR would generate discussion. Stopwatch is pretty common pattern, the implementation we need is small and it is only used for logging (so seems like a relatively low-risk change.) The only reason I started down the path of replacing Stopwatch is that it was the last usage of Guava within the `.params` package (besides the very pervasive usage of `Preconditions`.

msgilligan:
@schildbach And I thought about writing these unit tests in Kotlin haha.

schildbach:
Side note: from the experience of friend's projects I can say they got rid of Guava when they migrated their codebase to Kotlin.

schildbach:
About using the Java8 API on Android: it requires API level 26 (or the very new [API desugaring](https://developer.android.com/studio/releases/gradle-plugin#j8-library-desugaring)). My estimate is that `minSdkVersion=26` is feasible from about 2022 on. The desugaring unfortunately has very high tool requirements so that won't be available immediately either. The good news is, the packaging of Kotlin in Debian [is being worked on again](https://wiki.debian.org/Kotlin) (being a pre-requisite to modern Gradle versions).

msgilligan:
> they got rid of Guava when they migrated their codebase to Kotlin.

I'm a big fan of Kotlin and Groovy, but I am reluctant to use them in OSS libraries that target a wide audience because they introduce dependencies on their standard libraries. You also need to be aware of [Java-to-Kotlin interop issues.](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html)

(I'm looking forward to using and learning more about Kotlin in unit tests, Android applications, and JavaFX applications.)

msgilligan:
> My estimate is that `minSdkVersion=26` is feasible from about 2022 on.

That seems reasonable to me. I certainly view migration away from Guava as a long-term/background project.

> The desugaring unfortunately has very high tool requirements so that won't be available immediately either.

I'm hoping to experiment with it in an Android app using **bitcoinj** later this summer.

> The good news is, the packaging of Kotlin in Debian [is being worked on again](https://wiki.debian.org/Kotlin) (being a pre-requisite to modern Gradle versions).

That's great news!

schildbach:
How do we continue with this? At the moment I'm tempted to close this, because there is no stopwatch replacement besides our own re-implementation.

msgilligan:
Let's go ahead and close it. We have more important things to work on. When we get to the point where the Stopwatch is the only thing preventing us from removing a Guava dependency somewhere we can re-examine it.

schildbach:
See https://github.com/bitcoinj/bitcoinj/pull/2731 for an alternative suggestion to remove usage of Guava Stopwatch.

msgilligan:
I'm approving PR #2731 to be merged. It seems the best approach at this time for replacing Guava stopwatch.

PR #2013, opened 2020-06-29
Coin: conversion between satoshi/long and bitcoin/BigDecimal

msgilligan:
* Add static utility methods: bitcoinToSatoshi(), satoshiToBitcoin()
* Add static factory method: ofBitcoin()
* Add instance method: toBitcoin()
* Add hamcrest-library to testImplementation for BigDecimal comparison

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Coin.java
+++ b/core/src/main/java/org/bitcoinj/core/Coin.java
@@ -119,6 +119,36 @@ public final class Coin implements Monetary, Comparable<Coin>, Serializable {
     }
 
     /**
+     * Convert a decimal amount of bitcoin into satoshis.
+     *
+     * @param coins number of coins
+     * @return number of satoshis
+     */
+    public static long bitcoinToSatoshi(BigDecimal coins) {
+        return coins.movePointRight(SMALLEST_UNIT_EXPONENT).longValueExact();
+    }
+
+    /**
+     * Convert an amount in satoshis to an amount in bitcoin.
+     *
+     * @param satoshis number of satoshis
+     * @return number of bitcoins
+     */
+    public static BigDecimal satoshiToBitcoin(long satoshis) {
+        return new BigDecimal(satoshis).movePointLeft(SMALLEST_UNIT_EXPONENT);
+    }
+
+    /**
+     * Convert a decimal amount of bitcoin into satoshis.
+     *
+     * @param coins number of coins
+     * @return {@code Coin} object containing value in satoshis
+     */
+    public static Coin ofBitcoin(BigDecimal coins) {
+        return  Coin.valueOf(bitcoinToSatoshi(coins));
+    }
+
+    /**
      * Parses an amount expressed in the way humans are used to.
      * 
      * @param str string in a format understood by {@link BigDecimal#BigDecimal(String)}, for example "0", "1", "0.10",
@@ -129,7 +159,7 @@ public final class Coin implements Monetary, Comparable<Coin>, Serializable {
      */
     public static Coin parseCoin(final String str) {
         try {
-            long satoshis = new BigDecimal(str).movePointRight(SMALLEST_UNIT_EXPONENT).longValueExact();
+            long satoshis = bitcoinToSatoshi(new BigDecimal(str));
             return Coin.valueOf(satoshis);
         } catch (ArithmeticException e) {
             throw new IllegalArgumentException(e); // Repackage exception to honor method contract
@@ -273,6 +303,15 @@ public final class Coin implements Monetary, Comparable<Coin>, Serializable {
      */
     public long longValue() {
         return this.value;
+    }
+
+    /**
+     * Convert to number of bitcoin
+     *
+     * @return decimal number of bitcoin
+     */
+    public BigDecimal toBitcoin() {
+        return satoshiToBitcoin(this.value);
     }
 
     private static final MonetaryFormat FRIENDLY_FORMAT = MonetaryFormat.BTC.minDecimals(2).repeatOptionalDecimals(1, 6).postfixCode();
--- a/core/src/main/java/org/bitcoinj/core/Coin.java
+++ b/core/src/main/java/org/bitcoinj/core/Coin.java
@@ -119,33 +119,43 @@ public final class Coin implements Monetary, Comparable<Coin>, Serializable {
     }
 
     /**
-     * Convert a decimal amount of bitcoin into satoshis.
+     * Convert a decimal amount of BTC into satoshis.
      *
      * @param coins number of coins
      * @return number of satoshis
      */
-    public static long bitcoinToSatoshi(BigDecimal coins) {
+    public static long btcToSatoshi(BigDecimal coins) {
         return coins.movePointRight(SMALLEST_UNIT_EXPONENT).longValueExact();
     }
 
     /**
-     * Convert an amount in satoshis to an amount in bitcoin.
+     * Convert an amount in satoshis to an amount in BTC.
      *
      * @param satoshis number of satoshis
-     * @return number of bitcoins
+     * @return number of bitcoins (in BTC)
      */
-    public static BigDecimal satoshiToBitcoin(long satoshis) {
+    public static BigDecimal satoshiToBtc(long satoshis) {
         return new BigDecimal(satoshis).movePointLeft(SMALLEST_UNIT_EXPONENT);
     }
 
     /**
-     * Convert a decimal amount of bitcoin into satoshis.
+     * Convert a decimal amount of BTC into satoshis.
      *
-     * @param coins number of coins
+     * @param coins number of coins (in BTC)
      * @return {@code Coin} object containing value in satoshis
      */
-    public static Coin ofBitcoin(BigDecimal coins) {
-        return  Coin.valueOf(bitcoinToSatoshi(coins));
+    public static Coin ofBtc(BigDecimal coins) {
+        return  Coin.valueOf(btcToSatoshi(coins));
+    }
+
+    /**
+     * Convert a long integer number of satoshis to a Coin
+     *
+     * @param satoshis number of satoshis
+     * @return {@code Coin} object containing value in satoshis
+     */
+    public static Coin ofSat(long satoshis) {
+        return  Coin.valueOf(satoshis);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/bitcoinj/core/Coin.java
```diff
@@ -118,6 +118,36 @@ public static Coin valueOf(final int coins, final int cents) {
...
+        return new BigDecimal(satoshis).movePointLeft(SMALLEST_UNIT_EXPONENT);
+    }
+
+    /**
+     * Convert a decimal amount of bitcoin into satoshis.
+     *
+     * @param coins number of coins
+     * @return {@code Coin} object containing value in satoshis
+     */
+    public static Coin ofBitcoin(BigDecimal coins) {
```
schildbach, line 147: I think this should be named `valueOf` too, for consistency.

core/src/main/java/org/bitcoinj/core/Coin.java
```diff
@@ -118,6 +118,36 @@ public static Coin valueOf(final int coins, final int cents) {
         return coin;
     }
 
+    /**
+     * Convert a decimal amount of bitcoin into satoshis.
+     *
+     * @param coins number of coins
+     * @return number of satoshis
+     */
+    public static long bitcoinToSatoshi(BigDecimal coins) {
```
schildbach, line 127: Hmm, do we need this and the counterpart method as _public_ methods? Actually the idea of the `Coin` class is that people don't fiddle around with monetary amounts themselves, so methods like this should not be needed.

Also I'm not so happy about the use of the word bitcoin in the method names – satoshis are Bitcoin too. Perhaps BTC would be more fitting, as that implies the denomination?

schildbach:
Just as a background: back when we introduced the `Coin` class one of our goals was to navigate users away from using the BigInteger/BigDecimal classes as they're slow and awkward to use (and I vaguely remember nasty implementation bugs/details).

core/src/main/java/org/bitcoinj/core/Coin.java
msgilligan: I think it needs to have a different name because it takes a "different" argument.

    long satoshis = 50000000
    BigDecimal bitcoin = new BigDecimal(satoshis)
    Coin coin = Coin.valueOf(bitcoin)

What is the value of `coin` at this point?

Because both methods take `Number` types but using different units (orders of magnitude) it doesn't seem intuitive to use method overloading (i.e. the same name with different types) of the same (static) method name.

Ideally, I'd have:

    public static Coin of(long satoshis)
    public static Coin ofBitcoin(BigDecimal bitcoin)

In theory we could also have:

    public static Coin of(BigDecimal satoshis)
    public static Coin ofBitcoin(long bitcoin)

but I don't think those methods are really necessary.

core/src/main/java/org/bitcoinj/core/Coin.java
msgilligan: > do we need this and the counterpart method as _public_ methods?

I thought about keeping them `private`, but ultimately decided they were useful as `public`.

> Actually the idea of the `Coin` class is that people don't fiddle around with monetary amounts themselves, so methods like this should not be needed.

 I view **bitcoinj** as the reference implementation of Bitcoin constants and types for the JVM ecosystem. There are clearly cases where people using **bitcoinj** will have `long` values of satoshi they need to convert to (big)decimal values of bitcoin. I think there should be a reference conversion function and that it should live in the `Coin` class.

There are use cases where people will be developing bitcoin apps where they're going to need to use a charting library that takes BigDecimal (or maybe float, ugh). They want to use a math library that works with BigDecimal, etc. They will need to convert back and forth to `Coin` and even to just `long` when they interact with transactions, etc.

In ConsensusJ I ended up writing this conversion code twice (because I didn't want the two libraries to have dependencies on each other.) And I didn't write the code as well as you and Mike Hearn did! I just discovered the shift point functions this weekend looking at `Coin.java`!

So it's nice to have the factory method and the `toBitcoin` instance method, but I also think bitcoinj clients will want to be able to convert `long` to `BigDecimal` and vice-versa without unnecessary object creation.


> 
> Also I'm not so happy about the use of the word bitcoin in the method names – satoshis are Bitcoin too. Perhaps BTC would be more fitting, as that implies the denomination?

I'd be ok with BTC, but (lower case) bitcoin is also a [denomination](https://en.bitcoin.it/wiki/Units#Table_of_all_units).

msgilligan: >the idea of the Coin class is that people don't fiddle around with monetary amounts themselves, so methods like this should not be needed.

I agree that we want to discourage users from doing this. Maybe I should add some text to the class-level JavaDoc comment talking about how to use `Coin`, to avoid using floating-point and to use `long` and `BigDecimal` when you absolutely need a primitive-ish type to interact with other libraries.

msgilligan:
> Just as a background: back when we introduced the `Coin` class one of our goals was to navigate users away from using the BigInteger/BigDecimal classes as they're slow and awkward to use (and I vaguely remember nasty implementation bugs/details).
> 
> And for Bitcoin they're not needed as the max number of satoshis nicely fits into a long.

This all makes sense. As I said above:

> Maybe I should add some text to the class-level JavaDoc comment talking about how to use Coin, to avoid using floating-point and to use long and BigDecimal when you absolutely need a primitive-ish type to interact with other libraries.

core/src/main/java/org/bitcoinj/core/Coin.java
schildbach: I see your point. I would make the denomination clearer however:

```
public static Coin ofSat(long satoshis)
public static Coin ofBtc(BigDecimal bitcoin)
```

Yes, "bitcoin" is also a denomination, but it's also a currency especially when written with an uppercase B (which we're forced to due to camel-casing).

core/src/main/java/org/bitcoinj/core/Coin.java
schildbach: What do you think about extending the Coin API to allow for this?

- BigDecimal→long: `long sat = Coin.ofBtc(btc).toSat()`
- long→BigDecimal: `BigDecimal btc = Coin.ofSat(sat).toBtc()`

msgilligan: > What do you think about extending the Coin API to allow for this?
> 
> * BigDecimal→long: `long sat = Coin.ofBtc(btc).toSat()`
> * long→BigDecimal: `BigDecimal btc = Coin.ofSat(sat).toBtc()`

This looks pretty good. But for high-performance or large numbers of conversions there is an object creation and disposal that could be avoided with pure `static` functions.

msgilligan: I renamed `Bitcoin` to `Btc` in the appropriate places and added the `ofSat` and `toSat` methods (don't know if you want to deprecate the duplicate `valueOf` and `longValue` methods)

msgilligan:
@schildbach I'm leaving the static conversion methods as `public` but can make them `private` if you insist. As I said above, I think there are use cases (processing millions of numbers) where the overhead of class creation and disposal should be avoided.

Let me know what you think the next step(s) should be.

msgilligan:
@schildbach Rebased. Can you take another look and continue the discussion?

core/src/main/java/org/bitcoinj/core/Coin.java
```diff
@@ -118,6 +118,46 @@ public static Coin valueOf(final int coins, final int cents) {
...
+     *
+     * @param satoshis number of satoshis
+     * @return number of bitcoins (in BTC)
+     */
+    public static BigDecimal satoshiToBtc(long satoshis) {
+        return new BigDecimal(satoshis).movePointLeft(SMALLEST_UNIT_EXPONENT);
+    }
+
+    /**
+     * Convert a decimal amount of BTC into satoshis.
```
schildbach, line 142: Shouldn't this read "…to a Coin", similar to the below method?

core/src/test/java/org/bitcoinj/core/CoinTest.java
```diff
@@ -80,6 +84,54 @@ public void testValueOf() {
...
+    public void testbtcToSatoshi() {
+        assertEquals(Long.MIN_VALUE, btcToSatoshi(new BigDecimal("-92233720368.54775808")));
+        assertEquals(0L, btcToSatoshi(BigDecimal.ZERO));
+        assertEquals(COIN.value, btcToSatoshi(BigDecimal.ONE));
+        assertEquals(Long.MAX_VALUE, btcToSatoshi(new BigDecimal("92233720368.54775807")));
+    }
+
+    @Test(expected = ArithmeticException.class)
+    public void testofBtcTooSmall() {
+        btcToSatoshi(new BigDecimal("-92233720368.54775809"));
```
schildbach, line 98: Perhaps add a comment like 'a bit smaller than Long.MIN_VALUE'?

core/src/test/java/org/bitcoinj/core/CoinTest.java
```diff
@@ -80,6 +84,54 @@ public void testValueOf() {
...
+    }
+
+    @Test(expected = ArithmeticException.class)
+    public void testofBtcTooBig() {
+        btcToSatoshi(new BigDecimal("92233720368.54775808"));
```
schildbach, line 103: …Long.MAX_VALUE…

schildbach (commented):
I've added a few comments. On top of that, I think all of the added test method should have the letter after 'test' capitalized. E.g. `testBtc…` rather than `testbtc…`.

core/src/main/java/org/bitcoinj/core/Coin.java
msgilligan: fixed in 165341fbf4a027f9375561964b52a0e88f9073d1 (and made similar changes to other static factory methods)

core/src/test/java/org/bitcoinj/core/CoinTest.java
msgilligan: Fixed in 165341fbf4a027f9375561964b52a0e88f9073d1

core/src/test/java/org/bitcoinj/core/CoinTest.java
msgilligan: Fixed in 165341fbf4a027f9375561964b52a0e88f9073d1

msgilligan:
> I think all of the added test method should have the letter after 'test' capitalized. E.g. `testBtc…` rather than `testbtc…`.

Fixed in 165341fbf4a027f9375561964b52a0e88f9073d1.

core/src/test/java/org/bitcoinj/core/CoinTest.java
```diff
@@ -116,10 +168,21 @@ public void testAdditionOverflow() {
     }
 
     @Test(expected = ArithmeticException.class)
-    public void testSubstractionUnderflow() {
+    public void testSubtractionUnderflow() {
```
msgilligan, line 171: Note: I fixed a typo here.

schildbach:
I changed the names of 4 test methods that were actually testing a different method, and merged.

PR #2015, opened 2020-08-01
Test improvement: removed sensitive equality (test smell)

eas5:
This is a test refactoring.

**Problem:**
The Sensitive Equality test smell occurs when the toString() method is used within a test method. Test methods verify objects by invoking the default toString() method of the object and comparing the output against a specific string. Changes to the implementation of toString() might result in failure. 

**Solution:**
The correct approach is to implement a custom method within the object to perform this comparison. In this case, no new method creation was required.

**Result (sample):**
_Before:_
`assertEquals("3N25saC4dT24RphDAwLtD8LUN4E2gZPJke", address.toString());`
_After:_
`assertEquals("3N25saC4dT24RphDAwLtD8LUN4E2gZPJke", address.toBase58());`


schildbach:
The `toString() / fromString()` API is actually the recommended API to use if you deal with addresses. Note that there is also SegwitAddress, which doesn't have `toBase58() / fromBase58()` but rather `toBech32() / fromBech32()`.

Perhaps we should document the `toString()` overrides to make it more clear.

schildbach:
I'm closing this, as I think testing `.toString()` is the right thing to do.

PR #2017, opened 2020-08-10
Add a (currently broken) test that demonstrates the compareTo collision

msgilligan:
This test illustrates the problem with compareTo not finding a difference between p2sh and non-p2sh addresses with the same `bytes[]`.


msgilligan:
Closing. This test is included in PR #2018

PR #2018, opened 2020-08-10
Add Comparable to Address remove from PrefixedChecksummedBytes

msgilligan:
This is a draft implementation of my proposed solution discussed in PR #1854.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
@@ -23,6 +23,7 @@ import java.util.Objects;
 
 import javax.annotation.Nullable;
 
+import com.google.common.primitives.UnsignedBytes;
 import org.bitcoinj.params.Networks;
 import org.bitcoinj.script.Script.ScriptType;
 
@@ -212,5 +213,21 @@ public class LegacyAddress extends Address {
     @Override
     public LegacyAddress clone() throws CloneNotSupportedException {
         return (LegacyAddress) super.clone();
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @param o other {@code Address} object
+     * @return comparison result
+     */
+    @Override
+    public int compareTo(Address o) {
+        int result = compareAddressPartial(o);
+        if (result != 0) return result;
+
+        // Compare version byte and finally the {@code bytes} field itself
+        result = Integer.compare(getVersion(), ((LegacyAddress) o).getVersion());
+        return result != 0 ? result : UnsignedBytes.lexicographicalComparator().compare(this.bytes, o.bytes);
     }
 }
--- a/core/src/main/java/org/bitcoinj/core/PrefixedChecksummedBytes.java
+++ b/core/src/main/java/org/bitcoinj/core/PrefixedChecksummedBytes.java
@@ -41,7 +41,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
  * keys exported using Bitcoin Core's dumpprivkey command.
  * </p>
  */
-public abstract class PrefixedChecksummedBytes implements Serializable, Cloneable, Comparable<PrefixedChecksummedBytes> {
+public abstract class PrefixedChecksummedBytes implements Serializable, Cloneable {
     protected final transient NetworkParameters params;
     protected final byte[] bytes;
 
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/bitcoinj/core/LegacyAddress.java
```diff
@@ -213,4 +214,20 @@ public String toString() {
...
         return (LegacyAddress) super.clone();
     }
+
+    @Override
+    public int compareTo(Address o) {
+        // First compare netParams
+        int result = this.params.getId().compareTo(o.params.getId());
+        if (result != 0) return result;
+
+        // Then compare Legacy vs Segwit
```
schildbach, line 224: Couldn't the Legacy/Segwit comparison go to Address.compareTo()?

core/src/main/java/org/bitcoinj/core/PrefixedChecksummedBytes.java
```diff
@@ -82,8 +82,10 @@ public PrefixedChecksummedBytes clone() throws CloneNotSupportedException {
 
     /**
      * This implementation uses an optimized Google Guava method to compare {@code bytes}.
+     * @deprecated Can't implement comparable because of Segwit changes (subclasses with
+     * added value components, eg. version field based on p2sh in LegacyAddress)
      */
-    @Override
+    @Deprecated
     public int compareTo(PrefixedChecksummedBytes o) {
```
schildbach, line 89: Do we still need this old impl? Since the `Comparable` interface has been removed from this class, I assume it would not serve backwards compatibility much?

msgilligan: It's not needed internally. I don't know if there are any external projects depending on it. I had no choice but to remove the `Comparable` interface (because I couldn't add it to `Address` otherwise) but I left this here just in case some other project needs the comparator code. But I like the idea of deleting it.

core/src/main/java/org/bitcoinj/core/LegacyAddress.java
msgilligan: Yeah, we could put that there and then call an abstract method with a different signature that the two subclasses would implement. I'm going to take a break from work for the next week, so I'll look at doing this when I return.

core/src/main/java/org/bitcoinj/core/PrefixedChecksummedBytes.java
msgilligan: Fixed in 2a204b79221be0292a7fe1ce3c8ed97c045b1501 -- I'll squash when everything else is resolved and we change the PR from "Draft" to "Ready to Merge".

core/src/main/java/org/bitcoinj/core/LegacyAddress.java
msgilligan: How about this: 62ab82fab214f112880ed7bccee0a7ab6b8dd2c7 ?

msgilligan:
@schildbach Can you review this one? If it looks good, I'll squash it, add a few more comments, and change it from "DRAFT" to "Ready to Merge".

core/src/main/java/org/bitcoinj/core/LegacyAddress.java
schildbach: Yes, that's the general idea. But is there a reason for the additonal method? Couldn't it all be contained in three `compareTo()` impls? The leaves would call `super.compareTo()` for the comparison on `Address` level.

msgilligan: > Yes, that's the general idea. But is there a reason for the additonal method? Couldn't it all be contained in three `compareTo()` impls? The leaves would call `super.compareTo()` for the comparison on `Address` level.

I thought about doing it that way (#1), but decided against it because (a) it seemed wrong to have an implementation of `compareTo` for `Address` that (as a standalone method, that would be inherited by subclasses) was incorrect. It also leaves no unimplemented `abstract` method that subclasses might implement.

Instead I implemented (#2) a `compa,,reTo` method that along with the `abstract` `compareAddressTo` method forces subclasses to implement comparison between two objects of the subclass type.

A third way of doing it (#3) would be to have an `abstract` `compareTo` in `Address` and provide a protected method that the subclass implementation of `compareTo` can call that will provide the ordering by netParams and subclass.

#3 might be a better way to go. I want to get this fixed, so I'm happy with any of the methods, but after implementing #2, I would say that #3 is my first choice, #2 my second choice, and #1 my last choice.

So let me know what you want me to do. I can implement #3 or #1 or leave it as it is.

schildbach: After your explanation I also think that (1) is clearly inferiour. As for picking (2) or (3), I think I have no preference so I'll leave it up to you.

msgilligan: Thanks @schildbach , (3) is implemented (along with some JavaDoc cleanup), squashed, and is passing tests.

msgilligan:
@schildbach This is ready for final review.

schildbach:
I removed a few unused imports from this PR and merged. Thanks for sticking to this!

Issue #1984, opened 2020-04-24
Resolve (MS-Windows) unit test issues.


msgilligan:
This is an umbrella issue/task for resolving problems with unit tests that are currently manifesting (mostly) under MS-Windows.

There are at least two existing issues that I believe are related to this task:

* Issue #840 - Unit tests that test NIO are flaky
* Issue #1549 - network abstraction tests issues

Now that we are able to run tests under Windows via Github Actions, it should be much easier to solve all the Windows-related testing problems. I believe there are (or were) related intermittent failures that can occur on macOS or Linux.

I believe the solution is non-trivial and will be broken down into a handful of sub-issues and will require multiple Pull Requests to resolve. This issue should remain open until we have all tests working reliable under Windows and Windows tests activated (re-activated) in the Github Actions test configuration file(s).

msgilligan:
**Action Plan**

- [x] Add unit tests under Windows on CI.
- [ ] Disable Windows unit tests until issues are resolved
- [ ] Fix Windows and intermittent unit-test failures
- [ ] Re-enable Windows unit tests.

This _comment_ will be updated as plans change and new information is discovered.

msgilligan:
I have a handful of ideas for how we can approach this and various subtasks that may be helpful:

1. Separate unit tests from integration tests (see [Configuring integration tests](https://docs.gradle.org/current/userguide/java_testing.html#sec:configuring_java_integration_tests))
1. Replace (or supplement) integration tests with true unit tests that test the same thing only faster and more reliably. Doing this may require creating test fixtures, creating mocks, adding a mocking library, etc.
1. Re-enable excluded tests `PeerTest*`, `TransactionBroadcastTest*`, `DnsDiscoveryTest*` once we figure out how to run them properly/reliably.
1. Not directly related to the main issue, but while we're considering test environment improvements it would be nice to add a CI job that runs the tests in an Android environment.
1. Consider Upgrading to JUnit 5 (see [Using JUnit 5 with Gradle](https://www.baeldung.com/junit-5-gradle) and [Android Unit Testing with JUnit5](https://medium.com/@boonkeat/android-unit-testing-with-junit5-d1b8f9c620b6))
1. Consider using Spock Framework -- it has powerful Mocking capabilities and a very expressive DSL for tests. (One downside is it doesn't work in the Android environment)
1. Consider using Kotlin language tests and tools.  This may offer some of the expressiveness of Groovy/Spock but still run in the Android environment.

[...]

PR #2022, opened 2020-08-18
Upgrade to JUnit 5 using “Vintage” engine

msgilligan:
Unfortunately, this requires Gradle 4.6 or later. Upgrading to JUnit 5
allows us to start using JUnit 5 features while still running JUnit 4
tests with the “vintage” engine.

Upgrading to JUnit 5 was one of the proposed steps to improving our testing capability in Issue #1984.   (See idea number 5 on this comment: https://github.com/bitcoinj/bitcoinj/issues/1984#issuecomment-619239187 )

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -23,6 +23,7 @@ dependencies {
     testImplementation 'com.h2database:h2:1.3.176'
     testImplementation 'nl.jqno.equalsverifier:equalsverifier:3.9'
     testImplementation 'org.hamcrest:hamcrest-library:2.2'
+    testRuntimeOnly "org.junit.vintage:junit-vintage-engine:5.8.2"
 }
 
 sourceCompatibility = 8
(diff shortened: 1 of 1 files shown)


schildbach:
That's the way to go imho! – as soon as we can meet the Gradle dependency.

msgilligan:
We should consider moving the "integration" tests to a new subproject -- we can call it integration-test -- and have that module only run if Gradle 4.6 or later. That way we can still build the core ( and run "unit" tests) on Debian, but begin converting some tests to JUnit 5 (and possibly Kotlin)

schildbach:
That's basically the currently disabled tests, right? Sounds fine to me then.

msgilligan:
I rebased and force-pushed. Now that `master` is already using JDK 8+ for `core`, all this PR needs to do is add the `useJUnitPlatform()` to the Gradle test configuration block and add the dependency for `junit-vintage-engine`:

```
testRuntimeOnly "org.junit.vintage:junit-vintage-engine:5.8.2"
```

msgilligan:
Since this PR was created have created the `integration-test` module which uses JUnit 5 and contains both `junit-jupiter-api` tests and older tests that rely on the `junit-vintage-engine`.

The current plan is to keep `core` tests on JUnit 4 and use JUnit 5 (for both "vintage" and new Jupiter API) integration tests.

We can open a new PR in a few years when/if we're ready to upgrade `core` to JUnit 5.

PR #2027, opened 2020-08-22
Remove unused mock sleep support from Utils

msgilligan:
This is apparently unused, so I think it should be removed. If we need such a capability I recommend injecting it via a constructor rather than using a Utils object and static variables/methods.

(The mock time capability is causing problems in the unit tests that I'm currently debugging and it should probably also be removed/refactored, but it is more widely used so I intend to fix/workaround the issue instead)


schildbach:
I agree. Merged!

PR #2028, opened 2020-08-22
Fix bogus test failures using @BeforeClass initializer for Utils.resetMocking

msgilligan:
This fixes (or works around) the failures we're currently seeing on GitHub Actions Ubuntu and most (but not all) of the failures we were seeing on Github Actions Windows before we disabled testing on Windows.


schildbach:
Sounds great! I'll take a look at this tomorrow.

schildbach:
Thanks for figuring this out!

I must admit I don't really understand the difference between a static initializer and `@BeforeClass`, but I think the change certainly can't do any harm and it seems to be the way JUnit is supposed to be used.

Merged.

schildbach:
We've got one failing performance test: https://github.com/bitcoinj/bitcoinj/runs/1025378727
Later builds succeed, so I believe the failing test is a separate issue.

PR #2030, opened 2020-08-22
Github Actions: Re-enable windows-latest

msgilligan:
This is a draft PR that we can keep re-basing against `master` until all the Windows-blockers are fixed.

PR #2031, opened 2020-08-22
Fix incorrect AppDataDirectory tests on Windows

msgilligan:
These tests were failing on Windows. This PR, plus PR #2028 will fix all the failures I'm seeing on `windows-latest` except one (`SPVBlockStoreTest.oneStoreDelete()` which has a comment "Used to fail on Widows")

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -7,9 +7,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


schildbach:
I merged the two assert message fixes right away, but can you rebase the rest on your PR #2030 so we can actually see the improvement of this PR?

msgilligan:
Rebased onto PR #2030

schildbach:
Verified that the fix actually works and merged (just the fix).

PR #2036, opened 2020-08-29
Add integration tests subproject: with JUnit 5 & Kotlin

msgilligan:
Add integration tests subproject: `integtest`

* JUnit 5 (API & Engine)
* Tests compile/run as JDK 8+
* Tests require Gradle 4.6+ 
* Tests can be in either Java or Kotlin
* One sample test — in Kotlin

The Kotlin plugin requires Gradle 5.3 or newer. So:

* settings.gradle checks for 5.3+ before loading integtest
* .travis.yml adds jdk8/Gradle5.3.1 and jdk11/Gradle5.3.1
* Gitlab CI upgrades GRADLE_VERSION to 5.3.1


schildbach:
It is my understanding that we create this subproject mainly for tests that currently sporadically fail, because they are dependent on external factors like the network. Your sample test however feels much like a normal unit test that should go to (or stay in) the test class path of core. (I'm aware that lots of the current "unit tests" are not really unit tests, but still.)

Maybe "integration test" is not the right name, would e.g. "networked test" or "smoke test" be better? As a side note, I never encountered the "integ test" abbreviation, personnaly I would prefer to use the full name if no universally accepted abbreviation is at hand.

btw. Debian will be aiming at Gradle 6.4.1 and Kotlin 1.3.30 for Bullseye (Debian 11). They have been making great progress in the past weeks.

msgilligan:
> It is my understanding that we create this subproject mainly for tests that currently sporadically fail, because they are dependent on external factors like the network.

I think the sporadically failing (especially I/O using) tests may need significant rewrites and using JUnit 5 and JUnit 5-based libraries will give us more powerful tools to do so. I also think we may want to define some flags (tags?) that control which group of tests run and I would like to implement this with current generation (JUnit 5) technology. We are also interested in using Kotlin for writing tests, so (given our builds-only-with-Debian-packages goal) a separate subproject for test seems like the best approach.

So, I was imagining this subproject would be for "integration tests" and we would start migrating some of our existing integration tests to this subproject as well as creating new tests here. Although "unit testing" is commonly used as a catch-all term for all automated tests, the most common technical definition of a Java "unit" test is one that tests an individual class or method, and an "integration" test is one that tests the interaction between multiple classes or methods.

But practically speaking, what I think I really want is a place to *begin* refactoring all the tests to JUnit 5.

> Your sample test however feels much like a normal unit test that should go to (or stay in) the test class path of core. (I'm aware that lots of the current "unit tests" are not really unit tests, but still.)

I see your point, here. This was a test that I initially wanted to write with Spock and decided to try it with JUnit 5 and Kotlin (as a learning experience.) So writing this test was what pushed me to try to implement the new subproject. And, of course, by at least one definition, this test is an integration tests (it tests the interaction of 3 of our classes, `List`, and `sort`.)

> Maybe "integration test" is not the right name, would e.g. "networked test" or "smoke test" be better? As a side note, I never encountered the "integ test" abbreviation, personnaly I would prefer to use the full name if no universally accepted abbreviation is at hand.

Yeah, I agree the `integtest` abbreviation sucks.

I think we need to come to an agreement as to what type of tests we'll put in the project and chose a name around that. (Also, I think the idea of this module is tests of the integration of components inside `core`, not integration of  `core` with other modules.)

Maybe it's `core-integration-tests`, maybe it's `core-network-tests`, or maybe the broadest category `core-junit5-tests`.

> btw. Debian will be aiming at Gradle 6.4.1 and Kotlin 1.3.30 for Bullseye (Debian 11). They have been making great progress in the past weeks.

That's great news!

msgilligan:
So aside from the directory name and `AddressComparisonTest.kt`, are you happy with the Gradle and CI file changes?

msgilligan:
> I think we need to come to an agreement as to what type of tests we'll put in the project and chose a name around that. 

Let's continue this part of the discussion in Issue #1984.

msgilligan:
I created a branch from this PR and switched from Java to Kotlin test, here's the WIP PR: https://github.com/bitcoinj/bitcoinj/pull/2381

I upgraded:

* JDK requirement for integration tests to JDK 11 (if we can't use Kotlin, let's at least get latest Collections and Streams)
* JUnit Jupiter and SLF4J dependencies to the latest

I down-graded:

* The minimum Gradle from 5.3 (needed for Kotlin) back to 4.6 (needed for JUnit 5)

msgilligan:
I went ahead and migrated the address comparison/sorting test all the way back to Java 8 and JUnit 4 and put it in `core/src/test`.  See PR #2384.


PR #2381 and PR #2384 replace everything useful in this PR, so I'm going to close it.

PR #2043, opened 2020-09-07
Java level migrations

schildbach:
Apply several automated Java level migrations, one per commit:

* Explicit generic type can be replaced with <>
* try finally replaceable with try with resources
* Anonymous type can be replaced with lambda
* Anonymous type has shorter lambda alternative
* Statement lambda can be replaced with expression lambda
* Lambda can be replaced with method reference
* for loop replaceable with enhanced for loop
* Identical catch branches in try statement
* Replace indexOf() with contains()


schildbach:
I just noticed that I didn't have wallettemplate on my class path. I fixed that and added some refactorings.

msgilligan (approved):
Looks great!

msgilligan:
-217 lines!

schildbach:
Merged.

PR #2044, opened 2020-09-08
wallettemplate: upgrade JavaFX to version 15

msgilligan:
JavaFX 15 is released and is the now the latest supported version. It supports JDK 11 and later — same as the previous release.

Release notes are here:
https://github.com/openjdk/jfx/blob/jfx15/doc-files/release-notes-15.md#release-notes-for-javafx-15


schildbach:
Merged.

PR #2046, opened 2020-09-14
WalletAppKit create fallback keychain when using restoreFromKey

oscarguindzberg:


core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
```diff
@@ -133,6 +133,40 @@ public Builder fromSeed(DeterministicSeed seed, Script.ScriptType outputScriptTy
             return this;
         }
 
+        /**
+         * <p>Add chain from a given key.</p>
```
schildbach, line 137: This is an account key, right? Maybe name it like that here?

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
```diff
@@ -133,6 +133,40 @@ public Builder fromSeed(DeterministicSeed seed, Script.ScriptType outputScriptTy
...
+         * <p>In the case of P2PKH, just a P2PKH chain is created and activated which is then the default chain for fresh
+         * addresses. It can be upgraded to P2WPKH later.</p>
+         * <p>In the case of P2WPKH, both a P2PKH and a P2WPKH chain are created and activated, the latter being the default
+         * chain. This behaviour will likely be changed with bitcoinj 0.16 such that only a P2WPKH chain is created and
+         * activated.</p>
+         * @param key deterministic key to derive all keys from
```
schildbach, line 143: …and here? Maybe also in the parameter name?

schildbach (commented):
Looks good to me. I've added one inline comment.

schildbach:
Thanks! Merged.

PR #2047, opened 2020-09-15
Make Wallet.SendResult fields final

msgilligan:
* Make the fields final
* Add a constructor
* Simplify sendCoins

PR #2048, opened 2020-09-16
TestWithPeerGroup — increase timeout to 15 seconds

msgilligan:
PeerGroup tests intermittently failing with timeout errors on CI. [example](https://github.com/bitcoinj/bitcoinj/pull/2047/checks?check_run_id=1120297345)

Let's increase the timeout from 10 seconds to 20 seconds to reduce or eliminate these random failures.

### DIFF
--- a/core/src/test/java/org/bitcoinj/testing/TestWithPeerGroup.java
+++ b/core/src/test/java/org/bitcoinj/testing/TestWithPeerGroup.java
@@ -49,7 +49,7 @@ public class TestWithPeerGroup extends TestWithNetworkConnections {
     }
 
     @Rule
-    public Timeout globalTimeout = Timeout.seconds(10);
+    public Timeout globalTimeout = Timeout.seconds(15);
 
     @Override
     public void setUp() throws Exception {


schildbach:
Hu – we have the `Block too far in future: 2020-09-16T20:49:13Z (1600289353) vs allowed 1970-01-01T09:00:08Z` error again. Haven't we fixed this?

msgilligan:
I thought we had, but another case popped up.  The best long-term solution is to not use a global   for mocking time.

schildbach:
It might be bikeshedding, but I'd rather try with a more moderate timeout first, e.g. 15 seconds. Those 20s timeouts can pile up pretty fast since a lot of tests derive from TestWithPeerGroup.

Also could you rebase on current master to get the other test failures cleared?

msgilligan:
> I'd rather try with a more moderate timeout first, e.g. 15 seconds. 

Fixed via `push --force`.

> Those 20s timeouts can pile up pretty fast since a lot of tests derive from TestWithPeerGroup.

Well, not directly -- because increasing the timeout doesn't affect how long the tests take to run -- only whether they fail when they exceed the threshold. There could be an indirect effect if we fail to fix a performance issue that is hidden by the longer timeout.

> Also could you rebase on current master to get the other test failures cleared?

Also included in the `push --force`.

msgilligan:
... and for some reason the current JDK11/Windows build is hanging: https://github.com/bitcoinj/bitcoinj/pull/2048/checks?check_run_id=1150790436

schildbach:
Merged anyway.

Issue #2050, opened 2020-09-18
TransactionBroadcast::future() does not complete when run against local RegTest mode


msgilligan:
Change e3513964ed2fcbd0f5e9edb6e160094061a46225 in 0.15.8 prevents the following (Groovy) code from running:

```
        Transaction tx = new Transaction(params)
        tx.addOutput(amount, rpcAddress)
        SendRequest request = SendRequest.forTx(tx)
        wallet.completeTx(request)  // Find an appropriate input, calculate fees, etc.
        wallet.commitTx(request.tx)
        log.info("Sending Tx: {}", tx)
        TransactionBroadcast broadcast = peerGroup.broadcastTransaction(request.tx)
        log.info("Waiting for completion of broadcast for txid: {}", broadcast.tx.getTxId())
        Transaction sentTx = broadcast.future().get()   // This future never completes (successfully or exceptionally)
```

The future never completes (successfully or exceptionally).

Reverting to 0.15.7 or reverting change e3513964ed2fcbd0f5e9edb6e160094061a46225 fixes the problem.

We can either make a "micro" fix/workaround (e.g. reverting e3513964ed2fcbd0f5e9edb6e160094061a46225) or we could consider some (possibly breaking) changes to `TransactionBroadcast` to perhaps create two futures:

1. Indicates Transaction was sent to the required number of peers
2. Indicates `TransactionConfidence.Listener ` confirms that the correct threshold of remote peers have relayed the transaction.

Note that (1) is complicated somewhat by the fact that `MessageWriteTarget::writeBytes` only indicates that the transaction was buffered, not that it was received by the server.

msgilligan:
>we could consider some (possibly breaking) changes to `TransactionBroadcast` to perhaps create two futures:
> 
> 1. Indicates Transaction was sent to the required number of peers
> 2. Indicates `TransactionConfidence.Listener ` confirms that the correct threshold of remote peers have relayed the transaction.

I'm working on a Pull Request to implement the above. Stay tuned...

msgilligan:
PR #2583 and PR #2584 are merged, closing.

PR #2054, opened 2020-09-22
Always use ConfidenceChange listener

oscarguindzberg:
If connected to 1 peer, it will be disconnected and reconnected after the broadcast, so we in fact will hear the inv from that peer.

Otherwise, when connected to just 1 peer, the future is never set.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java
@@ -141,9 +141,7 @@ public class TransactionBroadcast {
             // a big effect.
             List<Peer> peers = peerGroup.getConnectedPeers();    // snapshots
             // Prepare to send the transaction by adding a listener that'll be called when confidence changes.
-            // Only bother with this if we might actually hear back:
-            if (minConnections > 1)
-                tx.getConfidence().addEventListener(new ConfidenceChange());
+            tx.getConfidence().addEventListener(new ConfidenceChange());
             // Bitcoin Core sends an inv in this case and then lets the peer request the tx data. We just
             // blast out the TX here for a couple of reasons. Firstly it's simpler: in the case where we have
             // just a single connection we don't have to wait for getdata to be received and handled before


msgilligan:
This looks like a better fix for #2050 then reverting e3513964ed2fcbd0f5e9edb6e160094061a46225 -- if it works with all the use cases.  @schildbach ?

oscarguindzberg:
@msgilligan I am not sure either what would be the best fix. I just tested this fix in regtest with bitcoin core 0.20.1 on localhost and "sometimes" the future is not set (did not investigate the cause)

msgilligan:
Eventually, I'd like to see two futures in `TransactionBroadcast` one that indicates that the message was successfully sent to one or more Peers and one that indicates the broadcast message was seen by other peers. See my suggestion in Issue #2050. I'm thinking we could do some refactoring for version 0.17 once we have lambdas/CompletableFuture. It seems reasonable for the API to let clients find the difference between something that failed to send vs. something that failed to relay.

I'll try your patch in my RegTest environment where I first discovered the issue. Maybe we should test on MainNet or TestNet with the single/trusted peer use case. I don't (yet) understand Bitcoin at the P2P layer well enough to understand all the use cases and failure modes.

chimp1984:
> it will be disconnected and reconnected after the broadcast,

Why is that the case?

msgilligan:
> > it will be disconnected and reconnected after the broadcast,
> 
> Why is that the case?

In normal operation bitcoinj disconnects from peers immediately after broadcasting for privacy reasons (to minimize the amount of information a particular peer can gather on the SPV wallet) and should then connect to new peers. My assumption is that this behavior is essentially being replicated in the single-peer case for consistency/simplicity reasons.

msgilligan:
@oscarguindzberg Your PR fixes the bug as it has been manifesting in my RegTest environment on my development system.

When I run the ConsensuJ `WalletSendSpec` test with your branch (`./gradlew bitcoinj-rpcclient:regTest --include-build ../bitcoinj`) and  `workaroundBitcoinJ_015_8_Issue = false` ([this line](https://github.com/ConsensusJ/consensusj/blob/master/bitcoinj-rpcclient/src/integ/groovy/com/msgilligan/bitcoinj/integ/WalletSendSpec.groovy#L38)), the futures complete correctly.

The lines that would otherwise fail are [here](https://github.com/ConsensusJ/consensusj/blob/master/bitcoinj-rpcclient/src/integ/groovy/com/msgilligan/bitcoinj/integ/WalletSendSpec.groovy#L109) and [here](https://github.com/ConsensusJ/consensusj/blob/master/bitcoinj-rpcclient/src/integ/groovy/com/msgilligan/bitcoinj/integ/WalletSendSpec.groovy#L144).

chimp1984:
> In normal operation bitcoinj disconnects from peers immediately after broadcasting for privacy reasons (to minimize the amount of information a particular peer can gather on the SPV wallet) and should then connect to new peers. My assumption is that this behavior is essentially being replicated in the single-peer case for consistency/simplicity reasons.

But it also has the down side to connect to more nodes and increases the chance to connect to a chainanalysis node. I think privacy with bloomfilters is anyway broken to an extent that cannot be fixed. But now we decrease network stability by adding a questionable privacy improvement.

Does the code not consider the use case to be connected to a single localhost node? That is anyway the only case how to not lose privacy.

schildbach:
Indeed the idea is to get rid of the special casing for the 1 peer case. @chimp1984 It's unrelated to bloom filtering, bitcoinj can be run in "full blocks" mode (not to be mistaken with "full validation").

@oscarguindzberg I think this is a great catch! I certainly have missed that 'if' when I came up with a4f241ef29cdc58d9525357716278fa758d4ab4f.

schildbach:
I just merged this. I keep this issue open as a reminder to backport to 0.15.x.

PR #2056, opened 2020-10-02
Prevent Infinite Loop

oguzhancevik:


msgilligan:
Thanks for this! 

It is an improvement in coding style, but unless I'm missing something it doesn't change the behavior in any way. It just uses a `break` instead of a `while` condition to end the loop. In either case if the condition `cursor.equals(lower)` never occurs it will eventually throw a `NullPointerException` due to the `checkNotNull`.

@schildbach I'm ok with merging this as it makes the code more readable.

msgilligan (approved):
This makes the code more readable, but (unless I'm missing something) doesn't actually prevent any infinite loop from occurring.

But I say merge it for the readability improvement.

oguzhancevik:
This coding approach is defensive programming. You can check https://stackoverflow.com/a/52268446/7927573

I agree with you. This commit looks like more readable than prevents the infinite loop.

schildbach:
Merged.

PR #2057, opened 2020-10-03
prevention null pointer exception

Sangaibisi:
Dereference of block.getTransactions() may produce null pointer exception.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -28,6 +28,7 @@ import org.slf4j.*;
 
 import javax.annotation.*;
 import java.util.*;
+import java.util.Objects;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/wallet/WalletProtobufSerializer.java
```diff
@@ -707,17 +707,16 @@ private WalletTransaction connectTransactionOutputs(final NetworkParameters para
         final WalletTransaction.Pool pool;
         switch (txProto.getPool()) {
             case DEAD: pool = WalletTransaction.Pool.DEAD; break;
-            case PENDING: pool = WalletTransaction.Pool.PENDING; break;
```
schildbach, line 710: I think for clarity it's better to keep the two deprecated values at the bottom, even if it's duplicating one statement.

We could maybe think about removing the values though, I'm not sure.

PR #2059, opened 2020-10-06
Add TransactionWitness.redeemP2WSH()

oscarguindzberg:


schildbach:
First of all, I think this really needs a unit test.

When it comes to P2WSH, it is my understanding that it is a successor to P2SH, right? This is why I'm always looking for how P2SH did things API-wise. In this case, I could not find an equivalent. Do you know any?

core/src/main/java/org/bitcoinj/core/TransactionWitness.java
```diff
@@ -41,6 +41,20 @@ public static TransactionWitness redeemP2WPKH(@Nullable TransactionSignature sig
         return witness;
     }
 
+    /**
+     * Creates the stack pushes necessary to redeem a P2WSH output.
+     */
+    public static TransactionWitness redeemP2WSH(byte[] witnessScript, TransactionSignature... signatures) {
```
schildbach, line 47: Would it maybe make sense to express the witnessScript as type Script?

oscarguindzberg: fixed

oscarguindzberg:
Added a unit test.

In the future it would be good to write tests for the examples described in the BIP https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#Native_P2WSH .

The P2SH way of doing this is ScriptBuilder.createP2SHMultiSigInputScript()

schildbach:
Merged.

schildbach:
> In the future it would be good to write tests for the examples described in the BIP https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#Native_P2WSH .

Yes, I tried to follow those example with P2WPKH too.

PR #2060, opened 2020-10-08
key chains: Keep listeners when decrypting/encrypting

oscarguindzberg:


core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
```diff
@@ -475,6 +479,9 @@ public BasicKeyChain toEncrypted(KeyCrypter keyCrypter, KeyParameter aesKey) {
                     throw new KeyCrypterException("The key " + key.toString() + " cannot be successfully decrypted after encryption so aborting wallet encryption.");
                 encrypted.importKeyLocked(encryptedKey);
             }
+            for (ListenerRegistration<KeyChainEventListener> listener : listeners) {
+                encrypted.listeners.add(listener);
```
schildbach, line 483: You could use the newly added `addEventListener(ListenerRegistration)` here, no?

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
```diff
@@ -500,6 +507,9 @@ public BasicKeyChain toDecrypted(KeyParameter aesKey) {
             for (ECKey key : hashToKeys.values()) {
                 decrypted.importKeyLocked(key.decrypt(aesKey));
             }
+            for (ListenerRegistration<KeyChainEventListener> listener : listeners) {
```
schildbach, line 510: …and here?

schildbach (commented):
Looks like a good catch!

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
oscarguindzberg: done.

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
oscarguindzberg: done.

schildbach:
I changed the visibility of the method to package-private and merged.

PR #2078, opened 2021-01-06
WalletTool, BuildCheckpoints, FetchBlock: Migrate parsing of command …

schildbach:
…line options from JOpt to picocli.

JOpt is not maintained any more and picocli is much more powerful.


schildbach:
Merged.

PR #2088, opened 2021-02-10
Wallet: Fix adjustOutputDownwardsForFee() to require a fee per kb…

schildbach:
…rather than an absolute fee if ensureMinRequiredFee is true.

Currently, if someone creates a small tx with one input and one output emptying the wallet, an absolute fee of 1000 sat is applied – maybe overpaying.


schildbach:
Merged.

PR #2096, opened 2021-02-26
DeterministicKey: ensure depth < 256. depth is serialized as uint8

afk11:
Since the depth serialization is a single byte deriving more than 255 levels is causing problems. It's not explicitly limited to 255 in the BIP, but this check seems worth it to rule out mistakes or denial of service

### DIFF
--- a/core/src/test/java/org/bitcoinj/crypto/HDKeyDerivationTest.java
+++ b/core/src/test/java/org/bitcoinj/crypto/HDKeyDerivationTest.java
@@ -16,19 +16,18 @@
 
 package org.bitcoinj.crypto;
 
+import java.math.BigInteger;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
-import java.math.BigInteger;
-
+import org.bitcoinj.core.NetworkParameters;
 import org.bitcoinj.core.Utils;
 import org.bitcoinj.crypto.HDKeyDerivation.PublicDeriveMode;
+import org.bitcoinj.params.TestNet3Params;
 import org.bouncycastle.crypto.params.KeyParameter;
 import org.junit.Test;
-
-import com.google.common.collect.ImmutableList;
 
 /**
  * @author Andreas Schildbach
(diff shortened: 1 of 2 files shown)


schildbach:
Actually the BIP does state "1 byte": https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#Key_identifiers

In code, we simply cast it to Java "byte", which is signed. I wonder if this already fails with values >127. 

I wonder if there is a test vector availble for such edge cases.

afk11:
@schildbach unlikely, the only tests I recall there are positive cases - that a derivation succeeds and results in abc.

Sometimes such bugs get discovered and test cases shared, but they don't get the attention they need really. (bitcore + btcd both had a bug in bip32 where if a 32byte ByteArrays left most bytes were 0x00 it'd accidentally trim to 31.. so a non-portable wallet)

I think a community project that defined a testing interface, and contains test fixtures.. if the long running bitcoin libraries implemented the interface, each time a new fixture is added they could all be checked. A simple start might just be BIP39/BIP32. Minimal interface to test, but covers a lot of operations where things can go wrong that we hope they'd detect.

core/src/test/java/org/bitcoinj/crypto/HDKeyDerivationTest.java
```diff
@@ -100,7 +99,30 @@ public void testDeriveFromPublicParent() {
         assertTrue(fromPublicWithInversion.isPubKeyOnly());
         assertFalse(fromPublicWithInversion.isEncrypted());
     }
+    @Test (expected = java.lang.IllegalArgumentException.class)
```
schildbach, line 102: I think these two new tests don't work as expected. They actually throw IllegalArgumentException due to a different reason than expected, and earlier than expected (which I understand is the last line of each test).

core/src/test/java/org/bitcoinj/crypto/HDKeyDerivationTest.java
```diff
@@ -100,7 +99,30 @@ public void testDeriveFromPublicParent() {
...
+        String xprv = "tprv8ZgxMBicQKsPdAoL5dkkJ2rLrYcYssH1UXBtY7N6zJQA99CuwR3t817dmKHNP3RqRsLVE2p1HAyTdXK1xjLBj9vr3xsN2p8sMoRUu4yBQzP";
+        DeterministicKey child = DeterministicKey.deserializeB58(xprv, params);
+        for (int i = 0; i < 255; i++) {
+            child = HDKeyDerivation.deriveChildKey(child, new ChildNumber(i, true));
+        }
+        assertEquals(255, child.getDepth());
 
+        HDKeyDerivation.deriveChildKey(child, new ChildNumber(0, true));
+    }
+    @Test (expected = java.lang.IllegalArgumentException.class)
```
schildbach, line 114: Nitpick: Can you use an empty line between methods?

core/src/test/java/org/bitcoinj/crypto/HDKeyDerivationTest.java
```diff
@@ -16,20 +16,19 @@
...
 package org.bitcoinj.crypto;
 
+import java.math.BigInteger;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
-import java.math.BigInteger;
-
```
schildbach, line 25: Nitpick: Please don't touch imports other than those you need to add or remove.

schildbach (changes requested):
I just came back to your fix and wanted to finally merge it for inclusion in 0.16. However I noticed that the tests are actually not working as expected. See my review comments, and don't worry about the two nitpicks.

PR #2102, opened 2021-03-07
Cut short script execution for the standard P2PKH and P2PK cases

schildbach:
These two commits also remove some (previously valid) edgy test cases from tx_valid.json, because the new code is more strict than the old.


schildbach:
Merged.

PR #2001, opened 2020-05-28
Work-in-progress: Replace usages of Guava Immutable collections with JDK unmodifiable

msgilligan:
* Mostly replacing usage of Guava Immutable collections
* A few other Guava Collections replacements (Set.newHashSet())

Issue #2000, opened 2020-05-28
Migrate away from usage of Guava


msgilligan:
This is an umbrella-Issue for discussion of migrating away or reducing our usage of the Guava library. (Similar to Issue #1984: Resolve (MS-Windows) unit test issues)

There are a handful of goals and objectives that _might_ be achieved by a reduction in the usage of Guava:

1. Create a library of Bitcoin Java protocol definitions with no dependencies save the JDK. In theory this could even be submitted to the JCP (Java Community Process) as a JSR (Java Specifications Request.) A JSR might be wishful thinking, but a library with no dependencies it could be very broadly used in the Bitcoin JDK ecosystem.
2. Make migrating to different/future versions of Guava easier for both bitcoinj and bitcoinj-based-applications
3. Reduce occurrences of "jar dependency hell" for users of bitcoinj.
4. Possible smaller applications after running optimizing build tools (such as Android optimizers and JDK native-compilation tools.)
5. Less code to review/inspect.

There is also a discussion around the scope of the usage of Guava and scope of efforts to reduce it:

1. Remove Guava dependency from the public API. (e.g. Guava would no longer be an `api` dependency in Gradle, this could also allow creation of a bitcoinj  "uber" JAR that hides/bundles its Guava dependencies -- allowing client applications to use a different version of Guava if needed.)
2. Remove usages of Guava that are `@deprecated` or informally deprecated in the Guava JavaDoc comments.
3. Remove usage of Guava collections like ImmutableList, ImmutableSet, etc. (the JDK has been adding equivalent methods in every release. The `.of`methods added in JDK 9, for example -- See PR #2001 for some work-in-progress.)
4. Remove usage of Guava futures (requires upgrade to JDK8/`CompletableFuture`, see Issue #2254)
5. Replace usage of Guava Preconditions (or figure out a build tool that creates a dependency jar that is just the preconditions subset)

The scope of usage of Guava can decrease over time, especially as newer bitcoinj versions migrate to requiring JDK8, JDK9, etc.

And a discussion of which current and future modules of bitcoinj might have dependencies on Guava.

1. existing `core` subproject
2. A subset of `core` for definitions and types (see above)
3. `tools` subproject (or possible future individual tools subprojects)
4. `samples` subproject
5.  `wallettemplate` subproject

Existing, related issues:

1. Issue #1874: Consider splitting the `core` module into `core` and `base` modules
2. Issue #1873: Public methods whose parameters and return types are not public
[...]

PR #2105, opened 2021-04-04
Remove guava from core

fireuse:


schildbach:
Also see #2000 and #2001.

fireuse:
@schildbach Could you possibly review changes in the last commit?

core/build.gradle
```diff
@@ -25,7 +25,7 @@ dependencies {
     testImplementation 'org.hamcrest:hamcrest-library:1.3'
 }
 
-sourceCompatibility = 1.7
+sourceCompatibility = 1.8
```
schildbach, line 28: At the moment, we're still targeting JDK 7/Android 6.0. We have talked about raising this to something like JDK 8/Android 7.0 for version 0.17, but it's not decided yet.

core/src/main/java/org/bitcoinj/core/Bech32.java
```diff
@@ -103,10 +101,18 @@ public static String encode(final Bech32Data bech32) {
         return encode(bech32.hrp, bech32.data);
     }
 
+    private static void checkHRPLen(String hrp){
```
schildbach, line 104: I'm not convinced this is preferable to helpers like `checkArgument()`. Is there no JDK equivalent available?

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -17,21 +17,21 @@
 
 package org.bitcoinj.core;
 
-import com.google.common.annotations.*;
```
schildbach, line 20: Please do not reorder imports if the topic of your PR is actually something else.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -470,8 +469,8 @@ public String toString() {
         s.append(" block: \n");
         s.append("   hash: ").append(getHashAsString()).append('\n');
         s.append("   version: ").append(version);
-        String bips = Joiner.on(", ").skipNulls().join(isBIP34() ? "BIP34" : null, isBIP66() ? "BIP66" : null,
-                isBIP65() ? "BIP65" : null);
+        String bips = ", " + (isBIP34() ? "BIP34" : null) + (isBIP66() ? "BIP66" : null) +
```
schildbach, line 472: This is not equivalent! See the documentation of `skipNulls()`.

core/src/main/java/org/bitcoinj/core/BlockLocator.java
```diff
@@ -18,30 +18,35 @@
...
     }
 
     /**
      * Add a {@link Sha256Hash} to a newly created block locator.
      */
-    public BlockLocator add(Sha256Hash hash) {
-        return new BlockLocator(new ImmutableList.Builder<Sha256Hash>().addAll(this.hashes).add(hash).build());
+    public BlockLocator add(final Sha256Hash hash) {
+        return new BlockLocator(new LinkedList<Sha256Hash>(){{
+            addAll(hashes);
```
schildbach, line 47: This is scary, a static block in an anonymous class…

schildbach (commented):
Why do you want to "remove Guava from core"? I'm certainly not against replacing individual Guava idioms by their JDK counterpart, if there is one (this is usually indicated in the Guava JavaDoc). But why copy and paste Guava code?

In general for this kind of PR I'd prefer individual PRs (or at least individual commits) per kind of refactoring, rather than a commit per class. That way, each idea could be discussed and merged separately. Just as an example, your `ListenableFuture` to `CompletableFuture` refactoring would go into one dedicated commit.

msgilligan:
I missed this one while away from the project temporarily. Nice to see the effort!

PR #2107, opened 2021-04-15
WalletTool: Implement two coin selection options for sending

schildbach:
WalletTool: Implement two coin selection options for sending, `--select-addr=<address>` and `--select-output=<txid>:<index>`.


schildbach:
Merged.

PR #2109, opened 2021-04-21
Message hierarchy: Remove constructors that take a separate protocolVersion argument.

schildbach:
It's contained in the MessageSerializer since a while.


schildbach:
Merged.

PR #2110, opened 2021-04-22
VersionMessage: Remove support for protocol versions older than 106.

schildbach:
The minimum is 70000 anyway, and I'm pretty sure the code path wasn't properly tested.


schildbach:
Merged.

PR #2111, opened 2021-04-22
VarInt: Introduce intValue() and longValue() accessors and use them, …

schildbach:
…deprecating access to the field.


schildbach:
Merged.

PR #2112, opened 2021-04-26
build.gradle: Migrate from maven to maven-publish plugin.

schildbach:
To publish to the local Maven repository, use `gradle publishToMavenLocal` rather than `gradle install`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.10'
+        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.16'
     }
 }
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.10'
+        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.16'
```
schildbach, line 7: I wonder is it possible to apply different versions on the condition of Gradle version? @msgilligan

msgilligan: I might be able to do that...

msgilligan:
See PR #2134 and PR #2135

build.gradle
msgilligan: @schildbach See: https://github.com/bitcoinj/bitcoinj/pull/2140

schildbach:
Merged the maven change via 8a775a37d8d2f7edb0b7b9cf5c615867062f0f5e, and dropped the protobuf change in favor of #2140.

PR #2117, opened 2021-05-12
MonetaryFormat: Support satoshi denomination.

schildbach:


msgilligan:
Looks good to me.

For reference: https://medium.com/coinranking/satoshi-deserves-its-own-symbol-43138fe1ddf9

I'm not sure how widely adopted and recognized ș is. But we can always change it if something else becomes more popular/recognized.

Personally, I like the _bits_ proposal, as does Adam Back: https://twitter.com/adam3us/status/1335354258858905604?lang=en
Although that proposal seems to have lost traction.

schildbach:
@msgilligan Thanks for your review!

Merged.

PR #2124, opened 2021-06-22
 fix error: unterminated inline tag

newbull:
When you make a full build with "gradle clean build", you will got:
"error: unterminated inline tag
Execution failed for task ':bitcoinj-core:javadoc'.
> Javadoc generation failed."

This commit fix the error.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/SendAddrV2Message.java
+++ b/core/src/main/java/org/bitcoinj/core/SendAddrV2Message.java
@@ -18,7 +18,7 @@ package org.bitcoinj.core;
 
 /**
  * <p>Represents the {@code sendaddrv2} P2P protocol message, which indicates that a node can understand and prefers
- * to receive {@code addrv2] messages instead of {@code addr} messages.</p>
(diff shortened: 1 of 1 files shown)


msgilligan (approved):

msgilligan:
@schildbach This commit is good and fixes the build for me in local testing.

schildbach:
Merged. Thanks for spotting this!

msgilligan:
@schildbach This should have fixed the Github Actions build, but for some reason the jobs didn't run. Can you try using your admin privileges to re-run or restart the "Queued" jobs?

schildbach:
I think the GitHub actions builds fail because they upgraded to Gradle 7+, and our build is not compatible to that version. Imho if possible, we should use an action that uses anything from Gradle 4.4 to Gradle 6.x.

Nevertheless, I looked at the builds and noticed that the most recent 3 builds are "queued" for 2 days, and cancelling fails. So I guess this is something that needs to be fixed by GitHub.

PR #2132, opened 2021-08-09
Removed the limit of parsing blocks of max 2,000,000 bytes. Some bloc…

lvca:
…ks from 2020 are larger and weren't imported

### DIFF
--- a/core/src/main/java/org/bitcoinj/utils/BlockFileLoader.java
+++ b/core/src/main/java/org/bitcoinj/utils/BlockFileLoader.java
@@ -158,7 +158,7 @@ public class BlockFileLoader implements Iterable<Block>, Iterator<Block> {
                 currentFileStream.read(bytes, 0, 4);
                 long size = Utils.readUint32BE(Utils.reverseBytes(bytes), 0);
                 // We allow larger than MAX_BLOCK_SIZE because test code uses this as well.
-                if (size > Block.MAX_BLOCK_SIZE*2 || size <= 0)
+                if (size > Block.MAX_BLOCK_SIZE*5 || size <= 0)
                     continue;
                 bytes = new byte[(int) size];
                 currentFileStream.read(bytes, 0, (int) size);


schildbach:
(FYI: If you rebase a PR branch, there is no need to open a new PR. Just force-push over the old remote branch, GitHub will update the PR accordingly.)

As I wrote in the other PR, I wonder if we maybe should remove the check entirely. What's its purpose? It's certainly not a consensus rule.

lvca:
I realized I was testing a different fork. Better to stay with the official repo. About the issue, since it's coming from the .dat files, there should not be a wrong size from attacks (like it could arrive from the network). So I think it's relatively safe to just load the block, but leaving a number there assures you it would never OOM in case a crazy number of block size is found. With Segwit the max block size is around 8MB (https://jimmysong.medium.com/understanding-segwit-block-size-fd901b87c9d4), so I'd keep around 10MB for now. Or we could totally remove it.

msgilligan:
Is there no bitcoinj implementation of the actual (weighted, virtual) block size limit for SegWit?

schildbach:
> Is there no bitcoinj implementation of the actual (weighted, virtual) block size limit for SegWit?

I don't think so. Nobody cared about the consensus for the last few years, and actually I proposed to remove full verification entirely for 0.16.

lvca:
Ok, so we could completely remove that check.

msgilligan:
I care about the consensus and would like to see it maintained and updated. It's not something I could do single-handedly without funding, though. At the very least we could put it in a separate module and mark it as "incubating", "under-maintained" or something.

Some kind of an upper-bound would might still be a useful protection against a malicious node.

lvca:
Once the blocks are on file means they pass basic checks about malicious attacks, so I don't think you'll ever read an overside block from the .dat. But leaving that check costs nothing just in case.

schildbach:
I've decided to remove the check via a61080ebebaf57e00736da026e0a7a600ac570a2. I agree with @lvca that we can trust Bitcoin Core enough to not DoS our loader. Plus, if we just adapt the limit, it will fall on our feet at some point in the future again.

A proper consensus implementation would not go into this loader. Rather, I'd expect it somewhere within `AbstractBlockChain.add(Block)`.

PR #2134, opened 2021-08-14
Github Actions: Build with Gradle 6.9

msgilligan:
Use the `gradle/gradle-build-action` to build with a specified
version of Gradle. In our case, let’s use Gradle 6.9.


msgilligan:
This PR gets us to a working build with the latest Gradle we support and allows us to control which version of Gradle is used for our build.

Note: It also opens the door for adding the Gradle version used for the build into the build matrix, but that should be a separate PR (and might not be as easy as it looks.)

schildbach:
Cool, thanks! Merged.

PR #2141, opened 2021-08-24
Add “Legacy Gradle CI” Github Actions workflow

msgilligan:
This workflow currently builds with: JDK 8 (Temurin), Gradle 4.4.1
on macOS, Windows, and Linux.

It is meant to be a test of what a Debian package targeted for
inclusion in Debian would use.

Note: This comment was updated to match force push that uses JDK 8/Gradle 4.4.1

### DIFF
--- a/.github/workflows/legacy.yml
+++ b/.github/workflows/legacy.yml
@@ -1,4 +1,5 @@
-name: Gradle CI
+name: Legacy Gradle CI
+# Build with JDK 8 and Gradle 4.4.1 (for Debian package support)
 
 on: [push, pull_request]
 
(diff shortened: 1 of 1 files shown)


msgilligan:
We could possibly remove the macOS and Windows versions of this build, but I thought they might be useful just to make sure developers targeting the "legacy" Gradle version can build on their respective platforms.

msgilligan:
@schildbach I updated and force-pushed to use Java 8 and Gradle 4.4.1. Tests are passing and this should be ready to merge.

msgilligan:
Actually, wait. I just found a better way.

PR #2143, opened 2021-08-25
Github Actions: Add Ubuntu/JDK 11/Gradle 4.10.3

msgilligan:
Github Actions allows adding special cases to the matrix.  See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

Using the `include:` option, we can add a special case for Ubuntu, JDK 11, and Gradle 4.10.3

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -11,6 +11,12 @@ jobs:
         java: [ '8', '11']
         distribution: ['temurin']
         gradle: ['7.2']
+        include:
+          # Special case to test something close to Debian config with the oldest supported (standard) Gradle on Ubuntu
+          - os: ubuntu-latest
+            java: '11'
+            distribution: 'temurin'
+            gradle: '4.10.3'
       fail-fast: false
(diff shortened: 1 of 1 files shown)


schildbach:
Hmm, wouldn't that be Java 8 and Gradle 4.4.1? I know this will skip `wallettemplate`, but we have plenty of tests where it is included.

msgilligan:
> Hmm, wouldn't that be Java 8 and Gradle 4.4.1? 

No, because Debian has included Java 11 for the last two releases. It also contains a specially patched version of Gradle 4.4.1 that works with Java 11.

schildbach:
Ah ok, and Debian doesn't come with `openjdk-8-jdk` any more. (Ubuntu [does](https://packages.ubuntu.com/search?keywords=openjdk-8-jdk&searchon=names&suite=all&section=all).)

I guess this kind of depends on the discussion started in #2149.

msgilligan:
> I guess this kind of depends on the discussion started in #2149.

Yes, although if we decide against requiring JDK 11 for now, we could modify this PR to use JDK 8 + (standard) Gradle 4.4.1 as our minimum baseline -- but either way we can't exactly match the Debian config.

schildbach:
Merged.

PR #2150, opened 2021-08-30
Inconsistent wallet bug fix

sqrrm:
### Wallet: Clone transaction before committing

#### Problem
A transaction received from the network is added to all wallets
that find it relevant. If two wallets find the same transaction relevant
the same Transaction instance is added to both wallets.

Spending the outputs from this transaction can cause consistiency
issues, in particular if the outputs are spent in the same transaction,
as shown in WalletTest.oneTxTwoWallets. There are probably more issues
with having the same Transaction instance handled by two different
wallets.

#### Fix
Clone the transaction before adding it to the wallet.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -1893,6 +1893,7 @@ public class Wallet extends BaseTaggableObject
                 // two wallets depend on the same transaction.
                 Transaction cloneTx = tx.getParams().getDefaultSerializer().makeTransaction(tx.bitcoinSerialize());
                 cloneTx.setPurpose(tx.getPurpose());
+                cloneTx.setUpdateTime(tx.getUpdateTime());
 
                 riskDropped.put(cloneTx.getTxId(), cloneTx);
                 log.warn("There are now {} risk dropped transactions being kept in memory", riskDropped.size());
@@ -1911,6 +1912,7 @@ public class Wallet extends BaseTaggableObject
             // two wallets depend on the same transaction.
             Transaction cloneTx = tx.getParams().getDefaultSerializer().makeTransaction(tx.bitcoinSerialize());
             cloneTx.setPurpose(tx.getPurpose());
+            cloneTx.setUpdateTime(tx.getUpdateTime());
 
             // If this tx spends any of our unspent outputs, mark them as spent now, then add to the pending pool. This
             // ensures that if some other client that has our keys broadcasts a spend we stay in sync. Also updates the
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -1888,7 +1888,13 @@ public class Wallet extends BaseTaggableObject
                 return;
             if (isTransactionRisky(tx, dependencies) && !acceptRiskyTransactions) {
                 // isTransactionRisky already logged the reason.
-                riskDropped.put(tx.getTxId(), tx);
+
+                // Clone transaction to avoid multiple wallets pointing to the same transaction. This can happen when
+                // two wallets depend on the same transaction.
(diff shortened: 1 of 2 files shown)


msgilligan:
@schildbach I took a quick look at the code and I think it is ok to "approve running [Github] workflows".

@sqrrm Thanks for the PR!

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -1888,7 +1888,13 @@ public void receivePending(Transaction tx, @Nullable List<Transaction> dependenc
                 return;
             if (isTransactionRisky(tx, dependencies) && !acceptRiskyTransactions) {
                 // isTransactionRisky already logged the reason.
-                riskDropped.put(tx.getTxId(), tx);
+
+                // Clone transaction to avoid multiple wallets pointing to the same transaction. This can happen when
+                // two wallets depend on the same transaction.
+                Transaction cloneTx = tx.getParams().getDefaultSerializer().makeTransaction(tx.bitcoinSerialize());
+                cloneTx.setPurpose(tx.getPurpose());
```
schildbach, line 1895: Maybe `updatedAt` should be cloned, too?

Just for the record, I see you're not cloning the `exchangeRate` and `memo` fields. This is probably ok, since they are kind of wallet specific (though they should be in the WalletTransaction class if that's true).

core/src/test/java/org/bitcoinj/wallet/WalletTest.java
```diff
@@ -3713,4 +3714,61 @@ public void roundtripViaMnemonicCode() {
...
+        inputW2.setWitness(TransactionWitness.redeemP2WPKH(txSig2, sigKey2));
+
+        wallet1.commitTx(sendReq.tx);
+        wallet2.commitTx(sendReq.tx);
+        assertEquals(ZERO, wallet1.getBalance());
+        assertEquals(ZERO, wallet2.getBalance());
+
+        assertTrue(wallet1.isConsistent());
+        assertTrue(wallet2.isConsistent());
+    }
```
schildbach, line 3773: Perhaps add a final test for "transaction equality but not identity"?

schildbach (commented):
I'd say the concept looks fine.

core/src/test/java/org/bitcoinj/wallet/WalletTest.java
sqrrm: Added in commit https://github.com/bitcoinj/bitcoinj/pull/2150/commits/cc14412b5c5b4c39b894d1bf2643a306dfc21573

core/src/main/java/org/bitcoinj/wallet/Wallet.java
sqrrm: Agreed, `updatedAt` makes sense to clone as well, fixed in https://github.com/bitcoinj/bitcoinj/pull/2150/commits/804612ed35856ae04fd02718a1466959a78331df

I think it doesn't make that much sense to clone `exchangeRate` and `memo` since the cloning is done on transactions received from the network. Probably won't hurt either, and it might make sense for Transaction to have a `clone()` method that would handle that. Probably for another PR as you say.

sqrrm:
@schildbach Are there any outstanding issues with this PR? I think I addressed your concerns.

schildbach:
I think it's fine. I just wanted to give others the chance for review.

msgilligan:
>  I just wanted to give others the chance for review.

I just saw this. I'm wrapping up for the day, but will try to look at it in more detail on Monday. Can you "request" that I review it, so it will be easier for me to remember?

sqrrm:
@msgilligan I seem to be unable to request anyone in particular for review.

schildbach:
> @msgilligan I seem to be unable to request anyone in particular for review.

Weirdly, same for me.

schildbach:
I squashed this and merged.

PR #2154, opened 2021-09-01
GithubCI: Upgrade to Gradle 7, add JDK 16

msgilligan:
Also parameterize `distribution` and `gradle` to allow for adding
future special cases with the `include: ` keyword.

PR #2155, opened 2021-09-01
Github Actions: Add Zulu JDK ea-17

msgilligan:
Depends upon PR #2154 (Temurin doesn't provide 17-ea yet, but should have 17 in September)

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -11,6 +11,12 @@ jobs:
         java: [ '8', '11']
         distribution: ['temurin']
         gradle: ['7.2']
+        include:
+  # Special case to test Zulu JDK 17-ea (17-ea not available on Temurin yet)
+          - os: ubuntu-latest
+            java: '17-ea'
+            distribution: 'zulu'
+            gradle: '7.2'
       fail-fast: false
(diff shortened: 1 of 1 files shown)


schildbach:
Ok, we now have 2 PRs that add a new Java version to the build matrix (this one and #2154).

Looking at https://www.oracle.com/java/technologies/java-se-support-roadmap.html, Java 17 LTS should be GA very soon. Maybe we wait until it is released and add that one then? Until then, the pipelines run on these PRs will cover the prereleases.

msgilligan:
I think it might make sense to continue to run one job on the latest JDK pre-release. It would be nice to have advanced notice of any issues.

So my preference is that we merge this now. When JDK 17 goes GA we update the `java` matrix to replace `'16'` with `'17'` and then replace the `'17-ea'` in this `include` case with a `'18-ea'`. It will only add one more job, but will be enough to alert us of any JDK related issues coming with the next release.

msgilligan:
> Java 17 LTS should be GA very soon.

Java 17 LTS will be the current `'17-ea`' version (unless they find a critical bug in the next 6 days.)

msgilligan:
I'm going to close this as we won't be merging this PR as it stands.

As discussed in Issue #2189 and PR #2188, we will update the current config from Temurin JDK 16 to 17, as soon as JDK 17 "GA" is available from Temurin.

PR #2157, opened 2021-09-01
Add @VisibleForTesting to methods in Block.java

msgilligan:
These methods should only be used for testing and will hopefully be
deprecated and/or refactored to another class in the future.


schildbach:
Merged.

PR #2159, opened 2021-09-01
AbstractBitcoinNetParamsTest refactor anon class

msgilligan:
Refactor anonymous inner class to static inner class to prepare for forthcoming refactoring that will add a constructor.


msgilligan:
@schildbach Do you want to merge this one now or do you want me to make this change in the PR for the refactored constructors in `NetworkParameters` and subclasses?

schildbach:
I fear you'll hate me, but I think this particular change could go together with the change that makes it necessary. On its own, it has little to no value as far as I can tell.

msgilligan:
No, that's fine and what I was thinking myself when I looked at it.

Issue #2161, opened 2021-09-01
Unit test org.bitcoinj.store.WalletProtobufSerializerTest::oneTx failing intermittently


msgilligan:
The unit test `org.bitcoinj.store.WalletProtobufSerializerTest::oneTx` is failing _intermittently_.

```
 org.bitcoinj.store.WalletProtobufSerializerTest > oneTx FAILED
     java.lang.AssertionError: expected:<2> but was:<4>
         at org.junit.Assert.fail(Assert.java:89)
         at org.junit.Assert.failNotEquals(Assert.java:835)
         at org.junit.Assert.assertEquals(Assert.java:647)
         at org.junit.Assert.assertEquals(Assert.java:633)
         at org.bitcoinj.store.WalletProtobufSerializerTest.oneTx(WalletProtobufSerializerTest.java:146)
```
For an example, see: https://github.com/bitcoinj/bitcoinj/runs/3486526841?check_suite_focus=true

The relevant code is here:
```
        Coin v1 = COIN;
        Transaction t1 = createFakeTx(UNITTEST, v1, myAddress);
        t1.getConfidence().markBroadcastBy(new PeerAddress(UNITTEST, InetAddress.getByName("1.2.3.4")));
        t1.getConfidence().markBroadcastBy(new PeerAddress(UNITTEST, InetAddress.getByName("5.6.7.8")));
        t1.getConfidence().setSource(TransactionConfidence.Source.NETWORK);
        myWallet.receivePending(t1, null);
        Wallet wallet1 = roundTrip(myWallet);
        assertEquals(1, wallet1.getTransactions(true).size());
        assertEquals(v1, wallet1.getBalance(Wallet.BalanceType.ESTIMATED));
        Transaction t1copy = wallet1.getTransaction(t1.getTxId());
        assertArrayEquals(t1.unsafeBitcoinSerialize(), t1copy.unsafeBitcoinSerialize());
        assertEquals(2, t1copy.getConfidence().numBroadcastPeers());
```

The final `assertEquals()` in the snippet is failing. Apparently when the wallet is duplicated by `roundTrip()` the two entries in `TransactionConfidence.broadcastBy` become four entries.

I suspect this might be because`PeerAddress::equals` contains the comparison `other.time == time`.

[...]

Issue #1842, opened 2019-05-07
PeerGroup inactives duplicates


pvyhnal-generalbytes:
I'm using DNS seed discovery and I noticed `org.bitcoinj.core.PeerGroup#inactives` is getting bigger and bigger when running for some time. 

It seems like every few (tens? of) minutes the number of elements get approximately doubled. (I'm not sure about this exactly)

The duplicated entries in the `inactive` collection are the same instances of `PeerAddress`

![Screenshot from 2019-05-07 13-27-19](https://user-images.githubusercontent.com/44365000/57295982-13d6e300-70cc-11e9-8db7-31b3ec5eeb1e.png)

I wonder if this is the case: https://github.com/bitcoinj/bitcoinj/issues/525#issuecomment-111185243 
> [...] duplicates can occur, for example when one Peer is connecting (neither present in inactives nor in peers) and that same address is discovered at the same time.

msgilligan:
Issue #2161 may be related, especially if (as I suspect) the root cause is that `PeerAddress::equals` includes a comparison of `PeerAddress#time`.

PR #2163, opened 2021-09-01
PeerAddress::equals, PeerAddress::hashCode - don’t use `time`

msgilligan:
I'm not sure if this will break anything that depends on the old behavior. It doesn't seem to break the unit tests. It might fix Issue #2161 (and possibly Issue #1842)

A better (or perhaps) additional solution might be able to create a `PeerAddress` type that contains just the identifying fields of an actual "peer address" and create a separate `PeerAddressMessage` that is a subclass of `ChildMessage`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerAddress.java
@@ -331,12 +331,12 @@ public class PeerAddress extends ChildMessage {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         PeerAddress other = (PeerAddress) o;
-        return other.addr.equals(addr) && other.port == port && other.time == time && other.services.equals(services);
+        return other.addr.equals(addr) && other.port == port && other.services.equals(services);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(addr, port, time, services);
+        return Objects.hash(addr, port, services);
     }
     
     public InetSocketAddress toSocketAddress() {


core/src/main/java/org/bitcoinj/core/PeerAddress.java
```diff
@@ -331,7 +331,7 @@ public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         PeerAddress other = (PeerAddress) o;
-        return other.addr.equals(addr) && other.port == port && other.time == time && other.services.equals(services);
```
pvyhnal-generalbytes, line 334: Should `time` be removed from `hashCode` too?

msgilligan: Good catch!

schildbach: Yes, I think it should be removed from `hashCode()` too.

schildbach:
I don't think peer addresses are used very much, neither from within bitcoinj nor by its consumers. So this is probably low risk.

core/src/main/java/org/bitcoinj/core/PeerAddress.java
msgilligan: I force-pushed an update with `time` removed from `hashCode()` as well.

msgilligan:
> I don't think peer addresses are used very much, .... So this is probably low risk.

That's good to hear. Do you think it will fix Issue #2161?

msgilligan:
I force pushed an update removing time from `hashCode()` as well and removed draft status from the PR.

I suppose we can just merge it and after about a dozen CI runs we'll know if Issue #2161 has been fixed.

schildbach:
> That's good to hear. Do you think it will fix Issue #2161?

I think this thesis could be verified tested by deliberately fuzzing the time field for the test.

schildbach:
I've hit "rerun jobs" a couple of times, and I got one "java.net.BindException: Address already in use" but no failure of the "oneTx" test.

msgilligan:
> > That's good to hear. Do you think it will fix Issue #2161?
> 
> I think this thesis could be verified tested by deliberately fuzzing the time field for the test.

I quickly copied the `oneTx` test and modified it by adding:

```
        TimeUnit.SECONDS.sleep(1);
```

Just before:        

```
        Wallet wallet1 = roundTrip(myWallet);
```

And this causes the `expected:<2> but was:<4>` error to occur every time. I then used `cherry-pick` to add this PR and it fixed the problem.

schildbach:
Merged.

PR #2170, opened 2021-09-04
Address: make binary constructor protected, javadoc

msgilligan:
The binary constructor does no validation checking and should only be used by subclasses, so lets make it `protected`

Also, fix JavaDoc warnings and clean up some of the javadoc text.


schildbach:
Merged.

PR #2153, opened 2021-09-01
Make Block a little less mutable

msgilligan:
* Remove deprecated constructor
* Move difficultyTarget and time to constructors (where possible)
* Remove copyBitcoinHeaderTo(Block) method that requires a mutable
  Block. Instead do the copying inside cloneAsHeader().
* Mark some mutator methods as @VisibleForTesting

NetworkParameters and subclasses:

* Move Genesis Block parameters to constructors
* Refactor some code into new createGenesisTransaction() method
* Prefer static final fields (constants) to magic numbers
* Move id initialization to top of constructors

This code simplifies the setting of `Block.time` and may help with Issue #2147. It is also a step towards making Block more immutable/unmodifiable.

PR #2171, opened 2021-09-04
Make Block a little less mutable, take #2

msgilligan:
This is the second iteration of PR #2153 now that many of the other changes in that PR have been merged via smaller PRs/commits.

There are two things going on in this PR:

1. Replace various magic number values in block creation with `static final` constants
2. Place some of those constants into constructors for `NetworkParameters` and `Block`

We could break these two changes into separate PRs if you want @schildbach.

Also note that I just slipped in PR #2172

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -115,12 +115,12 @@ public class Block extends Message {
     protected int optimalEncodingMessageSize;
 
     /** Special case constructor, used for the genesis node, cloneAsHeader and unit tests. */
-    Block(NetworkParameters params, long setVersion) {
+    Block(NetworkParameters params, long setVersion, long difficultyTarget, long time) {
         super(params);
         // Set up a few basic things. We are not complete after this though.
         version = setVersion;
-        difficultyTarget = 0x1d07fff8L;
-        time = Utils.currentTimeSeconds();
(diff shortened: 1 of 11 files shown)


schildbach:
Indeed I think the introduction of constants could go into a first commit.

msgilligan:
Actually before introducing comments, I think it would be best if we reorganize the initialization of the constants in each NetworkParameters subclass so we can easily do diffs of `MainNetParms` vs `TestNet3Params` and easily see how they are the same and how they are different.

See PR #2173

msgilligan:
@schildbach I rebased and force-pushed and was about to make a PR for defining symbolic constants, but then I noticed the overridden `getGenesisBlock()` in `RegTestParams` and made PR #2176 to eliminate it.

msgilligan:
I force-pushed again now that PR #2176 is merged.

msgilligan:
> 1. Replace various magic number values in block creation with static final constants

See PRs #2179 #2180 and #2181 -- they implement "part 1" of this PR and with some improvement/simplification.

msgilligan:
After breaking this commit into smaller commits and submitting them separately, things have gone in a different (and better) direction.

I'll leave this commit open for a little while longer until the equivalent of everything in here has been merged...

msgilligan:
OK, with PR #2190 everything of importance has been either merged or submitted in another PR.

PR #2172, opened 2021-09-04
Block::cloneAsHeader cleanup

msgilligan:
* Fix redundant write of `block.version` in new block (use the constructor parameter properly)
* Re-order setting other fields more logically


schildbach:
Merged.

PR #2173, opened 2021-09-05
Reorganize NetworkParameters subclasses for easy diffs

msgilligan:
Reorganize the initialization of the constants in each NetworkParameters
subclasses so we can easily do diffs of `MainNetParms` vs `TestNet3Params`
and easily see how they are the same and how they are different.

I also reorganized the import statements so they “diff” better as well.

Other than those changes the only code I actually changed is
using the constant `PAYMENT_PROTOCOL_ID_UNIT_TESTS` instead of `“unittest”`
in `getPaymentProcolId()` in `UnitTestParams`.

This will help us with the next step which is to create symbolic constants
for some of the parameters for the genesis blocks.

### DIFF
--- a/core/src/main/java/org/bitcoinj/params/MainNetParams.java
+++ b/core/src/main/java/org/bitcoinj/params/MainNetParams.java
@@ -17,12 +17,15 @@
 
 package org.bitcoinj.params;
 
-import org.bitcoinj.core.*;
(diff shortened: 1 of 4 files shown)


schildbach:
Merged.

schildbach:
That's a great achievement, thanks!

PR #2176, opened 2021-09-06
RegTestParams: remove redundant init of genesis block

msgilligan:
Remove the override of `getGenesisBlock()` that mutates the existing genesis block to get the desired block for RegTest. Instead use those parameters directly when building the genesis block in the `RegTestParams` constructor.

(It looks like the override of getGenesisBlock() was added
by Mike Hearn back in 2013 when RegTestParams was a subclass
of TestNet2Params.)


schildbach:
Makes sense. Merged.

PR #2178, opened 2021-09-06
Fix JavaDoc comment on EASIEST_DIFFICULTY_TARGET

msgilligan:
It said “half of all possible hash solutions” which
led me to believe that decodeCompactBits() would
extend 1’s all the way out to the least significant
bit.

(This is a nitpick, but the error caused me to spend an hour researching how difficulty targets work.)


schildbach:
Merged.

PR #2179, opened 2021-09-06
Set UnitTest maxTarget to EASIEST_DIFFICULTY_TARGET

msgilligan:
The tests are currently using a custom value for maxTarget 
that can’t be generated via CompactBits format.

This PR, changes the string constant to match what would be
generated by using Utils.decodeCompactBits(EASIEST_DIFFICULTY_TARGET)


core/src/main/java/org/bitcoinj/params/UnitTestParams.java
```diff
@@ -36,7 +36,7 @@ public UnitTestParams() {
         id = ID_UNITTESTNET;
 
         targetTimespan = 200000000;  // 6 years. Just a very big number.
-        maxTarget = new BigInteger("00ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", 16);
+        maxTarget = new BigInteger("007fffff0000000000000000000000000000000000000000000000000000000000", 16);
```
schildbach, line 39: Perhaps add a comment explaining this magic number represents EASIEST_DIFFICULTY_TARGET? A simple end-of-line comment like the one above would do I'd say.

msgilligan: @schildbach I added the comment and force-pushed.

Of course, this change is a precursor to actually using a `maxTarget` generated from `EASIEST_DIFFICULTY_TARGET`.

msgilligan:
Note: The slight behavioral change in this PR, prepares for the changes in PR #2181 which (after this one is merged) will not change behavior.

PR #2180, opened 2021-09-07
Set RegTest MAX_TARGET to EASIEST_DIFFICULTY_TARGET

msgilligan:
Clear the lowest 232 bits so we're equivalent to EASIEST_DIFFICULTY_TARGET.


msgilligan:
Note: The slight behavioral change in this PR, prepares for the changes in PR #2181 which (after this one is merged) will not change behavior.

schildbach:
Merged.

PR #2182, opened 2021-09-07
ChainSplitTest: use getGenesisBlock() getter

msgilligan:
This was the only place where getGenesisBlock() was
not being used.


schildbach:
Merged.

PR #2183, opened 2021-09-07
Improve Genesis hash checking

msgilligan:
1. Define `Sha256` constants for expected hashes
2. Compare hashes in binary form (as `Sha256`, not `String`)
3. Move hash comparison to immediately follow `Block` creation 
4. Add error message string to the `checkState()` call


schildbach:
Merged.

PR #2184, opened 2021-09-07
NetParams: Move genesis block construction to lazy getter

msgilligan:
This PR will need rebasing after other pending PRs are merged.

This refactoring breaks the cyclic dependency of NetParams constructors
on the constructors in `Block` and will allow us to migrate the code
now in the `getGenesisBlock()` methods into (atomic) factories/constructors
in `Block` allowing us to create block objects that are unmodifiable/immutable.

Currently depends on PR #2183 which was submitted separately.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -256,7 +256,7 @@ public class Block extends Message {
         length = cursor - offset;
     }
 
-    static Block createGenesis(NetworkParameters n) {
+    public static Block createGenesis(NetworkParameters n) {
         Block genesisBlock = new Block(n, BLOCK_VERSION_GENESIS);
         Transaction t = createGenesisTransaction(n, genesisTxInputScriptBytes, FIFTY_COINS, genesisTxScriptPubKeyBytes);
         genesisBlock.addTransaction(t);
--- a/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+++ b/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
@@ -60,7 +60,6 @@ public abstract class NetworkParameters {
 
     // TODO: Seed nodes should be here as well.
 
-    protected final Block genesisBlock;
     protected BigInteger maxTarget;
     protected int port;
     protected long packetMagic;  // Indicates message origin network and is used to seek to the next message when stream state is unknown.
@@ -99,7 +98,6 @@ public abstract class NetworkParameters {
     protected volatile transient MessageSerializer defaultSerializer = null;
 
     protected NetworkParameters() {
-        genesisBlock = Block.createGenesis(this);
     }
 
     public static final int TARGET_TIMESPAN = 14 * 24 * 60 * 60;  // 2 weeks per difficulty cycle, on average.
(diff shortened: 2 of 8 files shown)


msgilligan:
Responding to @schildbach 's comment from a previous commit that has been overwritten by force-push...

> I think the commit message could be clarified in that the cyclic dependency is based on static initializers (if I understand the problem correctly).

The cyclic dependency is:

`final Block GenesisBlock` -> NetworkParameters() (constructor) -> `Block.createGenesis()` (static factory method) -> `Block(n, BLOCK_VERSION_GENESIS)` (constructor)

The code as is works correctly because the initialization of everything occurs in the `NetworkParameters()` constructor, but is fragile and limiting. For example my first thought was to pass the Genesis block in to the `NetworkParameters()` constructor rather than pass in a handful of parameters for the Genesis block, but that is not allowed because of this "cyclic" dependency.

Does the above make sense? Is the above what I should update the commit comment to contain?

schildbach:
Am I correct if I say it's the same issue as this one? https://stackoverflow.com/questions/6416408/static-circular-dependency-in-java

If that's the case, I would perhaps suggest using the term "static circular dependency".

schildbach:
The code changes look good to me!

msgilligan:
> Am I correct if I say it's the same issue as this one? https://stackoverflow.com/questions/6416408/static-circular-dependency-in-java
> 
> If that's the case, I would perhaps suggest using the term "static circular dependency".

It's not the same issue. There are no static variables involved.

In our cases there's a `final` `genesisBlock` member of `NetworkParameters` that must be initialized in the constructor, which calls `Block.createGenesis(this)` because you can't create a `Block` without a `NetworkParameters`.

The main constraints are:

1. The `final` keyword means that `genesisBlock` must be created in the `NetworkParameters` constructor
2. Because a `Block` must have a `NetworkParameters` this means we can't create a genesis block before the `NetworkParameters` constructor is called.

This means we can't create a genesis block and pass it into the `NetworkParameters` constructor and we can't delay the creation until after the  `NetworkParameters` constructor returns. So essentially, as the code is currently structured, the two object must both be created/initialized inside the `NetworkParameters` constructor.

My original plan, which I mentioned before was to just pass the network id string into `Block` and lazy-initialize the `params` field on first use. But I thought about it more and decided it would be better to lazily created the genesis `Block` inside `NetworkParameters` (because this will only happen once) rather than doing the lazy init of `params` for every `Block that we created.

msgilligan:
Maybe I should just leave mention of the "cyclic dependency" out. I may be misusing the term.

The simplest way to say it is that in the current code, a `NetworkParameters` can't be constructed unless it contains a `Block` and a `Block` can't be constructed unless it contains a `NetworkParameters`.  The current code avoids the issue by creating the `Block` inside the constructor for `NetworkParameters`.

This "circular" dependency was forcing me to push more and more parameters into the `NetworkParameters` constructor in order to avoid mutating the `Block` after construction.

schildbach:
Thanks for your thorough explanations! I just had to make my head wrap around it. Guess it was too late yesterday.

Merged.

msgilligan:
It took me a while to get my head wrapped around it, too. Thanks, as always, for your questions, your feedback, and your thorough reviews.

PR #2185, opened 2021-09-08
UtilsTest: resetMocking() after MockClock test

msgilligan:
This should fix #2049. (We'll have to check what happens GitLab CI to know for sure)

Notice that the current time in the failures is: `1970-01-01T09:00:08Z` this definitely points at the test that this PR fixes.
Not just 1970, but notice the 8 seconds past the hour!

Update: Gitlab tests are passing on my fork: https://gitlab.com/ConsensusJ/bitcoinj/-/pipelines/367361062

PR #2186, opened 2021-09-08
Use no-args Utils.setMockClock() where possible

msgilligan:
Replace a few calls to:

```
Utils.setMockClock(Utils.setCurrentTimeSeconds())
```
with

```
Utils.setMockClock()
```

The behavior of the former is slightly different and possibly pathological (i.e. if mocking wasn't reset by the previous test, it won't actually be setting to the "current" time)  However, I believe that in all three cases in this PR, it is the  later behavior that is intended.

Note that this PR, in conjunction with PR #2185 seems to make it clear that none of our blockchain unit tests are using MockTime to set the time in the past.


schildbach:
Merged.

PR #2187, opened 2021-09-08
GithubCI: parameterize JDK distro & gradle version

msgilligan:
Add the parameters, but don’t change what tests
are run (there are two separate PRs that will have those changes)


schildbach:
Merged.

PR #2190, opened 2021-09-09
Block: refactor constructors and factory methods

msgilligan:
* Move construction code from *NetParams into Block
* Create private constructors for special purposes
* Remove duplication in factories / constructors
* Mark apparently unused deserializing constructs as @Deprecated

Draft: Looking for preliminary feedback. After that I'll add comments and update unit tests to use the new methods/constructors.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -117,7 +117,8 @@ public class Block extends Message {
     // of the size of the ideal encoding in addition to the actual message size (which Message needs)
     protected int optimalEncodingMessageSize;
 
-    /** Special case constructor, used for the genesis node, cloneAsHeader and unit tests. */
+    /** Special case constructor, used for unit tests. */
+    @VisibleForTesting
     Block(NetworkParameters params, long setVersion) {
         super(params);
         // Set up a few basic things. We are not complete after this though.
@@ -138,6 +139,7 @@ public class Block extends Message {
      * as the length will be provided as part of the header.  If unknown then set to Message.UNKNOWN_LENGTH
      * @throws ProtocolException
      */
+    @Deprecated
     public Block(NetworkParameters params, byte[] payloadBytes, MessageSerializer serializer, int length)
             throws ProtocolException {
         super(params, payloadBytes, 0, serializer, length);
@@ -171,6 +173,7 @@ public class Block extends Message {
      * as the length will be provided as part of the header.  If unknown then set to Message.UNKNOWN_LENGTH
      * @throws ProtocolException
      */
+    @Deprecated
     public Block(NetworkParameters params, byte[] payloadBytes, int offset, @Nullable Message parent, MessageSerializer serializer, int length)
             throws ProtocolException {
         // TODO: Keep the parent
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -256,11 +280,15 @@ protected void parse() throws ProtocolException {
...
+        Transaction t = createGenesisTransaction(n, genesisTxInputScriptBytes, FIFTY_COINS, genesisTxScriptPubKeyBytes);
+        return new Block(n, BLOCK_VERSION_GENESIS, time, difficultyTarget, nonce, Collections.singletonList(t));
+    }
+
+    @VisibleForTesting
+    public static Block createGenesisUnitTest(NetworkParameters n) {
         Transaction t = createGenesisTransaction(n, genesisTxInputScriptBytes, FIFTY_COINS, genesisTxScriptPubKeyBytes);
-        genesisBlock.addTransaction(t);
-        return genesisBlock;
+        return new Block(n, BLOCK_VERSION_GENESIS, Utils.currentTimeSeconds(), Block.EASIEST_DIFFICULTY_TARGET, Collections.singletonList(t));
```
msgilligan, line 291: After this refactoring (I don't want to change the behavior in the refactoring) we should probably try replacing `Utils.currentTimeSeconds()` with a call to directly get (un-mockable) system time.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -256,11 +280,15 @@ protected void parse() throws ProtocolException {
         length = cursor - offset;
     }
 
-    public static Block createGenesis(NetworkParameters n) {
-        Block genesisBlock = new Block(n, BLOCK_VERSION_GENESIS);
+    public static Block createGenesis(NetworkParameters n, long time, long difficultyTarget, long nonce) {
+        Transaction t = createGenesisTransaction(n, genesisTxInputScriptBytes, FIFTY_COINS, genesisTxScriptPubKeyBytes);
+        return new Block(n, BLOCK_VERSION_GENESIS, time, difficultyTarget, nonce, Collections.singletonList(t));
```
schildbach, line 285: We could use the appropriate TARGET constant here, rather than passing the value in.

schildbach (commented):
Concept looks good.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -190,15 +193,36 @@ public Block(NetworkParameters params, byte[] payloadBytes, int offset, @Nullabl
...
         this.version = version;
         this.prevBlockHash = prevBlockHash;
         this.merkleRoot = merkleRoot;
         this.time = time;
         this.difficultyTarget = difficultyTarget;
         this.nonce = nonce;
-        this.transactions = new LinkedList<>();
-        this.transactions.addAll(transactions);
+        this.hash = hash;
+        if (transactions != null) {
```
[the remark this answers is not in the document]
msgilligan: Good question. Generally its a best practice to use empty lists instead of null lists.

My (probably flawed) thinking was that `transactions == null` for header-only, and that in a real block transactions would be present. Of course if `Block` were truly immutable there would never be a valid block without at least one transaction, right?

core/src/main/java/org/bitcoinj/core/Block.java
msgilligan: We could do that, but there could conceivably be use cases (test cases) where we want the difficulty to be different.

The TARGET constants from the NetworkParameters subclasses are no accessible here, but that could be changed, of course.

core/src/main/java/org/bitcoinj/core/Block.java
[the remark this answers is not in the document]
msgilligan: Yes, the comment for `transactions`  says `// If null, it means this object holds only the headers.` -- but maybe we should change that (probably later, though.)

msgilligan:
@schildbach I responded to your two comments above. Are you ok with leaving both of those things as they are?

My next steps would be to add some comments and update the unit tests to use the new constructors.

msgilligan:
Closing given that other changes and in particular PR #3135 have made this redundant.

PR #2194, opened 2021-09-17
Upgrade to JavaFX 17.0.0.1

msgilligan:
JavaFX 17 is an LTS release and works with JDK 11 or later.

### DIFF
--- a/wallettemplate/build.gradle
+++ b/wallettemplate/build.gradle
@@ -13,7 +13,7 @@ dependencies {
 }
 
 javafx {
-    version = '15'
+    version = '17.0.0.1'
     modules = [ 'javafx.controls', 'javafx.fxml' ]
 }
 


schildbach:
This seems to introduce new requirements to the version of Gradle used. I get

```
WARNING: You use Gradle 4.10.2. The minimum version supported (with some limitations) by this plugin is 5.1.  It is strongly recommended to use at least Gradle 5.6.
```

msgilligan:
I don't think the Plug-in upgrade is strictly necessary. I'll resubmit without the plug-in upgrade.

msgilligan:
Alternatively, we could bump `minFxGradleVersion` since this module is an optional add-on anyway.

schildbach:
I think for now I'd prefer to only update JavaFX.

msgilligan:
> I think for now I'd prefer to only update JavaFX.

OK, I force-pushed an update with just the JavaFX version upgrade.

schildbach:
Merged.

PR #2196, opened 2021-09-17
wallettemplate: move wallettemplate.utils to org.bitcoinj.walletfx.utils

msgilligan:
This is a step towards (a renewed effort at) separating the wallettemplate into a JavaFX wallet library and a template application.

This is a simple package move refactoring. I also moved `bitcoin_logo_plain.png` to its own `images` resource directory since it is shared between two `.fxml` files.


schildbach:
Merged.

PR #2198, opened 2021-09-17
Wrap OkHttp -- make it an `implementation` not `api` dependency

msgilligan:
1. Make usage of OkHttpClient internal to HttpDiscovery.java
2. Deprecate one constructor in HttpDiscovery 
3. Use wrapper in MultiplexingDiscovery
4. Change build.gradle dependency for OkHttp to `api`

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -11,7 +11,7 @@ dependencies {
     api 'org.bouncycastle:bcprov-jdk15to18:1.68'
     api 'com.google.guava:guava:30.0-android'
     api 'com.google.protobuf:protobuf-java:3.13.0'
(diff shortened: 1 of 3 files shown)


schildbach:
Concept looks good to me.

schildbach:
We could also just drop the HttpDiscovery. It's only used by Bitcoin Wallet I assume, and probably my seed is the only one?

I think we should finally implement discovery of new peers via P2P protocol.

msgilligan:
Assuming PR #2279 gets merged we can close this one.

schildbach:
#2279 is merged, so I'm closing this.

PR #2209, opened 2021-09-22
walletfx: Extract `MainWindowController` abstract base class

msgilligan:
* Extract `MainWindowController` from `wallettemplate.MainController`
* WalletApplication: `wallettemplate.MainController` with `MainWindowController`


schildbach:
Merged.

PR #2210, opened 2021-09-22
WalletApplication: improve encapsulation

msgilligan:
* Make all public fields private and wrap with accessors
* Make `WalletAppKit` (was confusingly named `bitcoin`) field a member not a static
* Rename `bitcoin` to `walletAppKit`


schildbach:
Merged.

PR #2214, opened 2021-09-23
BasicKeyChain, DeterministicKeyChain, WalletTool: only use protobuf API that is available also on protobuf-lite.

schildbach:
This is meant to aid migration to protobuf-lite.


schildbach:
Merged.

PR #2212, opened 2021-09-22
walletfx: factor out WalletApplication::startWalletAppKit

msgilligan:
In WalletApplication:

* rename `realStart` to `startImpl`
* Factor out `WalletAppKit` initialization code to `startWalletAppKit` method

Rationale:

* Trying to separate `WalletAppKit` related startup from general JavaFX-related startupand more clearly identify what is depending upon having a `WalletAppKit` instance (this separation is still ongoing) 
* This helps identify items for further improvement like `WalletSetPasswordController.estimateKeyDerivationTimeMsec()` (see forthcoming PR)

PR #2215, opened 2021-09-23
walletfx: refactor initialization of `estimatedKeyDerivationTime`

msgilligan:
These changes should not change behavior of the application

* make `estimatedKeyDerivationTime` private
* Rename `WalletSetPasswordController.estimateKeyDerivationTimeMsec()` to `WalletSetPasswordController.initEstimatedKeyDerivationTime()`
* `WalletSetPasswordController.initEstimatedKeyDerivationTime()` returns `void` (result was previously unused)
* Separate the calculation into a synchronous `estimateKeyDerivationTime()` with no side effects, and `initEstimatedKeyDerivationTime()` which provides the asynchrony via `supplyAsync` and uses `thenAccept` to update the `estimatedKeyDerivationTime ` field.
* In `initEstimatedKeyDerivationTime()` actually implement the comment:  "If we haven't recorded it before"

Rationale:

* Better encapsulation
* Use `CompletableFuture` in an easier-to-read, more idiomatic way
* Separate computation of the value from setting the global static
* Remove confusing (and unnecessary) use of `Platform::runLater`

Will need to rebase after #2212 is merged.

### DIFF
--- a/wallettemplate/src/main/java/org/bitcoinj/walletfx/application/WalletApplication.java
+++ b/wallettemplate/src/main/java/org/bitcoinj/walletfx/application/WalletApplication.java
@@ -18,7 +18,6 @@ package org.bitcoinj.walletfx.application;
 
 import com.google.common.util.concurrent.Service;
 import javafx.application.Platform;
-import javafx.scene.Scene;
 import javafx.scene.input.KeyCombination;
 import javafx.stage.Stage;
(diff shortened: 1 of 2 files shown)


msgilligan:
@schildbach rebased and ready to merge

schildbach:
Merged.

PR #2218, opened 2021-09-28
build.gradle: fix declarations of sourceCompatibility

schildbach:


schildbach:
Looks like this isn't really a fix :cry:

PR #2219, opened 2021-09-28
build.gradle: produce Java 8 bytecode for the core module

schildbach:
It will still only use the Java 7 API.


schildbach:
Merged.

Issue #1874, opened 2019-05-31
Consider splitting the `core` module into `core`, `wallet`, and `net` modules


msgilligan:
It _might_ make sense to split the `core` module into two modules the new module being called something like `base` and ideally having zero dependencies besides the JRE, but perhaps might end up having 1 or 2 dependencies (e.g. protobuf or guava) if unavoidable.

**NOTE:** The proposal has changed to three modules, and the title has been changed. The above is the original proposal unmodified.

msgilligan:
As @Polve said over in issue #1877, this module would contain:

>  just crypto stuff and parsing classes for base bitcoin objects: no network code or logic to handle the blockchain and wallets.

schildbach:
I actually would call it 'core', and try to find other names for the stuff that we exclude. As a step further out, we could also extract a 'wallet' module.

msgilligan:
Yeah, `core` was my first thought, but it seems like the migration would be easier if we called it something else, so we could start moving code into that module gradually as we refactor. That's why I suggested `base`  (which mirrors `java.base` and `javafx.base`)  Backward compatibility would be provided by `core` depending on `base`.

While refactoring we could create an `org.bitcoinj.base` package (with sub packages like `org.bitcoinj.base.hd`, for example) and leave it in the `core` JAR and automatic JPMS module. Later we could put the `org.bitcoinj.base` package in its own JAR and JPMS module.

So we could start with the `hd` sub package [as we discussed](https://github.com/bitcoinj/bitcoinj/pull/1877#issuecomment-501605765).

What do you think?

msgilligan:
I made an attempt at extracting a wallet module (well just refactoring packages within the existing module to eliminate circular dependencies at the package level by introducing a _temporarily-named_ `WalletIF` class)

See PR #2222

msgilligan:
I changed the title of the issue to match my (our*) latest thinking. 

I think we could break into three modules as follows:

`core`: most of existing "core", defines wallet and network API but not implementation. Looses dependencies on okHttp and (hopefully) protobuf.
`wallet`: current protobuf-based wallet implementation, protobuf dependency
`net`: current networking implementation, okhttp and other networking clients using older JDK APIs

This would allow us to develop experimental implementations of updated wallet and network modules, perhaps one or more of the following:

1. A SQLite-based wallet (perhaps even file compatible with the SQLite-based wallet files in Bitcoin Core)
2. A reimplemented network layer using Java 8 APIs
[...]

PR #2222, opened 2021-09-30
Add WalletIF interface for Wallet

msgilligan:
* Remove most direct dependencies on Wallet
* Create WalletIF and supporting classes in org.bitcoinj.core.wallet

Major remaining issue is splitting out pieces of Proto

This is a proof of concept for discussion only. This would be targeted to a future release 0.17 or later.


msgilligan:
Ideally we would rename `WalletIF` to `Wallet` and the existing `Wallet` class would be `DefaultWallet` or `WalletImpl` or something like that.

msgilligan:
This PR was just an experiment and is very stale at this point. A similar refactoring should be attempted in the future, but we'll track this under PR #1874.

PR #2224, opened 2021-10-06
make sure to only use Buffer.position(int), Buffer.flip() and Buffer.rewind()

schildbach:
Without these casts, on some JDKs it links against overloaded methods in MappedByteBuffer and Andoid doesn't have them.

Specifically, if bitcoinj-core is built using OpenJDK 11 from Ubuntu 20.04 LTS, and if the resulting jar is then used in an Android build (AGP 3.1.4, Gradle 4.4.1 from Ubuntu), ProGuard issues the following warnings and breaks the build:

```
Warning: org.bitcoinj.core.CheckpointManager: can't find referenced method 'java.nio.ByteBuffer position(int)' in library class java.nio.ByteBuffer
Warning: org.bitcoinj.core.PeerSocketHandler: can't find referenced method 'java.nio.ByteBuffer position(int)' in library class java.nio.ByteBuffer
Warning: org.bitcoinj.net.BlockingClient: can't find referenced method 'java.nio.ByteBuffer flip()' in library class java.nio.ByteBuffer
Warning: org.bitcoinj.net.ConnectionHandler: can't find referenced method 'java.nio.ByteBuffer flip()' in library class java.nio.ByteBuffer
Warning: org.bitcoinj.store.SPVBlockStore: can't find referenced method 'java.nio.MappedByteBuffer position(int)' in library class java.nio.MappedByteBuffer
```

ProGuard can be instructed to ignore the warnings, but then the Android app will fail at runtime with an exception like this:

```
java.lang.NoSuchMethodError: No virtual method position(I)Ljava/nio/MappedByteBuffer; in class Ljava/nio/MappedByteBuffer; or its super classes (declaration of 'java.nio.MappedByteBuffer' appears in /apex/com.android.art/javalib/core-oj.jar)
	at org.bitcoinj.store.SPVBlockStore.getChainHead(SPVBlockStore.java:259)
	[...]
```


schildbach:
Further investigation into presence of `MappedByteBuffer.position(int)`:

- OpenJDK 8 – not present: https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/nio/MappedByteBuffer.java
- OpenJDK 9 – present: https://hg.openjdk.java.net/jdk9/jdk9/jdk/file/65464a307408/src/java.base/share/classes/java/nio/MappedByteBuffer.java#l218
- OpenJDK 10 – present: https://hg.openjdk.java.net/jdk10/jdk10/jdk/file/777356696811/src/java.base/share/classes/java/nio/MappedByteBuffer.java#l218
- OpenJDK 11 – present: https://hg.openjdk.java.net/jdk/jdk11/file/1ddf9a99e4ad/src/java.base/share/classes/java/nio/MappedByteBuffer.java

I hope I picked the correct source code repositories.

msgilligan:
This looks like a reasonable workaround to me.

There is some risk that in a future JDK it becomes necessary to use the overridden methods.

I approve merging, but I definitely think we should do a "beta" release of 0.16 and see what feedback we get from various developers.

schildbach:
I think it's unlikely the overriding methods will always just call though to the overridden methods. It would be weird if they had a different semantics. So I think that risk is almost non-existent. Or do you have something different in mind?

+1 to the testing.

msgilligan:
Yes, it's unlikely, but still possible. I'm probably overthinking this.

schildbach:
Merged.

PR #2227, opened 2021-10-12
TransactionOutput: fix regression with calculating getMinNonDustValue() for the multisig case

schildbach:


schildbach:
@msgilligan Just checking. This fix also needs to be allied to master, right? Afaict, it's not on master yet.

msgilligan:
I should be able to make a PR in about a week. Anyone feel free to remind me if it hasn't happened by then.

schildbach:
No need to PR this. I can just cherry-pick, but I wanted to make sure it really needs the fix on master too (I tend to say yes).

PR #2228, opened 2021-10-17
README.md: visually separate two command lines

schildbach:


msgilligan:
I'm not sure what problem you are trying to fix here or if the PR actually fixes anything...

PR #2234, opened 2021-11-05
Script: add opcode name to disabled opcode exception message

mlazaric:
This PR adds the name of the specific disabled opcode that led to the `ScriptException` to the exception message. The current message can be a bit vague, especially for beginners who are unfamiliar with which opcodes are disabled, and since the specific opcode which caused the exception is available at the point of the exception being thrown, the opcode name can easily be included in the exception message. 

I was working on a university assignment to create a simple bitcoin script and I ended up using `OP_XOR` while not realizing it was disabled. After I ran it, I got the current exception message without the opcode name which did not help narrow it down to which opcode was causing problems and although I figured out it was `OP_XOR` by looking at https://en.bitcoin.it/wiki/Script I think it makes sense to just include the opcode name in the exception message.

Example of exception messages:
 - Before: `Script included a disabled Script Op.`
 - After: `Script included disabled Script Op XOR`


schildbach:
Thanks! Merged.

PR #2236, opened 2021-11-16
Wallet: Fix vbyte estimation for signing

alexandrupele:


schildbach:
Looks good to me.

I wonder if we could unit-test this. Something like, build a transaction with a fee of 1 sat/vbyte, sign it x times and check if on each occasion the fee is 1 sat/vbyte or (a tiny bit) higher.

schildbach:
Ah, some fee-related unit tests need to be adapted too.

alexandrupele:
> Ah, some fee-related unit tests need to be adapted too.

alexandrupele:
> Ah, some fee-related unit tests need to be adapted too.

Ah, I was hoping I won't have to touch the tests for this change. Thanks for spotting it. I'll try to see if I can run them locally.

schildbach:
I had a second look. I think in fact I *have* coded a "round up" in the original code: `(x + 3) / 4` is a round-up-division by 4, isn't it? I think I was trying to prevent the usage of `Math.ceil()` and the corresponding cast to float and back.

schildbach:
For the record: The "round up" comes from this part of the spec: https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-size-calculations

schildbach:
…and there is another instance of the code in `Transaction.getVsize()`.

alexandrupele:
From what I can tell the expression `(x + 3) / 4` is not rounding up. It is always rounding down. Maybe we need to fix `Transaction#getVsize()` as well?

alexandrupele:
Closing as the rounding up is not the issue. I misunderstood what the intention of (x + 3) / 4 was. It's basically the same as Math.ceil(x/4).

PR #2238, opened 2021-11-16
Wallet, Transaction: use Guava IntMath.divide() for the round-up divisions

schildbach:
The previous trick was confusing.

@alexandrupele This will not fix the fee issue, but it makes the division clearer, without casting to float and back.


schildbach:
Merged.

PR #2239, opened 2021-11-16
Script: take into account the various witness length fields in getNumberOfBytesRequiredToSpend()

### LINKED ISSUE #2235: 0.16 Min relay fee issue caused by vsize miscalculation
In the latest 0.16 version, the fee is calculated at a "per virtual kilobyte" rate which is great, but the actual vsize value seems to be off by 1 vbyte (less than it should be). The problem seems to be in `Wallet#estimateVirtualBytesForSigning`

`
vsize += (script.getNumberOfBytesRequiredToSpend(key, redeemScript) + 3) / 4; // round up
`

The value is not actually rounded up! There is a comment about it but I think the implementation was forgotten.

**What this means in practice is that given a minimum fee rate of 1000 sats/vbyte, the transaction will most likely be ignored or not even relayed by most applications because it's not passing the min relay fee filter.**

schildbach:
This will (hopefully entirely) fix the slight fee underspending for spends from P2WPKH.

Fixes #2235.

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/Script.java
+++ b/core/src/main/java/org/bitcoinj/script/Script.java
@@ -606,7 +606,12 @@ public class Script {
             // scriptSig is empty
             // witness: <sig> <pubKey>
             int compressedPubKeySize = 33;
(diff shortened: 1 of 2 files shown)


schildbach:
Merged.

PR #2244, opened 2021-12-22
Add id() and networkParameters() methods to NetworkEnum

msgilligan:
Use these new methods to simplify BuildCheckpoints and WalletTool. (If/when we upgrade these tools to JDK 17 we'll be able to use [switch expressions](https://www.baeldung.com/java-switch#switch-expressions))

The next step is to move this to the `org.bitcoinj.utils` package in the `bitcoinj-core` module...


msgilligan:
@schildbach this one is ready to merge and independent of any others. It is a simple refactoring that improves `NetworkEnum` and simplifies its use.

schildbach:
We've got this class in the examples, too. Maybe we should develop it unisono until we decide to move it to core.

If we move to to core, I'd vote for renaming it at the same timeto get rid of "Enum" in its name.

msgilligan:
> We've got this class in the examples, too.  Maybe we should develop it unisono until we decide to move it to core?

It is in examples, but I just checked and it is not used there, so we can just delete that one.

> If we move to to core, I'd vote for renaming it at the same time to get rid of "Enum" in its name.

I would prefer to rename it in this PR, so that the move to core only changes package name.

So, I propose the following two changes to this PR:

1. Delete the NetworkEnum in examples
2. Rename `NetworkEnum` to `Network`

schildbach:
Good catch! I went ahead and deleted the unused class right away.

I'm fine with the second change.

msgilligan:
> I'm fine with the second change.

Force pushed!

schildbach:
Merged.

PR #2245, opened 2021-12-24
README.md: update Android runtime requirement to Android 7.0

schildbach:
The Java runtime requirement is left unchanged at Java 8.

This would allow us to use more of the Java 8 API in core. Note we can always only use the subset of the two requirements, as we use one binary for both platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The bitcoinj library is a Java implementation of the Bitcoin protocol, which all
 
 ### Technologies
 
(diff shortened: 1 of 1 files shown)


schildbach:
I should note that Bitcoin Wallet 9.x, which already requires Android 7.0 but is based on bitcoinj 0.16, seems to run without problems.

msgilligan:
This is great news! This gives us (among others) CompletableFuture.

PR #2247, opened 2021-12-28
Add testOnJdk8 task (if Gradle > 6.7)

msgilligan:
* Task `testOnJdk8` is not run by default
* If `testJdk8` Gradle property is set, run `testOnJdk8` as part of `check`
* Set `testJdk8` in `gradle.yml` Github Actions workflow

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -30,7 +30,7 @@ jobs:
         uses: gradle/gradle-build-action@v1
         with:
           gradle-version: ${{ matrix.gradle }}
-          arguments: build --stacktrace
+          arguments: -PtestJdk8=true build --stacktrace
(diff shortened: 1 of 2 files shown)


msgilligan:
@schildbach OK, this is ready for review and or merge.

To run the `testOnJdk8` task you must be using Gradle 6.7 or later. There are two ways to invoke the task:

1. Explicitly by using `gradle testOnJdk8`
2. Implicitly by setting `-PtestJdk8=true` and invoking `check`, `build` or another task that executes tests

I've also added some informative output to show which JDK is being used for each test task.

schildbach:
Looks good! Tests on JDK8 are picked up by our CI builds.

schildbach:
Merged.

PR #2248, opened 2021-12-29
Move Network (enum class) from tools to core (org.bitcoinj.utils package)

msgilligan:


msgilligan:
This PR is dependent on PR #2244

schildbach:
Would you like to rebase now that #2244 is merged?

msgilligan:
rebased and force-pushed.

schildbach:
Generally looks good to me, but I wonder if the `org.bitcoinj.core` package would perhaps be the better target, next to `NetworkParameters`? What do you think is the vision for this class?

msgilligan:
> Generally looks good to me, but I wonder if the `org.bitcoinj.core` package would perhaps be the better target, next to `NetworkParameters`? What do you think is the vision for this class?

The main reason I moved it to the `core` submodule is because it is used by two different command-line tools and I wanted to migrate one of them to a new submodule. I also thought it would be nice to encapsulate the code that maps to `NetworkParameters` inside the `enum` as it cleaned up both the tools that are using it.

I'm not sure I have a vision for it that is worthy of `core/core`. I would like to see a canonical string representation of the concept of a Bitcoin/PoW network, but we aren't quite there yet, IMO. 

The issues I see are:

1. We now have two String representations the dotted "network id" string and the String representation of the current enum constants.
2. There is not a bidirectional 1-1 mapping between NetworkParameters singletons and the enums.
3. We haven't thought through the ramifications to other PoW chains that are based on bitcoinj.

(I'm not quite a Bitcoin maximalist, but I generally think that Bitcoin/BTC should be the only Bitcoin-style/PoW chain. However, if someone is going to develop an alternate chain -- e.g. Doge -- I would prefer if they would contribute upstream to bitcoinj)

I'm not sure any of these issues should be prioritized right now, but for these reasons I'm reluctant to add `Network` to `core/core` but think it is ok to have in `core/utils`. Does this make sense?

schildbach:
Fair enough. Thanks for the explanation.

Merged!

PR #2250, opened 2021-12-31
ECKey: Remove unused `public static boolean FAKE_SIGNATURES` 

msgilligan:
`FAKE_SIGNATURES` is a `public static boolean` that is `@VisibleForTesting` in `ECKey.java`. It is currently unused by any tests, so should be removed.


schildbach:
Good catch!

Issue #2254, opened 2022-01-01
Migrate from Guava ListenableFuture to JDK8+ CompletableFuture


msgilligan:
(This is an expansion of Issue #2101 to include all usage of `ListenableFuture` in the **bitcoinj** API to `CompletableFuture`.)

`CompletableFuture` is the JDK 8 equivalent of Guava's `ListenableFuture` and now that `bitcoin-core` is using JDK 8/Android API 24, we should convert all usage of `ListenableFuture` in our API to `CompletableFuture`. I propose doing this in two steps:

1. Replace all returns of `ListenableFuture` with `ListenableCompletableFuture` which is a subclass of `CompletableFuture` that implements the `ListenableFuture` interface. This will be backward-compatible, but allow clients to migrate to `CompletableFuture`.  Targeted for bitcoinj 0.17 release.  (`ListenableCompletableFuture` will probably be marked as `@Deprecated` since it will be removed in a future release)

2. Replace `ListenableCompletableFuture` with `CompletableFuture` (proposed for bitcoinj 0.18 release)

I'll be submitting a PR with implementation of `ListenableCompletableFuture` and conversion of `AbstractBlockChain` to use this class shortly.

schildbach:
That sounds like a good migration strategy to me.

schildbach:
I just noticed I should have commented this here:

Looks like we need replacements for the various `Futures` helpers. There is a [nice cheat sheet](https://gist.github.com/kofemann/a721af324f8c35f6e9840f583ca40f05), but the replacement methods operate on `CompletableFuture` not `ListenableCompletableFuture`.

msgilligan:
> [nice cheat sheet](https://gist.github.com/kofemann/a721af324f8c35f6e9840f583ca40f05), but the replacement methods operate on `CompletableFuture` not `ListenableCompletableFuture`.


* All of the methods that operate on a `CompletableFuture` instance, (`future` in the cheat sheet) will work on `ListenableCompletableFuture`.
* The two entries in the table that call `new CompletableFuture()` constructor will need to use `new ListenableCompletableFuture()`
* We can add a `static` method `ListenableCompletableFuture.completedFuture(value)`

I don't think we need to create static/factor methods for these cases just change the code as is described in the cheat sheet.

msgilligan:
> I don't think we need to create static/factory methods for all these cases just change our code as to use the inline code in the right column of the cheat sheet.

Although, if we do it might make it easier when we:

[...]

PR #2255, opened 2022-01-01
AbstractBlockChain: migrate to CompletableFuture

msgilligan:
This pull request introduces two new utility classes:

* `ListenableCompletionStage` an interface with a single default method that implements Guava `ListenableFuture` using JDK8's `CompletionStage` 
* `ListenableCompletableFuture` a subclass of `CompletableFuture` that implements `ListenableCompletableFuture` (and therefore `ListenableFuture`)

`AbstractBlockchain::getHeightFuture` is updated to return `ListenableCompletableFuture`.

`BlockChainTest` is updated to use `CompletableFuture`  (although it would continue to work without the change)

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -18,8 +18,6 @@
 package org.bitcoinj.core;
 
 import com.google.common.base.*;
-import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.SettableFuture;
 import org.bitcoinj.core.listeners.*;
 import org.bitcoinj.script.ScriptException;
 import org.bitcoinj.store.*;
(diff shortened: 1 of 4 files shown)


schildbach:
If the other occurences can be migrated as easily as this, I think we can simply migrate everything in one bigger PR.

msgilligan:
> If the other occurrences can be migrated as easily as this, I think we can simply migrate everything in one bigger PR.

As I have available time, I will add commits to this PR. Once we've migrating everything and the code looks good we can squash and merge.

schildbach:
If you'd like I can take that task. What do you think?

msgilligan:
> If you'd like I can take that task. What do you think?

That would be great. I'd be happy to review the PR, of course! Don't know if you can add to this same PR or whether you'd just have to start with this branch, but either way that would be great!

schildbach:
Having looked at a few other cases, seems like most migrations will not be as straghforward as the one you picked. So I guess it makes sense to migrate on a case-by-case basis.

schildbach:
Looks like we need replacements for the various `Futures` helpers. There is a [nice cheat sheet](https://gist.github.com/kofemann/a721af324f8c35f6e9840f583ca40f05), but it operates on `CompletableFuture` not `ListenableCompletableFuture`.

msgilligan:
So merge this one?

schildbach:
Hmm, as this affects API (and not just impl detail like #2257 and #2256) I would rather wait until we're sure we don't get stuck with the migration half-way. Atm I'm afraid we might reconsider our strategry outlines in #2254.

msgilligan:
> as this affects API  ...I would rather wait until we're sure we don't get stuck with the migration half-way.

Sounds reasonable. (If we want to create multiple PRs for each use case of futures, we could merge the two `utils` methods into `master` so those PRs could be independent.)

msgilligan:
@schildbach I just rebased on current master.

Also, I'm trying to figure out how to move forward with this. I'm probably more confident than you that we can convert to `CompletableFuture` relatively easily, but I do understand your concern.

Given that our migration strategy is to use `ListenableCompletableFuture`, I also don't think it would be that bad of an outcome if we decide to release 0.17 without migration all `ListenableFuture` return types.

So, there are a handful of ways we could proceed:

1. Merge this PR as is and then start on some of the other use cases as separate PRs.
2. Split this PR into two PRs: (a) the utility interface + class (b) the change to `AbstractBlockChain`.  We'd then merge (a) and develop PRs for all the other use cases. Once we get PRs for the complete migration we merge them all.
3. Create a branch for the completable future work and develop the entire migration on that branch.

Personally, I prefer #1 as it is the least amount of work (and as I said above I'm confident this path will work out)

schildbach:
Ok, I agree. Let's go with (1) and continue on best-effort. I'll merge this in a moment.

schildbach:
Merged.

PR #2257, opened 2022-01-01
Improve Address comparison using JDK 8 Comparator function composition

msgilligan:


msgilligan:
@schildbach Can you take a quick look at this one?

schildbach:
Impl looks good to me.

I wonder do we have any tests that ensure the ordering doesn't change? I do see some "equals/not equals" tests but not of type "greater/less than".

msgilligan:
> I wonder do we have any tests that ensure the ordering doesn't change? I do see some "equals/not equals" tests but not of type "greater/less than".

There are a few in `LegacyAddressTest`. I also wrote some ordering tests in Kotlin and submitted in this ill-fated PR: https://github.com/bitcoinj/bitcoinj/pull/2036/  

Maybe we should reconsider some of the JUnit upgrades we were looking at now that we're on JDK8/JDK11. I think we may still have some problems with older Gradle, but maybe our usual "progressive enhancement" techniques can be used here.  (e.g. as long as all the tests run on CI and "official" builds, maybe the Debian builds can skip some tests)

schildbach:
Merged.

PR #2260, opened 2022-01-05
Replace `import com.google.common.util.concurrent.*` with single class imports

schildbach:
This is meant to prepare for migrations like #2255.


msgilligan (approved):

schildbach:
Merged.

PR #2264, opened 2022-01-22
Update Comparator anonymous impl. with lambda when possible.

yonetmen:


schildbach:
Looks good to me. There are actually 121 ocurrences of the "Anonymous type can be replaced with lambda" refactoring. Shall we perhaps replace them all at the same time?

yonetmen:
> Looks good to me. There are actually 121 ocurrences of the "Anonymous type can be replaced with lambda" refactoring. Shall we perhaps replace them all at the same time?

Done.

schildbach:
Because of the conflicts, I simply applied the refactoring again, but I credited the authorship to you.

PR #2265, opened 2022-01-23
TransactionTest: rewrite testHashForSignatureThreadSafety() to use an executor

schildbach:
This will actually fail the test if an assert in the worker threads fails.
Previously, it was just printing the exception to standard output.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/TransactionTest.java
+++ b/core/src/test/java/org/bitcoinj/core/TransactionTest.java
@@ -29,6 +29,11 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.math.BigInteger;
 import java.util.*;
+import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


schildbach:
@msgilligan I'm interested in your advice. Is this the way we juggle with threads on Java 8?

msgilligan:
> Is this the way we juggle with threads on Java 8?

I'm pretty sure there is a cleaner implementation for typical use. In this case we are deliberately trying to find thread safety issues and I'll need to think about it a little more carefully.

I'll try to code something up over the next few days.

schildbach:
Thanks for looking into this. Actually don't worry about coding this up – I'd do it. I just wanted to make sure I'm using the desired API here.

msgilligan:
Please see PR #2291 which uses Streams and `CompletableFuture`.

PR #2268, opened 2022-01-27
Utils: add bytesToBigInteger() helper and use it

schildbach:
This is the antagonist to Utils.bigIntegerToBytes().


schildbach:
Merged.

PR #2269, opened 2022-01-28
SigNetParams: support signet

schildbach:


schildbach:
Note: for now, it's missing the checking of the challenge as per BIP-325

msgilligan:
I'm happy to see this work. Looks good after a quick scan...

schildbach:
I should note that this PR doesn't use your fancy new `createGenesisTransaction()` introduced in 56a1e24ed2009dde793c60fe39aed70cea0f21bd. Why? Because the spec only includes the genesis block as a blob.

PR #2277, opened 2022-02-12
core/build.gradle: Set max javadoc warnings to 10,000

msgilligan:
This is for reference. I don't think we want to merge this change.


schildbach:
From the CI output: We've got 4381 warnings.

msgilligan:
> Just an idea: shall we perhaps merge this line in commented form, together with a "uncomment this to see all JavaDoc warnings" comment?

Good idea. I force pushed the change and change the PR from "draft" to "ready to review".

PR #2274, opened 2022-02-10
Transaction: Tests and fix for building simple Segwit transactions (without wallet or network)

msgilligan:
This PR contains a test that reproduces issue #2120 and also provides a possible fix.

* Commit fd49bf263370d91e9bba15c182eadc73ee28bfef adds two tests: one for Legacy addresses that works and one for Segwit addresses that fails (reproduces Issue #2120)
* Commit b926cb8c2b0d82c1473966e2f6ec7a80081796d2 provides a possible fix.

@schildbach if this looks good to you, let me know and I'll add JavaDoc and squash.

PR #2278, opened 2022-02-13
global replace SegWit by segwit

schildbach:
I propose to use consistent capitalization of the word segwit.

As noted in #2274, usage of "segwit" outnumbers "SegWit" by about 5-10x in the source code of Bitcoin Core, the BIPs and the bitcoin.org website.


schildbach:
Merged.

Issue #1978, opened 2020-04-21
Remove dependency on OkHttp in bitcoinj-core


msgilligan:
If I inspected the code correctly it is only used by an experimental feature (`org.bitcoinj.net.discovery.HttpDiscovery` and `org.bitcoinj.net.discovery.MultiplexingDiscovery`)

There are at least 4 ways we could do this:

1) Move experimental features to another module/JAR.
2) Remove the experimental feature all together (or put it on a branch) 
3) Replace the direct dependency on OkHttp with an interface that is implemented elsewhere
4) Mark the dependency as optional in Gradle (I've never done this, but would be willing to try it out)

schildbach:
Hmm, what's so bad about OkHttp? I can't think of any app that could do without it nowadays.

`HttpDiscovery` is in production use since a couple of years. Maybe we should reword the comment in its class JavaDoc?

Isn't OkHttp also used in the payment protocol implementation?

schildbach:
> Isn't OkHttp also used in the payment protocol implementation?

Actually it uses `HttpURLConnection` (in `PaymentSession`). It's a long-time todo to migrate that to OkHttp. We could think about splitting off payment protocol support to a submodule though.

msgilligan:
> Hmm, what's so bad about OkHttp? I can't think of any app that could do without it nowadays.

OkHttp is great. However, there are plenty of applications using **bitcoinj** in environments with other HTTP implementations, running offline, or needing to use a different version of OkHttp. As you know I'm working on packaging apps using `jpackage` and also on native compilation with Graal's `native-image`. In general, I'm trying to minimize dependencies as much as possible,  which is always a good practice for library developers.

(In one particular case, I'm using the `java.net.http` implementation -- see [Introduction to the Java HTTP Client](https://openjdk.java.net/groups/net/httpclient/intro.html) -- that was added in JDK 11. I had problems trying to run Retrospect/OkHttp inside a Java Module and to resolve the issue I created an abstract base class that uses `CompletableFuture`, an Android/JDK8 subclass using Retrospect/OkHttp and and a JDK11+ subclass using `java.net.http`. This allows my `jpackage`-d JavaFX app to use the `java.net.http` and only depend on built-in JDK networking. When I port the app to Android, I'll use the OkHttp subclass/JAR.)

In the short-term I'm looking to eliminate whatever dependencies I can. (For example, I've been trying to replace Guava objects/methods with standard JDK APIs wherever possible.)

[...]

PR #2279, opened 2022-02-16
HttpDiscovery: remove the protocol to discover seeds via HTTP

schildbach:
I propose to remove the `HttpDiscovery`. I appear to be the only entity to run such a seed and to be honest I'm not keen on continuing to do so. Anton is listed in the sourcecode as well, but his seed refuses connections and so far he has not replied to my email.

Nowadays one might better use DoH (DNS over HTTP) anyway.

This PR gets rid of the dependency to `OkHttp`. Also see #1978. An alternative to this PR could be #2198.

Furthermore, though not strictly necessary, I propose to finally implement seeding via P2P protocol before/if we merge this.


msgilligan:
Looks good to me.

gruve-p:
ping @btcontract

schildbach:
Nobody objected, so I went ahead and merged this.

PR #2280, opened 2022-02-16
GetUTXOsMessage, UTXOsMessage: remove support for querying Bitcoin XT nodes for UTXOs

schildbach:
Bitcoin XT is not maintained since years and I haven't seen any running nodes for a while. Afaik there is no other implementation of the protocol extension.

We could try to deprecate first, but I'm not sure this is worth the effort.


msgilligan:
Seems like a lot of unnecessary complexity that nobody is using. Since it's not really an API change we should definitely remove without deprecating.

schildbach:
Well, I think the removal of the two `ListenableFuture<UTXOsMessage> getUTXOs()` methods *is* an API change, but I believe the only user of that API was Mike's Lighthouse platform.

msgilligan:
I'm currently reading: https://github.com/bitcoin/bips/blob/master/bip-0064.mediawiki

msgilligan:
>  _is_ an API change, but I believe the only user of that API was Mike's Lighthouse platform.

Yeah, my language "it's not really an API change" was imprecise and somewhat tongue-in-cheek. My point was that anything using this code is not going to work, therefore removing the code/API will not break a working app.

So I think we can remove it without consequence.

schildbach:
Merged.

PR #2282, opened 2022-02-17
Update SLF4J to 1.7.36

msgilligan:
> Starting with version 1.7.36, slf4j releases will be reproducible. By reproducible we mean that anyone checking out the code corresponding to the release version from source code repository and building that local copy, will obtain an identical binary to the published binary.


schildbach:
Merged.

PR #2286, opened 2022-02-18
TransactionTest: rename the two recently added addSignedInput() tests and verify the signature

schildbach:
I thought it might be a good idea to validate the signature created by `addSignedInput()`.

Also, I thought that because the two tests were added specifically to test `addSignedInput()` we should reflect  that in the tests name – so I renamed them.

@msgilligan what do you think?


schildbach:
I just remembered we can just use `Script.correctlySpends()` instead.

PR #2291, opened 2022-02-22
TransactionTest: Use `CompletableFuture` in `testHashForSignatureThreadSafety`

msgilligan:
This is a follow-up to my comments on PR #2265.

This code has several advantages:

1. It uses CompletableFuture which we are hoping to standardize on
2. `Supplier` type is more general and interoperable than `Callable`
3. No intermediate collection with `nCopies` of the `Callable`
4. No synchronous looping for `result.get()`

It has one big wart, which is that `allOf` only takes an `Array` and because of type erasure, we have to deal with `SuppressWarnings`. We could have put the `SuppressWarnings` directly on the line being used, but I prefer a separate function because it makes the actual usage more readable and can potentially be reused in other tests, etc.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/TransactionTest.java
+++ b/core/src/test/java/org/bitcoinj/core/TransactionTest.java
@@ -29,10 +29,13 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.math.BigInteger;
 import java.util.*;
-import java.util.concurrent.Callable;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
+import java.util.function.IntFunction;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static org.bitcoinj.core.Utils.HEX;
 
@@ -589,8 +592,7 @@ public class TransactionTest {
                 new byte[0],
                 Transaction.SigHash.ALL.byteValue())
                 .toString();
-
-        final Callable<Void> c = () -> {
+        final Supplier<Void> supplier = () -> {
             // ensure the transaction object itself was not modified; if it was, the hash will change
             assertEquals(txHash, tx.getTxId());
             assertEquals(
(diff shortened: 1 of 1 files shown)


msgilligan:
It would also be possible to use `CompletableFuture.supplyAsync(supplier)` and eliminate the creation and shutdown of the `ExecutorService`. The default is to use the "fork-join pool" which by default has 1 thread per processor. I'm not sure if having 100 threads creates any more contention on thread safety, but I decided to not include that change in behavior in this PR.

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -589,8 +592,7 @@ public void testHashForSignatureThreadSafety() throws Exception {
                 new byte[0],
                 Transaction.SigHash.ALL.byteValue())
                 .toString();
-
-        final Callable<Void> c = () -> {
+        final Supplier<Void> supplier = () -> {
```
schildbach, line 595: I wonder if we should go back to a Runnable at this point? Would have the advantage of not having to return a dummy `null` value.

msgilligan: Well it's only a test, so it's not that important how we write it.

But you had specifically asked me what was the best way to do it with JDK 8, so I tried my best to do it with CompletableFuture.

Personally, I would go with CompletableFuture despite the warts. We could move the genericArray method to a utility class in "core" and/or add an allOf(List) method to one of our CompletableFuture utility classes.

msgilligan: And the Void/null thing is "idiomatic" for CompletableFuture -- I've seen it before and I'm used to it now. (Though it did bite me in a generic whenComplete lambda that I wrote)

msgilligan: Oops, we can use `Runnable` with `CompletableFuture`

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#runAsync-java.lang.Runnable-

schildbach: Yes, that's what I was thinking of. Recently I saw [this table](https://stackoverflow.com/a/65772832) and thought maybe those 4 interface types are now the types to go. Do you think otherwise?

msgilligan: The table looks good! 

Supplier is preferable to Callable because it doesn't throw checked exceptions.

I didn't realize Runnable was so officially lambda-friendly.

schildbach:
Merged, after changing from `Supplier` to `Runnable` to get rid of the dummy `returrn null`.

PR #2295, opened 2022-02-23
Convert TransactionBroadcast, Wallet doMaintenance to CompletableFuture

msgilligan:
* Use `ListenableCompletableFuture` on all public APIs
* Use `CompletableFuture` internally, in example, and in tests
* Add new `FutureUtils` class
* Add 2 new static methods to `ListenableCompletableFuture`


msgilligan:
> I guess you have deliberately not used the [completable-futures](https://github.com/spotify/completable-futures) library by dependency?

Yes, I'm trying to avoid adding a dependency -- especially since some/most of our utility classes and methods will be deleted in the following release.

I think "core" should be ruthless about avoiding dependencies. (When hyperbitcoinization happens every Java app will be including bitcoinj for one reason or another 😉)

schildbach:
I rebased this and merged.

Issue #1869, opened 2019-05-29
BIP 44 Keychain support


msgilligan:
The `DeterministicKeyChain` class JavaDoc says that it implements BIP 32. I've been using it with BIP 44 paths and it seems to be working correctly, but the class could be modified/extended to better support BIP 44 (or possibly a subclass could be created) as there are some missing constants and behaviors that could be built-in.

schildbach:
I'm not sure if `DeterministicKeyChain` is the right place to implement BIP44/49/84. It should work on any path. The only assumption is the internal/external separation of addresses which is currently hardcoded. And the lookahead.

Maybe we should move the `*_PATH` constants (but not the `*_SUBPATH` constants) to `KeyChainGroup` or to `KeyChainGroupStructure`?

I'm also not sure if we _need_ to support BIP49 – I'd say we _could_. Personally I hope we can skip it. P2WPKH is the way to go and P2PKH works as a fallback.

Regarding BIP84 … we could start by providing a `KeyChainGroupStructure` implementation that uses BIP44 for P2PKH and BIP84 for P2WPKH.

msgilligan:
> I'm not sure if DeterministicKeyChain is the right place to implement BIP44/49/84. The class should work on any path.

Agreed.

> Maybe we should move the *_PATH constants (but not the *_SUBPATH constants) to KeyChainGroup or to KeyChainGroupStructure?

Yes, that sounds about right.

> I'm also not sure if we need to support BIP49 – I'd say we could. Personally I hope we can skip it. P2WPKH is the way to go and P2PKH works as a fallback.

Yeah, I don't think it is very important (and I don't think I need it for anything I'm planning on doing) but we should probably provide the "purpose" constant and make sure we don't preclude anyone from using BIP 49 if they need to.

> Regarding BIP84 … we could start by providing a KeyChainGroupStructure implementation that uses BIP44 for P2PKH and BIP84 for P2WPKH?

That sounds like the right approach. I also have some ideas and code for an `HDPath` class that replaces (but is compatible with) the current use of `ImmutableList<ChildNumber>` that should make the code a little more readable and maintainable.

msgilligan:
PR #1877 adds the `HDPath` class which might be a good place to put some of the `*_PATH` constants.

HashEngineering:
[...]

PR #2297, opened 2022-02-24
Add BIP-44/BIP-84 KeyChainGroupStructure support

msgilligan:
See Issue #1869


core/src/main/java/org/bitcoinj/wallet/BipStandardKeyChainGroupStructure.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface BipStandardKeyChainGroupStructure extends KeyChainGroupStructure {
+    /**
+     *  Map desired output script type to an account path.
+     *  Default to MainNet and account 0.
+     * @param outputScriptType the script/address type
+     * @return account path
+     */
+    @Override
+    default HDPath accountPathFor(Script.ScriptType outputScriptType) {
```
schildbach, line 33: I still need to wrap my head around the new Java8 features.

This default method seems to be quite generic and not related to BIP-44/49/84 specifically. So wouldn't it make sense to add the default method to `KeyChainGroupStructure` instead and maybe not subclass the interface at all?

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
```diff
@@ -16,14 +16,54 @@
...
 package org.bitcoinj.wallet;
 
+import org.bitcoinj.core.NetworkParameters;
+import org.bitcoinj.crypto.ChildNumber;
 import org.bitcoinj.crypto.HDPath;
 import org.bitcoinj.script.Script;
 
 /** Defines a structure for hierarchical deterministic wallets. */
 public interface KeyChainGroupStructure {
+    ChildNumber PURPOSE_BIP44 = new ChildNumber(44, true);  // P2PKH
```
schildbach, line 26: Maybe we should (probably in another commit) also move the various constants from `DeterministicKeyChain` to here?

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
```diff
@@ -16,14 +16,54 @@
...
+    ChildNumber PURPOSE_BIP49 = new ChildNumber(49, true);  // P2WPKH-nested-in-P2SH
+    ChildNumber PURPOSE_BIP84 = new ChildNumber(84, true);  // P2WPKH
+
+    ChildNumber COINTYPE_BTC = new ChildNumber(0, true);
+    ChildNumber COINTYPE_TBTC = new ChildNumber(1, true);
+    ChildNumber COINTYPE_LTC = new ChildNumber(2, true);
```
schildbach, line 32: Support for litecoin in bitcoinj? Does this make sense?

schildbach (commented):
I think my main concern is that this PR adds another sub-interface. Is this really necessary?

core/src/main/java/org/bitcoinj/wallet/BipStandardKeyChainGroupStructure.java
schildbach: I guess the subclassing is because otherwise the change would be API breaking?

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
msgilligan: I recall the Litecoin value being in the BIP, but it's not there currently. SLIP-44 defines all the other coin types.

I put it there just to remind me that other coin types can go there. 

We can remove it if you want.

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
msgilligan: Sounds good, I'll take a look.

msgilligan:
> this PR adds another sub-interface. Is this really necessary?

I think it was necessary to be non-breaking. But I'm ok with making a breaking change and consolidating the interfaces if you think that's better. 

Make sure to look at how I used the code in the FX wallettemplate PR -- that's the use case I needed this PR for.

I definitely was hoping you might have some ideas for improving this -- and think making some minor breaking changes might be a good idea.

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
msgilligan: Shall I leave it or remove it?

core/src/main/java/org/bitcoinj/wallet/BipStandardKeyChainGroupStructure.java
msgilligan: I quickly liked these from the airport last week and commented on the main thread of this PR that, yeah I did it this way to be non-breaking, but think we should consider moving everything into `KeyChainGroupStructure` and making the breaking change.

msgilligan:
I looked into merging the constants from `DeterministicKeyChain` into `KeyChainGroupStructure` and eliminating the sub-interface and starting thinking that maybe we should simplify/refactor the way `KeyChainGroupStructure` works and is used completely. It seems like there is a lot of redundant code and maybe there is a way to simplify it by looking at the bigger picture. Any ideas @schildbach ?

core/src/main/java/org/bitcoinj/wallet/BipStandardKeyChainGroupStructure.java
moncefdhk999: msgilligan:msgilligan-BipStandardKeyChainGroupStructure

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
msgilligan: I actually think it might be better to put the ChildNumber constants in ChildNumber and the HDPath constants in HDPath.

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
msgilligan: I reluctantly agree with that strategy, but I sometimes think might get more upstream fixes if they could use our JAR directly.

msgilligan:
OK, I think some minor breaking changes would be the best approach here. I would like to consider Wallet Descriptors, but definitely after the 0.17 release.

core/src/main/java/org/bitcoinj/wallet/BipStandardKeyChainGroupStructure.java
msgilligan: OK, I force-pushed a new implementation with everything in `KeyChainGroupStructure` (i.e. without a subclass.)

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
msgilligan: I moved the definitions of the various constants into `ChildNumber` and `HDPath`. What do you think of that? (See new force-push)

Note that `DeterministicKeyChain.BIP44_ACCOUNT_ZERO_PATH` is actually a MainNet path and the path for Testnet is different.

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
msgilligan: `COINTYPE_LTC` removed in new force-push.

msgilligan:
> I ... and starting thinking that maybe we should simplify/refactor the way `KeyChainGroupStructure` works and is used completely. 

I ended up not doing that. I think today's force-push 5131a4b663c976c7d4d21a8204c858884ffa0bb0 is reasonable next-step.

core/src/main/java/org/bitcoinj/utils/Network.java
```diff
@@ -23,6 +23,7 @@
  */
 public enum Network {
     MAIN(NetworkParameters.ID_MAINNET),
+    @Deprecated
```
schildbach, line 26: This small change seems unrelated. I'm not opposed to it, shall we merge this right away?

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
schildbach: Yes, those constants in `ChildNumber` and `HDPath` look much cleaner.

core/src/main/java/org/bitcoinj/utils/Network.java
schildbach: Or maybe it is related, I'm not sure.

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
```diff
@@ -16,21 +16,115 @@
...
 package org.bitcoinj.wallet;
 
+import org.bitcoinj.core.NetworkParameters;
+import org.bitcoinj.crypto.ChildNumber;
 import org.bitcoinj.crypto.HDPath;
 import org.bitcoinj.script.Script;
+import org.bitcoinj.utils.Network;
 
-/** Defines a structure for hierarchical deterministic wallets. */
+/**
```
schildbach, line 25: This comment could perhaps be a bit more comprehensible if it would first state everything for BIP32 wallets, and then everything for BIP44 wallets.

Alternatively, at least I would add: **In the BIP-43 case,** currently only BIP-44…

schildbach:
Looks good! I think this is almost ready to merge.

core/src/main/java/org/bitcoinj/utils/Network.java
msgilligan: It's not really related, but if we're gong to use `Network` enum more seriously, it would be a good idea if it doesn't have two different values that are "the same".

msgilligan: See PR #2434

core/src/main/java/org/bitcoinj/wallet/KeyChainGroupStructure.java
msgilligan: Fixed and force-pushed.

core/src/main/java/org/bitcoinj/utils/Network.java
msgilligan: Removed in favor of PR #2434.

msgilligan:
@schildbach OK, remaining issues addressed and ready for merge.

schildbach:
Merged via 6ab20a098b56b095a56484f9f882f0577e459048 and 406d976fd614f365294b1e511eb908ec48b7eff3.

PR #2299, opened 2022-02-24
WalletFX: Add BIP-44/BIP-84 support

msgilligan:
This depends on PR #2297 and should be rebased once that one is merged.

### DIFF
--- a/wallettemplate/src/main/java/org/bitcoinj/walletfx/application/WalletApplication.java
+++ b/wallettemplate/src/main/java/org/bitcoinj/walletfx/application/WalletApplication.java
@@ -29,6 +29,7 @@ import org.bitcoinj.utils.AppDataDirectory;
 import org.bitcoinj.utils.BriefLogFormatter;
 import org.bitcoinj.utils.Threading;
 import org.bitcoinj.wallet.DeterministicSeed;
+import org.bitcoinj.wallet.KeyChainGroupStructure;
 import org.bitcoinj.walletfx.utils.GuiUtils;
 import wallettemplate.WalletSetPasswordController;
 
@@ -46,16 +47,22 @@ public abstract class WalletApplication implements AppDelegate {
     private WalletAppKit walletAppKit;
     private final String applicationName;
     private final NetworkParameters params;
+    private final KeyChainGroupStructure keyChainGroupStructure;
     private final Script.ScriptType preferredOutputScriptType;
     private final String walletFileName;
     private MainWindowController controller;
 
-    public WalletApplication(String applicationName, NetworkParameters params, Script.ScriptType preferredOutputScriptType) {
+    public WalletApplication(String applicationName, NetworkParameters params, Script.ScriptType preferredOutputScriptType, KeyChainGroupStructure keyChainGroupStructure) {
         instance = this;
         this.applicationName = applicationName;
         this.walletFileName = applicationName.replaceAll("[^a-zA-Z0-9.-]", "_") + "-" + params.getPaymentProtocolId();
         this.params = params;
         this.preferredOutputScriptType = preferredOutputScriptType;
+        this.keyChainGroupStructure = keyChainGroupStructure;
+    }
+
+    public WalletApplication(String applicationName, NetworkParameters params, Script.ScriptType preferredOutputScriptType) {
+        this(applicationName, params, preferredOutputScriptType, KeyChainGroupStructure.BIP43);
     }
 
     public static WalletApplication instance() {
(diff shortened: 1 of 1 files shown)


schildbach:
I guess this can be rebased now.

msgilligan:
@schildbach rebased and ready-for-review

schildbach:
Just for my understanding: This causes new wallets to be created with BIP43 structure, while old wallets will still be loaded as BIP32?

I wonder how migration will work. What happens if we load a BIP32 wallet with only P2PKH chain and then a migration to P2WPKH requested? Will we end up with a mixture of BIP32 and BIP43 paths?

msgilligan:
> Just for my understanding: This causes new wallets to be created with BIP43 structure, while old wallets will still be loaded as BIP32?

Yes.

> I wonder how migration will work.

I haven't thought about migration much. Since WalletTemplate is (still) just a sample/template app, I'm assuming it doesn't support migration.

> What happens if we load a BIP32 wallet with only P2PKH chain and then a migration to P2WPKH requested?

How would you "request" migration in WalletTemplate?

> Will we end up with a mixture of BIP32 and BIP43 paths?

I'm not sure, but I don't think there is inherently a problem with a mixture of BIP32 and BIP43 paths in the same wallet.

msgilligan:
Perhaps (for WalletTemplate) we should encourage/hardcode BIP43? I'm assuming that the "installed base" is effectively zero and we should probably view this a forward-looking app/demo and not try to support BIP32 in this application.

We need to start thinking about migrating to TapRoot. To upgrade to TapRoot, BIP43/P2WKH would migrate to BIP43/P2TR. Perhaps BIP32/P2WKH should migrate to BIP43/P2TR?

schildbach:
> How would you "request" migration in WalletTemplate?

You call `Wallet.upgradeToDeterministic()`. Because it needs the password for encrypted wallets, it's best to do this e.g. when the user is sending or changing the password. Of course, also an explicit "Upgrade" option, asking for the password, is possible.

A somewhat related case is `Wallet.doMaintenance()`.

> I'm not sure, but I don't think there is inherently a problem with a mixture of BIP32 and BIP43 paths in the same wallet.

If we want to allow this, I think we need extensive tests for these cases. Hell can break loose if something with the wallet persistence breaks. But to be honest, at the moment I think we should encourage developers to pick a structure and then stick with it for the entire lifetime of the app.

> Perhaps (for WalletTemplate) we should encourage/hardcode BIP43? I'm assuming that the "installed base" is effectively zero and we should probably view this a forward-looking app/demo and not try to support BIP32 in this application.

Maybe. You've got a point here, but on the other hand if we encourage devs to stay with a structure then we should not state a bad example.

> We need to start thinking about migrating to TapRoot. To upgrade to TapRoot, BIP43/P2WKH would migrate to BIP43/P2TR. Perhaps BIP32/P2WKH should migrate to BIP43/P2TR?

I was going to propose m / 2' for P2TR on BIP32.

msgilligan:
> > I don't think there is inherently a problem with a mixture of BIP32 and BIP43 paths in the same wallet.
> 
> If we want to allow this, I think we need extensive tests for these cases.

Sounds reasonable. 

> I think we should encourage developers to pick a structure and then stick with it for the entire lifetime of the app.

I think that is reasonable. And that is why I'm pushing for BIP-43 for the WalletTemplate. (It's not a published/released/supported app and probably has a user base of _zero_. So I think it's reasonable that we can "switch" to BIP-43 at this point and not consider it a breaking change.)

>> for WalletTemplate we should encourage/hardcode BIP43? 

> Maybe. You've got a point here, but on the other hand if we encourage devs to stay with a structure then we should not state a bad example.

I do think we want to encourage people to use BIP-43 for interoperability reasons. And I don't think anyone reasonable would view `WalletTemplate` as an example of a production/supported **bitcoinj** wallet.

> I was going to propose m / 2' for P2TR on BIP32.

I think that is a reasonable solution. But it's not going to be interoperable (as in restore from keywords) with any other wallets, right?

msgilligan:
> I'm skeptical about changing the default. However if we do, I think it should change at all places. `KeyChainGroupStructure.DEFAULT` comes to mind, as well as some `Wallet` constructors.

I suppose we could deprecate `KeyChainGroupStructure.DEFAULT` and replace it with `KeyChainGroupStructure.BIP32`, `KeyChainGroupStructure.BITCOINJ`, `KeyChainGroupStructure.LEGACY` or something like that.

I do think it is OK to have a separate default for WalletTemplate then for the **bitcoinj** library. And maybe the **bitcoinj** library should have no default at all. But I also feel we should probably encourage people to use the BIP43-family and probably make that the default at some point.

Obviously **Android Wallet** and its large user base is yet another decision.

msgilligan:
I think there should be a progression of steps in moving to BIP43, maybe not in this exact order, but similar:

1. Add basic BIP43 support to `core` -- done
2. Add BIP43 support (and make it default) to `WalletTemplate` -- this PR
3. Add BIP43 option to `wallet-tool`
4. More BIP43 tests
5. Add BIP43 option to `wallet-tool`
5. Make BIP43 the default for new wallets in `wallet-tool`
6. Make BIP43 the default in `WalletAppKit` -- this can be done in conjunction with making the nullable constructor parameter for `structure` non-nullable.
7. Update constructors in `core` to favor BIP43 -- I don't think we should break backward compatibility, so I'm not sure what we would do here.

msgilligan:
@schildbach I just forced-pushed a squash.

schildbach:
Merged.

PR #2302, opened 2022-02-25
[DRAFT] ECKey: when signing, grind for low R values

schildbach:
Our goal is to make transaction sizes more predictable, simplifying the fee calculation.
Also see https://github.com/bitcoin/bitcoin/pull/13666

~FIXME unit tests~ (fixed)
    TODO extract HMacDSAKCalculatorWithEntrophy to top level class?
    TODO make encodeToCompact() its own commit?
    TODO more test vectors from other projects
    TODO should sigs with DER encodings *less* than 70 bytes also be retried?

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/ECKey.java
@@ -47,9 +47,11 @@ import org.bouncycastle.asn1.DLSequence;
 import org.bouncycastle.asn1.x9.X9ECParameters;
 import org.bouncycastle.asn1.x9.X9IntegerConverter;
 import org.bouncycastle.crypto.AsymmetricCipherKeyPair;
+import org.bouncycastle.crypto.Digest;
 import org.bouncycastle.crypto.digests.SHA256Digest;
 import org.bouncycastle.crypto.ec.CustomNamedCurves;
 import org.bouncycastle.crypto.generators.ECKeyPairGenerator;
+import org.bouncycastle.crypto.macs.HMac;
 import org.bouncycastle.crypto.params.ECDomainParameters;
 import org.bouncycastle.crypto.params.ECKeyGenerationParameters;
 import org.bouncycastle.crypto.params.ECPrivateKeyParameters;
@@ -71,6 +73,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.nio.charset.StandardCharsets;
 import java.security.SecureRandom;
 import java.security.SignatureException;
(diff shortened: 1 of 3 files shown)


msgilligan:
Concept ACK.

schildbach:
Rebased to remove one of the two impediments.

The tests need a closer look though.

schildbach:
Rebased.

schildbach:
Major progress: I was able to reproduce a test vector from rust-secp256k1. I added it as a unit test. But I think we need to collect and reproduce more test vectors from projects that claim to be compatible to each other. I just checked Bitcoin Core again, and it seems they still simply disable grinding for during tests.

I think we should decide on what we do if DER encodings happen to be *shorter* than expected. Currently, the grinding only focusses on "too large" encodings of 71 bytes (without the sighash flags byte), but with a chance of 1/256 it can be just 69 bytes or shorter, with a chance of 1/65536 it can be 68 bytes or shorter, and so on. With our goal of always constant length DER encoding in mind, we might want to re-grind those rare "too short" cases as well. However I believe Bitcoin Core doesn't do this atm, so we'd become somewhat incompatible. What's the exact requirement of Sparrow?

Also I think we should not grind at all in the SegWit case. SegWit uses constant length raw encoding rather than DER for sigs in witnesses afaik, so grinding is unnessecary. Grinding has downsides: key material is handled longer/more often (lengthening the window for side channel attacks), more resource intensive signing in general, and ~1 bit of entrophy is lost.

Finally, here is a stack trace of the currently failing unit test:

```
org.bitcoinj.core.TransactionTest > testWitnessSignatureP2WPKH FAILED
    org.junit.ComparisonFailure: expected:<304[50221008b9d1dc26ba6a9cb62127b02742fa9d754cd3bebf337f7a55d114c8e5cdd30be022040529b194ba3f9281a99f2b1c0a19c0489bc22ede944ccf4ecbab4cc618ef3ed]01> but was:<304[40220414110cf9b6d81fe3f22a8cc8bf867c1bbe0672ccfb2e8338db9d6f907105b2802201cfee6a5cb8a6239572963de01744ae66dd4a3af5b1413b489c165bf14ce2976]01>
        at org.junit.Assert.assertEquals(Assert.java:117)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at org.bitcoinj.core.TransactionTest.testWitnessSignatureP2WPKH(TransactionTest.java:322)
```

schildbach:
The above test fails because it contains a signature that needs to be ground and the vector was created before low-R-grinding was invented. However, the signature is for an P2PK input which isn't the focus of this test vector. It's about SegWit-specific inputs, outputs and witnesses. So I decided to change the affected signature to the new value (keeping the old in comments) and call it a day.

msgilligan:
> should sigs with DER encodings _less_ than ~~70~~ [71] bytes also be retried?

No. For two reasons:

1. Retrying shorter signatures to get longer ones does not save fees -- you pay the same either way. But with the shorter signature you might get confirmed slightly faster (because you slightly "overpaid") _and_ you use less space on chain.  See [gmaxwell's comment](https://github.com/bitcoin/bitcoin/pull/13666#issuecomment-405116700) on the Bitcoin Core PR.

2. We should try to use the exact same deterministic algorithm as Bitcoin Core does.

msgilligan:
> TODO make encodeToCompact() its own commit?

Looking at the Bitcoin Core code, I see ` static constexpr unsigned int COMPACT_SIGNATURE_SIZE = 65` and apparently there is information in the first byte to recover the public key from the signature. So we should either implement that feature, restrict access to `encodeToCompact()` (e.g. `private` or package-private), or more carefully document its purpose.

msgilligan:
I am currently working on implementing low-R grinding in secp256k1-jdk, so this PR has been very helpful to me!

When we get an implementation working in secp-jdk, we can either try to update this PR or wait for bitcoinj to switch to secp-jdk to implement low-R.

msgilligan:
The secp256k1-jdk implementation of Low-R signing includes both Bouncy Castle and libsecp256k1/FFM implementations and is in https://github.com/bitcoinj/secp256k1-jdk/pull/321

Issue #2300, opened 2022-02-24
Plugin [id: 'org.openjfx.javafxplugin', version: '0.0.8'] was not found


schildbach:
We do have this issue with all our GitHub Actions builds:

```
org.gradle.api.plugins.UnknownPluginException: Plugin [id: 'org.openjfx.javafxplugin', version: '0.0.8'] was not found in any of the following sources:

- Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
- Plugin Repositories (could not resolve plugin artifact 'org.openjfx.javafxplugin:org.openjfx.javafxplugin.gradle.plugin:0.0.8')
  Searched in the following repositories:
    Gradle Central Plugin Repository
```

At the moment, I assume that we don't have this in our hands, but I thought I'll file it anyway.

schildbach:
GitLab CI works fine.

msgilligan:
Maybe we should try bumping the version.

msgilligan:
@schildbach I haven't seen this failure. Was this intermittent?

I created PR #2306 which upgrades to the latest. This might help, but we should update anyway.

schildbach:
The issue was just intermittent, but I merged #2306 anyway.

PR #2306, opened 2022-03-04
wallettemplate: javafxplugin to 0.0.10

msgilligan:
This might help with Issue #2300, but is worth upgrading anyway.


schildbach:
Merged.

PR #2308, opened 2022-03-04
GithubCI gradle.yml: add name to “Git checkout” step

msgilligan:
This is minor, but IMO improves readability and also matches what I've got in the new `graalvm.yml` "workflow".


schildbach:
Merged.

PR #2312, opened 2022-03-05
Convert (remaining non-Peer-related) to CompletableFuture

msgilligan:
This converts all/most of the remaining non Peer-related code in `core` to use `CompletableFuture`.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/ECKeyTest.java
+++ b/core/src/test/java/org/bitcoinj/core/ECKeyTest.java
@@ -26,10 +26,7 @@ import org.bitcoinj.params.MainNetParams;
 import org.bitcoinj.params.TestNet3Params;
 import org.bitcoinj.params.UnitTestParams;
 import com.google.common.collect.Lists;
-import com.google.common.util.concurrent.Futures;
-import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.ListeningExecutorService;
-import com.google.common.util.concurrent.MoreExecutors;
+import org.bitcoinj.utils.FutureUtils;
 import org.junit.Before;
 import org.junit.Test;
 import org.slf4j.Logger;
@@ -39,12 +36,13 @@ import org.bouncycastle.crypto.params.KeyParameter;
 import java.io.InputStream;
 import java.math.BigInteger;
 import java.security.SignatureException;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Random;
-import java.util.concurrent.Callable;
-import java.util.concurrent.Executors;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import static org.bitcoinj.core.Utils.HEX;
 import static org.bitcoinj.core.Utils.reverseBytes;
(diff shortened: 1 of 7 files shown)


core/src/test/java/org/bitcoinj/core/ECKeyTest.java
```diff
@@ -72,15 +70,18 @@ public void setUp() throws Exception {
     public void sValue() throws Exception {
         // Check that we never generate an S value that is larger than half the curve order. This avoids a malleability
         // issue that can allow someone to change a transaction [hash] without invalidating the signature.
-        final int ITERATIONS = 10;
```
schildbach, line 75: Why this change to `byte`? Feels weird to keep a count in a byte.

core/src/test/java/org/bitcoinj/core/ECKeyTest.java
```diff
@@ -72,15 +70,18 @@ public void setUp() throws Exception {
...
-            sigFutures.add(executor.submit(() -> key.sign(hash)));
-        }
-        List<ECKey.ECDSASignature> sigs = Futures.allAsList(sigFutures).get();
+
+        Function<Byte, ECKey.ECDSASignature> signer = b -> key.sign(Sha256Hash.of(new byte[]{b}));
+        Function<Byte, CompletableFuture<ECKey.ECDSASignature>> asyncSigner = b -> CompletableFuture.supplyAsync(() -> signer.apply(b));
+
+        List<CompletableFuture<ECKey.ECDSASignature>> sigFutures = IntStream.range(0, ITERATIONS - 1)
+                .mapToObj(i -> (byte) i)
+                .map(asyncSigner)
```
schildbach, line 81: Seems to me that instead if signing `hash(i)` we're now signing just `i`. Is that intentional? It probably doesn't matter much though.

core/src/test/java/org/bitcoinj/core/ECKeyTest.java
```diff
@@ -72,15 +70,18 @@ public void setUp() throws Exception {
     public void sValue() throws Exception {
         // Check that we never generate an S value that is larger than half the curve order. This avoids a malleability
```
schildbach, line 71: For both the old and the new code I miss where the actual "check" is executed. Do you know?

schildbach (commented):
I merged most of your changes as separate commits, or will merge soon. For ECKeyTest, I've added a few review comments.

core/src/test/java/org/bitcoinj/core/ECKeyTest.java
msgilligan: Because the code that generates the hashes uses a single-byte buffer and this value shouldn't be made larger than a byte without some consideration.

core/src/test/java/org/bitcoinj/core/ECKeyTest.java
msgilligan: The `.map(asyncSigner)` step calls the `asyncSinger` function, which calls the `signer` function which does the hash.

core/src/test/java/org/bitcoinj/core/ECKeyTest.java
msgilligan: The checks are performed _after_ the async calculations complete and the results of the completed futures are assembled in a list. This starts at line 85 in the "after" code of this PR:

```
for (ECKey.ECDSASignature signature : sigs) {
     assertTrue(signature.isCanonical());
}
```

core/src/test/java/org/bitcoinj/core/ECKeyTest.java
schildbach: Ah, I missed the Sha256Hash in the function.

core/src/test/java/org/bitcoinj/core/ECKeyTest.java
```diff
@@ -72,15 +70,18 @@ public void setUp() throws Exception {
...
-        for (byte i = 0; i < ITERATIONS; i++) {
-            final Sha256Hash hash = Sha256Hash.of(new byte[]{i});
-            sigFutures.add(executor.submit(() -> key.sign(hash)));
-        }
-        List<ECKey.ECDSASignature> sigs = Futures.allAsList(sigFutures).get();
+
+        Function<Byte, ECKey.ECDSASignature> signer = b -> key.sign(Sha256Hash.of(new byte[]{b}));
+        Function<Byte, CompletableFuture<ECKey.ECDSASignature>> asyncSigner = b -> CompletableFuture.supplyAsync(() -> signer.apply(b));
+
+        List<CompletableFuture<ECKey.ECDSASignature>> sigFutures = IntStream.range(0, ITERATIONS - 1)
```
schildbach, line 79: I think no `- 1` needed here. The end is exclusive, just as with the loop before.

core/src/test/java/org/bitcoinj/core/ECKeyTest.java
schildbach: Ok, I see.

Ultimately as I understand the code it just feeds pseudo-random data into the signer.

But for the purpose of this refactoring, it's fine I guess.

core/src/test/java/org/bitcoinj/core/ECKeyTest.java
schildbach: You're right, thanks for pointing this out!

schildbach:
Merged.

PR #2313, opened 2022-03-05
examples: more CompletableFuture conversion

msgilligan:
This should be the last of the non-Peer related conversion in examples.


schildbach:
Merged.

PR #2322, opened 2022-03-07
DeterministicKeyChainTest: use JDK8 in readResourceFile

msgilligan:
This replaces “Beta” classes in Guava with JDK 8 classes/methods.


schildbach:
Merged.

PR #2324, opened 2022-03-08
Peer: Simplify needToRequest set of hashes

msgilligan:
* Make “more immutable” by using stream/toSet
* Eliminate unnecessary use of CopyOnWriteArraySet

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -45,10 +45,10 @@ import javax.annotation.Nullable;
 import java.net.SocketAddress;
 import java.util.*;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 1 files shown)


msgilligan:
First attempt had a silly mistake (that broke tests.)  I force-pushed a corrected version immediately.

schildbach:
Looks good!

Just to clear this up: unnecessary because it's a local variable, so there is no concurrency – right?

msgilligan:
> Just to clear this up: unnecessary because it's a local variable, so there is no concurrency – right?

It's both a local variable and once created it's not mutated. Note that when we upgrade to JDK 11 we can replace `Collectors.toSet` with `Collectors.toUnmodifiableSet` to make the immutability more clear. There is a way to do this in JDK 8 but I think it's not worth the extra complexity because this is a local.

schildbach:
Merged.

PR #2326, opened 2022-03-14
BasicKeyChainTest: fix spontaneous failure of bloom() because of false positive

### LINKED ISSUE #1804: KeyChainGroupTest.bloom does not adequately test bloom filters
Even if the code in question is being tested correctly, the bloom filter will cause tests to fail, on rare occasion, through false positives:

https://github.com/bitcoinj/bitcoinj/blob/59a6407e68288dc28ecc583e20d112f7f8882aa3/core/src/test/java/org/bitcoinj/wallet/KeyChainGroupTest.java#L340

A correct test should actually hit the bloom filter with enough fuzzed data to make a statistically valid determination regarding its false-positive rate, and ensure zero false-negative rate.

schildbach:
This should hopefully fix #1804, or at least make it very rare.

### DIFF
--- a/core/src/test/java/org/bitcoinj/wallet/BasicKeyChainTest.java
+++ b/core/src/test/java/org/bitcoinj/wallet/BasicKeyChainTest.java
@@ -252,13 +252,22 @@ public class BasicKeyChainTest {
         chain.importKeys(key1, key2);
         assertEquals(2, chain.numKeys());
         assertEquals(4, chain.numBloomFilterEntries());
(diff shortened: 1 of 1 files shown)


schildbach:
I must admit that the required tolerance seems excessive. A factor of 4 wasn't enough, I just doubled it to 8.

Maybe I'm not understanding bloom filters well enough )-:

msgilligan:
Looks good to me.

Of course we could try to remove randomness from these tests entirely...

PR #2327, opened 2022-03-16
Simplify Peer.downloadDependenciesInternal

msgilligan:
See the comments on the individual commits for details.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -49,6 +49,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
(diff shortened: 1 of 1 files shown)


schildbach:
Merged.

At first, was going to discuss about the "final" change, but since the entire method was rewritten I simply rolled the commit into the "make functional" commit.

msgilligan:
> At first, was going to discuss about the "final" change

We should talk about that at some point

My take on where _not_ to use final:

 I wish that variables were `final` in Java by default or that Java had something like the `val` keyword like Kotlin. I generally like to make things final/immutable wherever possible - especially with member variables. Java does have the concept of "effectively final"  which is used to determine if a local variable can be used in a Lambda.

When a local variable is "effectively final" (and especially when the code is simple enough for the reader to tell it is effectively final), I prefer _not_ to use the `final` modifier because of the verbosity.

I think that parameters should always be treated as if they are final, so I feel that the `final` keyword is redundant.

Sometimes, when I'm not sure if something is effectively final in code that I am working on, I add the `final` keyword to see if the IDE's compiler reports any errors. If I find it necessary to do this, then I consider that an indication that maybe adding the `final ` keyword should be committed to the codebase. I realize there is no clear rule here, and I'm not dogmatic about leaving or adding the `final` in these borderline cases.

So that's why I removed the final keywords in that particular commit. Especially because they made reading all the method parameters more difficult.

I definitely think we don't want to go through our existing codebase and make method parameters and local variables final wherever it is possible.

schildbach:
I tend to prefer the compile time safety of the `final` keyword to the verbosity it adds. I also would prefer if fields were final by default, but unfortunately this doesn't seem to happen with Java. In my own projects, I've been using `final` whereever possible. In bitcoinj, I tried to maintain the status quo – which seems to be what you describe.

PR #2332, opened 2022-03-21
Peer: Finish CompletableFuture conversion

msgilligan:
* private connect and handshake futures
* PendingPing
* Remaining public methods return ListenableCompletableFuture


schildbach:
Merged.

PR #2335, opened 2022-03-22
PaymentSession: Use CompletableFuture internally (rather than ListenableCompletableFuture)

msgilligan:
Also in support of / as a result of this change:

* remove supplyAsync from ListenableCompletableFuture
* move ThrowingSupplier to FutureUtils

This change is to make future deprecation and/or removal of `ListenableCompletableFuture` easier.


msgilligan:
The problem here is that `sendPayment()` can sometimes return `null`: https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/protocols/payments/PaymentSession.java#L317

A method that returns `CompletableFuture` (or any `Future` subclass) should never return null. It can return an already completed future or a future the completed exceptionally. I'll need to understand the use case better in order to fix this.

schildbach:
I think finally we can get back to this (-:

msgilligan:
> I think finally we can get back to this (-:

I'm back to it but getting an NPE that I currently don't understand. I'll keep debugging for a while...

msgilligan:
@schildbach Once PR #2346 and PR #2347 are merged I will rebase and remove "DRAFT" status.

msgilligan:
@schildbach OK, this one is rebased and ready to merge.

schildbach:
I separated this in 3 commits (63d9b07c8021d577dd5056892dc4458cf01e9bfa, d0db98a462ee2cae2cd578159d99128b0a2a8d41 and 061dfd45a1f2cc3165fcf8ec9a7efa7f145d265f) and merged.

PR #2336, opened 2022-03-22
Wallet: Use CompletableFuture internally

msgilligan:
`BalanceFutureRequest` is a private nested class. It should use `CompletableFuture` and any public API methods should convert to `ListenableCompletableFuture` before returning those futures.


schildbach:
Sorry, I forgot to mark this as merged.

PR #2338, opened 2022-03-22
PaymentSession.sendPayment proper future-style error handling

msgilligan:
This is a breaking change, but the existing code is not using futures properly and I think this fix is worth the breakage.

* Instead of returning `null`, return a failed future
* Instead of throwing errors, return failed futures
* Improve JavaDoc
* Fix tests
* Improve error handling for sendPayment in WalletTool (needs testing)

### DIFF
--- a/core/src/main/java/org/bitcoinj/protocols/payments/PaymentSession.java
+++ b/core/src/main/java/org/bitcoinj/protocols/payments/PaymentSession.java
@@ -21,6 +21,7 @@ import org.bitcoinj.crypto.TrustStoreLoader;
 import org.bitcoinj.params.MainNetParams;
 import org.bitcoinj.protocols.payments.PaymentProtocol.PkiVerificationData;
 import org.bitcoinj.uri.BitcoinURI;
+import org.bitcoinj.utils.FutureUtils;
 import org.bitcoinj.utils.ListenableCompletableFuture;
 import org.bitcoinj.utils.Threading;
 import org.bitcoinj.wallet.SendRequest;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/bitcoinj/protocols/payments/PaymentSession.java
```diff
@@ -304,35 +305,39 @@ public SendRequest getSendRequest() {
      * NOTE: This does not broadcast the transactions to the bitcoin network, it merely sends a Payment message to the
      * merchant confirming the payment.
      * Returns an object wrapping PaymentACK once received.
-     * If the PaymentRequest did not specify a payment_url, returns null and does nothing.
+     * If the PaymentRequest did not specify a payment_url, completes exceptionally.
```
schildbach, line 308: payment_url is optional, so I think completing exceptionally is not the right thing to do.

schildbach: But then again in this state of the protocol, maybe it is an exception.

schildbach:
I'm not concerned about the breaking aspect, since I think the chances this method is used are slim. Bitcoin Wallet only uses the message constructors, not the methods that do the actual network IO (it uses its own).

core/src/main/java/org/bitcoinj/protocols/payments/PaymentSession.java
msgilligan: `payment_url` may be optional, but this method call is `sendPayment` and it is supposed to generate a payment request, send it to the server, and return an ACK. So sending trying to send a payment when there is no URL to send it to seems exceptional to me.

And, in the general case of a `CompletableFuture` it is supposed to either:

a. Complete with a value
b. Complete with an exception

A method that returns a `CompletableFuture` should generally not return `null` because they are supposed to be chain able, etc. I made some changes elsewhere and they were broken by this method returning `null`. I have never seen a method that is supposed to return a `CompletableFuture` (or `Future`) return `null`.

So if this method can't return a Future that completes with an ACK it should complete with an exception.

msgilligan: (This may be preaching to the choir after your second message, but I thought you might appreciate the full rationale)

msgilligan: To drone on a little bit further, the current `sendPayment` has 4 possible behaviors:

1. Returns a Completable future that complete with an ACK
2. Returns a Completable future that completes exceptionally
3. Returns `null`
4. Throws an exception

After this PR only (1) or (2) should occur.

msgilligan: @schildbach is this explanation good enough to consider this issue resolved?

schildbach: > this method call is `sendPayment` and it is supposed to generate a payment request,

It's sending a **payment** message, rather than a **payment request** message, right?

schildbach: > @schildbach is this explanation good enough to consider this issue resolved?

I guess yes (-:

schildbach:
Which of the changes to WalletTool are required to adapt to the breaking change? It seems to me the "Improve error handling for sendPayment in WalletTool (needs testing)" aspect of this PR should be split out.

core/src/main/java/org/bitcoinj/protocols/payments/PaymentSession.java
msgilligan: > > this method call is `sendPayment` and it is supposed to generate a payment request,
> 
> It's sending a **payment** message, rather than a **payment request** message, right?

Yes. I edited the above to say: ~~request~~ message

I confusing used the word "request" as network-level terminology, as in request/response protocol.

msgilligan:
> Which of the changes to WalletTool are required to adapt to the breaking change? It seems to me the "Improve error handling for sendPayment in WalletTool (needs testing)" aspect of this PR should be split out.

I found that the changes required to adapt to the breaking change made the code look a little awkward (and I generally find it confusing to have really long `try` blocks that throw multiple exceptions. I prefer if the the `catch` is more closely linked with the `try` block) so I did the cleanup as part of this PR.

I can split it out, but just be warned that the code might look a little awkward.

schildbach:
Perhaps it helps to merge the error handling improvements first? That's mainly the blocks you commented with "Complete the transaction" and  "Setup network communication (but not PeerGroup)".

msgilligan:
> Perhaps it helps to merge the error handling improvements first? That's mainly the blocks you commented with "Complete the transaction" and "Setup network communication (but not PeerGroup)".
> 
> I can do that if you think it's a way to go.

I thought it would be easier to do it the other way around. So PR #2344 has the changes related to `sendPayment` exceptions. I made this PR a draft and can rebase it once PR #2344 is merged.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -884,36 +885,64 @@ private void send(PaymentSession session) {
...
                 peerGroup.start();
-                peerGroup.broadcastTransaction(req.tx).future().get();
-            } else {
-                PaymentProtocol.Ack ack = future.get();
-                wallet.commitTx(req.tx);
-                System.out.println("Memo from server: " + ack.getMemo());
+                TransactionBroadcast broadcast = peerGroup.broadcastTransaction(req.tx);
+                try {
+                    // Wait for broadcast to be sent
+                    broadcast.future().get();
```
msgilligan, line 924: @schildbach Maybe this peer broadcast code should be moved into a separate, `private` method. Let me know if you want me to do that and then force-push.

msgilligan (commented):
I think the broadcast via peerGroup code should be moved into a separate, `private` message. So, yes, I'm requesting changes in my own PR (although Github won't let me click the "Request Changes" button on my own PR)

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
msgilligan: Oops, I meant to make this comment on the _new_ PR...

schildbach:
Ready for rebase, I guess?

msgilligan:
This PR has been split into PR #2344 and PR #2345. When/if PR #2345 is merged, I will close this one.

PR #2339, opened 2022-03-23
refactoring of code and implemented extracted method, rename variable…

Kavyaraval:
Below are the refactoring done in the code:

1. Extract Method 
I performed extract method refactoring at two places in NativeSecp256k1.java at line numbers 251 and 298.
I just extracted the method into the same class only which was counting byte array for bitcoin transactions and based on the result it was returning the array of bytes. So, I extracted the code of block and define it as privateByte and I made a call at Two places that could increase its maintainability

2. Rename variable
In the chosen project I performed Rename variable refactoring at line number 275 in the WalletTool.java file. Where the name of a variable was s so I have to change it to the comparison string as the variable inside the method was storing the result of compared data

3. Extract Class
Extracted the methods into the other class and simply make a call from that extracted class which will decompose the responsibilities into two classes. In our code, I performed this refactoring in the NativeSecp251k1.java file at line numbers 331, 375, 419, 485. Where I called the method which I extracted in the class RateByteArray

4. Replace conditional with polymorphism
In our project, I made changes to the WalletTool.java file at line number 923 there was a big switch case that was performing comparison for different cases so to hide complexity from users I made an abstract class for each case and each class has defined method wait2() with corresponding parameters. This method was performing conditional operations without defining them explicitly.

5. Push Down Refactoring
In our code, we have the file OverlayableStackPaneController.java this class was inherited by many different subclasses but the method done() was only inherited by four classes so I performed push down refactoring in that four classes I just moved done() method into that subclasses where it was needed and I removed it from the super class.
In sendMoneyController.java the done() method was used at three various places at line numbers 103, 125, 146

### DIFF
--- a/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
+++ b/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
@@ -268,21 +268,21 @@ public class WalletTool implements Callable<Integer> {
             else if (from.startsWith(">")) type = Type.GT;
             else throw new RuntimeException("Unknown operator in condition: " + from);
 
-            String s;
+            String ComparisonString;
             switch (type) {
                 case LT:
                 case GT:
                 case EQUAL:
-                    s = from.substring(1);
+                    ComparisonString = from.substring(1);
                     break;
                 case LTE:
                 case GTE:
-                    s = from.substring(2);
+                    ComparisonString = from.substring(2);
                     break;
                 default:
                     throw new RuntimeException("Unreachable");
             }
-            value = s;
+            value = ComparisonString;
         }
 
         public boolean matchBitcoins(Coin comparison) {
--- a/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
+++ b/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
@@ -176,7 +176,7 @@ public class WalletTool implements Callable<Integer> {
     @CommandLine.Option(names = "--force", description = "Overrides any safety checks on the requested action.")
     private boolean force = false;
     @CommandLine.Option(names = "--wallet", description = "Specifies what wallet file to load and save.")
-    private File walletFile = null;
+    private static File walletFile = null;
     @CommandLine.Option(names = "--seed", description = "Specifies either a mnemonic code or hex/base58 raw seed bytes.")
     private String seedStr = null;
     @CommandLine.Option(names = "--watchkey", description = "Describes a watching wallet using the specified base58 xpub.")
@@ -1105,7 +1105,7 @@ public class WalletTool implements Callable<Integer> {
         wallet.saveToFile(walletFile);
     }
 
-    private void saveWallet(File walletFile) {
+    private static void saveWallet(File walletFile) {
         try {
             // This will save the new state of the wallet to a temp file then rename, in case anything goes wrong.
             wallet.saveToFile(walletFile);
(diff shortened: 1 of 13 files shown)


msgilligan:
Thanks for the pull request! We appreciate your desire and efforts to help **bitcoinj**.

This being a cryptocurrency project that powers wallet implementations, The bar for acceptance of pull requests is very high. A few general suggestions about creating PRs.

1. Before starting look to see if there are any existing issue or PRs related to the code you will be modifying.
2. If possible, do one ore more of the following:
   a. Open an issue suggesting the change
  b. Propose the idea for the change in the Google Group or on the Matrix Chat and get some feedback
3. Submit small PRs that don't bundle un-related changes together. That makes the review, modify, rewrite cycle easier for all of us.

Now for some feedback on the individual changes:

> 1. Extract Method
>    I performed extract method refactoring at two places in NativeSecp256k1.java 


There is an existing PR (that will almost certainly be merged) removing this entire class: https://github.com/bitcoinj/bitcoinj/pull/2267

> 2. Rename variable
>    In the chosen project I performed Rename variable refactoring at line number 275 in the WalletTool.java file. Where the name of a variable was s so I have to change it to the comparison string as the variable inside the method was storing the result of compared data

I don't see this change in the pull request.

> 3. Extract Class
>    refactoring in the NativeSecp251k1.java file 

There is an existing PR (that will almost certainly be merged) removing this entire class: https://github.com/bitcoinj/bitcoinj/pull/2267


> 4. Replace conditional with polymorphism
>    In our project, I made changes to the WalletTool.java file at line number 923 there was a big switch case

I don't see any change at line 923 in `WalletTool.java` but I will say that I'm reluctant at this time to approve any change that spits `WalletTool.java` into multiple files.

> 5. Push Down Refactoring
>    In our code, we have the file OverlayableStackPaneController.java this class was inherited by many different subclasses but the method done() was only inherited by four classes so I performed push down refactoring in that four classes I just moved done() method into that subclasses where it was needed and I removed it from the super class.

Every method of the parent class doesn't have to be used by all subclasses. I think this particular refactoring increases complexity and duplication and would not approve it (at least in its current iteration.)

My suggestion is to try submitting just (2) as a separate pull request. I can't promise we will accept it, but I promise I will at least give it a thoughtful review.

Thanks again for your efforts!

msgilligan (commented):
As per my comments, I think this PR should be closed. The changes discussed in item (2) "Rename variable" (and apparently not present in the submitted code) might be worth submitting as a separate pull request.

schildbach:
I think Sean has summed it up quite nicely. Especially the bundling of multiple unrelated changes into one commit/PR makes it unnecessary hard to review and merge individually. I'm closing this for now.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -268,21 +268,21 @@ public Condition(String from) {
             else if (from.startsWith(">")) type = Type.GT;
             else throw new RuntimeException("Unknown operator in condition: " + from);
 
-            String s;
+            String ComparisonString;
```
msgilligan, line 271: I think this could/should be simplified even further by eliminating this local variable altogether and just assigning `value` directly in the two `case` statements. If you want to submit that change as a separate PR, I would approve it. (But @schildbach 's approval is required for a merge.)

Also, standard Java coding conventions require that local variable start with a lowercase letter, so this would be `comparisonString`. 

This would also be a great candidate for a [switch expression](https://stackoverflow.com/questions/65657169/what-are-switch-expressions-and-how-are-they-different-from-switch-statements) at some time in the future when we might be using JDK 14 or later.

PR #2344, opened 2022-03-25
PaymentSession.sendPayment proper future-style error handling (with minimal WalletTool changes)

msgilligan:
This is a breaking change, but the existing code is not using futures properly and I think this fix is worth the breakage.

* Instead of returning `null`, return a failed future
* Instead of throwing errors, return failed futures
* Improve JavaDoc
* Fix tests
* Update sendPayment in WalletTool to work with the change

(This is similar to a previous PR, but without additional exception-handling changesin the WalletTool)

### DIFF
--- a/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
+++ b/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
@@ -64,6 +64,7 @@ import org.bitcoinj.core.PeerGroup;
 import org.bitcoinj.core.SegwitAddress;
 import org.bitcoinj.core.StoredBlock;
 import org.bitcoinj.core.Transaction;
+import org.bitcoinj.core.TransactionBroadcast;
 import org.bitcoinj.core.Utils;
 import org.bitcoinj.core.VerificationException;
 import org.bitcoinj.core.listeners.DownloadProgressTracker;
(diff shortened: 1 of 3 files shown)


msgilligan:
The `ı` failure is very frustrating. It happens about 1 out of 25 builds on my Mac.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -892,23 +893,37 @@ private void send(PaymentSession session) {
...
-                wallet.commitTx(req.tx);
-                System.out.println("Memo from server: " + ack.getMemo());
+                TransactionBroadcast broadcast = peerGroup.broadcastTransaction(req.tx);
+                try {
+                    // Wait for broadcast to be sent
+                    broadcast.future().get();
+                } catch (InterruptedException | ExecutionException be) {
+                    System.err.println("Failed to broadcast payment " + be.getMessage());
+                    System.exit(1);
+                }
```
msgilligan, line 911: I think the broadcast via peerGroup code should be moved into a separate, private message. So, yes, I'm requesting changes in my own PR (although Github won't let me click the "Request Changes" button on my own PR)

msgilligan (commented):
I would click "Request Changes" in this review, but Github doesn't allow me to request changes in my own PR...

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
msgilligan: Yeah, give me 5-10 minute to change and force-push...

schildbach: > I think the broadcast via peerGroup code should be moved into a separate, private message. So, yes, I'm requesting changes in my own PR (although Github won't let me click the "Request Changes" button on my own PR)

It would be an improvement, yes. Like with the error handling improvement in the other PR. What I usually do – if possible – commit the improvements first (two commits in this case) and then rebase this PR on the improvements.

msgilligan: OK, I'll try to create two new PRs, one with each of the two improvements.

schildbach: Turns out my advice doesn't apply to this situation, since the fallback code also changes śignificantly. Sorry for the noise.

schildbach:
I've merged this now, including the extracted method.

PR #2345, opened 2022-03-25
WalletTool: Improve sendPayment exception handling

msgilligan:
* Split the long `try` block into multiple `try/catch` blocks
* Consistently use System.err.println and System.exit on errors
* System.out.println an information message before falling back to `broadcastPayment`

This contains the changes from DRAFT PR #2338 that were not included in merged PR #2344.


schildbach:
Merged,

PR #2346, opened 2022-03-25
PaymentSessionTest: Minor improvements

msgilligan:
* Make nested classes `static`
* Make members `final` in a few places
* Remove a throws IOException where it can’t happen

### DIFF
--- a/core/src/test/java/org/bitcoinj/protocols/payments/PaymentSessionTest.java
+++ b/core/src/test/java/org/bitcoinj/protocols/payments/PaymentSessionTest.java
@@ -37,7 +37,6 @@ import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 
-import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.util.ArrayList;
@@ -60,14 +59,14 @@ public class PaymentSessionTest {
     private ECKey serverKey;
     private Transaction tx;
     private TransactionOutput outputToMe;
-    private Coin coin = COIN;
+    private final Coin amount = COIN;
 
     @Before
     public void setUp() throws Exception {
         new Context(TESTNET);
         serverKey = new ECKey();
         tx = new Transaction(TESTNET);
-        outputToMe = new TransactionOutput(TESTNET, tx, coin, serverKey);
+        outputToMe = new TransactionOutput(TESTNET, tx, amount, serverKey);
         tx.addOutput(outputToMe);
     }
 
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/bitcoinj/protocols/payments/PaymentSessionTest.java
```diff
@@ -216,8 +215,8 @@ public void testWrongNetwork() throws Throwable {
                 .build();
     }
 
-    private class MockPaymentSession extends PaymentSession {
```
schildbach, line 219: I ran this refactoring on the entire project via 7f38154244315cfcc869f4bd72daea9498567b3a.

core/src/test/java/org/bitcoinj/protocols/payments/PaymentSessionTest.java
```diff
@@ -60,7 +59,7 @@
     private ECKey serverKey;
     private Transaction tx;
     private TransactionOutput outputToMe;
-    private Coin coin = COIN;
```
schildbach, line 63: I think we should just inline this field. Otherwise it's just a "copied constant".

msgilligan: Actually, I think we should rename it to `amount`. It's not a constant that specifies 1 coin, it's a constant that specifies a test amount that must match in the before and after state.

core/src/test/java/org/bitcoinj/protocols/payments/PaymentSessionTest.java
msgilligan: Thanks!

core/src/test/java/org/bitcoinj/protocols/payments/PaymentSessionTest.java
msgilligan: I rebased, renamed `coin` to `amount`, and force-pushed. I'll mark this as resolved and await your review of the updated push.

schildbach:
Merged in two commits (9156dbef36c4a7e66638ded5a96dcfec2f28a099, caa057e40989165926e43bd89fe77ddf26717c0d).

I left out the single "throws cleanup". I wonder if we should uniformly make test methods throw `Exception`, regardless of what they throw and if they throw anything at all. Since test methods are not an interface of anything (except the test framework), it would make our lives easier.

msgilligan:
> I left out the single "throws cleanup". I wonder if we should uniformly make test methods throw `Exception`, regardless of what they throw and if they throw anything at all. Since test methods are not an interface of anything (except the test framework), it would make our lives easier.

While it's true that any unexpected/uncaught exception is a test failure, I like to see accurate and informative `throws` clauses, even in tests, because it help me read the code.

It tells me everything that can go wrong and helps me understand what is going on. This is also helpful for people who look at the tests for sample code when trying to implement something.

I don't think we need to go through and update every test to declare the minimal/exact exceptions thrown, but when working on a test I think it makes sense to improve the `throws` clauses when appropriate.

PR #2347, opened 2022-03-25
PaymentSessionTest.MockPaymentSession: Don’t return a null Future

msgilligan:
* Use `ListenableCompletableFuture(null)` instead.

If I had been more observant, this would have been included in PR #2344.


schildbach:
Merged.

PR #2350, opened 2022-03-25
Peer: rename `needToRequest` to `txIdsToRequest`

msgilligan:
This `Set` is used to filter out duplicate transactions from the `rootTx` inputs and desperately needs a better name.


schildbach:
I understand the old name is not ideal, but the new name doesn't express what these IDs are for. The "unique" can be derived from the type (Set).

What about `txIdsToRequest`?

msgilligan:
> What about `txIdsToRequest`?

Sounds good to me. I'll amend and force-push.

msgilligan:
@schildbach Changed and force-pushed. Ready to merge.

schildbach:
Merged.

PR #2352, opened 2022-03-26
Convert ImmutableList to unmodifiable List

msgilligan:
This converts all remaining instances of `ImmutableList` (except those in core tests) to use unmodifiable `List`


schildbach:
Merged.

PR #2355, opened 2022-03-26
ListenableCompletableFuture: Fix error in comment, rename lambda parameter

msgilligan:
* Fix a comment that said `not-null` where it should have said `null`
* Further clarify that same comment
* Rename lambda parameter from `val` to `value`


schildbach:
Merged.

PR #2357, opened 2022-03-30
PaymentSessionTest: Remove a throws IOException where it can’t happen

msgilligan:
* Remove a throws IOException where it can’t happen.

This is a resubmit of the portion of PR #2346 that was not merged.

See my comment https://github.com/bitcoinj/bitcoinj/pull/2346#issuecomment-1083429123 on the closed PR.

In the case of this specific test, the minimal/specific `throws` clause makes it more clear that the test expects a `PaymentProtocolException.Expired` exception, but the code under test can theoretically throw the more general `PaymentProtocolException` exception which is uncaught by the test.

I can go either way on this, but I wanted @schildbach to consider my argument and take a second look at this one change in isolation.

### DIFF
--- a/core/src/test/java/org/bitcoinj/protocols/payments/PaymentSessionTest.java
+++ b/core/src/test/java/org/bitcoinj/protocols/payments/PaymentSessionTest.java
@@ -37,7 +37,6 @@ import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 
-import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.util.ArrayList;
@@ -119,7 +118,7 @@ public class PaymentSessionTest {
     }
 
     @Test
-    public void testExpiredPaymentRequest() throws PaymentProtocolException, IOException {
+    public void testExpiredPaymentRequest() throws PaymentProtocolException {
         MockPaymentSession paymentSession = new MockPaymentSession(newExpiredPaymentRequest());
         assertTrue(paymentSession.isExpired());
         // Send the payment and verify that an exception is thrown.


schildbach:
> I like to see accurate and informative throws clauses, even in tests, because they help me read the code. They tell me everything that can go wrong and help me understand the bigger picture. Specific exceptions in a throws clause also help your IDE/compiler give you a warnings/errors when you add code that throws a new exception.

I would agree with those arguments if it were production code (e.g. the classes to test, rather than the test itself). For tests, anything that deviates from the expected path/behaviour distracts me. I'd actually prefer to omit the throws clause entirely for tests if it were possible in plain Java.

> They are also helpful for people who look at the tests for sample code when trying to implement something.

Imho tests are not an ideal sample. They _don't_ show the big picture and don't want to, at least not with a single test. If you want the big picture you'd need to aggregate information from a couple of tests (for example see MonetaryFormatTest) – especially those that test (and expect) the exception cases.

> In the case of this specific test, the minimal/specific throws clause makes it more clear that the test expects a PaymentProtocolException.Expired exception, but the code under test can theoretically throw the more general PaymentProtocolException exception which is uncaught by the test.

That's true, but again the PaymentProtocolException exception case isn't important for this particular test. If it would happen it would be unexpected and be handled by the test framework.

I'm not adamant on this one. If you think it's really important I can also live with specific exceptions everywhere. I just think it's not necessary for test methods.

msgilligan:
>For tests, anything that deviates from the expected path/behaviour distracts me. I'd actually prefer to omit the throws clause entirely for tests if it were possible in plain Java.

Fair enough. As you know, I really like writing tests in Spock/Groovy and in that case the `throws` clause is omitted.

> Imho tests are not an ideal sample.

I agree. But nevertheless, sometimes I find myself looking at the tests to understand how the code works.

> I'm not adamant on this one. If you think it's really important I can also live with specific exceptions everywhere. I just think it's not necessary for test methods.

I'm not adamant, either. And I'm certainly not saying we should make a concerted effort to change it everywhere. But when I'm working on a test using the IDE and thinking about exception handling, I tend to make these kind of changes and want to push them upstream. What should I do in the future?

schildbach:
Here is my "competing" PR: #2378

schildbach:
I just merged this via 9669c07a99c5fe2dcdc59ac80348fe1a8e7de7a0.

PR #2358, opened 2022-03-30
WalletTool: remove `throws Exception` from main

msgilligan:
`main()` currently throws no checked exceptions. This is a good thing and
means all checked exceptions are caught and (hopefully properly) handled by
outputting error messages and return a non-zero exit code.


schildbach:
Merged.

PR #2359, opened 2022-03-30
WalletProtobufSerializer: add WalletFactory::DEFAULT

msgilligan:
This makes the default implementation more obvious and easier to reference.

It will also will help us make the `walletFactory` member of `WalletAppKit` not nullable in a dependent PR. (see PR #2361)


schildbach:
Looks good to me.

However I wonder if we should apply https://github.com/bitcoinj/bitcoinj/pull/2370 first and then merge this ("extract method reference as a constant")?

schildbach:
Looks good to me.

However I wonder if we should apply https://github.com/bitcoinj/bitcoinj/pull/2370 first and then merge this ("extract method reference as a constant")?

PR #2360, opened 2022-03-30
WalletProtobufSerializer: update obsolete comment

msgilligan:
`writeWallet` and `readWallet` must have been static methods at one time, but they are instance methods now. Update this misleading comment.


schildbach:
Merged.

PR #2366, opened 2022-03-31
Github gradle.yml workflow: add JDK 18

msgilligan:
Add JDK 18 (Temurin) to the existing build matrix.

I believe this will fail a unit test because of a minor, breaking cryptography change in JDK 18.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        java: ['11', '17']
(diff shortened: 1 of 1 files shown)


schildbach:
#2368 is merged now.

msgilligan:
I think we should merge this and start testing every commit on JDK 18, too. I realize the increased chance of false negatives will be annoying, but maybe that will encourage us to fix more of those cases.

schildbach:
Luckily, GitHub has recently added the action to retry only the failed builds. Previously, you always had to retry the entire matrix.

schildbach:
Merged.

PR #2368, opened 2022-03-31
X509Utils: swallow CertificateParsingException in getDisplayNameFromCertificate() if certificate has no SubjectAltName extension

schildbach:
This fix should make the method compatible with JDK 18.


msgilligan (approved):
Looks good to me!

PR #2369, opened 2022-03-31
Replace Guava Joiner/Splitter with native JDK equivalent

msgilligan:
JDK 8 added capabilities roughly equivalent to what is provided by the Guava `Joiner` and `Splitter` classes. Migrate all internal usage to use the standard JDK 8 equivalent and deprecate the two public declarations of these classes.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AddressV1Message.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressV1Message.java
@@ -92,6 +92,6 @@ public class AddressV1Message extends AddressMessage {
 
     @Override
     public String toString() {
-        return "addr: " + Utils.SPACE_JOINER.join(addresses);
+        return "addr: " + Utils.SPACE_JOINER_FUNC.join(addresses);
     }
 }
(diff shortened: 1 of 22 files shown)


msgilligan:
After a suggestion by @schildbach, I added a second commit (2628a5b8200ce57672d07e6d0255afcb1c19e928) that moved the new/replacement functionality into `InternalUtils` in a `.core.internal` package.

In the future this means we could deprecate other members of `Utils` and/or move them to `InternalUtils` if they are still needed internally.

schildbach:
Looks even better! (-:

One last question: I see a duplicated `commaJoin()` method. Wouldn't it be sensible to move it to `InternalUtils` too (or some other shared class)?

msgilligan:
> Looks even better! (-:
> 
> One last question: I see a duplicated `commaJoin()` method. Wouldn't it be sensible to move it to `InternalUtils` too (or some other shared class)?

`InternalUtils` sounds like the correct place. I'll update the PR to move it there.

msgilligan:
@schildbach: Ready to merge. I can squash and force-push if you like...

schildbach:
Merged!

PR #2371, opened 2022-04-01
WalletProtobufSerializer: declare WalletFactory a @FunctionalInterface

schildbach:
I would actually locate those cases on our entire codebase, but there doesn't seem to be a refactoring for that in IntelliJ (though it seems to know about and verify the annotation).


msgilligan:
> I would actually locate those cases on our entire codebase

I'm not 100% sure that every effective SAM should be declared a `@FunctionalInterface` -- it would be interesting to see the PR if such a refactoring existed.

msgilligan (approved):
Looks good. Let's merge it!

schildbach:
Merged.

PR #2373, opened 2022-04-01
Update a few dependencies

schildbach:
Weirdly, BouncyCastle doesn't appear to be available on Maven Central for me, although IntelliJ alert that there is a new version, and according to bouncycastle.org it was released on Mar 31.

So I'm putting this to draft for now.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -11,7 +11,7 @@ version = '0.17-SNAPSHOT'
 
 dependencies {
     api 'org.bouncycastle:bcprov-jdk15to18:1.71'
-    api 'com.google.guava:guava:31.0.1-android'
+    api 'com.google.guava:guava:31.1-android'
     api 'com.google.protobuf:protobuf-javalite:3.19.4'
     api 'com.squareup.okhttp3:okhttp:3.14.9'
     implementation 'org.slf4j:slf4j-api:1.7.36'
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -10,7 +10,7 @@ plugins {
 version = '0.17-SNAPSHOT'
 
 dependencies {
-    api 'org.bouncycastle:bcprov-jdk15to18:1.70'
+    api 'org.bouncycastle:bcprov-jdk15to18:1.71'
     api 'com.google.guava:guava:31.0.1-android'
     api 'com.google.protobuf:protobuf-javalite:3.19.4'
     api 'com.squareup.okhttp3:okhttp:3.14.9'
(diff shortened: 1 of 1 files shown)


msgilligan:
Yeah, it looks like Bouncy Castle 1.71 has not been released on Maven Central yet.

msgilligan:
@schildbach A related question. Could we switch to the `bcprov-jdk18on` JAR?

From https://www.bouncycastle.org/latest_releases.html :

> **Java Version Details** With the arrival of Java 15. jdk15 is not quite as unambiguous as it was. The **jdk18on** jars are compiled to work with **anything** from Java 1.8 up. They are also multi-release jars so do support some features that were introduced in Java 9, Java 11, and Java 15. If you have issues with multi-release jars see the jdk15to18 release jars below.

> **Packaging Change (users of 1.70 or earlier)**: BC 1.71 changed the jdk15on jars to jdk18on so the base has now moved to Java 8. For earlier JVMs, or containers/applications that cannot cope with multi-release jars, you should now use the jdk15to18 jars.

core/build.gradle
```diff
@@ -10,18 +10,18 @@ plugins {
 version = '0.17-SNAPSHOT'
 
 dependencies {
-    api 'org.bouncycastle:bcprov-jdk15to18:1.70'
-    api 'com.google.guava:guava:31.0.1-android'
+    api 'org.bouncycastle:bcprov-jdk15to18:1.71'
```
msgilligan, line 13: If you revert this to 1.70 the rest of the PR works.

msgilligan (changes requested):

schildbach:
I should have noted that I filed this upstream: https://github.com/bcgit/bc-java/issues/1139

core/build.gradle
schildbach: If the missing BouncyCastle turns out to be more than a mistake, I will remove the BC update from this PR.

schildbach:
Note it's still bundled, the jar can be downloaded here: https://www.bouncycastle.org/download/bcprov-jdk15to18-171.jar

schildbach:
> A related question. Could we switch to the `bcprov-jdk18on` JAR?

Not without the guarantee that they don't use API that is not available in the Android API we want to support.

schildbach:
Their release announcement which I just got via email says:

> Finally, for users of the maven repositories, 1.71 should appear in the
next few days.

msgilligan:
> > A related question. Could we switch to the `bcprov-jdk18on` JAR?
> 
> Not without the guarantee that they don't use API that is not available in the Android API we want to support.

Right. Even if `bcprov-jdk18on` works with Android, there's no guarantee an updated version won't use an unsupported API.

schildbach:
BC 1.71 is now on Maven Central, so I merged this.

PR #2375, opened 2022-04-01
core: switch from BouncyCastle bcprov-jdk15to18 to bcprov-jdk15on

msgilligan:
This will pave the way for switching to the `bcprov-jdk18on` JARs once
they make it to Maven Central. The build and tests seem to work,
but @schildbach can you try it on Android to make sure that your
Android build tooling doesn’t have any issues with multi-release jars?

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -10,7 +10,7 @@ plugins {
 version = '0.17-SNAPSHOT'
 
 dependencies {
-    api 'org.bouncycastle:bcprov-jdk15to18:1.70'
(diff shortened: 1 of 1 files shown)


msgilligan:
Closing. If we get a `bitcoinj.api.crypto` interface we can look at using the JDK18on versions of Bouncy Castle.

msgilligan:
Closing. If we get a `bitcoinj.api.crypto` interface we can look at using the JDK18on versions of Bouncy Castle.

PR #2378, opened 2022-04-01
if tests throw a checked exception, only declare a simple `throws Exception`

schildbach:
For tests that don't throw a checked exception, the `throws` clause is
removed entirely.

There is no need to declare a proper API, as test frameworks are the
only consumer. Rather, keep the tests simple.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/CheckpointManagerTest.java
+++ b/core/src/test/java/org/bitcoinj/core/CheckpointManagerTest.java
@@ -33,28 +33,28 @@ public class CheckpointManagerTest {
     NetworkParameters params;
 
     @Test(expected = NullPointerException.class)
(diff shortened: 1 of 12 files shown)


msgilligan (commented):
I like the removal of `throws Exception` from the methods that don't throw exceptions, but I don't favor the IMO oversimplification to `Exception` in the other tests.

We've discussed this elsewhere, and ultimately I'm ok with the change, just not enthusiastic about it.

schildbach:
I merged the uncontroversial part of this via 9669c07a99c5fe2dcdc59ac80348fe1a8e7de7a0.

I'll keep the "oversimplification" part open for a while, hoping others might chime in.

schildbach:
I'm closing this as there seems to be no support for my idea.

PR #2379, opened 2022-04-01
Wallet: add loadFromFile method that supports WalletTool

msgilligan:
* Add `Wallet.loadFromFile` that takes options needed by WalletTool
* Add `Wallet.loadFromFileStream` that supports the extra options
* Update `WalletTool` to use the new `loadFromFile` method

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -1663,17 +1663,32 @@ public class Wallet extends BaseTaggableObject
     }
 
     /**
-     * <p>Returns a wallet deserialized from the given file. Extensions previously saved with the wallet can be
-     * deserialized by calling @{@link WalletExtension#deserializeWalletExtension(Wallet, byte[])}}</p>
+     * Returns a wallet deserialized from the given file. Extensions previously saved with the wallet can be
+     * deserialized by calling @{@link WalletExtension#deserializeWalletExtension(Wallet, byte[])}}
      *
      * @param file the wallet file to read
      * @param walletExtensions extensions possibly added to the wallet.
+     * @return The Wallet
+     * @throws UnreadableWalletException if there was a problem loading or parsing the file
      */
     public static Wallet loadFromFile(File file, @Nullable WalletExtension... walletExtensions) throws UnreadableWalletException {
-        try {
-            try (FileInputStream stream = new FileInputStream(file)) {
-                return loadFromFileStream(stream, walletExtensions);
-            }
+        return loadFromFile(file, false, false, walletExtensions);
+    }
+
+    /**
+     * Returns a wallet deserialized from the given file. Extensions previously saved with the wallet can be
+     * deserialized by calling @{@link WalletExtension#deserializeWalletExtension(Wallet, byte[])}}
+     *
+     * @param file the wallet file to read
+     * @param forceReset force a reset of the wallet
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -1668,12 +1668,25 @@ public Context getContext() {
...
      */
     public static Wallet loadFromFile(File file, @Nullable WalletExtension... walletExtensions) throws UnreadableWalletException {
-        try {
-            try (FileInputStream stream = new FileInputStream(file)) {
-                return loadFromFileStream(stream, walletExtensions);
-            }
+        return loadFromFile(file, false, false, walletExtensions);
+    }
+
+    /**
```
schildbach, line 1678: There is an incomplete JavaDoc here.

schildbach (commented):
Generally looks good. Except:

* The JavaDocs need work
* It would be great to have unit tests for the new API

Also I wonder if we should perhaps drop wallet extensions?

msgilligan:
> * The JavaDocs need work
> * It would be great to have unit tests for the new API


Yes, that's why it's a DRAFT 😁. I will add them.

> Also I wonder if we should perhaps drop wallet extensions?

I think there is one use of wallet extensions in the JavaFX wallet. We'll need to remove that first. Are there any other existing uses that you know of?

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: Fixed.

msgilligan:
@schildbach Actually, I think it would be better to add the tests in a separate PR. I'm hoping we can develop these tests with JUnit 5, so we need to pick up that discussion and see if we can figure out a way to use JUnit 5 for at least some subset of tests.  See https://github.com/bitcoinj/bitcoinj/pull/2022 -- but I think what we should do is add a integration-tests subproject to do this in. We could then check in a few `Wallet` data files and use them for testing these new methods _and_ for testing `WalletTool`.

schildbach:
> I think there is one use of wallet extensions in the JavaFX wallet. We'll need to remove that first. Are there any other existing uses that you know of?

I'm not aware of any.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: It's still missing the overall description and the description of the `ignoreMandatoryExtensions` parameter.

msgilligan: That should be resolved. For some reason that latest force push is not showing here. I'm going to mark resolved.

msgilligan (commented):
OK, the remaining JavaDoc issues have been resolved and force-pushed. Should be ready re-review/merge.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: Ignore my previous confusion, I'll fix and force push...

schildbach:
Merged, except the one small change to the `call()` exception declaration. It doesn't feel related to this PR to me.

msgilligan:
> except the one small change to the `call()` exception declaration. It doesn't feel related to this PR to me.

OK, here's that change in a separate PR: https://github.com/bitcoinj/bitcoinj/pull/2387

PR #2382, opened 2022-04-07
WalletTool: remove unnecessary `null` setting

msgilligan:
Remove unnecessary setting of variables to `null` (IDE warning)


schildbach:
Merged.

PR #2381, opened 2022-04-06
WIP: Integration test submodule without Kotlin

msgilligan:

PR #2384, opened 2022-04-09
core tests: add AddressComparatorSortTest

msgilligan:
This is a version of the Kotlin test from PR #2036 but ported to Java 8, JUnit 4, and included in the main `core` unit test tree.

It's not quite as elegant as the Kotlin version, but it still is useful test for making sure comparison/sorting works correctly across network, address type, and the actual Address data.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/AddressComparatorSortTest.java
+++ b/core/src/test/java/org/bitcoinj/core/AddressComparatorSortTest.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright by the original author or authors.
+ *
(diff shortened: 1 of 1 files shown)


schildbach:
I admit I prefer this one to #2381 (or #2036). Somehow I feel the included test is a unit test, not an intergration test. It should be run on (reasonably) any build of the core module, and there is no external dependency that could make it break, right?

Examples for integration test might be networking tests against test servers, using the real network. Tests that allocate ports, like those that fail now and then in our current test suite.

So if you don't mind, I'd merge this one.

msgilligan:
I agree! Please merge.

schildbach:
Merged.

PR #2386, opened 2022-04-12
add first integration test: WalletLoadTest.java

msgilligan:
Depends on PR #2381, will be rebased after that is merged.


integration-test/src/test/java/org/bitcoinj/wallet/WalletLoadTest.java
```diff
@@ -0,0 +1,27 @@
+package org.bitcoinj.wallet;
```
schildbach, line 1: I'm afraid this is missing the license.

integration-test/src/test/java/org/bitcoinj/wallet/WalletLoadTest.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Basic test of loading a wallet from a known test file
+ */
+public class WalletLoadTest {
+    private static final File walletFile = new File("src/test/resources/org/bitcoinj/wallet/panda-test-wallet.wallet");
+    private static final String testWalletMnemonic = "panda diary marriage suffer basic glare surge auto scissors describe sell unique";
+    private static final long testWalletCreation = 1554102000;
+
+    @Test
+    void basicWalletLoadTest() throws UnreadableWalletException {
```
schildbach, line 18: Is there a reason this test isn't public?

msgilligan: In JUnit 5 tests don't need to be public and the recommendation is to use default (package-private) visibility:

> It is generally recommended to omit the public modifier for test classes, test methods, and lifecycle methods unless there is a technical reason for doing so 

See: https://junit.org/junit5/docs/current/user-guide/#writing-tests-classes-and-methods

I don't have a strong preference one way or the other.

integration-test/src/test/java/org/bitcoinj/wallet/WalletLoadTest.java
msgilligan: Fixed. Sorry I keep doing this.

schildbach:
Merged!

PR #2389, opened 2022-04-13
NetworkParameters,AbstractBitcoinNetParams: fix JavaDoc warnings

msgilligan:


schildbach:
Merged.

It's great to see the number of warnings reduce!

msgilligan:
> It's great to see the number of warnings reduce!

Every once in a while, I decide to eliminate JavaDoc warnings in another couple files. It's not fun work, but I want to keep making progress on this.

One thought/suggestion: Maybe we should use a tool (IDE?) to automatically add all the missing `@return` and `@param` fields but leave them blank and check that in as a single commit. It won't change the number of warnings, but it will do two things:

1. reduce the amount of manual editing required to eliminate the warnings altogether
2. Highlight the missing documentation more clearly in IntelliJ (and other IDEs I assume)

PR #2393, opened 2022-04-14
HDPath: deprecate list constructor

msgilligan:
This constructor is/was used in HDUtils methods that were deprecated/removed (removed in PR #2392)
The .M(list) and .m(list) factory methods should be used instead.

Also add a mild warning to the (boolean, List) constructor.


schildbach:
Merged.

PR #2394, opened 2022-04-14
Improve HDPath deserialization in DeterministicKeyChain

msgilligan:
This is WIP. I don't fully understand how the code currently works. In theory, this PR shouldn't change behavior, but I'm looking for input on whether there is a better solution.


schildbach:
Merged.

PR #2396, opened 2022-04-14
HDPathTest: Add 2 more asserts

msgilligan:
Test that basePath is initialized correctly and that `hasPrivateKey()` is tested somewhere.


schildbach:
Merged.

PR #2399, opened 2022-04-14
DeterministicKeychain: replace PeekingIterator with standard Iterator

msgilligan:
No features of the PeekingIterator were being used.


schildbach:
Merged.

PR #2402, opened 2022-04-16
CoinSelection: make fields final (minor breaking)

msgilligan:
This could be breaking if anyone is setting these fields directly, but they really shouldn't be doing that. Making these fields immutable is a good practice and also a first step towards other possible improvements to `CoinSelector` and `CoinSelection`


schildbach:
Merged happily (-:

PR #2403, opened 2022-04-16
HDPath: add .parent() method.

msgilligan:
The motivation for this addition can be seen in this commit from a dependent draft PR: 15cccc34c820c41bc550b4343e76f2fff66805ec

### DIFF
--- a/core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
+++ b/core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
@@ -59,6 +59,21 @@ public class HDPathTest {
     }
 
     @Test
+    public void testParent() {
+        HDPath path1 = HDPath.parsePath("m/0H/1H");
+
+        assertEquals(HDPath.parsePath("m/0H"), path1.parent());
+
+        HDPath path2 = HDPath.parsePath("m/0H");
+
+        assertEquals(HDPath.parsePath(""), path2.parent());
+
+        HDPath path3 = HDPath.parsePath("");
+
+        assertEquals(HDPath.parsePath(""), path3.parent());
+    }
+
+    @Test
     public void testAncestors() {
         HDPath path = HDPath.parsePath("m/0H/1H/0H/1/0");
 
--- a/core/src/main/java/org/bitcoinj/crypto/HDPath.java
+++ b/core/src/main/java/org/bitcoinj/crypto/HDPath.java
@@ -40,6 +40,8 @@ import java.util.stream.Stream;
  * where necessary in your code. Although it is recommended to use the {@code HDPath} type for clarity and for
  * access to {@code HDPath}-specific functionality.
  * <p>
+ * Note that it is possible for {@code HDPath} to be an empty list.
+ * <p>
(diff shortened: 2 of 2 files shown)


core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
```diff
@@ -58,6 +58,22 @@ public void testExtendVarargs() {
...
 
+    @Test
+    public void testParent() {
+        HDPath path1 = HDPath.parsePath("m/0H/1H");
+
+        assertEquals(HDPath.parsePath("m/0H"), path1.parent());
+
+        HDPath path2 = HDPath.parsePath("m/0H");
+
+        assertEquals(HDPath.parsePath(""), path2.parent());
```
schildbach, line 69: I'm not too familiar with these paths, but should this parent not be 'm'?

core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
```diff
@@ -58,6 +58,22 @@ public void testExtendVarargs() {
...
+
+        HDPath path3 = HDPath.parsePath("");
+
+        assertEquals(HDPath.parsePath(""), path3.parent());
```
schildbach, line 73: We could also throw an exception in this case? What's the reason for making this edge case legal?

core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
msgilligan: It's actually an empty path at that point. The string representation is a little confusing. But thing of the underlying `List<ChildNumber>`. That is an empty list.

core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
msgilligan: I generally try to avoid exceptions and nulls, when possible. In this case, I think it make sense to return an empty list. In most use cases I expect the code will check before calling `.parent()`, but it is also possible to check for empty list after calling parent.

msgilligan: If look at https://github.com/bitcoinj/bitcoinj/commit/15cccc34c820c41bc550b4343e76f2fff66805ec (lines 904-905 in the changed file) you'll see the check for `!path.isEmpty()` before calling `.parent()`.

core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
schildbach: Ok, agreed.

core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
msgilligan: JavaDoc updated to explain this behavior more clearly.

msgilligan:
@schildbach comments added, force-pushed, and ready to re-review/merge.

msgilligan:
One more force push, I discovered a typo (extra '*') in a comment and removed it.

schildbach:
Merged.

PR #2405, opened 2022-04-16
BIP32Test: Use HDPath

msgilligan:
Use HDPath to simplify the tests and also be more consistent in using HDPath.


schildbach:
Merged.

PR #2408, opened 2022-04-17
DeterministicKeyChain.get*Keys(): Use Stream and Predicate to simplify

msgilligan:
* add `getKeys(Predicate<DeterministicKey> keyFilter)` method that gets keys filtered by a predicate
* add `filterKeys(boolean includeLookahead, boolean includeParents)` to return a key-filtering predicate
* Rewrite existing methods to use these two building blocks


schildbach:
Merged!

Its interesting to see the Streams coming to fruition.

PR #2409, opened 2022-04-18
ECKey, DumpedPrivateKey: move from core to crypto (minor breaking)

msgilligan:
This was an experiment to move `ECKey` from `core` to `crypto`. This would be a first step of moving all Bouncy Castle code to `crypto` and then wrapping all Bouncy Castle classes behind interfaces so they are no longer visible to `core`. The step after  that would be to split the Bouncy Castle implementation into a separate module/JAR.


msgilligan:
After this PR the remaining reference to `org.bouncycastle.crypto` outside of the `crypto` package seem to be limited to 3 cases:

1. Use of `RIPEMD160Digest` in calculating hash address. This can easily be wrapped in an interface in the `crypto` package.
2. Use of `digest.SHA3` for calculating `onionChecksum()`. This can also easily be wrapped in an interface in the `crypto` package.
3. Use of `org.bouncycastle.crypto.params.KeyParameter`. This occurs in both `core` and in `wallet`. This needs further study.

schildbach:
I generally like the idea of modularization. So I'm not opposed to moving the class to the crypto package if we perceive this class as a crypto API (as opposed to a handle for a "key that can hold Bitcoin").

At the moment, I'm not convinced we need pluggable crypto implementations. However, of course being able to switch it out quickly (e.g. if BC would turn hostile) would be nice to have.

This PR seems to have some imports organized that are not relevant to this PR. Note I've cherry-picked the DumpedPrivateKeyTest accessor fix right away.

msgilligan:
> I generally like the idea of modularization.

Me, too. If we can make **bitcoinj** more modular it will simplify the code. Even if we decide not to make pluggable crypto implementations, moving in that direction will help us structure the code better.

>  I'm not opposed to moving the class to the crypto package if we perceive this class as a crypto API (as opposed to a handle for a "key that can hold Bitcoin").

From looking at the code and looking at the changes in `import` statements in this PR, it seems like `ECKey` functions well as a wrapper around Bouncy Castle details. It looks like `KeyParameter` is the only class from BC that is present in the public API of `ECKey`.


> 
> At the moment, I'm not convinced we need pluggable crypto implementations. However, of course being able to switch it out quickly (e.g. if BC would turn hostile) would be nice to have.

I agree. It's not a "need" but more of a "nice to have". Of course that could change in the future...

> This PR seems to have some imports organized that are not relevant to this PR.

It looks like IntelliJ did that automatically as part of the refactoring. I can break them into a separate commit if you like.

> Note I've cherry-picked the DumpedPrivateKeyTest accessor fix right away.

I've rebased/pushed to account for that and some other conflicts from recent merges.

msgilligan:
> > This PR seems to have some imports organized that are not relevant to this PR.
> 
> It looks like IntelliJ did that automatically as part of the refactoring. I can break them into a separate commit if you like.

I used the IntelliJ Optimize Imports command to optimize imports for the entire `core` module and submitted as PR #2417. If you agree that this is the "correct" canonical format, I think we should make the change in a single PR to avoid these kinds of problems in the future. We were also wanting to eliminate wildcard imports and PR #2417 does that, too.

msgilligan:
This has been implemented by other PRs that have already been merged.

PR #2213, opened 2021-09-23
PeerGroup: migrate to BlockchainDownloadEventListener (breaking)

msgilligan:
* Deprecate `PeerDataEventListener` (aggregate of 4 listeners)
* Remove long-deprecated `AbstractPeerDataEventListener` (4 no-op listeners)
* Add `BlockchainDownloadEventListener` (aggregate of 2 listeners)
* `PeerGroup`: replace `PeerDataEventListener` with `BlockchainDownloadEventListener` (BREAKING!)
* `DownloadProgressTracker`: implement `BlockchainDownloadEventListener` instead of `AbstractPeerDataEventListener`

This is a breaking change and will require 2 simple, one-line changes in `BlockchainDownloadListener.java` 
in Bitcoin-wallet. However, it removes the long-deprecated `AbstractPeerDataEventListener` and simplifies
the listeners for Blockchain downloading. 

After another release the deprecated `PeerDataEventListener` can be removed.

PR #2410, opened 2022-04-18
All modules: add detailed deprecation warnings

msgilligan:
The only module with deprecation warnings is `core` and it has two:

1. `AbstractPeerDataEventListener` which is fixed in PR #2213.
2. `getObject() in ASN1TaggedObject` which we should try to fix. I took a quick look, but I'm not knowledgable enough on ASN1 to know how to fix it.

Given that we only have 1 deprecation warning (after PR #2213 is merged) we should display it to remind us to fix it and make it more obvious if another deprecation occurs.


schildbach:
Merged.

PR #2412, opened 2022-04-18
TransactionInput/TransactionOutput: sum(), sumSats()

msgilligan:
* Add sum() and sumSats() static methods to TransactionInput and TransactionOutput
* Use them in Transaction.getInputSum() and Transaction.getOutputSum()
* Use TransactionOutput.sum() in Wallet.completeTx()
* Add TODO in Wallet.completeTx() about possibly using Transaction.getInputSum()

This is a draft PR to allow people to comment on the TODO


msgilligan:
@schildbach I didn't test this locally before submitting because I was confident 😇 my code was a "pure" refactor and should have behaved exactly the same. It would have worked, if it weren't for a bug in `Wallet.getTotalSent()` that is fixed in PR #2413.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4184,6 +4181,8 @@ public void completeTx(SendRequest req) throws InsufficientMoneyException, BadWa
                     value.toFriendlyString(), req.feePerKb.toFriendlyString());
 
             // If any inputs have already been added, we don't need to get their value from wallet
+            // TODO: Consider using Transaction.getInputSum() here, but is there a difference between
+            // `input.getConnectedOutput().getValue()` and `input.value`? and what about the warning?
```
msgilligan, line 4185: @schildbach I want your feedback on this, please.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4184,6 +4181,8 @@ public void completeTx(SendRequest req) throws InsufficientMoneyException, BadWa
...
             Coin totalInput = Coin.ZERO;
             for (TransactionInput input : req.tx.getInputs())
                 if (input.getConnectedOutput() != null)
```
msgilligan, line 4188: @schildbach What is the difference (if any) between `input.getConnectedOutput != null` and `input.value != null`.  Are the cases where these are null the same. Is it safe to replace the logic in this code with the logic in `Transaction.getInputSum()`?

core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -549,4 +555,38 @@ public String toString() {
...
+     * @return the sum (in satoshis)
+     * @throws ArithmeticException if overflows occurred in signed {@code long} arithmetic
+     */
+    public static long sumSats(Collection<TransactionInput> inputs) {
+        return inputs.stream()
+                .peek(input -> {
+                    if (input.value == null) {
+                        log.warn("input with unknown value: {}", input);
+                    }
+                })
```
msgilligan, line 586: I put this logging here after seeing the logging in Wallet.java on line 4192.

msgilligan (commented):
I posted some comments on question at 3 places in the code for @schildbach. I think these questions should be answered (and possibly a revision and force-push) before merging.

schildbach:
Just to make sure we're all on the same page: Inputs in a transaction message don't contain values, only a reference to the connected output (the output the input spends) which in turn contains the value. Since as an SPV client we don't get (and persist) each and every transaction, there is the situation we have an input, but not its connected output.

In commit 780be05260cdf30d7a90e21a79ede22475a41d6e we decided to persist an inputs value because it's so important to know it, even if we have thrown away the connected output. The value field also serves as an optimization (cache), because looking up the connected transaction in the wallet by txid is costly.

I think there are three cases to consider:

1. We never learned about the connected output. In this case both the connected output and the cached value are null.
2. We learned about the connected output at some point but since thrown it away. The connected output should be null, but the value should represent the correct amount.
3. We still have the connected output. Both fields are non-null.

Does this clear things up?

core/src/main/java/org/bitcoinj/core/TransactionInput.java
schildbach: I assume you mean the
`log.warn("SendRequest transaction already has inputs but we don't know how much they are worth - they will be added to fee.");`
line? So what kind of comment are you asking for here?

schildbach: The message explains that if you pick the output but don't know its value (and assume zero), you'll implicitly add its value to the fee (since you are sending less to its outputs). This might lead in an unexpected high fee and thus loss of Bitcoins.

However, I think since SegWit this situation is quite unlikely as the input value is part of the sigHash. So you need to cryptographically prove to know it.

For me, the question becomes: why don't we throw an exception at this point? I think this is because the SendRequest API somehow supports passing in half-baked transactions received from elsewhere (e.g. multisig protocol) and let bitcoinj add the missing bits. However I don't understand how those usecases work to prevent the "loosing Bitcoins to fee" problem.

schildbach:
It's also worth knowing that a blockchain replay can yield less information than if you follow the tip of the chain, and the mempool, closely. For example, for risk analysis of pending transactions we might opt to query for ancestors and/or descendants, and that information might not be available any more if we zip over the transaction history years later using bloom filters.

msgilligan:
> Does this clear things up?

Thanks for the detailed response. I'm going to read it and try to understand it next week.

msgilligan:
> I think there are three cases to consider:
> 
> 1. We never learned about the connected output. In this case both the connected output and the cached value are null.
> 2. We learned about the connected output at some point but since thrown it away. The connected output should be null, but the value should represent the correct amount.
> 3. We still have the connected output. Both fields are non-null.

So there's no case where we have the connected output and the cached value is null?

msgilligan:
118f2d2849a03c24428a7c592e0c3f7457b73b55 updates `CoinSelection` to use `TransactionOutput.sum()` -- so we're up to 3 uses of that method and  I think there are probably other places in the code where it can be used.

PR #2413, opened 2022-04-18
Wallet: Fix !equals() vs != bug in getTotalSent()

msgilligan:
Coin values should be compared with .equals() not == or !=


msgilligan:
@schildbach: This bug was found by the failing tests for PR #2412. Note that if PR #2414 had been applied beforehand, I don't think we would have caught the failure.  The changes in PR #2412 resulted in two different instances of zero value coins being compared.

schildbach:
Great catch!

Merged.

PR #2414, opened 2022-04-18
Coin: only allocate Coin with zero value once

msgilligan:
* In `Coin.valueOf()` reuse `Coin.ZERO` if `satoshis == zero`
* In other methods prefer `Coin.valueOf()` over `new Coin()`

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Coin.java
+++ b/core/src/main/java/org/bitcoinj/core/Coin.java
@@ -43,7 +43,7 @@ public final class Coin implements Monetary, Comparable<Coin>, Serializable {
     /**
      * Zero Bitcoins.
      */
-    public static final Coin ZERO = Coin.valueOf(0);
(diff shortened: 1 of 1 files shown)


schildbach:
Looks like a sensible optimization.

But have you actually indentified this as a bottleck or something?

msgilligan:
> have you actually indentified this as a bottleck or something?

I haven't made any performance or memory usage measurements. But I have been using **bitcoinj** for applications (e.g. rich-list analysis) where there can potentially be millions of balances in memory, the majority of which will be zero. So I think the cost of the conditional is well worth the potential memory savings.

And I recently discovered (see PR #2413) that changes I made in summation calculations where producing duplicate zero-valued coin objects, so this change was partly intended to correct for that.

schildbach:
Ok, merged. The optimisation furtunately isn't any less readable than before, I'd say.

PR #2418, opened 2022-04-19
DeterministicKeyChain: add private putKey(key) method

msgilligan:
To add a key it must be added to both the `hierarchy` and `basicKeyChain` this
should be wrapped in a method and used whenever a key is addded.


schildbach:
Merged.

PR #2426, opened 2022-04-20
DeterministicKeyChain: remove SuppressWarnings(PublicStaticCollectionField)

msgilligan:
The warning is no longer present so we don't need to suppress.


msgilligan:
This was merged: 333cf03e7ec1529fe88e2448a57224ccb6d3e423

PR #2427, opened 2022-04-20
DeterministicKeyChain: add and use static cloneKey() method

msgilligan:
simplify two for loops using new `cloneKey()` method containing common code.


schildbach:
Merged.

PR #2434, opened 2022-04-28
Network.java: remove PROD (alias for MAIN)

msgilligan:
This will affect the command-line UI of some tools, but since the
`Network` enum hasn't been released yet, we can remove it without deprecating 
first.


schildbach:
Merged.

PR #2435, opened 2022-04-29
WIP: Add P2TR support to deterministic keychains

msgilligan:


core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
```diff
@@ -146,6 +146,17 @@ public Builder fromSeed(DeterministicSeed seed, Script.ScriptType outputScriptTy
                 this.chains.clear();
                 this.chains.add(fallbackChain);
                 this.chains.add(defaultChain);
+            } else if (outputScriptType == Script.ScriptType.P2TR) {
```
schildbach, line 149: Probably at this point we'd also add a P2PKH chain. After all, there are still services that cannot receive to Bech32. So we'd have two fallback chains and the P2TR default chain.

If we decide P2PKH is not important any more, we'd remove it from the implicit creation of fallback chains in all cases.

However, before we take this step I believe we need to extend the `DeterministicSeed` and `MnemonicCode` classes to include a version number (and while we're at it also a birthdate, and thus probably more words), so that we can know what's the oldest standard the wallet supports. Currently, we can always assume P2PK.

msgilligan: > ... there are still services that cannot **receive** to Bech32 ...

I assume you mean "send to Bech32"

schildbach: Yes, sorry.

PR #2437, opened 2022-04-30
WalletTool/Wallet: add BIP43 support to WalletTool

msgilligan:
Wallet: add createDeterministic overloads with KeyChainGroupStructure parameter
WalletTool: Add --structure option and use it to configure KeyChainGroupStructure


wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -1088,6 +1096,10 @@ private void shutdown() {
     }
 
     private void createWallet(NetworkParameters params, File walletFile) throws IOException {
+        KeyChainGroupStructure keychainStructure = (structure.equals(Structure.BIP32)) ?
+                KeyChainGroupStructure.DEFAULT :
```
schildbach, line 1100: Shouldn't that be the explicit `KeyChainGroupStructure.BIP32`?

schildbach:
I think we need to make clear in the documentation to the `upgrade` action that using a `--structure` that doesn't match the given wallet might have unindended consequences.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -174,6 +175,8 @@ public class WalletTool implements Callable<Integer> {
     private String actionStr;
     @CommandLine.Option(names = "--net", description = "Which network to connect to. Valid values: ${COMPLETION-CANDIDATES}. Default: ${DEFAULT-VALUE}")
     private Network net = Network.MAIN;
+    @CommandLine.Option(names = "--structure", description = "Which HD hierarchy structure to use. Valid values: ${COMPLETION-CANDIDATES}. Default: ${DEFAULT-VALUE}")
+    private Structure structure = Structure.BIP43;
```
schildbach, line 179: Since this is the default, I think it should be in sync with `KeyChainGroupStructure.DEFAULT`.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
msgilligan: Yes, it should. That didn't exist when I made this PR.

msgilligan:
> we need to make clear in the documentation to the `upgrade` action that using a `--structure` that doesn't match ...

`--structure` is currently only used for `create`. At the very least we should document that. But it sounds like we need to think through what should happen with the combination of `--upgrade` and `--structure`.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
msgilligan: I thought I had force-pushed a revision to make this `BIP32`. I'll check and see what happened.

I think we agreed elsewhere that `wallet-tool` should for now default to `BIP32` and I'll find that commit and force-push it. But, in general, I think **bitcoinj** applications should not use `DEFAULT` and potentially have their behavior changed by an upgrade. In fact I think that `DEFAULT` should be deprecated in either this release or the next one.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
msgilligan: Fixed and force-pushed.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
msgilligan: I force-pushed that the default for this app is now `BIP32`.

Do you agree that this app should not use `KeyChainGroupStructure.DEFAULT` at all? Should we make a separate PR to deprecated `KeyChainGroupStructure.DEFAULT`?

msgilligan:
This is WIP until we decide what to do about the `upgrade` action and implement it.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
schildbach: > Do you agree that this app should not use KeyChainGroupStructure.DEFAULT at all?

Yes, I think that's reasonable.

> Should we make a separate PR to deprecated KeyChainGroupStructure.DEFAULT?

We could, but we also have constructors with defaults for the structure. Are they all deprecated as well?

> applications should not use DEFAULT and potentially have their behavior changed by an upgrade

That's the reason I'd be, to say the least, very reluctant to change the default.

msgilligan: > > Should we make a separate PR to deprecated KeyChainGroupStructure.DEFAULT?
> 
> We could, but we also have constructors with defaults for the structure. Are they all deprecated as well?

I've thought about this more and think we should definitely deprecate it and submitted PR #2443 to do so. Take a look at the PR, I think you'll agree it's the right way to go. The existing constructors, static constructors, and builders that use `DEFAULT` have been changed to use `BIP32` and I think that makes the code more readable. I don't think we need to deprecate any of the existing constructors, static constructors, and builders that use BIP32 (formerly DEFAULT) in this release, but we might choose to do so at some point in the future.

> 
> > applications should not use DEFAULT and potentially have their behavior changed by an upgrade
> 
> That's the reason I'd be, to say the least, very reluctant to change the default.

Agree, we should not change the behavior of `KeyChainGroupStructure.DEFAULT`. We should deprecate it, but we should never change behavior.

msgilligan:
I submitted the overloads of `Wallet.createDeterministic()` as a separate PR, see PR #2444. This PR should be rebased when/if that PR is merged.

schildbach:
Both #2443 and #2444 are now merged.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -1088,6 +1097,10 @@ private void shutdown() {
     }
 
     private void createWallet(NetworkParameters params, File walletFile) throws IOException {
+        KeyChainGroupStructure keychainStructure = (structure == Structure.BIP32) ?
```
schildbach, line 1100: I wonder if this logic shouldn't be moved more global, given that we need the structure for more than just creating wallets?

schildbach:
I've just quickly scanned through the actions and believe the structure is relevant for:

- `create` – creates chain(s)
- `marry` – maybe, not sure how it works exactly
- `upgrade` – potentially creates chain(s)
- `rotate` – maybe – in theory it would copy the parameters from the chain to replace; at least we need to check it works as expected

schildbach:
I wonder if it makes sense to implement a logic to detect the current structure from a wallet. That method could be used to derive defaults for `upgrade` and `rotate`, or maybe even to block those operations from executing on wallets with mismatching structure. One downside of this idea is that it probably can't take custom (user-defined) structures into account, should they ever exist.

Also, the idea might simply be overkill and we're better off letting users decide (after providing them warning about mxing structures, whereever we can).

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
msgilligan: Probably.

msgilligan: Probably.

msgilligan:
So it seems adding "full" BIP43 support to WalletTool is significantly more complicated than just adding it for `create`. I think there are (at least) 2 paths for moving forward:

Path A:
1. Add BIP-43 `create` support to `wallet-tool` which will make it easy to create BIP43 wallet files for testing.
2. Add support and tests for the other use cases (upgrade/marry/etc) to `core`
3. Update `wallet-tool` to support BIP-43 for all operations.

Path B:
1. Add support and tests for the other use cases  (upgrade/marry/etc) to `core`
2. Update `wallet-tool` to support BIP-43 for all operations.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
schildbach: I think this conversation can be resolved?

schildbach:
Marrying wallets has been removed from bitcoinj and thus the "marry" action doesn't exist any more in wallet-tool.

schildbach:
I move this to 0.18. While it would be nice to have full structure support in wallet-tool, it's not essential for a 0.17 release.

PR #2439, opened 2022-04-30
KeyChainGroupStructure: add BIP32 constant, DEFAULT = BIP32

msgilligan:
KeyChainGroupStructure:

* rename `DEFAULT` to `BIP32`
* add `DEFAULT = BIP32`

This will allow code to be more clear as to what it intends to use, and potentially allows us to deprecated `DEFAULT` in the future.


schildbach:
Merged.

PR #2440, opened 2022-04-30
examples/tools: Use explicit structure, scriptType

msgilligan:
* Explicitly use P2WPKH (this changes behavior to segwit)
* Explicitly use BIP32 (this is the same behavior as before)

This depends on PR #2439


msgilligan:
The destination addresses in these examples should probably also be updated to segwit, but that's for another PR. Volunteers?

schildbach:
> The destination addresses in these examples should probably also be updated to segwit, but that's for another PR. Volunteers?

I'm on it.

schildbach:
Merged.

PR #2443, opened 2022-05-01
Deprecate KeyChainGroupStructure.DEFAULT

msgilligan:
Deprecated DEFAULT and use BIP32 internally where DEFAULT had been used before. 

This makes the code more clear. I think it makes sense to be explicit when specifying a KeyChainGroupStructure and not have a DEFAULT symbol (which sort of implies the default might change.)

Various constructors, static constructors and builders having a default is OK, but it should say `BIP32` in the code. Perhaps in the future we might choose to deprecate some of them, but I don't think we should do that in this release.


schildbach:
Merged.

PR #2444, opened 2022-05-02
Wallet: add createDeterministic() overload with KeyChainGroupStructure parameter

msgilligan:
Add `Wallet.createDeterministic()` overload with KeyChainGroupStructure parameter and update `setUp()` in `TestWithWallet` to make sure it is used in some tests. (And to remind us that we need to work on parameterizing tests with structure and scriptType parameters)


schildbach:
Merged.

PR #2441, opened 2022-04-30
WIP: integration-tests - move TestWithNetworkConnections and subclasses from core tests

msgilligan:
This is Work-in-Progress. For the first attempt at moving these classes to `integration-test` I decided to COPY rather than move them, so I could have the old (working) tests and the new tests running side-by-side.

There are 4 concrete test classes that inherit from `TestWithNetworkConnections`. The current status of those tests is:

* `PeerGroupTest` -- all tests pass (use commit 76c17cae83ce010a87a32d8f653aab10e73189d8 to verify)
* `FilteredBlockAndPartialMerkleTreeTest` -- all tests pass (use commit 76c17cae83ce010a87a32d8f653aab10e73189d8 to verify)
* `TransactionBroadcastTest` -- 2 failures (use df5b9d506c1e0ed916a4e1986c5c1148a982ad8f to demonstrate)
* `PeerTest` -- many failures (use dab2ee383b321f737ad95367a91281eba4a2ba52 to demonstrate)

I'm not very familiar with the internals of these tests (and I haven't yet spent much time learning/debugging), so I thought I'd post my WIP in case anyone else wants to take a look. @schildbach any ideas?

PR #2446, opened 2022-05-02
Move FakeTxBuilder from `src/test` to `src/main`

msgilligan:
This will allow it to be used in both core tests and integration-test or in tests for any other module.

I'm also imagining some of this functionality might get merged into the "immutable" transactions capability we plan on adding.

### DIFF
--- a/core/src/main/java/org/bitcoinj/testing/FakeTxBuilder.java
+++ b/core/src/main/java/org/bitcoinj/testing/FakeTxBuilder.java
@@ -17,6 +17,7 @@
 
 package org.bitcoinj.testing;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.bitcoinj.core.Address;
 import org.bitcoinj.core.Block;
 import org.bitcoinj.core.Coin;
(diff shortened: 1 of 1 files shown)


schildbach:
I see the point of this. However, the downside of testing classes in main is they get distributed and may inadvertently get part of our API. Do we actually have any test that needs this?

> I'm also imagining some of this functionality might get merged into the "immutable" transactions capability we plan on adding.

I agree.

msgilligan:
> the downside of testing classes in main is they get distributed and may inadvertently get part of our API.

There are three simple things we can do to protect against this:

1. Put in separate package (`org.bitcoinj.testing`) - DONE
2. Tag with `@VisibleForTesting`
3. Add comments warning against its use

The name `FakeTxBuilder` is also a fairly good warning against using it.

Alternatively, I suppose we could create another module called `test-support` (for example) and have both `core` and `integration-test` depend on it. (Personally, I think the new module would be overkill. Especially since we are planning on creating the immutable transaction building classes which hopefully largely or completely replace `FakeTxBuilder`)

> Do we actually have any test that needs this? The tests in #2441 maybe?

I believe they do, but I will double-check.

msgilligan:
> 2. Tag with `@VisibleForTesting`
> 3. Add comments warning against its use

Updated and force-pushed.

schildbach:
Merged.

PR #2447, opened 2022-05-02
WalletAccountPathTest: use Wallet.fromSeed() to simplify private createWallet()

msgilligan:
I discover that the private method `createWallet()` duplicated code from `Wallet.fromSeed()`. This PR fixes that.


schildbach:
Merged.

PR #2448, opened 2022-05-02
WalletTool: explicitly specify BIP32 in createWallet()

msgilligan:
This is dependent on PR #2444


schildbach:
I'm not sure if this is actually a step forward, given that we want to make the structure user-selectable in #2437.

msgilligan:
Rebased.

> I'm not sure if this is actually a step forward, given that we want to make the structure user-selectable in #2437.

I think it is a step forward. We can view PR #2437 as two steps:

1. Create a "constant" for structure and use it to explicitly specify structure in calls to the wallet.
2. Change the "constant" to a variable that is set with command-line arguments.

If you compare this rebase to PR #2437 you'll hopefully agree with me.

schildbach:
Thanks for the explanation!

Merged.

PR #2449, opened 2022-05-02
WalletAppKit: replace null with BIP32 in constructor

msgilligan:
This makes the code for this constructor more explicit/readable and prepares the way for making the parameter in the "full" constructor non-nullable (in a minor breaking change) in the future.


schildbach:
Merged.

PR #2450, opened 2022-05-03
Deterministic wallets.txt: update with developments of the last years

schildbach:
* Support for different wallet structures
* Support for multiple output script types
* Wallet upgrade from basic to deterministic was removed


designdocs/Deterministic wallets.txt
```diff
@@ -1,9 +1,15 @@
...
 Goals:
-- Wallets to derive new keys deterministically using the BIP32 algorithm.
-- Seamless/silent upgrade of old wallets to use the new scheme
+- Wallets to derive new keys deterministically using the BIP32 algorithm
+- Support different wallet structures:
+  - The default BIP32 structure, described in the later part of the document
+  - BIP43/44/84/86
+  - Custom, by implementing the KeyChainGroupStructure interface
+- Support multiple output script types (P2PKH, P2WPKH)
+  - One active chain per script type, so that compatible addresses can be generated
```
msgilligan, line 11: Should say one active "keychain" (I interpreted that as blockchain on first read and was momentarily confused)

designdocs/Deterministic wallets.txt
```diff
@@ -81,10 +87,8 @@ it's less convenient than an integrated solution would be.
 Chain structure
 ---------------
 
-We follow the suggested chain structure outlined in BIP32 in which there is the notion of a top-level account, though
-in bitcoinj this will always be zero as it won't be exposed in the API for now. Under the account are two keys, one
-for receiving of funds (i.e. exposed to the use) and one which is used for "internal purposes", typically change keys
-though it may also be used for things like micropayment channels and, in future, de/refragmentation.
```
msgilligan, line 87: Shouldn't this behavior still be documented somewhere?

designdocs/Deterministic wallets.txt
```diff
@@ -198,27 +202,6 @@ that encode the index of each child as the tree is traversed downwards, with the
...
-128 bits is more than enough, and people like to write down their seeds on paper. 128 bit seeds using the BIP 39
-mnemonic code specification yields 12 words, which is a convenient size.
-
-As part of migrating to deterministic wallets, if the wallet is encrypted the wallet author is expected to test
-after load whether the wallet needs an upgrade, and call the upgrade method explicitly with the password.
-Note that attempting to create a spend will fail if the wallet is not upgraded, because it will attempt to retrieve a
-change key which is done deterministically in the new version of the code: thus an non-upgraded wallet is not very
-useful for more than viewing (unless the API caller explicitly overrides the change address behaviour using the
-relevant field in SendRequest of course).
-
```
msgilligan, line 221: Why are you removing this? Are we going to just write entirely new text to replace it?

msgilligan (changes requested):

msgilligan:
Also, after updating the text should we convert this to `.md` or `.adoc`?

designdocs/Deterministic wallets.txt
schildbach: Good point. I added a commit that uses the term 'keychain' everywhere in the document.

designdocs/Deterministic wallets.txt
schildbach: Somehow I thought it's better if people read up the mentioned BIPs (32/43/44/84/86). I think the idea about internal and external addresses is covered there very well.

Also I feel talking too much about features that _could_ be implemented, but after all these years haven't been, distracts the reader from what bitcoinj can actually do.

designdocs/Deterministic wallets.txt
schildbach: Are you referring to the entire "Upgrade" section?

It mainly describes the upgrade from random keys to HD, and this upgrade doesn't exist any more. But I admit I was lazy, I should have documented the P2PKH to P2WPKH upgrade instead, which is very similar. Is that what you were thinking about?

msgilligan: Yes.

PR #2452, opened 2022-05-24
core/build.gradle: reenable TransactionBroadcastTest and try to fix (experimental WIP PR)

msgilligan:
(this should run the test class with 2 failing tests)


msgilligan:
@schildbach rebased to apply to the `integration-test` subproject. (the rebase also squashed the re-enable and the work-around into a single commit)

I need to look at why the tests are failing on Windows. Do you have any idea why removing the `ver` parameter fixes the problem?

PR #2458, opened 2022-06-07
Remove all usage of Java Serialization

msgilligan:
* Remove all implementations of `Serializable`
* Remove unit tests of Java serialization

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Coin.java
+++ b/core/src/main/java/org/bitcoinj/core/Coin.java
@@ -19,7 +19,6 @@ package org.bitcoinj.core;
 import com.google.common.math.LongMath;
 import org.bitcoinj.utils.MonetaryFormat;
 
-import java.io.Serializable;
 import java.math.BigDecimal;
 
 import static com.google.common.base.Preconditions.checkArgument;
@@ -27,7 +26,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 /**
  * Represents a monetary Bitcoin value. This class is immutable.
  */
-public final class Coin implements Monetary, Comparable<Coin>, Serializable {
+public final class Coin implements Monetary, Comparable<Coin> {
 
     /**
      * Number of decimals for one Bitcoin. This constant is useful for quick adapting to other coins because a lot of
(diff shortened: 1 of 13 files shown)


schildbach:
Years ago, when we got rid of most Java serialization, I remember we agreed to keep the Monetary/Coin/Fiat triple serializable. But we could decide otherwise, I'm not opposed to the idea.

I just grepped through Bitcoind Wallet, and Coin seems to be the only class expected to implement Serializable. I guess it would be trivial to short-time persist a long instead.

schildbach:
I just tried to get rid of Serialization in Bitcoin Wallet, but I'd need to basically re-implement Serialization there. I'm short-time persisting a `Monetary` – meaning I'd have to persist class, amount and – in case of a `Fiat` – currency code.

msgilligan:
> I just tried to get rid of Serialization in Bitcoin Wallet, but I'd need to basically re-implement Serialization there. I'm short-time persisting a `Monetary` – meaning I'd have to persist class, amount and – in case of a `Fiat` – currency code.

We could add `serialize` and `deserialize` methods to `Monetary` (or alternatively, perhaps just to the subclasses depending upon your use case.) We would serialize to an amount and a currency code (`BTC` for Bitcoin) and provide the reverse.

Or we could change this PR to remove serialization from everything but `Monetary` and its subclasses. I'd really like to see it removed from `Coin`, though. Maybe we could leave it just in `Fiat`.

What do you think?

msgilligan:
@schildbach rebased now that PR #2459 is merged.

msgilligan:
To summarize the options I presented above:

1) Add `serialize` and `deserialize` methods to `Monetary`. We would serialize to an amount and a currency code (`BTC` for Bitcoin) and provide the reverse. (would not be too different from`Fiat::parseFiat`.) 

1a) Add `serialize` and `deserialize` methods to `Fiat`.

2) Remove serialization from everything except `Monetary` and its subclasses 

2a) Remove serialization from everything except `Fiat`

3) Remove serialization from `PrefixedChecksummedBytes` only.

@schildbach What would you like me to do?

schildbach:
I managed to keep the "re-implementation" in Bitcoin Wallet quite compact, so I think it's fine for now to remove Serialization entirely. Maybe some point in future there will be a better generic serialization mechanism.

Merged.

PR #2459, opened 2022-06-08
PrefixedChecksummedBytes: remove Cloneable

msgilligan:
* Remove the `implements Cloneable` marker
* Remove the `@Override` of `clone()`
* Remove tests of `clone()` for the implementing classes

Not logically dependent upon PR #2458, but does modify some of the same lines. So I'll make this a DRAFT PR and we can rebase (or modify) after #2458 is merged (or rejected).

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Coin.java
+++ b/core/src/main/java/org/bitcoinj/core/Coin.java
@@ -19,7 +19,6 @@ package org.bitcoinj.core;
 import com.google.common.math.LongMath;
 import org.bitcoinj.utils.MonetaryFormat;
 
-import java.io.Serializable;
 import java.math.BigDecimal;
 
 import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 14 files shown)


schildbach:
Is `Cloneable` bad?

msgilligan:
> Is `Cloneable` bad?

Yes, although not as bad as `Serializable`. According to _Effective Java, 3rd Edition_:

> In order [to provide a properly functioning public `clone` method], the class and all its superclasses must obey a complex, unenforceable, thinly documented protocol. The resulting mechanism is fragile, dangerous, and extralinguistic: it creates objects without calling a constructor. (p. 58)

(Basically, the design of `Cloneable` was botched by Java's designers)

It goes on to say that:

> immutable classes should never provide a `clone` method because it would encourage wasteful copying.

and for non-immutable classes:

> A better approach is to provide a _copy constructor_ or a _copy factory_.

I believe all subclasses of `PrefixedChecksummedBytes` are immutable, but will double-check.

schildbach:
Thanks!

I went ahead and merged this already.

PR #2462, opened 2022-06-12
DeterministicKey: deprecate serializeP*B58 methods with outputScriptType parameter

msgilligan:
The reasons are explained in the added JavaDoc

* Add deprecation and JavaDoc comments
* Remove usage of deprecated methods

Note: DeterministicKeyChain.toString() will no longer output the non-standard "ypub", etc formats.

Draft because dependent upon PR #2461

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -479,6 +479,7 @@ public class DeterministicKey extends ECKey {
         return serialize(params, pub, Script.ScriptType.P2PKH);
     }
 
+    // TODO: remove outputScriptType parameter and merge with the two-param serialize() method. When deprecated serializePubB58/serializePrivB58 methods are removed.
     private byte[] serialize(NetworkParameters params, boolean pub, Script.ScriptType outputScriptType) {
         ByteBuffer ser = ByteBuffer.allocate(78);
         if (outputScriptType == Script.ScriptType.P2PKH)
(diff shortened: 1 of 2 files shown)


msgilligan:
I just force-pushed a fix for a typo in a comment (it was "tprv" where it should've been "tpub")

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -496,20 +497,52 @@ else if (outputScriptType == Script.ScriptType.P2WPKH)
         return ser.array();
     }
 
+    /**
```
schildbach, line 500: I think `@deprecated` and `@Deprecated` should be enough to mark a symbol as deprecated. No need for another marker in the text.

schildbach:
I'm generally fine with replacing ypub/zpub etc by output descriptors, but on the other hand I think output descriptors need to be well established both in bitcoinj and within the ecosystem first.

`toString()` is used in wallet dumps (in "key to watch") and for now, I need it to state the "correct" xpub variant.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
msgilligan: Yeah, I'm not sure why I did that. Will fix.

msgilligan:
> I'm generally fine with replacing ypub/zpub etc by output descriptors, but on the other hand I think output descriptors need to be well established both in bitcoinj and within the ecosystem first.

OK. Let's leave those methods deprecated to indicate that we're moving towards output descriptors and to encourage migration.

I've made further progress on output descriptors and expect to have the ability to parse and create keychains from (a supported subset of) output descriptors soon.

> `toString()` is used in wallet dumps (in "key to watch") and for now, I need it to state the "correct" xpub variant.

I force-pushed an update that continues to use the deprecated method in the `toString()` method. In a subsequent PR we can add one or both of the following to `toString()`:

1. The "standard" xpub/tpub in addition to the "extended standard" "x/y/zpub" format.
2. The output descriptor

msgilligan:
Here is Peter Wuille's explanation of why Bitcoin is moving to output descriptors:

https://bitcoin.stackexchange.com/questions/89261/why-does-importmulti-not-support-zpub-and-ypub/89281#89281

PR #2463, opened 2022-06-13
Move Monetary/Coin/etc to new org.bitcoinj.base package

msgilligan:
This PR explores the possibility of creating a new `org.bitcoinj.base` package that would provide fundamental bitcoin(j) types with minimal dependencies. This PR moves `Coin`, its parent `Monetary` and related types into the new `base` package.

The package comment proposes some (initial) rules for what acceptable dependencies would be:

```
/**
 * The {@code base} package provides fundamental types for <b>bitcoinj</b>. These types must have
 * minimal dependencies. The criteria for allowed dependencies for {@code base} types are:
 * <ul>
 *     <li>No dependencies on other packages of bitcoinj</li>
 *     <li>No API dependencies on external libraries other than the core JDK</li>
 *     <li>Implementation dependencies on Guava are allowed for now (but should be avoided as much as possible)</li>
 * </ul>
 * The base package will help us make bitcoinj more modular as we will use it to break circular dependencies
 * between existing packages. We are also considering splitting {@code base} into a separate JAR/module in a
 * future release.
 */
```


schildbach:
I agree with the rules.

If I understand correctly, this PR doesn't ensure the rules are followed. I know I might have made a different impression earlier, but now I wonder if we should start a new sub-project right away and declare the dependendies (ideally an empty list)?

msgilligan:
> I wonder if we should start a new sub-project right away and declare the dependencies?

At first I was happy to hear you say this, because I would like to get to a separate sub-project as soon as possible. But I also think it might make things more difficult when we are refactoring to create the wallet interface and separate the wallet. There may be some intermediate and temporary steps in the refactoring that might violate the rules.

It's fairly easy to check the `base` package for imports that break the rules, so my recommendation is that we wait until after  the "Wallet interface" refactoring to do this. 

>  (ideally an empty list)

I thought we were going to depend on Guava. (But I have thought that we could maybe just copy the "check" and "safe math" classes into a utility package. It would be really nice to have zero dependencies.)

msgilligan:
@schildbach rebased and force-pushed. (I found one file that was missing a final newline and fixed it.)

schildbach:
At first, I was going to propose to move only the `Monetary` hierarchy in this PR, but then I realized this isn't possible because of the circular dependency `Coin`→`MonetaryFormat`→`Coin` (and the same for `Fiat`). I wonder is this something we want to break up in future?

Should we move `CoinTest`, `FiatTest` and `MonetaryFormatTest` to the new package too? I'd say yes.

schildbach:
Merged.

msgilligan:
Until we move `base` to a subproject the following command-line query to look for files that break the rules:
```
grep import `find core/src/main/java/org/bitcoinj/base  -name '*.java'` | egrep -v 'import (static )?(com.google.common|java.|org.bitcoinj.base)'
```

PR #2464, opened 2022-06-14
VarInt: make `value` member private

msgilligan:
(it was marked as @Deprecated so nobody should be using it)


schildbach:
Merged.

Issue #2465, opened 2022-06-14
Deprecate and remove C++ JNI wallet interface


msgilligan:
@mikehearn created a C++ interface for *bitcoinj* wallets about 9 years ago (and I remain impressed by all the cross-language work that Mike did.)

There is some documentation here: https://bitcoinj.org/using-from-other-languages#c

The `org.bitcoinj.jni` package contains some classes to support this implementation.

As far as I know this has been unused and unmaintained for about 9 years. We should deprecate and remove all classes in this package (unless someone has a strong need for this _and_ steps forward to maintain it.)

Personally, I would like to see a modern native interface to **bitcoinj** developed after we finish a few rounds of refactoring, but in the meantime we should deprecate and remove these classes to avoid confusion and to speed refactoring.

msgilligan:
https://github.com/bitcoinj/bitcoinj/pull/2466 implements the deprecation.

msgilligan:
See related PR #2267

schildbach:
The interface has been deprecated and we'll remove it in the 0.18 cycle. I'm setting the milestone as a reminder.

msgilligan:
> The interface has been deprecated and we'll remove it in the 0.18 cycle. I'm setting the milestone as a reminder.

I changed the title to reflect that it has already been deprecated.

Note that we have begun work on a separate project to provide access to native secp256k1 functionality:https://github.com/bitcoinj/secp256k1-jdk

msgilligan:
Closed by #3847

PR #2466, opened 2022-06-14
org.bitcoinj.jni package: Deprecate all classes

msgilligan:
Deprecate all classes in the `org.bitcoinj.jni` package for removal. See Issue #2465


schildbach:
Merged.

PR #2467, opened 2022-06-18
Wallet.calculateAllSpendCandidates: make more functional

msgilligan:
Use a filtered stream to generate an unmodifiable list.


schildbach:
Merged.

PR #2468, opened 2022-06-18
Wallet: private listener lists as List type (was CopyOnWriteArrayList)

msgilligan:
This simplifies the declaration and verifies that we're using the standard List
interface in all our code. The implementation remains unchanged.


schildbach:
Merged.

PR #2470, opened 2022-06-19
WalletTool add listunspent command

msgilligan:
This is a proof-of-concept. It needs refactoring, review, and polish.


schildbach:
Generally I like the idea of a separate command for this.

Are you aware of the unspent amounts output that appears below wallet dumps?

msgilligan:
> Are you aware of the unspent amounts output that appears below wallet dumps?

Yes. I added this because I was seeing a discrepancy in the balances between a **bitcoinj** wallet and a **Bitcoin Core** wallet created by importing the descriptors.  (It turns out that I had sent two UTXOs to /44'/1'/0'/0 -- the parent for receiving addresses, rather than one if the indexed children)  I had a large number of addresses and balances that I needed to compare, so the format from `dump` wasn't very helpful.  

I took the output generated by the code in this PR and the output from `bitcoin-cli listunspent` and imported them into a spreadsheet, sorted by address and compared to locate the discrepancy.

This wasn't the first time during development that I've wanted a format the just shows address/balance (or address/path/balance) in a summary format.

msgilligan:
This should be rebased and completed.

schildbach:
In my own version of this (which is not yet published or merged), I aggregate the UTXO values by address, and sort the resulting table by values (highest first).

Do you think this would be a good default behaviour, or should it be locked behind a parameter?

Anyway, a rebase is needed.

msgilligan:
> Just for reference, my code is:
> 
> ```
>         Map<Address, Set<TransactionOutput>> a = new HashMap<>();
>         for (TransactionOutput output : wallet.calculateAllSpendCandidates()) {
>             Address addr = output.getScriptPubKey().getToAddress(params);
>             Set<TransactionOutput> outputs = a.get(addr);
>             if (outputs == null) {
>                 outputs = new HashSet<>();
>                 a.put(addr, outputs);
>             }
>             outputs.add(output);
>         }
> ```

Note that we could use `Collectors.groupingBy` to build a very similar (`List` instead of `Set`) `Map`, something like this:
```
Map<Address, List<TransactionOutput>> a =  wallet.getUnspents().stream()
        .collect(Collectors.groupingBy(o -> o.getAddress(net).get()));
```

msgilligan:
This is rebased and updated a bit.

msgilligan:
> I aggregate the UTXO values by address, and sort the resulting table by values (highest first).
> 
> Do you think this would be a good default behaviour, or should it be locked behind a parameter?

I think it is reasonable default behavior and maybe we can avoid any formatting options for the initial version of this command. That said, it might be nice to sort by HD path to see where the UTXOs are in the hierarchy.

Output format like this might be nice:

```
M/44H/1H/0H/0/0    mkrGoJYwpiGHtH8pnJS5nmkQF3EgG3CKoi   0.17422047 BTC
```
With a total amount at the bottom.

PR #2472, opened 2022-06-20
NetworkParameters: Use Network enum

msgilligan:
This change makes NetworkParameters use and depend upon Network instead
of the reverse. This will allow Network to be moved to `base` and simplifies
some comparisons (enums can use ==/!=)

Not addressed in this PR is support for non-Bitcoin networks.


schildbach:
Is there a reason you've migrated some of the String id equals checks to Network id identity checks and some not?

Are we going to deprecate String ids at some point (not necessarly in this PR). I think at this point there is no reason to support altcoins in the bitcoinj main branch, so a Network id would do for us?

msgilligan:
> Is there a reason you've migrated some of the String id equals checks to Network id identity checks and some not?

Yes. I changed most of the ones that were in lines that were updated by the refactoring. The ones I didn't change were (all/mostly?) ones that referenced the UNITTEST String/NetworkParameters. That case must be treated carefully.

> 
> Are we going to deprecate String ids at some point (not necessarly in this PR).

I don't know if deprecation is needed. I do think they will be use less and less over time. The String ids are still useful for serialization and they handle the unit test case as well.

>I think at this point there is no reason to support altcoins in the bitcoinj main branch, so a Network id would do for us?

I actually think we should try to create classes that will behave will if used in Bitcoin-based alt coin project. I do think we might see some contribution upstream if we provide this. I also think it just makes our code better to have the flexibility of adding other "networks" (e.g "unit test" or other test networks.

I think https://github.com/bitcoinj/bitcoinj/pull/2473 is a reasonable way to go and it is somewhat analogous to the way NetworkParameters supports alternative coins/networks.

msgilligan:
@schildbach Rebased, conflicts-resolved, force-pushed and ready for review/merge.

schildbach:
> > Is there a reason you've migrated some of the String id equals checks to Network id identity checks and some not?
> 
> Yes. I changed most of the ones that were in lines that were updated by the refactoring. The ones I didn't change were (all/mostly?) ones that referenced the UNITTEST String/NetworkParameters. That case must be treated carefully.

I'm not happy about the different treatment of networks. I think we need to decide about the way a unittest environment is established. Since unit testing is not a network, I think ideally we get rid of `UnitTestParams` entirely. But then we'd need to come with an alternative, e.g. a way to disable output shuffling for unit tests. I once implemented a way to set the default fee to 0 via context, we could maybe copy that approach.

Until then, since you stated this PR to be one of your top priorities, I propose to strip down this PR to the bare minimum you need for progressing with modularization.

msgilligan:
> > > Is there a reason you've migrated some of the String id equals checks to Network id identity checks and some not?
> > 
> > 
> > Yes. I changed most of the ones that were in lines that were updated by the refactoring. The ones I didn't change were (all/mostly?) ones that referenced the UNITTEST String/NetworkParameters. That case must be treated carefully.
> 
> I'm not happy about the different treatment of networks.

I think your uneasiness is well-founded.

> I think we need to decide about the way a unittest environment is established. Since unit testing is not a network, I think ideally we get rid of `UnitTestParams` entirely.

I agree. I'm working on a plan for how to do this. I'm thinking we should use the Java "Service Provider" (aka "Service Loader") mechanism to provide a `NetworkConfiguration` object. This would incorporate `Networks`, part of `NetworkParameters` and perhaps even parts of the `Context` object. The `NetworkConfiguration` would include/list all available/supported networks. This would allow for special test configurations, the addition of new networks (e.g. SIGNET when it was new), things like side chains, or even alt-chains. But more on this later (and elsewhere)...

> But then we'd need to come with an alternative, e.g. a way to disable output shuffling for unit tests. I once implemented a way to set the default fee to 0 via context, we could maybe copy that approach.

I just pushed a 2nd commit to this PR that uses a simple approach for doing this that works with the current `Network`/`NetworkParameters` setup: `params instanceof UnitTestParams`.  See 45c4e4d442dc53f6d6138401ea6c046b34fc0850

> Until then, since you stated this PR to be one of your top priorities, I propose to strip down this PR to the bare minimum you need for progressing with modularization.

Hopefully, 45c4e4d442dc53f6d6138401ea6c046b34fc0850 will resolve your concerns instead of "stripping down". What do you think?

msgilligan:
> Also, does this change work towards the goal of this PR?

I think the goal of this PR is to begin a migration from `NetworkParameters` to `Network` and to provide a mapping between them, with `Network` being the "preferred" mechanism wherever possible (and the one capable if being moved to `o.b.base`)

I suppose we could split this into 2 PRs if you insist (i.e. the changes to comparisons etc would be in the 2nd PR)

msgilligan:
@schildbach I removed the changes to NetworkParameters comparisons and force-pushed. Hopefully this is good-to-go now!

schildbach:
Merged.

> I'm wondering if it would be better if we change this to return an Optional. This would let callers handle "not found" errors without using exceptions.

What "this" are you referring to?

msgilligan:
> Merged.
> 
> > I'm wondering if it would be better if we change this to return an Optional. This would let callers handle "not found" errors without using exceptions.
> 
> What "this" are you referring to?

The `public static Network of(String idString)` method. I thought about this further and arrived at what I think is the best way to handle this: have two `static` methods that can be used to create a `Network`: one named `of` that throws a `RuntimeException` for an invalid `idString` and another named `find` that returns an `Optional<Network>`.

This is what I did in PR #2486 for `ScriptType`. If that approach seems reasonable to you, we can add a `find` method here.

I think there are two use-cases, one where you have a known good idString and you need to create a `Network` and the other where you have an id string that was input by a user or read from a file and you need to validate it first. I also think it is misleading to have an `of` method return an `Optional`.

schildbach:
> I think there are two use-cases, one where you have a known good idString and you need to create a `Network` and the other where you have an id string that was input by a user or read from a file and you need to validate it first. I also think it is misleading to have an `of` method return an `Optional`.

Sounds good to me. I'm not really used to Optional, but it seems preferable to returning `null` on the find operation.

PR #2476, opened 2022-06-21
move Script.ScriptType to a top level type in `o.b.base`

msgilligan:
`Script.ScriptType` is a fundamental type and used in many classes in **bitcoinj** that do not need to depend on `Script` or the `o.b.script` package. It should be a standalone type in `o.b.base`.


schildbach:
Rebased and merged.

PR #2477, opened 2022-06-21
(WIP) Address/PrefixedChecksummedBytes: Use Network not NetworkParameters

msgilligan:
This is one path towards moving Address to `o.b.base`. (Another path is to reduce dependencies in `NetworkParameters` and move it to `o.b.base`)

This gets really messy because there are unit tests that assume a `UNITTEST` `NetworkParameters` is stored inside Address (and other types) (We really shouldn't have our `Address` class have a dependency on unit test mechanisms)

I believe this needs to be fixed at some point, but it is not an easy change. This PR is just a record of what I tried and a reminder to try again later. It should also be easier after some other pending changes are merged.

**Note: This is raw work-in-progress and doesn't even pass tests**

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Address.java
+++ b/core/src/main/java/org/bitcoinj/core/Address.java
@@ -18,6 +18,7 @@ package org.bitcoinj.core;
 
 import org.bitcoinj.script.Script;
 import org.bitcoinj.script.Script.ScriptType;
+import org.bitcoinj.utils.Network;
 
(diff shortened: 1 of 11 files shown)


schildbach:
I think we've already merged a very similar PR. Is this still relevant?

msgilligan:
> I think we've already merged a very similar PR. Is this still relevant?

Yes, most of the ideas in this PR have been implemented by other PRs and there is nothing left in here of interest.

Closing...

PR #2478, opened 2022-06-22
A quick sketch of a possible near-future modular architecture

msgilligan:
This is a rough sketch/proposal of some of the architectural directions we have been discussing.

### DIFF
--- a/designdocs/modular-architecture.md
+++ b/designdocs/modular-architecture.md
@@ -1,0 +1,22 @@
+# A rough overview of a modular bitcoinj architecture
+
+Note: this diagram somewhat blurs the distinctions between modules/JARs and packages. Some of the "modules" would remain packages in the short-run.
+Other packages that are not significant or are not changing are not shown (and are generally expected to remain in the `core` module.)           The difference between API and implementation dependencies is (for now) omitted.
+
+````mermaid
+graph TD
+    B[base] --> G[guava]
+    B --> S[slf4j]
+    B --> A[jcip-annotations]
+    CR[crypto] --> BC[Bouncy Castle]
+    CR --> B
+    CO[core] --> CR
+    W --> P[ProtoBuf]
+    W[wallet] --> CO
+    I[integration-test] --> W
+    WT[wallet-tool] --> W
+    FX[wallettemplate] --> W
+    E[examples] --> W
+classDef external fill:#999;
+class G,S,A,BC,P external;
+````


schildbach:
How is `base` dependent on jcip? Do we want to allow this?

The dependence of `base` on slf4j is maybe something we want to add to the exceptions.

I take it you intentionally left out the dependencies that are covered by transitive deps already? For example, almost any module in the diagram is dependent on slf4j, no?

msgilligan:
> How is `base` dependent on jcip? Do we want to allow this?

I'm not sure it is, but we're using `@Nullable`, etc. throughout our codebase, so that's where this dependency comes from. I'd certainly be happy if we could not have that dependency.
 
> 
> The dependence of `base` on slf4j is maybe something we want to add to the exceptions.

Agreed.


> I take it you intentionally left out the dependencies that are covered by transitive deps already? For example, almost any module in the diagram is dependent on slf4j, no?

Yes.

schildbach:
For what I know, `@Nullable` uses `javax.annotation.Nullable`.

schildbach:
We do have some `import net.jcip.annotations.GuardedBy;` however, luckily not in any immediate candidates for `base`.

schildbach:
I wonder if those `@GuardedBy` annotations are checked somehow? If not, we could consider to migrate them to a comment and do away with the test dep.

kparmar1:
HI,

Just a thought — might not make any sense as just looked at this diagram (so I don’t have context — case is, or case wants to be (I think it’s that)?). Just that line between wallet and ProtoBuf (that’s a transitive dep), but wallet shouldn’t care about that. Should that not be all be via core?

Sorry if I jumped in and totally talking nonsense (or just plain wrong). Just that line bothered me :-)

Well unless wallet is intended to be stand alone (but I can’t see how that’s possible without core)

Kal.

On 7 Jul 2022, at 23:34, Sean Gilligan ***@***.******@***.***>> wrote:



I used the mmdc (aka mermaid-cli) command-line tool to convert the diagram to an SVG file and dragged it in to a comment here:

[bitcoinj-modules]<https://user-images.githubusercontent.com/61612/177882427-b7f4e9a4-daaf-4e51-a904-dc66f8eda650.svg>

msgilligan:
I previously said something incorrect:
 
> the `core` package ... uses ProtoBuf to de/serialize messages to and from the P2P network

ProtoBuf **is not used** to de/serialize messages to and from the P2P network.

This which means that :

> The `wallet` package and the `Wallet` class have a transitive dependency on ProtoBuf through the `core` package.

is also incorrect: `wallet`/`Wallet` still have a direct dependency on ProtoBuf, but they do not have a transitive dependency through core.

This means the dependency arrow from `core` to `ProtoBuf` can be removed.

msgilligan:
I pushed some updates to the diagram, it now looks like this:
test mmd

schildbach:
What do you think, is this ready to merge?

msgilligan:
> What do you think, is this ready to merge?

I do. I squashed and force-pushed, so it should be ready-to-merge.

schildbach:
Merged.

PR #2479, opened 2022-06-22
ScriptType: Add id() accessor, deprecate public id field

msgilligan:


schildbach:
Hmm I wonder if we want to continue this to be public API. It is currently only used for DB persistence, so theortically the numbers could be moved there.

msgilligan:
> I wonder if we want to continue this to be public API.

Hmm. If you're open to changing the public API, I think we might want to change from number id to a short string, ideally the same strings as are used in Bitcoin Output Descriptors (e.g. "sh", "pkh", etc.)  This way the string ID could be used for serialization and paired with a `ScriptType.of(String)` method for deserializing. This would certainly be useful in the OutputDescriptor parser...

msgilligan:
I recommend closing this PR and merging PR #2486 instead.

schildbach:
Closing, in favour of #2486.

PR #2482, opened 2022-06-22
PrefixedChecksummedBytes: remove unused transient keyword

msgilligan:
Now that serialization support has been removed from PrefixedChecksummedBytes, we no longer need the `transient` keyword.


schildbach:
Merged.

PR #2483, opened 2022-06-23
NetworkParameters: deprecate/move fromID() to AbstractBitcoinNetParams

msgilligan:
`fromID(String)` and `fromPmtProtocolID(String)` only know about Bitcoin networks so are deprecated and
moved to AbstractBitcoinNetParams.

When they are removed from `NetworkParameters` it will no longer be dependent on its subclasses.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+++ b/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
@@ -19,6 +19,7 @@ package org.bitcoinj.core;
 
 import org.bitcoinj.base.Coin;
 import org.bitcoinj.base.Sha256Hash;
+import org.bitcoinj.params.AbstractBitcoinNetParams;
 import org.bitcoinj.params.MainNetParams;
 import org.bitcoinj.params.RegTestParams;
 import org.bitcoinj.params.SigNetParams;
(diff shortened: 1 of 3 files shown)


msgilligan:
> Could `fromPmtProtocolID(String)` be moved into the payment protocol classes, and then at some point could the payment protocol support be moved into a separate sub-project?

Yes! Decoupling payment protocol from NetworkParameters is a good idea. I moved all payment protocol related code from the NetworkParams hierarchy into `o.b.protocols.payments` and left `@Deprecated` members that reference it that can be removed in the release after next. This should move us closer towards payment protocol support being in its own sub-project. (Note that the deprecation also found a minor payment protocol dependency in `wallettemplate` that shouldn't have been there.)

schildbach:
Picked the WalletApplication change as 3732a6179592e8a60cf871c5e375b93cb51f3c8a.

schildbach:
Picked payment protocol changes as ce103ed2c354b4cf2b7c86664f39e95b0c20f290.

I think the rest needs a hearty rebase.

msgilligan:
@schildbach Hearty rebase complete (and  also rebased after PR #2496)

schildbach:
Merged.

PR #2486, opened 2022-06-24
ScriptType: add ID string, deprecate numeric ID.

msgilligan:
`ScriptType` is an ideal enum to represent the [BIP 380](https://github.com/bitcoin/bips/blob/master/bip-0380.mediawiki)-standard names for the script types.

This PR deprecates the numeric `id` field and also creates a deprecated `numericId()` accessor. I propose that we make the `id` field private in the next release, but leave the `numericId()` accessor until the database SQL and code is updated to use the string ID (or provide it's own mapping from `String` to `int`)

This PR IMO should replace PR #2479 .

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/ScriptType.java
+++ b/core/src/main/java/org/bitcoinj/base/ScriptType.java
@@ -16,20 +16,85 @@
 
 package org.bitcoinj.base;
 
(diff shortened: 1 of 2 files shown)


schildbach:
Good idea with the BIP 380 ids. Merged.

PR #2487, opened 2022-06-24
Github graalvm.yml: Run `test` task

msgilligan:
This makes sure that unit and integration tests are run before doing the native compile.

The GraalVM tasks should fail if the tests are failing.


schildbach:
Merged.

PR #2488, opened 2022-06-24
Context: fix implicit context propagation by constructor

schildbach:
We used to do this a long time ago, but switched the codebase to explicit propagation.
Generally, constructors should not have any side-effects.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/TxConfidenceTableTest.java
+++ b/core/src/test/java/org/bitcoinj/core/TxConfidenceTableTest.java
@@ -46,6 +46,7 @@ public class TxConfidenceTableTest {
     public void setup() throws Exception {
         BriefLogFormatter.init();
         Context context = new Context(UNITTEST);
+        Context.propagate(context);
         table = context.getConfidenceTable();
 
         Address to = LegacyAddress.fromKey(UNITTEST, new ECKey());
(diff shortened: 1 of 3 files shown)


schildbach:
Ready for review, as integration tests are happy.

core/src/test/java/org/bitcoinj/core/TxConfidenceTableTest.java
```diff
@@ -46,6 +46,7 @@ public class TxConfidenceTableTest {
     public void setup() throws Exception {
         BriefLogFormatter.init();
         Context context = new Context(UNITTEST);
+        Context.propagate(context);
```
msgilligan, line 49: Why did you make this change here and nowhere else?

msgilligan:
> Generally, constructors should not have any side-effects.

Well to be fair, setting the class variable `lastConstructed` is also a side-effect.

core/src/test/java/org/bitcoinj/core/TxConfidenceTableTest.java
schildbach: Because that test needs the correct context propagated. The others succeed without, so I didn't take any steps. Would you prefer me explicitly propagating context where one is instantiated?

schildbach:
I think Mike stated that if you want to support multiple networks at the same time, you'd need to make sure threads and thread factories stay separated. Generally speaking, we were not really considering this usecase.

I agree about `lastConstructed` – maybe we can get rid of that too at some time.

schildbach:
Merged.

PR #2492, opened 2022-06-24
Address: Implement PrefixedChecksummedBytes inline in Address rather than extend a base class

msgilligan:
In preparation for future refactoring of Address, we don't want Address
tied to a base class with dependencies on o.b.core or NetworkParameters.

(Instead we provide the NetworkParametersSupplier interface -- removed in force-push)

We also want the option of making Address itself an interface in the future.

(I made PrefixedChecksummedBytes implement NetworkParametersSupplier, though
this was not strictly necessary. This will remind us to consider similar refactoring
for other subclasses of PrefixedChecksummedBytes -- also removed in force-push)

Yes, this results in some code duplication but there is no need or use of polymorphism in the subclasses
of PrefixedChecksummedBytes.

Since equals() was copied from NetworkParameters into Address, create AddressTest
and copy the equals() test from PrefixedChecksummedBytesTest.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Address.java
+++ b/core/src/main/java/org/bitcoinj/core/Address.java
@@ -20,7 +20,11 @@ import org.bitcoinj.base.ScriptType;
 import org.bitcoinj.base.exceptions.AddressFormatException;
 
 import javax.annotation.Nullable;
+import java.util.Arrays;
 import java.util.Comparator;
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Base class for addresses, e.g. native segwit addresses ({@link SegwitAddress}) or legacy addresses ({@link LegacyAddress}).
@@ -28,7 +32,9 @@ import java.util.Comparator;
  * Use {@link #fromString(NetworkParameters, String)} to conveniently construct any kind of address from its textual
  * form.
  */
-public abstract class Address extends PrefixedChecksummedBytes implements Comparable<Address> {
+public abstract class Address implements Comparable<Address> {
+    protected final NetworkParameters params;
+    protected final byte[] bytes;
 
     /**
      * Construct an address from its binary form.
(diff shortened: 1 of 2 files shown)


schildbach:
I know you're doing this for modularization, but this PR seems to complicate what I previously called the "PrefixedChecksummedBytes hierarchy" by kind of breaking it into two hierarchies. This seems somehow backwards, given that e.g. `DumpedPrivateKey` and `LegacyAddress` have so much in common.

Did you consider moving the entire hierarchy to `base`?

msgilligan:
> I know you're doing this for modularization, but this PR seems to complicate what I previously called the "PrefixedChecksummedBytes hierarchy" by kind of breaking it into two hierarchies. This seems somehow backwards, given that e.g. `DumpedPrivateKey` and `LegacyAddress` have so much in common.
> 
> Did you consider moving the entire hierarchy to `base`?

Yes, I would like to eventually move the entire hierarchy to base and to complete eliminate `PrefixedChecksummedBytes`.  There really is no need for these classes to share a common ancestor. All `PrefixedChecksummedBytes` "says" really is that "this class has a byte array and a NetworkParameters".  And we're not (that I know of) ever using `PrefixedChecksummedBytes` for polymorphism, so I think it's best to just inline it into all the classes that inherit from it.

We can leave the `NetworkParametersSupplier` shared interface if we want, but even that isn't strictly necessary.

I know your instinct is to probably refactor this hierarchy at the same time, but I've been very focused on `Address` and decided to break the problem up as `Address` is a fairly complicated refactoring as you can see from all the related PRs.

schildbach:
Quickly glancing at Bitcoin Wallet, I'm using `PrefixedChecksummedBytes` at various places, though I agree it looks like I should be using the more specific classes instead. I vaguely remember I've been using polymorphism for the `InputParser` (mainly used for the generic QR code scanner).

I'll try to get rid of `PrefixedChecksummedBytes` and will report back.

schildbach:
Hmm, ok, `PrefixedChecksummedBytes` seems to only be used to represent private keys. Both `DumpedPrivateKey` and `BIP38PrivateKey` are not touched by this PR, right?

msgilligan:
> Hmm, ok, `PrefixedChecksummedBytes` seems to only be used to represent private keys. Both `DumpedPrivateKey` and `BIP38PrivateKey` are not touched by this PR, right?

Right. (I made PR #2500 to do that.)

msgilligan:
@schildbach I removed `NetworkParametersSupplier` and force-pushed. Should be ready for merge.

schildbach:
Merged.

PR #2493, opened 2022-06-24
Use (new and deprecated) NetworkParameters.fromAddress to isolate incorrect usage

msgilligan:
* Add NetworkParameters.fromAddress() method to consolidate the use cases
  where an Address is used to get a NetworkParameters
* Deprecate NetworkParameters.fromAddress() to indicate code must be rewritten
* This also removes one dependency on NetworkParameters in Address and helps
  prepare for moving Address (perhaps as an interface) to `o.b.base`
  
  Draft PR since it depends on PR #2492 (though it might be possible to cherry-pick)


schildbach:
Merged.

PR #2496, opened 2022-06-25
Fix minor JavaDoc error that broke build.

msgilligan:
Removing the whole line because the suggested method doesn't exist yet.
(It's in PR #2483 and merging 2483 will also fix the broken build.)


schildbach:
Merged.

Sorry for the error; it didn't break locally.

msgilligan:
> Merged.
> 
> Sorry for the error; it didn't break locally.

No problem. Maybe a JavaDoc rebuild wasn't being triggered locally?

PR #2497, opened 2022-06-26
WIP on using AddressKey to remove ECKey dependency from Address

msgilligan:
The AddressKey interface allows a key implementation (in our case just `ECKey`) to create an `Address`.

The idea here is to remove the ECKey-dependent constructors and factory methods in `Address` and subclasses and move that logic to `ECKey`. This simplified the address implementation and removes their dependency on `ECKey` which brings in many other dependencies from `o.b.core` and `o.b.crypto`.

I believe this set of changes is the last one needed to move `Address` and subclasses to `o.b.base`


msgilligan:
Most of the ideas/changes in this PR have been merged via other PRs.

PR #2491, opened 2022-06-24
Move Network and BitcoinNetwork to org.bitcoinj.base

msgilligan:

PR #2499, opened 2022-06-26
Move Network to `o.b.base`

msgilligan:
This replaces PR #2491. We can worry about supporting alternative networks after we finish migrating `Address` and related classes to base.


schildbach:
Merged.

PR #2500, opened 2022-06-26
DumpedPrivateKey,BIP38PrivateKey: inline PrefixedChecksummedBytes

msgilligan:
This is a draft because:

a. `PrefixedChecksummedBytes` is not removed yet
b. The tests in `PrefixedChecksummedBytesTest` should be replicated in `DumpedPrivateKey` & `BIP38PrivateKey`.


schildbach:
I think that's too much of separation. It makes sense to handle both standards for encoded private keys within one class hierarchy. However, at some point I'd rename the base class.

schildbach:
I think this is now obsolete?

msgilligan:
> I think this is now obsolete?

Yes. Closed.

PR #2490, opened 2022-06-24
rename Network to BitcoinNetwork create new Network interface (improved PR)

msgilligan:
This change give us flexibility and preservers the useful abstraction that we currently have with `NetworkParameters`.

This PR replaces PR #2473. It has some minor improvements, but more importantly it contains two commits which preservers the file history of `Network` -> `BitcoinNetwork`.

PR #2473, opened 2022-06-20
rename Network to BitcoinNetwork create new Network interface

msgilligan:
This is related to PR #2472 and should probably be rebased after that one is reviewed/merged.

PR #2501, opened 2022-06-27
use Network interface to implement partial alt-net support

msgilligan:
-  Distinguish between `Network` and `BitcoinNetwork` in the code (parts of our implementation supports alt-networks better than others and this commit doesn't try to fix them all, just make it more clear what is going on. Some may never be made 'generic')

The following two items were originally part of this PR, but were implemented via separate PRs and this PR rebased:

- Add `MockAltNetParams` and `MockAltNetwork` classes for testing
- Partially update BitcoinURI to support alt-networks via `Network`

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Address.java
+++ b/core/src/main/java/org/bitcoinj/core/Address.java
@@ -17,6 +17,7 @@
 package org.bitcoinj.core;
 
 import org.bitcoinj.base.BitcoinNetwork;
+import org.bitcoinj.base.Network;
 import org.bitcoinj.base.ScriptType;
 import org.bitcoinj.base.exceptions.AddressFormatException;
 
@@ -30,7 +31,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 /**
  * Base class for addresses, e.g. native segwit addresses ({@link SegwitAddress}) or legacy addresses ({@link LegacyAddress}).
  * <p>
- * Use an implementation of {@link AddressParser#parseAddress(String, BitcoinNetwork)} to conveniently construct any kind of address from its textual
+ * Use an implementation of {@link AddressParser#parseAddress(String, Network)} to conveniently construct any kind of address from its textual
  * form.
  */
 public abstract class Address implements Comparable<Address> {
@@ -61,7 +62,7 @@ public abstract class Address implements Comparable<Address> {
      *             if the given string doesn't parse or the checksum is invalid
      * @throws AddressFormatException.WrongNetwork
      *             if the given string is valid but not for the expected network (eg testnet vs mainnet)
-     * @deprecated Use {@link org.bitcoinj.wallet.Wallet#parseAddress(String)} or {@link AddressParser#parseAddress(String, BitcoinNetwork)}
+     * @deprecated Use {@link org.bitcoinj.wallet.Wallet#parseAddress(String)} or {@link AddressParser#parseAddress(String, Network)}
      */
     @Deprecated
     public static Address fromString(@Nullable NetworkParameters params, String str)
@@ -81,7 +82,7 @@ public abstract class Address implements Comparable<Address> {
      * @param outputScriptType
      *            script type the address should use
      * @return constructed address
-     * @deprecated Use {@link ECKey#toAddress(ScriptType, BitcoinNetwork)}
+     * @deprecated Use {@link ECKey#toAddress(ScriptType, Network)}
      */
     @Deprecated
     public static Address fromKey(final NetworkParameters params, final ECKey key, final ScriptType outputScriptType) {
@@ -146,7 +147,7 @@ public abstract class Address implements Comparable<Address> {
      * when you need to know what network an address is for.
      * @return the Network.
      */
-    public BitcoinNetwork network() {
+    public Network network() {
         return params.network();
     }
 
--- a/core/src/main/java/org/bitcoinj/core/AddressParser.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressParser.java
@@ -17,6 +17,7 @@
 package org.bitcoinj.core;
 
 import org.bitcoinj.base.BitcoinNetwork;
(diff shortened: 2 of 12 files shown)


msgilligan:
Note: This is the third attempted PR to implement this, for reference see:
* PR #2490 
* PR #2473

406712774:
Yo también los puedo agarrar Ono puedo

msgilligan:
> I've cherry-picked (and repaired) [60b622e](https://github.com/bitcoinj/bitcoinj/commit/60b622e57843579b8db077b7ec958decf39c45fc).

I've rebased on `master` after your cherry-pick, squashed into a single commit, and force-pushed .

msgilligan:
Rebased after PR #2502 was merged.

msgilligan:
@schildbach The next sub-step of this PR is:

> * Migrate certain NetworkParameters properties to Network

I'm going to break this off into a separate PR containing just this change, but before I do let's decide on which ones will migrate. The current state of this PR makes the following changes:

* `NetworkParameters.hasMaxMoney()` -> `Network.hasMaxMoney()` and deprecate
* `NetworkParameters.getMaxMoney()` -> `Network.maxMoney()` and deprecate
* `NetworkParameters.getUriScheme()` -> `Network.uriScheme()` and deprecate
* `NetworkParameters.getMonetaryFormat()` deprecated with recommendation to use `MonetaryFormat` directly

My rationale is that the first three seem like relatively simple properties that most Bitcoin-derived chains have implemented (or can easily false as in the case of `hasMaxMoney()`.)  `.getMonetaryFormat()` was not used elsewhere in **bitcoinj** and is not commonly accessed at the "network" level.

@schildbach has previously suggested that rather than migrate `uriScheme` to `Network` we simply deprecated it and define it in `BitcoinURI`. I think wallet applications may find it useful as a `Network` constant/attribute and it is simple, easy to move and easy to support, but don't feel too strongly that it's absolutely necessary.

So what should I do int the next PR? Make this sub-step of the changes as described above? Make it with the one change that @schildbach has suggested? Or something else?

I'd like to get the PR for that change submitted and reviewed and then move on, so hopefully we can decide quickly.

core/src/main/java/org/bitcoinj/testing/FakeAltNetParams.java
```diff
@@ -0,0 +1,75 @@
...
+import org.bitcoinj.core.NetworkParameters;
+import org.bitcoinj.core.StoredBlock;
+import org.bitcoinj.core.VerificationException;
+import org.bitcoinj.store.BlockStore;
+import org.bitcoinj.store.BlockStoreException;
+
+/**
+ * Fake Alt-net subclass of {@link NetworkParameters} for unit tests.
+ */
+public class FakeAltNetParams extends NetworkParameters {
```
schildbach, line 16: I wonder if "Mock" would perhaps be a better prefix for these two classes, rather than "Fake"?

schildbach:
- `has/getMaxMoney` is clearly a property of the network, so I'd migrate it as you suggested. (It could even be in its own commit if it helps speed up progress.)

- `uriScheme` seems more related to a "payment protocol", rather than the network itself. This is why I suggested moving it to the main BIP21 related class `BitcoinURI`. Bitcoin Wallet isn't even using the property from bitcoinj, because it's mainly needed in its static `AndroidManifest.xml`. Regarding the move to either `Network` or `BitcoinURI`: I can live with either of those. If there is no profit to be made in moving it to `BitcoinURI`, then maybe moving it to `Network` would make migration for bitcoinj users easier.

- `monetaryFormat`: For the records, the accessor was added in 7a3aa74c6ee23af5be461beab6f4ab4564f973d9. I'm fine with removing it, as I see no real usecase (after a deprecation cycle, if possible).

msgilligan:
We're almost finished with 

> - [ ] Migrate certain NetworkParameters properties to Network

PR #2515 is ready-to-merge.  We just need to decide what to do about `uriScheme`.

@schildbach What do you want to do? I prefer moving it to `Network` (for reasons given above), but am happy to move it to `BitcoinURI` if you prefer. Let me know and I'll make a PR based on your decision.

schildbach:
I'm fine with moving it to `Network` for now.

msgilligan:
PR #2593 is intended to reduce the number of classes affected by this pending changes. It's the next step in implementing this concept.

msgilligan:
PRs #2596 and #2597 are the next in the sequence.

core/src/main/java/org/bitcoinj/testing/FakeAltNetParams.java
msgilligan: Changed to "Mock" in latest rebase/push.

schildbach:
Generally I'm happy with this.

I'd wish we could chop this into at least two commits that make sense on its own. I noticed there are many generalizing changes (`BitcoinNetwork` → `Network`), and then there is a different change to `BitcoinURI`. Maybe we can merge the BitcoinURI changes first somehow – but only if it makes sense. If you think it belongs together, I'm fine with that as well.

msgilligan:
I think I can split it up into about 4 commits. Working on the first one now...

msgilligan:
The first three split-out PRs are:

* https://github.com/bitcoinj/bitcoinj/pull/2607
* https://github.com/bitcoinj/bitcoinj/pull/2608
* https://github.com/bitcoinj/bitcoinj/pull/2610

Once those are merged we can re-base and look at what's left.

schildbach:
> Once those are merged we can re-base and look at what's left.

Those are all merged. I think also the commit description now needs an update.

msgilligan:
> > Once those are merged we can re-base and look at what's left.
> 
> Those are all merged. I think also the commit description now needs an update.

Rebased, commit description updated, and force-pushed.  (I also updated the first comment on the PR to explain that two items were merged separately and a rebase occurred)

schildbach:
Merged.

PR #2504, opened 2022-06-29
PeerGroup: Use standard ScheduledExecutorService instead of Listenable

msgilligan:
Since we migrated to (Listenable)CompletableFuture in startAync/stopAsync we no longer need it.


schildbach:
Merged.

PR #2507, opened 2022-06-29
Wallet.SendRequest: construct from TransactionBroadcast only

msgilligan:
Simplify the constructor for Wallet.SendRequest to only take
a TransactionBroadcast object. The transaction was inside
the TransactionBroadcast object anyway, we just had to create
an accessor.

This paves the way for removing SendResult entirely and replacing
it with TransactionBroadcast. (The question is how to do this
smoothly -- i.e. what gets deprecated etc.)


schildbach:
Merged.

PR #2511, opened 2022-07-07
Network: hasMaxMoney(), maxMoney() moved from NetworkParameters

msgilligan:
* Network: add hasMaxMoney(), maxMoney()
* Deprecate existing equivalent methods in NetworkParameters/AbstractBitcoinNetParams

PR #2512, opened 2022-07-07
Coin: add ofBTC(long coins) factory method

msgilligan:
* Add overload `Coin.ofBTC(long coins)`
* Add overload `static long btcToSatoshi(long coins)`
* Use the new `Coin.ofBTC` where appropriate to improve readability

Draft: depends on PR #2511 (though it doesn't really _need_ to, but #2511 is what showed me the need for the overload)

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/Coin.java
+++ b/core/src/main/java/org/bitcoinj/base/Coin.java
@@ -47,7 +47,7 @@ public final class Coin implements Monetary, Comparable<Coin> {
     /**
      * One Bitcoin.
      */
-    public static final Coin COIN = Coin.valueOf(COIN_VALUE);
+    public static final Coin COIN = Coin.ofBtc(1);
 
     /**
      * 0.01 Bitcoins. This unit is not really used much.
@@ -69,7 +69,7 @@ public final class Coin implements Monetary, Comparable<Coin> {
      */
     public static final Coin SATOSHI = Coin.valueOf(1);
 
-    public static final Coin FIFTY_COINS = COIN.multiply(50);
+    public static final Coin FIFTY_COINS = Coin.ofBtc(50);
 
     /**
      * Represents a monetary value of minus one satoshi.
@@ -120,7 +120,7 @@ public final class Coin implements Monetary, Comparable<Coin> {
         checkArgument(cents < 100);
         checkArgument(cents >= 0);
         checkArgument(coins >= 0);
-        final Coin coin = COIN.multiply(coins).add(CENT.multiply(cents));
+        final Coin coin = ofBtc(coins).add(CENT.multiply(cents));
         return coin;
     }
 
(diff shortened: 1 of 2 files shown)


msgilligan:
PR #2511 was merged. This is rebased and ready-for-review/merge.

core/src/main/java/org/bitcoinj/base/Coin.java
```diff
@@ -154,6 +165,16 @@ public static Coin ofBtc(BigDecimal coins) {
         return Coin.valueOf(btcToSatoshi(coins));
     }
 
+    /**
+     * Create a {@code Coin} from a decimal amount of BTC.
+     *
+     * @param coins number of coins (in BTC)
+     * @return {@code Coin} object containing value in satoshis
+     */
+    public static Coin ofBtc(long coins) {
```
schildbach, line 174: There is already an `valueOf(coins, cents) ` method. Why not use that?

schildbach:
I agree some of the constants are populated a bit unconventionally (-:
We could fix that by using `valueOf(…)` everywhere.
However I'm not so happy about introducing another static constructor.

msgilligan:
> I agree some of the constants are populated a bit unconventionally (-:
> We could fix that by using `valueOf(…)` everywhere.

I'm not a fan of `valueOf(coins, cents)` because cents are a unit that almost nobody uses these days. I checked usages of that method and it's only used in `FakeTxBuilder`, 2 JavaDoc comments (in `TransactionOutput`), and the unit tests.

> However I'm not so happy about introducing another static constructor.

How about if we deprecate`valueOf(coins, cents)` and replace it with `ofBtc(long)`?

Also `ofBtc(long)` isn't _completely_ a new static constructor it's an _overload_ of `ofBtc(BigDecimal)` 😁 It can be used anytime someone needs a whole-number of coins and will also work for `int`.

schildbach:
Actually I think the already existing `ofBtc()` method is almost useless, except for esotheric purposes like unit tests with round values. As a user of Bitcoin, I have never sent (or received) a round Bitcoin value.

So if I were to decide, I'd deprecate and remove that method and only use `valueOf(sat)` or `valueOf(bitcoin, cent)` everywhere, also in unit tests.

msgilligan:
> Actually I think the already existing `ofBtc()` method is almost useless

You mean the existing `ofBTC(BigDecimal)` method? That method is really useful -- it allows any application that is using BigDecimal values to convert to `Coin`.  See the discussion in PR #2013.

>  except for … unit tests with round values. As a user of Bitcoin, I have never sent (or received) a round Bitcoin value.

Now it sounds like you're talking about the proposed new `ofBTC(long)`  method. I agree that there aren't many use cases for making `Coins` from an integer number of bitcoins, but I also view `Coin` as a numeric value-type and think it's entirely reasonable to have an overload of `ofBTC(BigDecimal)` that takes (`long`) integers. It's not a big deal, but it does improve the readability of some of our  code as you can see in this PR.

msgilligan:
> Side note: I also don't like the `ofBtc` naming.

That was actually your idea, see: PR #2013. I had original proposed `ofBitcoin`.

> Everywhere else we use the word "coins" for a round number of Bitcoins.

The problem with the word "coins" is that it implies the `Coin` class, so `Coin.ofCoins()` would be confusing.

PR #2515, opened 2022-07-08
NetworkParameters: deprecate getMonetaryFormat()

msgilligan:
`getMonetaryFormat()` is not used within **bitcoinj** and ``MonetaryFormat` isn't strictly tied to a "network" and can be instantiated or found via other mechanisms, so let's deprecate it as part of our effort to deemphasize `NetworkParameters`.


schildbach:
Merged.

PR #2519, opened 2022-07-09
ForwardingService: Use BitcoinNetwork where possible

msgilligan:
(Eventually, we should be able to eliminate the use of NetworkParameters entirely, but this is a first step)


schildbach:
Merged.

PR #2521, opened 2022-07-10
ForwardingService: Use BitcoinNetwork.fromString to parse network CLI parameter

msgilligan:


schildbach:
Doesn't this change the accepted values for ForwardingService again, re-introducing the regression fixed via bca5faaba22436c6e71dcdf865308cffd9e86a5a?

core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
```diff
@@ -76,17 +77,49 @@ public Coin maxMoney() {
         return MAX_MONEY;
     }
 
+    /**
+     * Get the {@code BitcoinNetwork} from a <i>validated</i> name String
+     * @param nameString A name string (e.g. "MAIN", "TEST", "REGTEST")
```
schildbach, line 82: I think here I'd prefer to see `SIGNET` as an example string, as it is more relevant than `REGTEST` these days.

core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
```diff
@@ -76,17 +77,49 @@ public Coin maxMoney() {
...
+     * @throws IllegalArgumentException if there is no matching enum
+     */
+    public static BitcoinNetwork of(String nameString) {
+        return find(nameString)
+                .orElseThrow(() -> new IllegalArgumentException("Unrecognized network name : " + nameString));
+    }
+
+    /**
+     * Find the {@code BitcoinNetwork} from a name String
+     * @param nameString A name string (e.g. "MAIN", "TEST", "REGTEST")
```
schildbach, line 93: Same here.

schildbach:
Generally I like the idea of having this methods and let tools use them.

msgilligan:
> Doesn't this change the accepted values for ForwardingService again, re-introducing the regression fixed via [bca5faa](https://github.com/bitcoinj/bitcoinj/commit/bca5faaba22436c6e71dcdf865308cffd9e86a5a)?

Yes it changes the accepted values which was my intention (I thought we decided that we wanted to make `ForwardingService` consistent with `WalletTool`.  However, your comment reminded me that I forgot to change the usage string accordingly. So I fixed the usage and force-pushed.

core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
msgilligan: Fixed.

core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
msgilligan: Fixed.

core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
```diff
@@ -76,17 +77,49 @@ public Coin maxMoney() {
...
+        return findById(idString)
+                .orElseThrow(() -> new IllegalArgumentException("Illegal network ID: " + idString));
+    }
+
+    /**
+     * Find the {@code BitcoinNetwork} from an ID String
+     * @param idString specifies the network
+     * @return An {@code Optional} containing the matching enum or empty
      */
-    public static BitcoinNetwork of(String idString) {
```
schildbach, line 86: I'm not sure if it's a good idea to change this "by ID" constructor from strict `of` to lenient `find`. After all, this is an ID which is likely not user facing (other than the freshly introduced "name"). Maybe restore the old error handling and name it `ofId()`, or keep the `of()` name and use `ofName()` for the other constructor?

msgilligan: The diff is a little hard to read, the strict 'of' became a strict `fromId` but the diff shows that as inserted above it.

This PR results in 4 static factory methods:

of(name)    -- strict: either returns a `BitcoinNetwork` or throws exception
fromId(id)   -- strict: either returns a `BitcoinNetwork` or throws exception
find(name)  -- lenient: returns an Optional<BitcoinNetwork>
findById(id) -- lenient: returns an Optional<BitcoinNetwork> 

I think this covers all the use cases. I have used network ID strings in configuration files for server apps and think others may do similar things. In that case you might view the ID string as "user facing". I thought the `of` should use the "name" and be the inverse of `.toString()` or `.name()`. We could rename `fromId()` to `ofId()` I suppose.

msgilligan:
@schildbach This is rebased (and the title has been changed.)  All that is left of this PR is using `BitcoinNetwork.fromString` to replace the existing method/switch.

schildbach:
Merged.

PR #2523, opened 2022-07-11
BitcoinNetwork: allow a number of common alternate network names in find()

schildbach:
Also adds tests for of() and find().

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
+++ b/core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
@@ -32,10 +32,10 @@ import static org.bitcoinj.base.Coin.COIN;
  * method for input of network values.
  */
 public enum BitcoinNetwork implements Network {
-    MAIN("org.bitcoin.production"),
-    TEST("org.bitcoin.test"),
-    SIGNET("org.bitcoin.signet"),
-    REGTEST("org.bitcoin.regtest");
+    MAIN("org.bitcoin.production", new String[] { "mainnet", "prod" }),
+    TEST("org.bitcoin.test", new String[] { "testnet" }),
+    SIGNET("org.bitcoin.signet", new String[] { "sig" }),
+    REGTEST("org.bitcoin.regtest", new String[0]);
 
     /**
      * The maximum number of coins to be generated
@@ -59,9 +59,11 @@ public enum BitcoinNetwork implements Network {
     public static final String ID_UNITTESTNET = "org.bitcoinj.unittest";
 
     private final String id;
+    private final String[] alternateNames;
 
-    BitcoinNetwork(String networkId) {
-        id = networkId;
+    BitcoinNetwork(String networkId, String[] alternateNames) {
+        this.id = networkId;
+        this.alternateNames = alternateNames;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
```diff
@@ -84,21 +86,33 @@ public Coin maxMoney() {
...
     }
 
     /**
-     * Find the {@code BitcoinNetwork} from a name String
+     * Find the {@code BitcoinNetwork} from a name string, or from a number of common alternate names
      * @param nameString A name string (e.g. "MAIN", "TEST", "SIGNET")
      * @return An {@code Optional} containing the matching enum or empty
      */
     public static Optional<BitcoinNetwork> find(String nameString) {
+        final Optional<BitcoinNetwork> canonicalName = findCanonicalName(nameString);
```
schildbach, line 99: This logic deserves a second look by someone used to Optionals. I figured the `.or()` method might be suitable here, but it's only available on Java9+. Converting the two Optionals into a stream and using `filter()/findFirst()` seemed like overkill to me.

msgilligan: Yeah, `or()` is what we really want here. I think the best we can do is:
```
        return canonicalName.isPresent()
                ? canonicalName
                : findAlternateName(nameString);
```

msgilligan: (Java 9 -- and to a lesser extent 10 and 11 -- really smoothed some of the rough edges in the new features of Java 8.)

core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
```diff
@@ -84,21 +86,33 @@ public Coin maxMoney() {
...
+        if (canonicalName.isPresent()) return canonicalName;
+        return findAlternateName(nameString);
+    }
+
+    private static Optional<BitcoinNetwork> findCanonicalName(String nameString) {
         return Arrays.stream(values())
                 .filter(n -> n.toString().equals(nameString.toUpperCase()))
```
msgilligan, line 106: I'm having second thoughts about the `.toUpperCase()` here and in `findAlternateName`. We are not providing an option for someone to find an exact, case-sensitive match. There are three alternatives:
1. Require the caller to use `.toUpperCase`, e.g. `var optionalNetwork = find(nameCandidate.toUpperCase())`
2. Provide an overloaded `find` method with a 2nd, boolean parameter called `ignoreCase`, eg. `var optionalNetwork = find(nameCandidate, true)`
3. Leave it as is and see if we really need it.

msgilligan (changes requested):
I guess I'm "requesting" the `return ? :` change -- I think that's a functional style we should use by default. As to an "ignore case" option, I'm on the fence.

core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
schildbach: Thanks! I've pushed the change.

core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
schildbach: I strongly dislke 2. Imho "canonical" should not allow for variance.

For similar reason, I dislike 3 as well because if we need it then we're faced with incompatible code.

I think we should make sure the `of()` methods require the exect name/ID. To me, the question is do we want all-uppercase or all-lowercase names (IDs are already all-lowercase).

msgilligan:
I think we made a mistake in b375a96565aa4c51587451cb7dea2654f37d5425. The `of()` method is a reimplementation of the compiler-generated `static valueOf(String)` method that all enums contain.

I've also just re-read **Item 34: Use enums instead of int constants** in _Effective Java (Third Edition)_ (see p. 164) and and had a Matrix conversation with @schildbach where he convinced me that we should strongly encourage the use of _lowercase_ names in user-facing input and output of these constants.

For these reasons, I'm going to open a new PR that re-implements b375a96565aa4c51587451cb7dea2654f37d5425 in a more standard/idiomatic way and takes @schildbach 's suggestion of canonical lowercase into account. Once that PR is reviewed and/or merged we can revisit support for alternative lowercase names.

schildbach:
I've changed all alternate names to lowercase, in harmony with our new toString() / fromString() values.

Kiler91daw (commented):
Brawo dobrze jeżeli widać efekty nie zniszczenia budowania 🙌👍👌

PR #2524, opened 2022-07-11
BitcoinNetwork: simplify/standardize and canonicalize lowercase for user-facing

msgilligan:
This PR corrects some issues with b375a96565aa4c51587451cb7dea2654f37d5425 and brings things more in alignment with Java conventions (see _Effective Java (Third Edition)_, **Item 34: Use enums instead of int constants** -- especially the discussion of `valueOf(String)`, `toString()` and `fromString(String)` on p. 164)

Specifically:
* Add an override of `toString()` to output the string in canonical, lower-case format.
* Remove the `.of(String)` method from b375a96565aa4c51587451cb7dea2654f37d5425 as it duplicates the built-in `.valueOf()`
* Remove the `ofId(String)` method from b375a96565aa4c51587451cb7dea2654f37d5425 in favor of the using the `fromIdString(String)` method with `orElseThrow()` or similar if throwing an exception is desired.
* Rename `.find(String)` to `.fromString(String)` to use the suggested name from _Effective Java_.
* Rename `.findById(String)` to `.fromIdString(String)` for consistency.
* Improve the JavaDoc.

In summary, this reduces the number of static factory methods from 5 to 3, conforms better with Java conventions, and clarifies our preference for lowercase strings as canonical for user-facing input/output.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
+++ b/core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
@@ -17,12 +17,19 @@
 package org.bitcoinj.base;
 
 import java.util.Arrays;
+import java.util.Locale;
 import java.util.Optional;
 
 import static org.bitcoinj.base.Coin.COIN;
 
 /**
- * A convenient {@code enum} representation of a network.
+ * A convenient {@code enum} representation of a Bitcoin network.
+ * <p>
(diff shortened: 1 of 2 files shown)


msgilligan:
> Looks good. Do you think you could add some tests for those 3 methods?

Yes. Will do.

msgilligan:
@schildbach Tests added.

schildbach:
Merged.

PR #2526, opened 2022-07-12
BitcoinNetwork: build name lists and static map at class initialization time

msgilligan:
* Specify altNames as list argument to constructor
* In constructor add canonical name to a temp list
* Build static map of names to BitcoinNetwork at class initialization time

This was inspired by an example in **Item 34: Use enums instead of int constants** in _Effective Java_.

This PR is dependent on PR #2523.


msgilligan:
@schildbach I don't know why this PR is showing 2 commits. I rebased on `master` now that PR #2523 was merged. GitHub bug?

schildbach:
I only see 1 commit.

msgilligan:
> I only see 1 commit.

That's because I rebased and force-pushed.

msgilligan:
> lots of crazy streaming magic going on (-:

It's not too crazy or magic if you've created streams of `Map.Entry` or collected streams into `Map` before.

I was able to simplify it a little in this commit: 311a7213ecc093bd597141c8f8954cdc747e06e2

There are features in JDK 11 and JDK 17 that could be used to simplify it further (and create an unmodifiable map), of course.

> However I wonder what do we gain? A bit of speed improvement when parsing CLI arguments?

* The code for fetching a `BitcoinNetwork` is much easier to read (one line)
*  Internally we have a `private final List<String>` of all names which is nice (though mostly unused at present)
* More immutability of our privates (except for the Map, which would could make unmodifiable with `collectingAndThen` if we want)

>On the negative side, it seems to me the code is much harder to read.

A little harder to read (311a7213ecc093bd597141c8f8954cdc747e06e2 helps) but not so bad if you are used to using `Stream` with `Map`.

It's not a big deal to me either way. What do you want to do: close or merge?

schildbach:
It's not so much about the API being used, I'm slowly getting used to it (-:

It's the indirection that bothers me a bit. When reading `fromString()`, the method itself doesn't contain any more what different sources are used for finding the appropriate network. Rather, I have to dive into the internals of the class and find out how the map is created at init time.

It's also little things like the alternate name arrays now being converted to collections with three different methods being used to create them. I personally would prefer if at least the "data section" of the class would stay as concise as possible (without distractions). Btw. is there a shorthand syntax for arrays in later Java versions, I'm thinking of type inference of the `new String[]`?

This seems more like a performance optimization to me than an improvement in readability or maintainability. And I admit, the old (naive) implementation is quite inefficient.

msgilligan:
> It's the indirection that bothers me a bit. When reading `fromString()`, the method itself doesn't contain any more what different sources are used for finding the appropriate network. Rather, I have to dive into the internals of the class and find out how the map is created at init time.

I suppose better documentation of what the Map looks like would help. Maybe even showing its exact state. In fact it might be good idea to just put a table in the JavaDoc that documents the strings (whichever way it is implemented.)

> 
> It's also little things like the alternate name arrays now being converted to collections with three different methods being used to create them.

I was using `Arrays.asList()` to create them, but IntelliJ was highlighting that in yellow for the "singleton" and "empty" cases, so I made that change to make IntelliJ happy. In Java 9 the `List.of()` methods could be used for all cases.

> Is there a shorthand syntax for arrays in later Java versions, I'm thinking of type inference of the `new String[]`?

I think we there is a simpler syntax than what you are using and it goes back to even earlier Java versions. Let me double-check and make a PR if possible.


> This seems more like a performance optimization to me than an improvement in readability or maintainability. And I admit, the old (naive) implementation is quite inefficient.

I don't think performance is really an issue here. I think code readability is the most important thing and your criticisms are valid.

msgilligan:
> This seems more like a performance optimization to me than an improvement in readability or maintainability. 

I'm not giving up, @schildbach! I just force-pushed another attempt at making this simpler and more readable. It's good practice with Streams (or maybe it's just bikeshedding.) We can (and probably should) move on at any point, but what do you think of this one?

schildbach:
Your new version certainly beats the old version at documentation. So I'm merging this :laughing:

PR #2527, opened 2022-07-13
WalletTool: Use BitcoinNetwork::toString for chain fileName

msgilligan:
This is a potentially breaking change and may require existing wallets managed with the `WalletTool` to have their chain files manually renamed:

`mainnet.chain` -> `main.chain`
`testnet.chain` -> `test.chain`

or use the `--chain` command line option to specify the name.

Given that this is a test tool and only distributed as source, I think this is an acceptable change in return for code simplification and naming standardization.


schildbach:
Merged.

PR #2529, opened 2022-07-13
Network: rename MAIN to MAINNET and TEST to TESTNET

schildbach:
The old, shorter names are added to alternate names so they will still be accepted for user-facing input.

The goal is to harmonize names within bitcoinj (we already have `SIGNET`) and also with other tools. With the exception of Bitcoin Core, most tools seem to use `*NET` constants.

Also, `TEST` is often used as a constant for app testing which is confusing since building an app in a test configuration and testing a (productive) app on TestNet are completely different things. A clear distinction between `TEST` and `TESTNET` helps.


msgilligan (approved):
Looks good.

schildbach:
Thanks! Merged.

PR #2530, opened 2022-07-13
wallettemplate: migrate from params to BitcoinNetwork in core classes

msgilligan:
When base/core adds the ability to create an `Address` from a `Network` we should be able to use `Network`/`BitcoinNetwork` for everything.


schildbach:
Merged.,

PR #2533, opened 2022-07-15
ForwardingService: simplify getPrefix()

msgilligan:
Use network.toString(). This also adds support for SIGNET. It changes the filename for MAINNET, but as this is just an example I don't think we really need to preserve backward-compatibility for the filename.


schildbach:
Merged.

PR #2534, opened 2022-07-15
ForwardingService: Refactor to CompletableFuture take 2

msgilligan:


examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
```diff
@@ -171,4 +163,29 @@ private void forwardCoins() {
             throw new RuntimeException(e);
         }
     }
+
+    static CliArguments parseArguments(String[] args) {
```
schildbach, line 179: I was already wondering if we should introduce picocli for the example(s)?

schildbach:
Cherry-picked aab6f6455540a51165a4ee4973e890450b54e68c.

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
msgilligan: I would recommend against it for the following reasons:

* Most examples only have one or two parameters (I think some have zero)
* **picocli** adds some cognitive load to people who aren't familiar with it and distracts from the example
* **picocli** will tempt us to try to get too fancy when the point of the examples is to demonstrate how to use **bitcoinj**

Note that I was experimenting with a Kotlin version of this app last night and in Kotlin we are able to use the `Pair` type and get:
```
val (network, forwardingAddress ) = parseArguments(args)
```

schildbach: …however I feel the new `CliArguments` class distracts too. Are you sure this commit is an enhancement?

msgilligan: Well, it shrinks `main()` by about 10 lines. Maybe the better solution is what I refer to here:

> a future PR that leaves the parameter parsing in main() and moves the actual async processing to another method

msgilligan: I think this one does the job: https://github.com/bitcoinj/bitcoinj/pull/2535

msgilligan:
Rebased after #2535.

schildbach:
Cherry-picked f6fceae8c6dbb37dae97a549a48b0d23a8278265.

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright 2013 Google Inc.
  * Copyright 2014 Andreas Schildbach
```
schildbach, line 3: Is it ok for you if we change to the generic copyright line here?

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
```diff
@@ -112,66 +135,59 @@ public void start() {
         // Download the blockchain and wait until it's done.
         kit.startAsync();
         kit.awaitRunning();
+        wallet = kit.wallet();
```
schildbach, line 138: This smells like a race to me. Peeking at the `WalletAppKit` code, wallet init happens in a separate thread.

Also, further down you're using the cached variable (I think) asynchronously, potentially introducing another race?

Generally I think there is no need to cache `wallet` at all. `WalletAppKit` is meant as a kit, shielding you from some internals. So you just save the kit, and use operations on it. Nothing wrong with `kit.wallet()` imho.

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
```diff
@@ -23,34 +24,37 @@
...
-    private final NetworkParameters params;
-    private final Address forwardingAddress;
     private final WalletAppKit kit;
+    private Wallet wallet;
 
-    public static void main(String[] args) throws Exception {
+    public static void main(String[] args) {
         // This line makes the log output more compact and easily read, especially when using the JDK log adapter.
         BriefLogFormatter.init();
+        PrintStream out = System.out;
```
schildbach, line 57: I'm not a fan of caching `System.out` like that, and (further down) pass it into methods. `System.out` is a global variable, and it has always been used like that.

If we decide to go for that style, we should apply it to the entire framework (where is applies, possibly only examples/tools).

msgilligan: We can drop this for now (and probably permanently.)

For the record, I like it for three reasons:

1. `System.out.println` gets verbose when used on many lines.
2. If you pass a `PrintStream` parameter to a method it allows that method to also write to other stream types.
3. In my refactoring I could use my `out` variable to easily check that all print output had been moved to a single method.

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
msgilligan: > This smells like a race to me. Peeking at the `WalletAppKit` code, wallet init happens in a separate thread.
> 
> Also, further down you're using the cached variable (I think) asynchronously, potentially introducing another race?

The `vWallet` member (returned by `wallet()`) is initialized during startup which is by definition complete after `.awaitRunning()`. So there is no race condition.

> Generally I think there is no need to cache `wallet` at all. `WalletAppKit` is meant as a kit, shielding you from some internals. So you just save the kit, and use operations on it. Nothing wrong with `kit.wallet()` imho.

`WalletAppKit` is basically a factory object for the `Wallet` class (that also handles shutdown). It does not shield you from any internals as you have to use `.kit.wallet()` for everything wallet related.  `vWallet`/`.wallet()` is not `final` but is essentially a Kotlin  `lateinit` that is initialized during the `STARTING` state of the Guava service.

I would like to see some refactoring of `WalletAppKit` because (as we're seeing) it _is_ a little bit confusing.  One idea is to make it more clearly an asynchronous `Wallet` factory (perhaps the `Wallet` instance could be the return value for a `CompletableFuture`.) Another is to have it directly implement a `Wallet` interface (once we have one.)

But for purposes of this PR, we can keep it simple and just use `.kit.wallet()`.

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
msgilligan: Yeah, that's fine. I wanted to get a `2022` in there somewhere.

schildbach: Well, the generic line doesn't have a year. See 5259cc1d72ca5de0a9a296510836aa1ced9a5581.

msgilligan:
Closing in favor of PR #2542

PR #2536, opened 2022-07-15
Transaction/TransactionOutput: move methods to TransactionBag

msgilligan:
* Move methods from Transaction/TransactionOutput that take a wallet/TransactionBag parameter
  to TransactionBag
* Deprecated existing methods
* Update existing code to use new calls

There are two outstanding issues (which is why I marked this as DRAFT):

1. TODO about commented out Android caching code in TransactionBag
2. @Ignore on isTxConsistentReturnsFalseAsExpected() method in WalletTest (EasyMock is being applied to `TransactionOutput` and so with the functionality moved to `TransactionBag` that seems to have broken the test)

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionBag.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionBag.java
@@ -17,19 +17,27 @@
 
 package org.bitcoinj.core;
 
+import org.bitcoinj.base.Coin;
 import org.bitcoinj.base.ScriptType;
 import org.bitcoinj.base.Sha256Hash;
 import org.bitcoinj.script.Script;
-import org.bitcoinj.wallet.Wallet;
+import org.bitcoinj.script.ScriptException;
+import org.bitcoinj.script.ScriptPattern;
 import org.bitcoinj.wallet.WalletTransaction;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 
 /**
- * This interface is used to abstract the {@link Wallet} and the {@link Transaction}
+ * This interface is used to abstract the {@link org.bitcoinj.wallet.Wallet} and the {@link Transaction}
  */
 public interface TransactionBag {
+    Logger log = LoggerFactory.getLogger(TransactionOutput.class);
+
     /**
      * Look for a public key which hashes to the given hash and (optionally) is used for a specific script type.
      * @param pubKeyHash hash of the public key to look for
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/bitcoinj/core/TransactionBag.java
```diff
@@ -49,4 +57,138 @@ public interface TransactionBag {
...
+        }
+
+        return walletOutputs;
+    }
+
+    /**
+     * Returns the difference of {@link #getValueSentToMe(Transaction)} and {@link #getValueSentFromMe(Transaction)}.
+     */
+    default Coin getValue(Transaction tx) throws ScriptException {
+        // TODO: Can we lose the commented-out caching code?
```
msgilligan, line 95: Issue #1

core/src/test/java/org/bitcoinj/wallet/WalletTest.java
```diff
@@ -743,10 +744,12 @@ public void isConsistent_spent() {
     }
 
     @Test
+    @Ignore
     public void isTxConsistentReturnsFalseAsExpected() {
         Wallet wallet = Wallet.createDeterministic(UNITTEST, ScriptType.P2PKH);
         TransactionOutput to = createMock(TransactionOutput.class);
         EasyMock.expect(to.isAvailableForSpending()).andReturn(true);
+        // TODO: Now that this is moved to Wallet (TransactionBag) mocking doesn't work here
```
msgilligan, line 752: This is issue #2

core/src/main/java/org/bitcoinj/core/TransactionBag.java
```diff
@@ -17,19 +17,27 @@
...
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 
 /**
- * This interface is used to abstract the {@link Wallet} and the {@link Transaction}
+ * This interface is used to abstract the {@link org.bitcoinj.wallet.Wallet} and the {@link Transaction}
  */
 public interface TransactionBag {
+    Logger log = LoggerFactory.getLogger(TransactionOutput.class);
```
schildbach, line 39: Is the class correct here?

core/src/main/java/org/bitcoinj/core/TransactionBag.java
schildbach, line 95: The hack was added via ece8d9a3477c0cc517f4dbf36d37e1e9441725de in 2015. In general I'd say mobile devices are quick enough that special cases should not be needed. However I don't know about this specific case.

schildbach: See #2538.

core/src/test/java/org/bitcoinj/wallet/WalletTest.java
schildbach: I've managed to fix the test by applying this patch:
```
-        EasyMock.expect(to.isMineOrWatched(wallet)).andReturn(true);
+        EasyMock.expect(to.getScriptPubKey()).andReturn(ScriptBuilder.createOutputScript(wallet.currentReceiveAddress()));
```
The idea is that if I convert one of the wallet's addresses to a script and return that in the mocked `getScriptPubKey()`, it should be mostly equivalent to the previous mocked `isMineOrWatched()` returning true.

schildbach:
I think this PR is an obvious structural improvement. I never liked the previous "backwards" API.

msgilligan:
Rather then rebase this PR in-place, I created PR #4273 to replace it.

core/src/test/java/org/bitcoinj/wallet/WalletTest.java
msgilligan: Resolved in PR #4273.

core/src/main/java/org/bitcoinj/core/TransactionBag.java
msgilligan: Resolved in PR #4273.

core/src/main/java/org/bitcoinj/core/TransactionBag.java
msgilligan: Resolved in PR #4273.

msgilligan: Resolved in PR #4273.

PR #2537, opened 2022-07-16
LinuxSecureRandom: don't automatically install RNG workaround any more

schildbach:
LinuxSecureRandom used to be initialized on Android because very old Android versions had a bug
in the random number generator. We don't support those old versions any more, so the workaround
has become unnecessary. It is still possible to initialize it manually if desired.


msgilligan (commented):
I trust your Android knowledge, @schildbach.

schildbach:
Merged.

PR #2539, opened 2022-07-16
ForwardingService: restructure in preparation for async refactoring

msgilligan:


schildbach:
I've cherry-picked into two commits: c9835eb6684cdaf0c30490aa35abd55b533f2134 and 95fb0ec3a4e927b98d616d9eb9b85a33fd1c18f6

The rest could probably go together with the CompletableFuture refactoring?

msgilligan:
> I've cherry-picked into two commits: [c9835eb](https://github.com/bitcoinj/bitcoinj/commit/c9835eb6684cdaf0c30490aa35abd55b533f2134) and [95fb0ec](https://github.com/bitcoinj/bitcoinj/commit/95fb0ec3a4e927b98d616d9eb9b85a33fd1c18f6)
> 
> The rest could probably go together with the CompletableFuture refactoring?

Let me rebase. I'd also like the `CompletableFuture` refactoring to be as minimal as possible for comparison purposes. So maybe one more commit before `CompletableFuture`.

schildbach:
Looking at your async commit, the extract variable seems to be independant. And I currently don't see why to extract it.

msgilligan:
> Looking at your async commit, the extract variable seems to be independant. And I currently don't see why to extract it.

You mean the `listener` variable? The reason for extracting it is that for proper cleanup (which isn't strictly necessary in an app where the entire process exists) the listener should be removed once it has received the single transaction it is waiting for.

schildbach:
I'm sorry, I just noticed the removal code.

I'm not sure if this is how ForwardingService is meant to work. My understanding was it is forwarding until you press ctrl-c. It won't stop after 1 forward.

But even if the requirement would say to stop listening. I think the extraction of the variable should go to the commit that changes this behaviour. Because on its own, the variable doesn't make much sense. But if it is actually needed at two locations, it makes sense.

msgilligan:
> I'm sorry, I just noticed the removal code.
> 
> I'm not sure if this is how ForwardingService is meant to work. My understanding was it is forwarding until you press ctrl-c. It won't stop after 1 forward.

I thought the intention was for it to forward once and it just lacked the proper structure to terminate after forwarding. So my changes to use `CompletableFuture` were partly designed around that. A `CompletableFuture` is by-definition a "1-shot" and I made the the `waitForCoins()` method a `CompletableFuture`. I thought having it terminate after receiving and forwarding was fixing a bug.

If we want it to operate a long-lived service, I'm going to have to rethink the async refactoring a little bit. (I hinted at this briefly in a private chat with @schildbach where I brought up the idea of using Reactive Streams for a transaction listener)

msgilligan:
Closing in favor of PR #2540 and #2542

PR #2541, opened 2022-07-16
ForwardingService: Save "coins received" listener in a variable

msgilligan:
This will make it easier for us to remove the listener in the future.


schildbach:
As I wrote in the other PR, I think the removal of the listener is probably a mistake.

msgilligan:
Closing because this change is included in PR #2542.

(Note that removing the listener is not a mistake **if** the app/service quits after forwarding one transaction)

PR #2542, opened 2022-07-16
ForwardingService: Convert to CompletableFuture, quit after forwarding 1 transaction

msgilligan:
* Use `CompletableFuture` for all asynchronous operations
* Use `thenCompose` to chain them
* Quit after forwarding one transaction

PR #2543, opened 2022-07-17
ForwardingService: convert back to multiple forwards

msgilligan:
This is dependent upon PR #2542.

This is just a proof-of-concept, there are a few issues to resolve first:

1. Only forward funds from incoming transaction to avoid race-conditions/double-spends.2. 
2. Add a termination handler to clean-up and remove the listener
3. Decide what is best for a "Getting Started" example

### DIFF
--- a/examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
+++ b/examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
@@ -34,7 +34,6 @@ import org.bitcoinj.wallet.listeners.WalletCoinsReceivedEventListener;
 
 import java.io.File;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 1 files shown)


schildbach:
From your issues list 1 and 2 seem to be a problem with current master too. So at least those are not a regression, right?

Ad 3, you mean if #2542 or #2543 is better?

msgilligan:
> From your issues list 1 and 2 seem to be a problem with current master too. So at least those are not a regression, right?

Right.

> Ad 3, you mean if #2542 or #2543 is better?

Yes, which is better from a beginner perspective. Obviously 2543 is better from a usefulness perspective.

msgilligan:
Closing in favor of PR #2546

PR #2544, opened 2022-07-17
ForwardingService: option to connect to specific peer, rather than using peer discovery

schildbach:


msgilligan (approved):
Looks good to me, though I'd get rid of the Nullable parameter.

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
```diff
@@ -103,6 +107,8 @@ public ForwardingService(Address forwardingAddress, BitcoinNetwork network) {
                 KeyChainGroupStructure.BIP32,
                 new File("."),
                 getPrefix(network));
+        if (peerAddress != null)
+            kit.setPeerNodes(peerAddress);
```
msgilligan, line 111: Actually, I don't think we're allowed to do this here. We have to wait for setup completed, I think.

msgilligan (changes requested):

schildbach:
My tests with this weren't really successful, so for now I'll keep this open.

I managed to get peer discovery find enough peers on testnet after lowering the max number of peers via commit 691cd462675143def984b69fb9906c758d27ec10.

PR #2545, opened 2022-07-17
ForwardingService: print the hash of the forwarding transaction earlier

schildbach:


msgilligan (approved):

msgilligan:
I still approve this PR and I think it's a good thing to have access to the `txId` of a transaction before it is broadcast. However, assuming we want to keep the separation between console output and the actual async operations (which I also think is a good thing) we're going to need to make further changes to support this. Those changes may even propagate to `Wallet` itself. All this is good and will improve our architecture.

schildbach:
Merged.

PR #2546, opened 2022-07-17
ForwardingService: CompletableFuture implementation that forwards multiple transactions

msgilligan:

PR #2547, opened 2022-07-17
Transaction: add CF-returning waitForConfirmations()

msgilligan:


schildbach:
In general, I like the idea. I just wonder how this will be changed if we ever manage to get rid of the depth in confidence.

msgilligan:
I've thought about it further and I think we should actually have a `.waitForTransactionConfirmation(Transaction, int)` as a method of either "Wallet" or "Blockchain".

msgilligan:
> So are you planning to ready this before we merge either #2542 or #2546?

No. Let's test the new approach in `ForwardingService`, merged that into `master`, and sometime later, let's consider our best option for how to provide equivalent functionality in `core`. I also have some architectural diagrams for further refactoring in mind and I think we should discuss them and hopefully make some decisions that will help us decide which classes some of this new functionality belongs in.

PR #2548, opened 2022-07-18
TransactionBroadcast.broadcast: return future that completes when broadcasts are sent (but not yet relayed by Peers)

msgilligan:
Previously broadcast() only returned when the broadcasts were seen by a certain number of remote peers, we now have two different futures one that completes when the sends are finished and one that completes when the sent transactions are seen.

This should fix Issue #2050 and perhaps some related issues.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java
@@ -20,6 +20,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.Uninterruptibles;
 import org.bitcoinj.core.internal.InternalUtils;
 import org.bitcoinj.core.listeners.PreMessageReceivedEventListener;
+import org.bitcoinj.net.MessageWriteTarget;
 import org.bitcoinj.utils.ListenableCompletableFuture;
 import org.bitcoinj.utils.Threading;
 import org.bitcoinj.wallet.Wallet;
@@ -48,6 +49,10 @@ import static com.google.common.base.Preconditions.checkState;
  */
 public class TransactionBroadcast {
     private static final Logger log = LoggerFactory.getLogger(TransactionBroadcast.class);
+
+    // As of bitcoinj 0.17 broadcast() now returns a future that completes when all broadcast messages were sent
+    // This future completes when all broadcast messages were sent and is equivalent to the future returned by broadcast()
+    private final CompletableFuture<TransactionBroadcast> sentFuture = new CompletableFuture<>();
 
     // This future completes when we have verified that more than numWaitingFor Peers have seen the broadcast
     private final CompletableFuture<Transaction> seenFuture = new ListenableCompletableFuture<>();
@@ -84,8 +89,8 @@ public class TransactionBroadcast {
     public static TransactionBroadcast createMockBroadcast(Transaction tx, final CompletableFuture<Transaction> future) {
         return new TransactionBroadcast(tx) {
             @Override
-            public ListenableCompletableFuture<Transaction> broadcast() {
-                return ListenableCompletableFuture.of(future);
+            public CompletableFuture<TransactionBroadcast> broadcast() {
+                return CompletableFuture.completedFuture(this);
             }
 
             @Override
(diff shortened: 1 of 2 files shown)


schildbach:
Oh sorry, I meant "both types of futures".

msgilligan:
One potentially issue in this commit: If someone uses the new future, they should be aware that it will (current state of the PR) fail if _any_ of the individual peer broadcasts fail. Maybe we want to define an m-of-n threshold for success. Note that prior to this PR there is really no way for a client to know what happened with the individual broadcasts (they fail silently) and the client detects the error when no peers are detected relaying the transaction.

We may want to consider (in a subsequent PR) doing one or more of the following:

1. Create a `TransactionBroadcast` _interface_ that hides some of the details of the current implementation (which could be renamed) and return the interface to clients.
2. Change the various broadcast methods to return `CompletableFuture<TransactionBroadcast>` (rather than a `SendResult` or a non-completed `TransactionBroadcast`) and maybe eliminate `SendResult` altogether.
3. Add information about the number of peers required, attempted, and successful and possibly even individual peer error results.

msgilligan:
Rebased on top of PR #2549 assuming that one will be merged before this one is finalized.

msgilligan:
Two "steps" from this PR have been "cherry picked" into their own PRs:

* PR #2549
* PR #2562

msgilligan:
This has been rebased now that PR #2549 and PR #2562 have been merged.

But I'll rebase one more time when PR #2563 is merged as that simplifies the long lambda that actually sends the broadcasts by a little bit.

schildbach:
https://github.com/bitcoinj/bitcoinj/pull/2563 is merged.

msgilligan:
This is rebased and force-pushed, but I've "cherry picked" two changes into separate PRs that we should review/merge first:

* PR #2579
* PR #2580

msgilligan:
I've thought about this further and realize it is pretty easy to make this change in a backward-compatible way. The next release should have 3 methods:

* `ListenableCompletableFuture <Transaction> broadcast()` - `@Deprecated`, behavior unchanged
* `CompletableFuture<TransactionBroadcast>  broadcastOnly()` - completes with this `TransactionBroadcast` when message is "sent" to _n_ peers.
* `CompletableFuture<TransactionBroadcast>  broadcastAndAwaitRelay()` - completes with this `TransactionBroadcast` when message is "seen" by _m_ peers.

The relationship between the methods should be roughly (some detail is left out here):

```
@Deprecated
ListenableCompletableFuture<Transaction> broadcast() {
   return broadcastAndAwaitRelay().thenApply(broadcast ->  broadcast.tx);
}

CompletableFuture<TransactionBroadcast> broadcastOnly() {
   // Essentially the current code in this PR;
}

CompletableFuture<TransactionBroadcast> broadcastAndAwaitRelay() {
   return broadcastOnly().thenCompose(broadcast -> broadcast.seenFuture());
}
```

msgilligan:
@schildbach - this is rebased, but I think we should discuss https://github.com/bitcoinj/bitcoinj/pull/2548#issuecomment-1198767858 and decide if that is the right approach before I take the next step.

schildbach:
https://github.com/bitcoinj/bitcoinj/pull/2548#issuecomment-1198767858 seems like a good plan to me. I only wonder what "sent to peer" means vs. "sent to OS network buffer"?

msgilligan:
> seems like a good plan to me.

OK, I'll make a new PR that implements a first step toward that goal.

>  I only wonder what "sent to peer" means vs. "sent to OS network buffer"?

Once we have a mechanism to return broadcast transmission results to clients, we can look at the `MessageWriteTarget::writeBytes` implementations and see if there's  way for us to return a future that completes or fails after the data is actually written to the remote-peer. (This is one reason why my "first step" will not expose the new `sentFuture` publicly)

Since `MessageWriteTarget::writeBytes` is async and returns a future, I don't think it will be a breaking behavioral change if it completes _after_ writing the bytes to the remote endpoint.

msgilligan:
Once PR #2583 and PR #2584 are reviewed/merged we can probably just close this one.

msgilligan:
> Once PR #2583 and PR #2584 are reviewed/merged we can probably just close this one.

Those PRs are merged. I just reviewed a diff and I don't see anything left in this PR that we want in master. Closing.

PR #2549, opened 2022-07-18
Peer/TransactionBroadcast: use thenRunAsync() instead of addListener()

msgilligan:
* Use CompletableFuture.thenRunAsync() rather than rely on the converter method ListenableCompletableFuture.addListener().

This is a minor change that reduces our reliance on our adapter class `ListenableCompletableFuture`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -1709,7 +1709,7 @@ public class Peer extends PeerSocketHandler {
             }
             // Ping/pong to wait for blocks that are still being streamed to us to finish being downloaded and
             // discarded.
-            ping().addListener(() -> {
+            ping().thenRunAsync(() -> {
                 lock.lock();
                 checkNotNull(awaitingFreshFilter);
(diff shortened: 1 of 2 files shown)


schildbach:
I wonder if before merging this, the listener is never removed?

msgilligan:
> I wonder if before merging this, the listener is never removed?

This is a Future listener (Guava) / completion-handler (Jdk8) so is inherently a 1-shot and does not need removal.

msgilligan:
This PR short-circuits the following adapter method:

https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/utils/ListenableCompletionStage.java#L31-L34

schildbach:
Quick code analysis reveals we could do the same refactoring on `TransactionBroadcast` as well.

msgilligan:
I ran code analysis and detected this as part of producing PR #2548. PR #2548 eliminates the usage in `TransactionBroadcaster`. I can add that and then rebase #2548. Should I do that?

msgilligan:
Ready for review/merge.

schildbach:
Merged.

PR #2550, opened 2022-07-18
ForwardingService: add nested ForwardingTool class

msgilligan:
Make the command-line tool itself a class. This will make writing integration tests easier, etc.

PR #2551, opened 2022-07-18
ForwardingService: AutoClosable, use try-with-resources, add stopAsync()/awaitTerminated()

msgilligan:
* ForwardingService  implement AutoClosable and use try-with-resources
* Add `kit.stopAsync()` and `kit.awaitTerminated()` in `close()`.

Draft because it depends upon PR #2550. (though it doesn't really need to)


schildbach:
If it doesn't need #2550 then can you base it on master? Makes it easier to merge independently.

schildbach:
I think this can be closed?

PR #2552, opened 2022-07-18
ForwardingService: infer network from address

schildbach:
This means for mainnet and testnet, you don't have to provide the network name.
It will be inferred from the address. For signet and regtest however the network
name is still needed.


schildbach:
Merged.

PR #2431, opened 2022-04-21
Functional-style CoinSelector WIP

msgilligan:
This is work-in-progress. Look at `WalletTool.java` to see how it is simplified.

TBD:
* Feedback from @schildbach 
* Rewrite/simplify existing `CoinSelector` implementations, etc.

PR #2554, opened 2022-07-18
ForwardingService: make the forwarding a safe operation

schildbach:
By using a CoinSelector that selects exactly the outputs that have been sent to us,
we make sure that

* we exactly send the value that has been sent to us, minus fee
* if the incoming transaction became invalid (e.g. due to a re-org),
  our outgoing transaction would become invalid, too

Strictly speaking, after this PR we would not need to wait for confirmations any
more.

TODO maybe extract selector to wallet package, to be used by everyone?

### DIFF
--- a/examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
+++ b/examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
@@ -18,6 +18,7 @@ package org.bitcoinj.examples;
 
 import org.bitcoinj.base.BitcoinNetwork;
 import org.bitcoinj.base.ScriptType;
+import org.bitcoinj.base.Sha256Hash;
 import org.bitcoinj.core.Address;
 import org.bitcoinj.base.Coin;
 import org.bitcoinj.core.Context;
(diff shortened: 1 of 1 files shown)


schildbach:
I've tested this with two concurrent incoming txns that were confirmed in the same block. The value was forwarded with two transaction, each picking the correct corresponding input.

schildbach:
I've cleared two of the TODOs, rebased against master and tested again (successfully). Marking this as ready for review.

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
```diff
@@ -243,4 +250,21 @@ public Address receivingAddress() {
...
+
+        @Override
+        public CoinSelection select(Coin target, List<TransactionOutput> candidates) {
+            var selected =
+                    candidates.stream()
+                            .filter(output -> output.getParentTransactionHash().equals(forwardTxId))
+                            .collect(Collectors.toSet());
+            return new CoinSelection(selected);
+        }
+    }
```
msgilligan, line 269: This can be simplified to a lambda after we complete PR #2431.

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
```diff
@@ -171,6 +177,7 @@ private void coinsReceivedListener(Wallet wallet, Transaction incomingTx, Coin p
                 System.out.printf("Incoming tx has received %d confirmations.\n", confidence.getDepthInBlocks());
                 // Now send the coins onwards by sending the entire contents of our wallet
```
msgilligan, line 178: This comment is incorrect now that we are adding the CoinSelector.

msgilligan (changes requested):

msgilligan (changes requested):

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
msgilligan: This isn't a suggested change, just a comment. The GitHub review UI is a little bit awkward.

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
```diff
@@ -171,6 +177,7 @@ private void coinsReceivedListener(Wallet wallet, Transaction incomingTx, Coin p
...
                 SendRequest sendRequest = SendRequest.emptyWallet(forwardingAddress);
+                sendRequest.coinSelector = new ForwardingCoinSelector(incomingTx);
```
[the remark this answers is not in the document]
msgilligan: This one is, in fact, a suggested change.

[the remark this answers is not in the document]
schildbach: Thanks! I amended the comment.

schildbach:
Merged.

msgilligan:
@schildbach I had a couple more thoughts on this one. I'm glad you merged it, but I think there are some things that can be improved...

PR #2555, opened 2022-07-18
CoinSelection: add a second constructor that doesn't require valueGathered

schildbach:
We'll add the output values up ourselves.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
+++ b/core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
@@ -33,8 +33,23 @@ public class CoinSelection {
     public final Coin valueGathered;
     public final Collection<TransactionOutput> gathered;
 
-    public CoinSelection(Coin valueGathered, Collection<TransactionOutput> gathered) {
-        this.valueGathered = valueGathered;
+    public CoinSelection(Collection<TransactionOutput> gathered) {
+        this.valueGathered = sumOutputValues(gathered);
         this.gathered = gathered;
+    }
+
+    /**
+     * @deprecated use {@link #CoinSelection(Collection<TransactionOutput>)}
+     */
+    @Deprecated
+    public CoinSelection(Coin valueGathered, Collection<TransactionOutput> gathered) {
+        // ignore valueGathered
+        this(gathered);
+    }
+
+    private static Coin sumOutputValues(Collection<TransactionOutput> outputs) {
+        return outputs.stream()
+                .map(TransactionOutput::getValue)
+                .reduce(Coin.ZERO, Coin::add);
     }
 }
--- a/core/src/main/java/org/bitcoinj/wallet/DefaultCoinSelector.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DefaultCoinSelector.java
@@ -61,7 +61,7 @@ public class DefaultCoinSelector implements CoinSelector {
         }
         // Total may be lower than target here, if the given candidates were insufficient to create to requested
         // transaction.
-        return new CoinSelection(Coin.valueOf(total), selected);
+        return new CoinSelection(selected);
     }
 
     @VisibleForTesting static void sortOutputs(ArrayList<TransactionOutput> outputs) {
--- a/core/src/main/java/org/bitcoinj/wallet/KeyTimeCoinSelector.java
+++ b/core/src/main/java/org/bitcoinj/wallet/KeyTimeCoinSelector.java
@@ -58,7 +58,6 @@ public class KeyTimeCoinSelector implements CoinSelector {
     public CoinSelection select(Coin target, List<TransactionOutput> candidates) {
         try {
             LinkedList<TransactionOutput> gathered = new LinkedList<>();
-            Coin valueGathered = Coin.ZERO;
             for (TransactionOutput output : candidates) {
                 if (ignorePending && !isConfirmed(output))
                     continue;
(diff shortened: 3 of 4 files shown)


schildbach:
We could also deprecate+remove the old constructor. It's currently only useful for CoinSelectors which must add up the amount anyway, so it doesn't get determined twice. However CoinSelections very rarely contain more than a handful of outputs.

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
```diff
@@ -33,8 +34,18 @@ public class CoinSelection {
...
+    }
+
     public CoinSelection(Coin valueGathered, Collection<TransactionOutput> gathered) {
         this.valueGathered = valueGathered;
         this.gathered = gathered;
     }
+
+    private static Coin sumOutputValues(Collection<TransactionOutput> outputs) {
+        final AtomicLong sats = new AtomicLong(0);
+        outputs.forEach(output -> sats.addAndGet(output.getValue().value));
```
msgilligan, line 48: This can be done better with a `Stream` IMO (no need for `AtomicLong` as a mutable accumulator). I'll post suggested code in a few minutes.

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
```diff
@@ -33,8 +34,18 @@ public class CoinSelection {
...
     public final Collection<TransactionOutput> gathered;
 
+    public CoinSelection(Collection<TransactionOutput> gathered) {
+        this(sumOutputValues(gathered), gathered);
+    }
+
     public CoinSelection(Coin valueGathered, Collection<TransactionOutput> gathered) {
         this.valueGathered = valueGathered;
         this.gathered = gathered;
     }
```
msgilligan, line 44: I've wondered in the past if this constructor can be eliminated. In some use cases you already have the sum collected, so recalculating for a long `List` of `TransactionOutput` feels wasteful, but maybe simplicity is more important than optimization.

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
```diff
@@ -33,8 +34,18 @@ public class CoinSelection {
...
+
+    private static Coin sumOutputValues(Collection<TransactionOutput> outputs) {
```
msgilligan, line 46: I've also thought adding a `static` `sum()` method to `TransactionOutput` (and `TranactionInput`) might be useful. We're summing them up in many places.

msgilligan (changes requested):

msgilligan:
There's some overlap with https://github.com/bitcoinj/bitcoinj/pull/2431 and this will make me have to rebase it, but that's ok.

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
schildbach: Yes, that's what I'm thinking in https://github.com/bitcoinj/bitcoinj/pull/2555#issuecomment-1188440661 too.

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
schildbach: I'll look for them, but I think unconditional sums are not that often.

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
msgilligan: I just noticed this comment https://github.com/bitcoinj/bitcoinj/pull/2555#issuecomment-1188440661. So maybe we should deprecate the two-arg constructor.

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
msgilligan: Here's the code:
```
        return outputs.stream()
            .map(TransactionOutput::getValue)
            .reduce(Coin.ZERO, Coin::add);
```

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
msgilligan: There's at least a handful. I've done this more than once, but I don't think it ever made it into a pull request.

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
msgilligan: And note that `sats.addAndGet` probably doesn't use the safe add that we're using in `Coin::add`.

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
msgilligan: Actually here's a PR: https://github.com/bitcoinj/bitcoinj/pull/2412

I actually realized that it's better to use `reduce` and `Coin::add` after making that PR. There's also the issue with nullable in the values for `TransactionInput`

schildbach:
Merged.

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
schildbach: Thanks, I merged your code via this PR. However I just noticed this approach also puts a higher load on the garbage collector by creating a `Coin` instance per stream element (if I'm reading this correctly). This might actually be more relevant than the increased runtime, compared to just adding to a long or AtomicLong. Hopefully still not relevant enough…

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
msgilligan: If you look at PR #2412 it avoids the garbage collection hit and creates a single new `Coin` for the total:

```
    public static Coin sum(Collection<TransactionOutput> outputs) {
        return Coin.ofSat(sumSats(outputs));
    }

    public static long sumSats(Collection<TransactionOutput> outputs) {
        return outputs.stream()
                .mapToLong(output -> output.value)
                .reduce(0, LongMath::checkedAdd);
    }
```

core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
msgilligan: If you look at PR #2412 it avoids the garbage collection hit and creates a single new `Coin` for the total:

```
    public static Coin sum(Collection<TransactionOutput> outputs) {
        return Coin.ofSat(sumSats(outputs));
    }

    public static long sumSats(Collection<TransactionOutput> outputs) {
        return outputs.stream()
                .mapToLong(output -> output.value)
                .reduce(0, LongMath::checkedAdd);
    }
```

PR #2556, opened 2022-07-18
ForwardingService: AutoClosable, use try-with-resources

msgilligan:
* ForwardingService  implement AutoClosable and use try-with-resources
* Add `kit.stopAsync()` and `kit.awaitTerminated()` in `close()`.

### DIFF
--- a/examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
+++ b/examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
@@ -41,13 +41,14 @@ import java.util.concurrent.CompletableFuture;
  * ForwardingService demonstrates basic usage of the library. It sits on the network and when it receives coins, simply
  * sends them onwards to an address given on the command line.
  */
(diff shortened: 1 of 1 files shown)


schildbach:
This looks good. Unfortunately I failed to test it. If I insert a `System.out.println()` into the start of `close()` it's never printed. I suspect IntelliJ isn't properly terminating the process. Or, of course, the auto-close mechanism doesn't work.

msgilligan:
Yeah, I actually think auto-close is not getting triggered by IntelliJ and probably wouldn't be with a control-c, either. I think an exception must occur within the main thread while it inside the "try-with-resources" scope.

Not swelling the `InterruptedException` might do the trick.

That doesn't mean we shouldn't use `AutoCloseable` anyway as an example/best practice.

schildbach:
Ok, so I merged. Can't hurt (-:

msgilligan:
PR #2557 improves this a little.

PR #2557, opened 2022-07-19
ForwardingService: Improve the implementation of close()

msgilligan:
* Don't try to remove listener (access .wallet()) unless kit.isRunning()
* Don't wait for termination, just call kit.stopAsync()
* Add some JavaDoc

@schildbach I actually tested this a little bit by adding code to throw interrupts within the _try-with-resources_ block. That's how I figured out to add the `.isRunning()`.


schildbach:
Merged.

PR #2559, opened 2022-07-19
CoinSelection: require a List, not a Collection

msgilligan:
The comment says that it is a "list" and I believe there are CoinSelectors that require an (ordered) list.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
+++ b/core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
@@ -19,6 +19,7 @@ package org.bitcoinj.wallet;
 import org.bitcoinj.base.Coin;
 import org.bitcoinj.core.TransactionOutput;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


schildbach:
I'd say the question is: does it make our lives easier if we pass around lists rather than sets?

msgilligan:
>...  most selectors impose some form of priority (e.g. time) and implement that by sorting a list ...
> 
>  What relevance does the order in a selection have?

As most of the common use cases have some kind of ordering to them and the proposed functional-style `CoinSelector` includes a sort function (see WIP PR #2431), I think simplifying to only `List` is a good idea. There is no reason it can't be an unsorted list in cases where sort-order is unimportant.

In fact I even considered (and rejected) the idea or replacing the `CoinSelection` class with `List<TransactionOutput>`.  `CoinSelection` offers two things over List<TransactionOutput>: (1) it pre-calculates the sum (though it also could be lazily calculated and then cached) and (2) it provides stronger typing than `List` (at least until Java provides reified generics.)

schildbach:
> As most of the common use cases have some kind of ordering to them and the proposed functional-style `CoinSelector` includes a sort function (see WIP PR #2431), I think simplifying to only `List` is a good idea. There is no reason it can't be an unsorted list in cases where sort-order is unimportant.

But do you agree that both the collection of candidates and the selection from these candidates doesn't require a list? It's just the implementation of the selection logic which (not always, but most of times) uses a list? I just want to make sure we're on the same page. I'm confused by the title of this PR – why _must_ this be a List?

msgilligan:
> But do you agree that both the collection of candidates and the selection from these candidates doesn't require a list?

Yes.


> I'm confused by the title of this PR – why _must_ this be a List?

I changed the title. The title is referring to the result of the change: requiring a more specific type.

msgilligan:
> I'd say the question is: does it make our lives easier if we pass around lists rather than sets?

I'm pretty confident the answer is "yes". If we're going to use functional-style and Streams for coin selection using an ordered collection (and also trying to define it is unmodifiable where practical) is going to simplify our lives.

msgilligan:
> One reason to enforce lists is reproducibility (mainly for unit tests).

And (immutable, unmodifiable) `List`s are the collection that maps most easily/logically to an from `Stream`.

msgilligan:
If it helps with the decision, PR #2561 is a logical next step.

msgilligan:
Converting to DRAFT until PR #2561 is merged.

schildbach:
I have this one in my workspace already?

schildbach:
I merged this anyway, as I was already through the final review.

PR #2561, opened 2022-07-19
CoinSelection: encapsulate fields

msgilligan:
* Encapsulate `valueGathered` as `total()`
* Encapsulate `gathered` as `outputs()`
* Deprecate public members
* Update clients to use accessors

This is dependent upon PR #2559 (thought it doesn't strictly need to be)

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
+++ b/core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
@@ -24,14 +24,20 @@ import java.util.Collection;
 import java.util.List;
 
 /**
- * Represents the results of a
- * {@link CoinSelector#select(Coin, List)} operation. A
- * coin selection represents a list of spendable transaction outputs that sum together to give valueGathered.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/bitcoinj/wallet/CoinSelection.java
```diff
@@ -53,4 +60,18 @@ private static Coin sumOutputValues(List<TransactionOutput> outputs) {
                 .map(TransactionOutput::getValue)
                 .reduce(Coin.ZERO, Coin::add);
     }
+
+    /**
+     * @return Total value of gathered outputs.
+     */
+    public Coin total() {
```
schildbach, line 67: `total` seems a bit vague to me. Maybe `totalValue` (or `totalAmount`)?

msgilligan: `.totalValue()` sounds good. I'll change it.

msgilligan (commented):
I'll make the change, rebase, and force-push.

PR #2564, opened 2022-07-20
StreamUtils: new util class with .toUnmodifiableList()

msgilligan:
This will also make it easier to find and upgrade these methods when we upgrade to newer JDKs.

I'm certain there are other methods we will want to add to this utility class.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/utils/StreamUtils.java
+++ b/core/src/main/java/org/bitcoinj/base/utils/StreamUtils.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright by the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/bitcoinj/base/utils/StreamUtils.java
```diff
@@ -0,0 +1,40 @@
...
+package org.bitcoinj.base.utils;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Stream Utilities. bitcoinj is moving towards functional-style programming, immutable data structures, and
```
schildbach, line 26: Nitpick: I think when the word "bitcoinj" is used as first word of a sentence it should read "Bitcoinj", like any other word.

schildbach:
Looks good!

core/src/main/java/org/bitcoinj/base/utils/StreamUtils.java
msgilligan: Fixed and force-pushed.

schildbach:
Removed some (now unused) imports and merged.

PR #2565, opened 2022-07-20
Coin: Improve JavaDoc

msgilligan:
* Add a more extensive explanation of the class purpose and how to convert to and and from other numeric types
* Explain that this is a "Value-based class" and provide a link to the JDK 17 [explanation](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/doc-files/ValueBased.html#Value-basedClasses) of this term.
* Add `@throws` explanations on the two "from BigDecimal" methods.
* Add `throws ArithmeticException` to match the `@throws` (not required since `RuntimeException` but a best practice)


schildbach:
Thanks for improving the documentation! Merged.

PR #2566, opened 2022-07-27
WalletAppKit: remove unnecessary try/catch in startUp()

msgilligan:
Remove the very long try/catch in startup that catches `BlockStoreException` and rethrows it wrapped in an `IOException`.

The Guava service base class (`AbstractIdleService`) that we are extending allows us to throw `Exception` so there is no reason to wrap the exception and it can just be thrown directly.

This is first step toward some further refactoring that will allow refactoring in `ForwardingService` to make it (and other apps that use `WalletAppKit`) more testable.


schildbach:
Merged.

PR #2567, opened 2022-07-27
WalletAppKit: make sure downloadListener is never null

msgilligan:
This simplifies the `if (blockingStartup)` and its `else` block. This change makes way
for further simplification.

Depends on PR #2566


schildbach:
Merged.

PR #2569, opened 2022-07-27
ForwardingService: Use kit.setBlockingStartup(false)

msgilligan:
This will speed up synchronization because the `.setMaxConnections()` call will occur before synchronization is finished and the necessary number of `Peers` will be found sooner.

Also IMO this a better example because most apps should not block while syncing the chain.

### DIFF
--- a/examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
+++ b/examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
@@ -124,9 +124,9 @@ public class ForwardingService implements AutoCloseable {
             kit.connectToLocalHost();
         }
 
-        // Download the blockchain and wait until it's done.
-        kit.startAsync();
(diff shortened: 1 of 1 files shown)


schildbach:
Merged.

We now have some redundant messages at startup though. For reference, here is the output:

```
Network: org.bitcoin.test
Forwarding address: tb1qemczz9rc49y5xzlmsr0r2qmhp9wsv9v3t5dtlv
Waiting to receive coins on tb1qh6wf2njcvzlcjhdzacsvd2kemf40l5dqwwchnq
Will send coins to tb1qemczz9rc49y5xzlmsr0r2qmhp9wsv9v3t5dtlv
Waiting for coins to arrive. Press Ctrl-C to quit.
```

msgilligan:
> We now have some redundant messages at startup though.

Fixed in PR #2575 which has been merged.

PR #2577, opened 2022-07-28
Upgrade JUnit 5 JARs to 5.9.0

msgilligan:
Maybe this will help with the `@TempDir` annotation issues we were seeing on Windows. (it doesn't see PR #2576)

We should upgrade to the latest version anyway.


schildbach:
Merged.

PR #2578, opened 2022-07-28
ForwardingService: static method to create CoinSelector

msgilligan:
Uses partial function application rather than creating a class to create the parent-transaction-only CoinSelector.


schildbach:
Merged.

PR #2579, opened 2022-07-28
TransactionBroadcast: rename private `future` to `seenFuture`, add comment.

msgilligan:
`future` is too generic and since we are planning on adding a second significant future, this one should have a more descriptive name.

Comments are also added to explain its purpose.


schildbach:
Merged.

PR #2580, opened 2022-07-28
TransactionBroadcast: make 2 private members final

msgilligan:
A minor change I made while refactoring, but it's alway helpful to be more clear about what's `final` (and to remove yellow in IntelliJ)


schildbach:
Merged.

PR #2581, opened 2022-07-29
WIP: Remove shutdownHook on shutdown

msgilligan:
WalletAppKit can shutdown without the VM shutting down, when the VM
shuts down later, an exception can occur because the Guava service was already
shut down. At least I think that can happen.

This is a quick stab at a solution. Definitely WIP.


schildbach:
At the moment, I'm missing the means to test this.

msgilligan:
We might need to create some tests that exec a new VM process to test this properly. There may be some libraries for this as well. Let's keep this open until we figure out how to test it and then add tests to the PR.

PR #2582, opened 2022-07-29
Context: re-add constructors that take `params` (as deprecated)

msgilligan:
There are apps and libraries that are using these, we need to keep them as `@Deprecated` for at least one release.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Context.java
+++ b/core/src/main/java/org/bitcoinj/core/Context.java
@@ -62,6 +62,14 @@ public class Context {
     }
 
     /**
+     * @deprecated Use {@link Context#Context()}
+     */
+    @Deprecated
+    public Context(NetworkParameters params) {
+        this();
+    }
+
+    /**
      * Creates a new custom context object. This is mainly meant for unit tests for now.
      *
      * @param eventHorizon Number of blocks after which the library will delete data and be unable to always process reorgs. See {@link #getEventHorizon()}.
@@ -75,6 +83,14 @@ public class Context {
         this.ensureMinRequiredFee = ensureMinRequiredFee;
         this.feePerKb = feePerKb;
         lastConstructed = this;
+    }
+
+    /**
+     * @deprecated Use {@link Context#Context(int, Coin, boolean)}
+     */
+    @Deprecated
+    public Context(NetworkParameters params, int eventHorizon, Coin feePerKb, boolean ensureMinRequiredFee) {
+        this(eventHorizon, feePerKb, ensureMinRequiredFee);
     }
 
     private static volatile Context lastConstructed;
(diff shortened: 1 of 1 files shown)


schildbach:
I evaluated adding the constructors back, but afaict there is no way to save the network params. As evidenced by your PR: In all three constructors, the params are just dropped.

schildbach:
Of course we can decide that this behavior is ok. But then I think we should document it somehow?

msgilligan:
To be honest, I'm still fuzzy as to how all this code really works (and I'm very happy to see that we are in the process of removing it.) 

> As evidenced by your PR: In all three constructors, the params are just dropped.
> Of course we can decide that this behavior is ok.

I got "constructor not found" with the ConsensusJ Bitcoin JSON-RPC client when I tried to run it with the code currently on `master`. If the deprecated constructor which ignores (doesn't save) the network params were present, the code would fail. So, at least in this use case, the answer is that this behavior is not ok.

However, what I really want to do is update ConsensusJ (`cj-bitcoin-jsronrpc' module) so that it can work with bitcoinj 0.16.1 _and_ not break when bitcoinj 0.17 is released. So I need to figure out if there's a way to do what I need with the current API on `master`. Let me check...

msgilligan:
`ForwardingService` has the same issue. It currently contains the following code:

```
        // This line makes the log output more compact and easily read, especially when using the JDK log adapter.
        BriefLogFormatter.init();
```

So `ForwardingService` or a similar app (unless the code uses reflection) has a dilemma: if it uses the no-args constructor it won't work with bitcoinj v0.16.1 and if it uses the `Context(NetworkParameters)` constructor it will break with 0.17.

How can we structure the handling of creating a `Context` so that code can be written that will work with both 0.16.1 and 0.17?

msgilligan:
One idea: we add a new static factory method in an 0.16.2 release:

```
    // Create a Context for an app that does not rely on params being in the context
    public static Context getOrCreateMyAppDoesntExpectParamsInContext(NetworkParameters params) {
```

Maybe there's a better/shorter name 🤣  but the idea is that apps and libraries that want to work with 0.16.2 and later can use this method to initialize a context after making sure they don't rely on having a `NetworkParameters` in the `Context`.

In release 0.16.2 this method store the `NetworkParameters` in the context, for the cases where it is needed by bitcoinj itself.

In release 0.17 this method would be deprecated and would not store the `NetworkParameters` in the context.

When apps/libraries are updated to require 0.17 or late, they can switch to another method.

schildbach:
What about simply adding a warning to the added-back constructors that the param is ignored?

schildbach:
I merged your PR, after adding a warning to each of the constructor JavaDocs.

I think it's enough for now. Apps that create a context this way and then feed it into e.g. Wallet will likely be alerted by the missing constructors there.

PR #2583, opened 2022-07-29
TransactionBroadcast: internally create/complete `sentFuture`

msgilligan:
* Add a private member called `sentFuture` that completes when the broadcast is "sent" (written to buffer) to the broadcast peers
* Update the `broadcast()` method to track success/failure of the individual broadcast messages and update `sentFuture` when the individual messages complete. (This is done by using `thenComposeAsync`/`whenComplete`)
* Add comments and TODOs

PR #2584, opened 2022-07-29
TransactionBroadcast: add broadcastOnly() and broadcastAndAwaitRelay(), deprecate broadcast()

msgilligan:
Add 2 new methods:

* `broadcastOnly()` - returns CF that completes when broadcast "sent" to all Peers
* `broadcastAndAwaitRelay()` - returns CF that completes when broadcast confirmed relayed by _m_ Peers

` broadcast()` is deprecated and calls broadcastAndAwaitRelay() and transforms result back to `Transaction`.

DRAFT -- depends on PR #2583

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerGroup.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerGroup.java
@@ -2122,7 +2122,7 @@ public class PeerGroup implements TransactionBroadcaster {
         // eventually be collected. This in turn could result in the transaction not being committed to the wallet
         // at all.
         runningBroadcasts.add(broadcast);
-        broadcast.broadcast();
+        broadcast.broadcastOnly();
         return broadcast;
     }
 
--- a/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java
@@ -41,7 +41,7 @@ import static com.google.common.base.Preconditions.checkState;
 
 /**
  * Represents a single transaction broadcast that we are performing. A broadcast occurs after a new transaction is created
- * (typically by a {@link Wallet} and needs to be sent to the network. A broadcast can succeed or fail. A success is
+ * (typically by a {@link Wallet}) and needs to be sent to the network. A broadcast can succeed or fail. A success is
  * defined as seeing the transaction be announced by peers via inv messages, thus indicating their acceptance. A failure
  * is defined as not reaching acceptance within a timeout period, or getting an explicit reject message from a peer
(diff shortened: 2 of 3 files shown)


schildbach:
Looks great, but is it possible to test `broadcastAndAwaitRelay()` too?

msgilligan:
> Looks great, but is it possible to test `broadcastAndAwaitRelay()` too?

`broadcastAndAwaitRelay()` is tested in the updated `TransactionBroadcastTest`

schildbach:
> `broadcastAndAwaitRelay()` is tested in the updated `TransactionBroadcastTest`

Sorry, then I probably mean `broadcastOnly()`.

msgilligan:
> > `broadcastAndAwaitRelay()` is tested in the updated `TransactionBroadcastTest`
> 
> Sorry, then I probably mean `broadcastOnly()`.

Well, `broadcastOnly()` is tested indirectly by `broadcastAndAwaitRelay()` (and by any tests that test `PeerGroup.broadcastTransaction()`.)

We can try to add some tests that test `broadcastOnly()` directly. The biggest thing missing IMO is error-handling tests. There is one significant behavior change in this code: the futures returned by `peer.sendMessage(Transaction)` are now aggregated and the `broadcastOnly()` CF will fail if _any_ of the `sendMessage()` futures fail. Since `broadcastAndAwaitRelay()` and the deprecated `broadcast()` are now chained handlers of the `broadcastOnly()` CF, we could see some unexpected behavior.

So there's several things we need to talk about doing:

1. Modifying this code so that it behaves more closely to the previous behavior.
2. Deciding what behavior we actually want. For example: if _zero_ of the broadcasts happen shouldn't we fail right away? Is there some default threshold for failed sends that should cause the "aggregated send" CF to fail? Should we make this threshold configurable?
3. Modify the `TransactionBroadcast` object to provide more diagnostic information
4. Writing tests for various failure cases.

I tried to look at the mock-Peer testing code before and found it very confusing so I might need some help with this.

schildbach:
I think the methods should try as hard as possible to broadcast the tx. Question to me is only for how long they should try. E.g. if the network isn't available for minutes or more. If there is network available, the assumption is that at some time the tx will be broadcast.

There is even the extreme edge case where we broadcast the tx but the first peer we broadcast turns out to be a miner who confirms the txn right away and sends it within a block. I would not count this as a failure, even though we've never seen this transaction from any other peer in its unconfirmed state.

schildbach:
Anyway, I merged this because imho it's a step in the right direction and my testing didn't reveal any regressions.

We could take some of the questions/discussion into future PRs however. Very valid points.

PR #2587, opened 2022-07-31
TransactionBroadcast: refactor loop that broadcasts to each peer

msgilligan:
There are three separate commits that could be reviewed and merged separately or reviewed once and merged without squash.

There is one (almost) behavior change in fe193d6acb8455107c7158c1a15b937226a7d97f -- an exception that was swallowed is now captured in the per-peer `CompletableFuture`. Of course, the per-peer CFs (in `sendFutures`) and the combined CF (`sentFuture`) are still private/inaccessible at this point so it doesn't really change any behavior. Once we capture the per-peer success/failure information we can decide how to report failures in a more logical way.


schildbach:
C'picked fe193d6acb8455107c7158c1a15b937226a7d97f

schildbach:
C'picked ac8aad67a55e909df7c147aa05664fd2aae728be

schildbach:
C'picked 47bd4051c1d5dd25bd9303de6395f30f8ed7a00c

PR #2591, opened 2022-07-31
Context: Have get() fallback to create a Context if missing

msgilligan:


schildbach:
That's what `.getOrCreate()` is meant for.

PR #2594, opened 2022-08-01
Context: introduce `relaxProofOfWork` property

schildbach:
This allows unit-tests to disable enforcement of proof of work, so that fake blocks
can be created and solved in short time.

This PR is working towards getting rid of `UnitTestParams`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Context.java
+++ b/core/src/main/java/org/bitcoinj/core/Context.java
@@ -52,13 +52,14 @@ public class Context {
     final private int eventHorizon;
     final private boolean ensureMinRequiredFee;
     final private Coin feePerKb;
+    final private boolean relaxProofOfWork;
 
     /**
      * Creates a new context object. For now, this will be done for you by the framework. Eventually you will be
      * expected to do this yourself in the same manner as fetching a NetworkParameters object (at the start of your app).
      */
     public Context() {
-        this(DEFAULT_EVENT_HORIZON, Transaction.DEFAULT_TX_FEE, true);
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/bitcoinj/core/Context.java
```diff
@@ -77,22 +78,23 @@ public Context(NetworkParameters params) {
      * @param feePerKb The default fee per 1000 virtual bytes of transaction data to pay when completing transactions. For details, see {@link SendRequest#feePerKb}.
      * @param ensureMinRequiredFee Whether to ensure the minimum required fee by default when completing transactions. For details, see {@link SendRequest#ensureMinRequiredFee}.
      */
-    public Context(int eventHorizon, Coin feePerKb, boolean ensureMinRequiredFee) {
+    public Context(int eventHorizon, Coin feePerKb, boolean ensureMinRequiredFee, boolean relaxProofOfWork) {
```
msgilligan, line 81: `relaxProofOfWork` should be added to JavaDoc

msgilligan (changes requested):

msgilligan:
I'm not happy to see functionality being added to `Context`, I would like to see us making progress towards eliminating it altogether, but if this helps us eliminate `UnitTestParams` it may be worth it.

core/src/main/java/org/bitcoinj/core/Context.java
schildbach: Done.

msgilligan:
> For now, the plan I'm following is getting rid of Context just for the standard usecase.

I think that is a worthy goal and perhaps we can remove `Context` altogether at some point in the future.

PR #2595, opened 2022-08-01
Rename AbstractBitcoinNetParams to BitcoinNetParams

msgilligan:
Temporarily insert a `@Deprecated` shim named `AbstractBitcoinNetParams` into the hierarchy to allow transition.

I don't think the `Abstract` qualifier is necessary in the class name. As I've been doing refactoring around `Network`, `BitcoinNetwork` and the `NetworkParameters` hierarchy, I've been thinking it would be nice to rename `AbstractBitcoinNetParams` to `BitcoinNetParams`. This PR would allow us to do that in a non-breaking way.


schildbach:
Seems sensible. However I'd take the opportunity and rename it to `BitcoinNet**work**Params`.

Also there is a `AbstractBitcoinNetParamsTest` we should rename in parallel.

Seems to me you're expecting the `NetworkParameters` hierarchy not to be removed in this cycle.

msgilligan:
> I'd take the opportunity and rename it to `BitcoinNet**work**Params`.
> 
> Also there is a `AbstractBitcoinNetParamsTest` we should rename in parallel.

Sounds good to me.

> Seems to me you're expecting the `NetworkParameters` hierarchy not to be removed in this cycle.

At a minimum we're going to seriously de-emphasize it, but I don't think we'll be able to remove it from all classes/use-cases. Even if we could do that, we'd want to deprecate and allow apps to migrate a little more smoothly.

My main goal is to remove it from `Address` and move `Address` to `o.b.base`. Next step would be to do the same for `Transaction` and related classes.

msgilligan:
Closing in favor of PR #2603.

PR #2596, opened 2022-08-01
AbstractBitcoinNetworkParams: add `of` method, delegate to it in NetworkParameters

msgilligan:
* Add `of(BitcoinNetwork)` `static` method in `AbstractBitcoinNetParams`
* In `NetworkParameters.of(Network)` delegate to `AbstractBitcoinNetParams` for type `BitcoinNetwork`


schildbach:
Is `of()` still a good name then? It doesn't return the type the method is contained in any more.

msgilligan:
> Is `of()` still a good name then? It doesn't return the type the method is contained in any more.

`AbstractBitcoinNetworkParams.of` returns an `AbstractBitcoinNetworkParams`
`NetworkParameters.of` returns a `NetworkParameters` (at least that's the type signature, the current implementation will actually only return subclasses of `AbstractBitcoinNetworkParams` but that may change in the future)

schildbach:
Merged.

PR #2498, opened 2022-06-26
Move Address and related classes to o.b.base

msgilligan:

PR #2601, opened 2022-08-02
Address: deprecate fromString, replace with AddressParser

msgilligan:
In the future DefaultAddressParser may be replaced by something loaded
via the ServiceLoader mechanism or other dynamically configured mechanism.

In most use cases, the AddressParser instance can be accessed through the Wallet,
which already knows the Network type and in this context validation for network makes
sense.

BitcoinURI allocates its own DefaultAddressParser for now, as do some other tests
and examples that don't have access to a Wallet

This is an iteration on the "Address.fromString" refactoring that is part of PR #2498.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AddressParser.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressParser.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright by the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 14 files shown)


schildbach:
The actual contribution looks good.

My only suggestion is maybe a bit about the commit comment: it would be nice to talk about the actual contribution, not so much about a (potential) future.

core/src/main/java/org/bitcoinj/core/AddressParser.java
```diff
@@ -0,0 +1,55 @@
...
+ * TODO: Move this class to o.b.base
+ */
+public interface AddressParser {
+    /**
+     * Parse an address that could be for any network
+     * @param addressString string representation of address
+     * @return A validated address object
+     * @throws AddressFormatException invalid address string
+     */
+    Address parseAddress(String addressString) throws AddressFormatException;
```
schildbach, line 34: Perhaps rename this to `parseAddressAnyNetwork`?

msgilligan: Yes! I'll change it.

msgilligan: Fixed, rebased, and force-pushed.

msgilligan:
`parseAddressAnyNetwork` definitely improves readability IMO.

schildbach:
Merged.

PR #2602, opened 2022-08-02
DatabaseFullPrunedBlockStore: remove all database backed block stores

schildbach:
All of them are unmaintained and their tests are in the way of refactorings.
MySQL and Postgres had been disabled for a while, and one seems to care.
H2 is tested, but pulls a driver dependency into the test classpath.

The removed unit tests account for 14% or 5s of our build time on my machine.

This PR is a requirement for removing `FullPrunedBlockStore`/`FullPrunedBlockChain` entirely, if so desired. Also, it's the path to removing full verification.


schildbach:
Note: The `MemoryFullPrunedBlockStore` will still be available after merging this change, for now.

msgilligan:
I have mixed feelings about this. I would like to see this stuff moved to a separate module (which would remove the extra dependencies and remove the tests from `core`) and maintained/modernized/improved.

On the other hand, I don't have the time to do this right now and I don't think we have any volunteers.

I suppose someone could pull this code out of the history and put it in a new module at some point in the future.

schildbach:
Yes, I was hesitating for a long time too. However a separate module doesn't help in this case, as it would still need to compile against core any time.

schildbach:
Merged.

PR #2603, opened 2022-08-03
Rename AbstractBitcoinNetParams to BitcoinNetworkParams

msgilligan:
I don't think the `Abstract` qualifier is necessary in the class name. We're also changing `Net` to `Network`.,

This PR does things in a non-breaking way, by temporarily inserting a `@Deprecated` shim class named `AbstractBitcoinNetParams` into the hierarchy to allow transition.

This is an improved version of PR #2595 and replaces it.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -23,7 +23,7 @@ import org.bitcoinj.base.Coin;
 import org.bitcoinj.base.Sha256Hash;
 import org.bitcoinj.base.utils.ByteUtils;
 import org.bitcoinj.core.internal.InternalUtils;
-import org.bitcoinj.params.AbstractBitcoinNetParams;
+import org.bitcoinj.params.BitcoinNetworkParams;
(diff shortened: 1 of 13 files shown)


schildbach:
Merged 900fbb393245d64a0b43d50d5bb2b713926b5e0d, but not 778b99b2022f80217e784e435dc282c62fab60f5 (after conversation).

msgilligan:
> Merged [900fbb3](https://github.com/bitcoinj/bitcoinj/commit/900fbb393245d64a0b43d50d5bb2b713926b5e0d), but not [778b99b](https://github.com/bitcoinj/bitcoinj/commit/778b99b2022f80217e784e435dc282c62fab60f5) (after conversation).

Yeah, 778b99b2022f80217e784e435dc282c62fab60f5 will not work as a shim. And in our conversation we checked some bitcoinj apps to see if they used it and (of the ones we checked) none did. So we're going to do the rename without a shim.

(We could make a shim, but it would have to be inserted between `BitcoinNetworkParams` and all of its subclasses. So given how little used this class is outside of `core`, we're going to skip it.)

PR #2609, opened 2022-08-09
NetworkParameters: make id final

msgilligan:
and initialize it from Network.id()

Note that I removed the following comment:

> This may be null for old deserialized wallets. In that case we derive it heuristically by looking at the port number.

I don't see the code that derives it heuristically, so I'm assuming it's ok to remove the comment and it's ok to make `id` final.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+++ b/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
@@ -83,10 +83,9 @@ public abstract class NetworkParameters {
     protected int majorityWindow;
 
     /**
-     * See getId(). This may be null for old deserialized wallets. In that case we derive it heuristically
-     * by looking at the port number.
+     * See getId()
      */
-    protected String id;
+    protected final String id;
     protected final Network network;
 
     /**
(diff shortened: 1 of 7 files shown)


schildbach:
> I don't see the code that derives it heuristically, so I'm assuming it's ok to remove the comment and it's ok to make `id` final.

The code was introduced via db60bfc731ca76e21cde0e6946c2cc70974f3cdd and reads:

```
        if (id == null) {
            // Migrate from old serialized wallets which lack the ID field. This code can eventually be deleted.
            if (port == 8333) {
                id = ID_PRODNET;
            } else if (port == 18333) {
                id = ID_TESTNET;
            }
        }
```

It was removed already a year later via fbc14c1040510176cf61e197eedd800362bc29ef.

schildbach:
Is f7581906949238709f6499281f4cffb2b3b28816 ready for merge?

msgilligan:
> Is [f758190](https://github.com/bitcoinj/bitcoinj/commit/f7581906949238709f6499281f4cffb2b3b28816) ready for merge?

Yes.

schildbach:
Merged.

PR #2612, opened 2022-08-09
Address/LegacyAddress/SegwitAddress: construct with and store Network (not NetworkParameters)

msgilligan:
* Address/LegacyAddress/SegwitAddress: construct with and store Network (not NetworkParameters)
* Replace static factories that take NetworkParameters
* Add static factory methods that take `Network`
* Deprecate static factory methods that take `NetworkParameters`
* Update all (non-test) usages to use the new methods

Note that we had to disable an alt-net test in LegacyAddressTest (we should bet able re-enable once PR #2608 is merged and we rebase on top of it.)

PR #2613, opened 2022-08-10
LegacyAddress/SegwitAddress: normalize network type

msgilligan:
normalize/canonicalize network member of Address types
 
`LegacyAddress`: map `SIGNET` and `REGTEST` to `TESTNET`
`SegwitAddress`: map `SIGNET` to `TESTNET`

I'm **not sure this is a good idea**, but I wanted to to at least discuss it. (**Update**: I now think it is a good idea.)

Currently...
`LegacyAddress` (Base58) supports encoding two network types in an address: `MAINNET` and `TESTNET`
`SegwitAddress` (Bech32) supports encoding three network types in an address: `MAINNET`,  `TESTNET`, and `REGTEST`.

The current code in `SegwitAddress.fromBech32(Net*, String)` will (probably?) throw an exception if given a RegTest Segwit address string. This is because `Networks.networks` contains the two entries for the networks supported by `LegacyAddress`.

This PR maps the network types that can't really be encoded into an address string into the type that a mapping from string to address would give. This might be considered a change in behavior. Although the existing behavior is a little strange.

This PR depends upon PR #2612, but would be DRAFT anyways because its main purpose is to start a discussion.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/LegacyAddress.java
@@ -20,6 +20,7 @@ package org.bitcoinj.core;
 
 import com.google.common.primitives.UnsignedBytes;
 import org.bitcoinj.base.Base58;
+import org.bitcoinj.base.BitcoinNetwork;
 import org.bitcoinj.base.Network;
 import org.bitcoinj.base.exceptions.AddressFormatException;
 import org.bitcoinj.params.Networks;
@@ -63,11 +64,22 @@ public class LegacyAddress extends Address {
      *            20-byte hash of pubkey or script
      */
     private LegacyAddress(Network network, boolean p2sh, byte[] hash160) throws AddressFormatException {
-        super(network, hash160);
+        super(normalizeNetwork(network), hash160);
         if (hash160.length != 20)
             throw new AddressFormatException.InvalidDataLength(
                     "Legacy addresses are 20 byte (160 bit) hashes, but got: " + hash160.length);
         this.p2sh = p2sh;
+    }
+
+    private static Network normalizeNetwork(Network network) {
+        // LegacyAddress does not distinguish between the different testnet types, normalize to TESTNET
+        if (network instanceof BitcoinNetwork) {
+            BitcoinNetwork bitcoinNetwork = (BitcoinNetwork) network;
+            if (bitcoinNetwork == BitcoinNetwork.SIGNET || bitcoinNetwork == BitcoinNetwork.REGTEST) {
+                return BitcoinNetwork.TESTNET;
+            }
+        }
+        return network;
     }
 
     /**
--- a/core/src/main/java/org/bitcoinj/core/SegwitAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/SegwitAddress.java
@@ -18,6 +18,7 @@ package org.bitcoinj.core;
 
 import com.google.common.primitives.UnsignedBytes;
 import org.bitcoinj.base.Bech32;
+import org.bitcoinj.base.BitcoinNetwork;
 import org.bitcoinj.base.Network;
 import org.bitcoinj.base.ScriptType;
 import org.bitcoinj.base.exceptions.AddressFormatException;
@@ -69,6 +70,17 @@ public class SegwitAddress extends Address {
         this(network, encode(witnessVersion, witnessProgram));
     }
 
+    private static Network normalizeNetwork(Network network) {
+        // SegwitAddress does not distinguish between the SIGNET and TESTNET, normalize to TESTNET
+        if (network instanceof BitcoinNetwork) {
+            BitcoinNetwork bitcoinNetwork = (BitcoinNetwork) network;
+            if (bitcoinNetwork == BitcoinNetwork.SIGNET) {
+                return BitcoinNetwork.TESTNET;
+            }
+        }
+        return network;
+    }
+
     /**
      * Helper for the above constructor.
      */
@@ -92,7 +104,7 @@ public class SegwitAddress extends Address {
      *             if any of the sanity checks fail
      */
     private SegwitAddress(Network network, byte[] data) throws AddressFormatException {
-        super(network, data);
+        super(normalizeNetwork(network), data);
         if (data.length < 1)
             throw new AddressFormatException.InvalidDataLength("Zero data found");
         final int witnessVersion = getWitnessVersion();
--- a/core/src/test/java/org/bitcoinj/core/LegacyAddressTest.java
+++ b/core/src/test/java/org/bitcoinj/core/LegacyAddressTest.java
@@ -25,6 +25,8 @@ import org.bitcoinj.base.exceptions.AddressFormatException;
 import org.bitcoinj.base.utils.ByteUtils;
 import org.bitcoinj.params.MainNetParams;
 import org.bitcoinj.params.Networks;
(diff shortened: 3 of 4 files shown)


msgilligan:
Related discussion from **Bitcoin Core**: https://github.com/bitcoin/bitcoin/issues/12314

msgilligan:
> The current code in `SegwitAddress.fromBech32(Net*, String)` will (probably?) throw an exception if given a RegTest Segwit address string. 

I verified that:
```
var address = SegwitAddress.fromBech32(null, "bcrt1qspfueag7fvty7m8htuzare3xs898zvh30fttu2");
```

will throw:

```
org.bitcoinj.base.exceptions.AddressFormatException$InvalidPrefix: No network found for bcrt1qspfueag7fvty7m8htuzare3xs898zvh30fttu2
```

msgilligan:
This is ready-to-review. After working on PR #2614 and PR #2617, I'm thinking we should probably merge this.

schildbach:
Somehow I wonder if `REGTEST` should be something that needs to be registered if needed, and is by default unregistered/unknown. Just throwing in the idea.

core/src/main/java/org/bitcoinj/core/LegacyAddress.java
```diff
@@ -63,13 +63,24 @@ public class LegacyAddress extends Address {
...
     private LegacyAddress(Network network, boolean p2sh, byte[] hash160) throws AddressFormatException {
-        super(network, hash160);
+        super(normalizeNetwork(network), hash160);
         if (hash160.length != 20)
             throw new AddressFormatException.InvalidDataLength(
                     "Legacy addresses are 20 byte (160 bit) hashes, but got: " + hash160.length);
         this.p2sh = p2sh;
     }
 
+    // LegacyAddress does not distinguish between the different testnet types, normalize to TESTNET
```
schildbach, line 73: I think the comment should be inside the method, just before the `if`.

core/src/main/java/org/bitcoinj/core/SegwitAddress.java
```diff
@@ -66,7 +66,18 @@ public class SegwitAddress extends Address {
      */
     private SegwitAddress(Network network, int witnessVersion, byte[] witnessProgram)
             throws AddressFormatException {
-        this(network, encode(witnessVersion, witnessProgram));
+        this(normalizeNetwork(network), encode(witnessVersion, witnessProgram));
+    }
+
+    // SegwitAddress does not distinguish between the SIGNET and TESTNET, normalize to TESTNET
```
schildbach, line 72: Same here.

schildbach (commented):
The normalization for `SIGNET` seems like a no-brainer to me. Similar for `REGTEST`, except I wonder if we should make registering it a requirement (see my previous comment).

schildbach:
It would be nice if this PR would already contain tests that "persist" the intended behaviour.

core/src/main/java/org/bitcoinj/core/LegacyAddress.java
msgilligan: Will fix.

core/src/main/java/org/bitcoinj/core/SegwitAddress.java
msgilligan: Will fix

msgilligan:
> It would be nice if this PR would already contain tests that "persist" the intended behaviour.

Agreed.

msgilligan:
> The normalization for `SIGNET` seems like a no-brainer to me. Similar for `REGTEST`

In general, I like to avoid global state and try to avoid it wherever possible.  I'm hoping we can move away (as much as possible) from a global list of networks.  See PR #2617.

At present we have three (that I know of) use cases for the `Networks.networks` global variable:

1. For parsing address strings (e.g. having a list of valid networks and the "HRP" and "version byte" info for each)
2. Parsing `DumpedPrivateKey` (very similar to address use case)
3. Mapping from `Network` to `NetworkParameters` (in `NetworkParameters.of(Network)`) for alt-networks. 

It's also worth considering that for the three use-cases the list of networks is different:

In (1) the list of networks differs for Segwit and Legacy addresses (See PR #2617). This list could also vary with context (e.g. one wallet/keychain/window may support different networks than another.)

In (2) we are only concerned about Base58 (right?) and the list is probably the same as the Base58 list from (1).

In (3) the list should include every valid `Network`/`NetworkParameters`: If we have a `Network`, we should be able to map it to a `NetworkParameters` (we can't provide the mapping as a property of `Network` for dependency reasons). For this use case there is no harm -- assuming a small number (less than 10, say) of registered networks -- in never unregistering them.

So my proposal is to do the following (PR #2617 is the first step in this direction):

1. Add 2 network lists to `DefaultAddressParser` that can be overridden in the constructor. Provide a deprecated mechanism to build `DefaultAddressParser` from `Networks` for backward-compatibility.
2. Create similar functionality (base58-only) to `DefaultAddressParser` for  `DumpedPrivateKey` 
3. Add a `static` `registerNetwork(NetworkParameters`) to `NetworkParameters` and use that list internally for `NetworkParameters.of()` (and ideally we'd figure out a way to migrate away from this entirely)
4. Deprecate `Networks`.

Does this make sense? Am I missing anything?

schildbach:
I've got nothing to add to cases (1) and (2).

Case (3) was just introduced because of the dependency reasons, so theroretically we could do away with it by finding another solution for the dependency problem. E.g. we manage to move `NetworkParameters` to `base`, or do away with `NetworkParameters`.

Fwiw, case (3) also has the anomaly that it's not a bi-directional mapping if you think of the Unit-Test "network".

Your proposal seems reasonable to me, but I also don't have the entire picture in my mind. I think it's not unlikely we'll run into some sub-issue while trying to fix the issue.

schildbach:
> 3\. Add a `static` `registerNetwork(NetworkParameters`)

However I think we should at least provide an `unregisterNetwork` too. When it comes to unit-tests, there are some different mocks being used, derived from UnitTestParams/TestNetParams.

core/src/main/java/org/bitcoinj/core/SegwitAddress.java
msgilligan: Fixed.

msgilligan:
> It would be nice if this PR would already contain tests that "persist" the intended behaviour. Basically what #2614 has, just with the parts marked with "passes, but shouldn't" adapted.

Added! (Adding the tests actually found a bug -- which I fixed.

This should be ready to merge.

schildbach:
Reviewed again and looks good.

One final thought: Are you sure we don't want to normalize for the comparison functions only?

Consider an app that works with mixed network and shows a MAINNET, TESTNET or SIGNET label alongside with each address, conveniently derived from Address.network(). After this PR, this label will read TESTNET in lots of cases, even if its a different network – right?

msgilligan:
> One final thought: Are you sure we don't want to normalize for the comparison functions only?

I'm sure (though I could be missing something.)

> Consider an app that works with mixed network and shows a MAINNET, TESTNET or SIGNET label alongside with each address

I don't think there is such an app, and if there were, there really is no such thing as a SIGNET address, right?

> After this PR, this label will read TESTNET in lots of cases, even if its a different network – right?

It would, but the address really is a TESTNET address, rather than a SIGNET address or a REGTEST (base58) address.

schildbach:
Ok thanks. Merged.

PR #2615, opened 2022-08-10
base/package-info.java: Add the "deprecation exception" rule to JavaDoc comment

msgilligan:
> In the 0.17 release, we are allowing some dependencies on other packages, e.g. to `org.bitcoinj.core.NetworkParameters` provided that those references are in **deprecated** methods. This will  smooth migration by allowing users to, for example, replace `import org.bitcoinj.base.Address` with `import org.bitcoinj.core.Address` as a first step of conversion and remove usages of the deprecated methods in a second step.


schildbach:
Merged.

PR #2620, opened 2022-08-12
PrefixedChecksummedBytes: rename to EncodedPrivateKey and update class-level JavaDoc

schildbach:
After 1d5e752917db7b2a0ef6e8cc767876f267e7f044, the hierarchy only has private keys:

```
PrefixedChecksummedBytes (now EncodedPrivateKey)
- DumpedPrivateKey
- BIP38PrivateKey
```

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/Base58.java
+++ b/core/src/main/java/org/bitcoinj/base/Base58.java
@@ -28,7 +28,7 @@ import java.util.Arrays;
  * <p>
  * Note that this is not the same base58 as used by Flickr, which you may find referenced around the Internet.
  * <p>
- * You may want to consider working with {@code org.bitcoinj.core.PrefixedChecksummedBytes} instead, which
+ * You may want to consider working with {@code org.bitcoinj.core.EncodedPrivateKey} instead, which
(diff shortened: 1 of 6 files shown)


schildbach:
I decided to prefix the class name with `Encoded` to avoid confusion with low-level keys.

msgilligan:
> I decided to prefix the class name with `Encoded` to avoid confusion with low-level keys.

I assume the `Encoded` as in Base58-encoded, not encrypted. It might be misleading (at first) to some...

msgilligan (approved):

schildbach:
The `Prefix` prefix would be misleading as well, because in fact there is no network-specific prefix at all. I was going to add this information to the other issue, but `DumpedPrivateKey` ignores the network, and `BIP38PrivateKey` encodes it deeply within its custom format (see source).

I specifically made the decision to not bake "base58" into the base class, because I'm hoping at some point we'll get a Bech32 (or better) based private key format.

I think the term "encoded" is usually just means transforming some data into a set of symbols (or readable characters in this case). I cannot remember seeing it in the context of encryption, except for the fact that ciphertext usually needs to be encoded if transferred via text.

schildbach:
Merged.

PR #2621, opened 2022-08-12
PrivateKey: provide abstract .toBase58() for existing DumpedPrivateKey and BIP38PrivateKey implementations

schildbach:
This PR depends on #2620, but could be rewritten to do without.


schildbach:
Closing this, as such an abstract method would be an obstacle for future (e.g. Bech32-based) private key formats.

PR #2622, opened 2022-08-12
EncodedPrivateKeyTest: move a test with non-standard version byte 0 to Base58EncodeCheckedTest

schildbach:
This allows us to get rid of mocking in EncodedPrivateKeyTest.

This PR depends on #2620 (merged), but could be rewritten to do without.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/Base58EncodeCheckedTest.java
+++ b/core/src/test/java/org/bitcoinj/core/Base58EncodeCheckedTest.java
@@ -26,6 +26,7 @@ import org.junit.runners.Parameterized.Parameters;
 import java.util.Arrays;
 import java.util.Collection;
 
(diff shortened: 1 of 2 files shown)


msgilligan:
Why not go one step further and eliminate the `stringification()` test altogether?

schildbach:
I think it's good to test the code path at least once, even if it's just a tiny wrapper around the method in Base58.

msgilligan (approved):

msgilligan:
> I think it's good to test the code path at least once, even if it's just a tiny wrapper around the method in Base58.

OK. This isn't that important because hopefully we'll be making further changes to how this works.

schildbach:
Looking at it again, you're right. It's only testing the codepath in the `EncodedPrivateKeyToTest` mock.

I moved both tests and merged.

PR #2625, opened 2022-08-16
PeerGroup: migrate constructors to Network from NetworkParameters

msgilligan:
* Add 3 new constructors which take network
* Deprecate (2 of 3) constructors that take NetworkParameters
* Mark 3-arg NetworkParameters constructor as @VisibleForTesting

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerGroup.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerGroup.java
@@ -26,6 +26,7 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.Runnables;
 import com.google.common.util.concurrent.Uninterruptibles;
 import net.jcip.annotations.GuardedBy;
+import org.bitcoinj.base.Network;
 import org.bitcoinj.core.listeners.AddressEventListener;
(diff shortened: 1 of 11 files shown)


schildbach:
If the internals of the class still work with NetworkParams, I think we don't win much here.

Imho we should start migrating from the bottom of the class hierarchy, meaning the Message hierarchy first (probably, I haven't extensively researched), then Peer and PeerDiscoveries, then PeerGroup.

msgilligan:
> If the internals of the class still work with NetworkParams, I think we don't win much here.

We are trying to convert the example apps to using `Network` to specify what network they are running on. In general the example apps use either `WalletAppKit` (if they use a wallet) or `PeerGroup` if they use a `PeerGroup`. What led me to make this PR was that I was reviewing `PeerMonitor`.

> 
> Imho we should start migrating from the bottom of the class hierarchy, meaning the Message hierarchy first (probably, I haven't extensively researched), then Peer and PeerDiscoveries, then PeerGroup.

Overall, I agree with this strategy, but I also think we should make sure the major/common API classes support `Network`.

schildbach:
Merged!

PR #2626, opened 2022-08-16
EncodedPrivateKey: Replace NetworkParameters with Network

msgilligan:
* Replace NetworkParameters with Network
* new constructor with `Network` and `int version`
* Remove old constructor (POTENTIALLY BREAKING)
* Mark `network` as transient (for EqualsVerifier, etc)
* Update subclasses to work correctly with new constructor and changed members
* Update tests

PR #2627, opened 2022-08-16
EncodedPrivateKey: migrate from NetworkParameters to Network

schildbach:


core/src/main/java/org/bitcoinj/core/EncodedPrivateKey.java
```diff
@@ -26,31 +28,46 @@
...
+    }
+
     /**
      * @return network this data is valid for
+     * @deprecated Use {@link #network()}
      */
+    @Deprecated
     public final NetworkParameters getParameters() {
-        return params;
+        return NetworkParameters.of(network);
```
msgilligan, line 58: This adds an indirect dependency on `Networks.get()`

msgilligan (approved):
I approve this PR as it follows our general pattern of moving from `NetworkParameters` to `Network`.

I had proposed PR #2626 which tried to move away from using any idea of "network" in `EncodedPrivateKey` (instead storing the `version` byte) and potentially making it easier for us to replace the dependency on `Networks.get()`, but those issues can be dealt with after this PR is merged.

core/src/main/java/org/bitcoinj/core/EncodedPrivateKey.java
schildbach: Yes, but it's deprecated so it will go away in the next release.

schildbach:
Merged.

Issue #703, opened 2015-06-11
null pointer handling in PeerAddress


mikehearn:
Originally reported on Google Code with ID 337

```
PeerAddress.addr and .services can be null.

However, this will lead to NPE in toString(), equals() and hashCode().

Here is one example of a user crash report:

java.lang.NullPointerException
at com.google.bitcoin.core.PeerAddress.hashCode(PeerAddress.java:248)
at java.util.HashMap.put(HashMap.java:390)
at java.util.HashSet.add(HashSet.java:95)
at com.google.bitcoin.core.PeerGroup.connectToAnyPeer(PeerGroup.java:479)
at com.google.bitcoin.core.PeerGroup.startUp(PeerGroup.java:508)
at com.google.common.util.concurrent.AbstractIdleService$1$1.run(AbstractIdleService.java:43)
at java.lang.Thread.run(Thread.java:856)
```

Reported by `andreas.schildbach` on 2013-03-06 19:56:20

msgilligan:
Null-handling is causing problems (NPE) for I2P addresses which are currently being parsed so that `addr` and `hostname` are both `null`.

msgilligan:
PR #2632 should resolve this issue.

msgilligan:
@schildbach I think we can close this now that PR #2632 is merged.

PR #2632, opened 2022-08-18
PeerAddress: clean up toString, equals, hashCode - handle nulls properly

msgilligan:
* Include `hostname` in equals and hashcode
* Fix NPE in toString() that occurs for (unsupported) I2P addresses (both addr and hostname are null)
* Better null-handling in equals
* Add EqualsVerify test
* Add comments

Addresses Issue #703


schildbach:
Great catch, thanks! Merged.

I guess this would be a candidate for backporting.

msgilligan:
> I guess this would be a candidate for backporting.

Is there a `backport` label on Github or some way of marking candidates for backporting?

PR #2633, opened 2022-08-18
PeerAddress: make onionChecksum static

msgilligan:
This onionChecksum method uses no instance variables and has no side-effects, it should be static.


schildbach:
Merged.

PR #2634, opened 2022-08-18
PeerAddress: create NetworkId enum, define I2P, CJDNS network IDs

msgilligan:
* Replace hardcoded integer values (e.g. `0x01`) with NetworkId enum.
* Define (currently) unsupported values: I2P(5), CJDNS(6)
* Use switch statement to handle Network IDs
* Add placeholder cases for I2P and CJDNS


schildbach:
Merged.

PR #2635, opened 2022-08-18
PeerAddressTest: add parameterized deserializeToString test

msgilligan:
* build.gradle: Add JUnitParams so we can have parameterized tests in existing test
* PeerAddressTest: Add deserializeToString method with three addresses to test


schildbach:
Merged.

PR #2638, opened 2022-08-18
PeerMonitor: Use CompletableFuture for lookupReverseDNS() & getAddr()

msgilligan:
* Create threadpool for reverseDns (rather than use new Thread() repeatedly)
* For getAddr() run completion handler on callback thread (and eliminate new Thread())
* Add getHostName() method that handles case where peerAddress.getAddr() is null
* Simplify getAddressForPeer() with trinary operator.


schildbach:
Merged via 39d799e58fbc29f565ace4084de70c7a75d4ad1f and 94ad7dd344490ce0c11921a076e78e1078dc3334.

PR #2640, opened 2022-08-19
PeerAddress: Add I2P hostname parsing (WIP)

msgilligan:
Note: this is not verified as correct encoding yet.

Depends on PR #2634


msgilligan:
This is rebased and ready-to-merge. I don't know if the deserialization is correct, though. I'm hoping someone can review it and tell me.

core/src/main/java/org/bitcoinj/core/PeerAddress.java
```diff
@@ -282,6 +282,11 @@ protected void parse() throws ProtocolException {
                         addr = null;
                         break;
                     case I2P:
+                        if (addrLen != 32)
+                            throw new ProtocolException("invalid length of I2P address: " + addrLen);
+                        hostname = BASE32.encode(addrBytes).replace("=","") + ".b32.i2p";
```
schildbach, line 287: Why not use `omitPadding()` instead of the `.replace()`?

schildbach (commented):
Looks fine, but we should add some tests – even if we make the vectors up ourselves.

core/src/main/java/org/bitcoinj/core/PeerAddress.java
msgilligan: > Why not use `omitPadding()` instead of the `.replace()`?

Because I didn't know it existed 🤦‍♂️

Also, I didn't know if any changes that affected the other address types should be made.

schildbach: If you use the `.omitPadding()` only on "your" code of block, it will stay local to that. Of course, #2642 suggests to use it in general in PeerAddress – also for `.onion`.

PR #2642, opened 2022-08-24
PeerAddress: addresses that make use of base32 in hostnames generally do not use padding, so omit it

schildbach:


msgilligan (approved):
Looks good to me!

schildbach:
Merged.

PR #2645, opened 2022-09-19
Docs: Fix invalid option for value --net

maksumic:


msgilligan (approved):
Approved. `TEST` is no longer a valid option for `--net`, but `TESTNET` is.

schildbach:
Good catch! Merged.

PR #2649, opened 2022-10-20
P2WPKH fallback for P2PKH Script

abhaygupta1999:
When we create a wallet with one keychain i.e of P2WPKH, and if we want to add the support for P2PKH for that same wallet. So, this code contains only fallback when for P2PKH it fails.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/Bech32.java
+++ b/core/src/main/java/org/bitcoinj/base/Bech32.java
@@ -144,6 +144,12 @@ public class Bech32 {
         return sb.toString();
     }
 
+    /** @deprecated use {@link #encode(Encoding, String, byte[])} */
+    @Deprecated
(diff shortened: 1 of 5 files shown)


JAYP1255:
> When we create a wallet with one keychain i.e of P2WPKH, and if we want to add the support for P2PKH for that same wallet. So, this code contains only fallback when for P2PKH it fails.

Awesome thanks

schildbach:
Why don't you create a `P2PKH` wallet to begin with? That way, your wallet will support `P2PKH`, `P2WPKH` out of the box, and in future `P2TR` (and any script type suitable for simple payments that might be invented in future).

By design, keychains are separated by script type. Unless there is a compelling reason, I'd like to keep it that way.

Btw. your PR seems to be based on an old release. Generally we prefer PRs rebased to master – it's easier to review.

PR #2659, opened 2022-11-21
SeedPeers: add a constructor for `InetSocketAddress`

schildbach:
The old `int` format for peer addresses is now deprecated, as it doesn't support IPv6.


schildbach:
We should probably also delete the `MainNetParams.addrSeeds` blob, as almost all of the seeds are disconnected and frankly I think we should not maintain this.

schildbach:
Merged.

PR #2663, opened 2023-01-30
SegwitAddress: include length check for Taproot witness programs

johnzweng:
Hi @schildbach ! I have another PR, this time a very little change, but a little bit more reasoning behind it. So I'm very curious to hear your feedback. 🙂 

## Overview

- This PR just adds a size check for Segwit v1 (Taproot) addresses to the `SegwitAddress` constructor so that it only accepts 32-byte long witness programs (the only valid length for Taproot as defined in [BIP 341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#script-validation-rules)). 
- Adding this check means that `AddressParser.parseAddress(...)` will in future throw an exception for too short or too long Segwit v1 addresses (as it already does today for Segwit v0 addresses with invalid lengths). Currently too short/long Taproot addresses are accepted as valid by `AddressParser.parseAddress(...)`.

## Current state

- Currently in bitcoinJ there [is already a check for witness program length for Segwit v0](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/core/SegwitAddress.java#L117-L120) in the constructor. So it is impossible to instantiate a v0 `SegwitAddress` with an invalid length.
- This is not the case for Taproot addresses (Segwit v1), so one can instantiate a `SegwitAddress` instance with too short or too long Taproot addresses without exception.
- I am aware of the fact that there is a semantic difference in v0 an v1 addresses with invalid lengths:
  - **Segwit v0:** [BIP 141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#witness-program) specifies that all v0 witness programs with invalid length (not 20 or 32 bytes) MUST fail (_"the script must fail"_). This means: any coins sent to such an address **will become unspendable**.
  - **Segwit v1**: This is different from v0: [BIP 341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#script-validation-rules) specifies that Taproot validation rules only apply to witness programs with a length of 32 bytes. This implies that if the witness program has any other length, then no rules will be applied to v1 witness scripts. In short: sending coins to too long or too short Taproot addresses **will become anyonce-can-spend** (unless changed in a future soft fork).
  - These BIP descriptions match the code in Bitcoin Core, where invalid length v0 scripts will immediately fail [here in the else branch](https://github.com/bitcoin/bitcoin/blob/58da1619be7ac13e686cb8bbfc2ab0f836eb3fa5/src/script/interpreter.cpp#L1880-L1902) and invalid v1 lengths are just not covered by [this if-condition (which includes length)](https://github.com/bitcoin/bitcoin/blob/58da1619be7ac13e686cb8bbfc2ab0f836eb3fa5/src/script/interpreter.cpp#L1903) so that they end up below in the [default else branch returning always `true`](https://github.com/bitcoin/bitcoin/blob/58da1619be7ac13e686cb8bbfc2ab0f836eb3fa5/src/script/interpreter.cpp#L1950-L1951).
- Nevertheless, v1 addresses with invalid lengths are still covered by transaction policy by marking them as non-standard (the flag `SCRIPT_VERIFY_DISCOURAGE_UPGRADABLE_WITNESS_PROGRAM` is set [here in `policy.h`](https://github.com/bitcoin/bitcoin/blob/306ccd4927a2efe325c8d84be1bdb79edeb29b04/src/policy/policy.h#L72-L90), which will lead to [returning an error on validation here](https://github.com/bitcoin/bitcoin/blob/58da1619be7ac13e686cb8bbfc2ab0f836eb3fa5/src/script/interpreter.cpp#L1947-L1949)). This makes it more difficult to spend such an invalid Taproot anyone-can-spend address as nodes will not accept such transactions in their mempool by default.

## Rationale

I was unsure if this PR is reasonable (and you might still argue about it) as BIP 351 not explicitly marks these scripts as invalid. But from a practical point of view I still would argue that we should treat them as invalid addresses. Implementors of Bitcoin services (ATMs, exchanges) use `AddressParser.parseAddress(...)` as an easy check to see if a given address is valid. And while invalid-length-v1 addresses technically still might be valid I think in most cases it is not what the user intends.

So if there are buggy wallets out there which create crap addresses (example: v1 addresses with length 20 like v0), it would be nice if such addresses would raise attention (to prevent services from just blindly sending coins there). Many callers of `AddressParser.parseAddress(...)` may not dive that deep into these semantics and just use it as a simple validation check (without the need for a deeper understanding).

I also understand if you don't agree with my reasoning and don't accept this PR. But then maybe we could add some "dangerous" flag, so that validation fails by default on such addresses (and also currently unused Segwit versions) but instantiation of such addresses can still be enforced by providing the "dangerous" flag (or something similiar).

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/SegwitAddress.java
+++ b/core/src/main/java/org/bitcoinj/core/SegwitAddress.java
@@ -113,11 +113,28 @@ public class SegwitAddress extends Address {
         byte[] witnessProgram = getWitnessProgram();
         if (witnessProgram.length < WITNESS_PROGRAM_MIN_LENGTH || witnessProgram.length > WITNESS_PROGRAM_MAX_LENGTH)
             throw new AddressFormatException.InvalidDataLength("Invalid length: " + witnessProgram.length);
-        // Check script length for version 0
+        // Check script length for version 0:
+        // BIP 141:
+        // "If the version byte is 0, but the witness program is neither 20 nor 32 bytes, the script must fail."
+        // In other words: coins sent to addresses with other lengths will become unspendable.
         if (witnessVersion == 0 && witnessProgram.length != WITNESS_PROGRAM_LENGTH_PKH
                 && witnessProgram.length != WITNESS_PROGRAM_LENGTH_SH)
             throw new AddressFormatException.InvalidDataLength(
                     "Invalid length for address version 0: " + witnessProgram.length);
+        // Check script length for version 1:
+        // BIP 341:
+        // "A Taproot output is a native SegWit output (see BIP141) with version number 1, and a 32-byte
+        // witness program. Any other outputs, including version 1 outputs with lengths other than 32 bytes,
+        // or P2SH-wrapped version 1 outputs, remain unencumbered."
+        // In other words: other lengths are not valid Taproot scripts but coins sent there won't be
+        // unspendable, quite the contrary, they will be anyone-can-spend. (Not that easy spendable, because still
+        // not-standard outputs and therefore not relayed, but a willing miner could easily spend them.)
+
+        // Rationale for still restricting length here: creating anyone-can-spend Taproot addresses is probably
+        // not that what callers expect.
+        if (witnessVersion == 1 && witnessProgram.length != WITNESS_PROGRAM_LENGTH_TR)
+            throw new AddressFormatException.InvalidDataLength(
+                    "Invalid length for address version 1: " + witnessProgram.length);
     }
 
     /**
--- a/core/src/test/java/org/bitcoinj/core/SegwitAddressTest.java
+++ b/core/src/test/java/org/bitcoinj/core/SegwitAddressTest.java
@@ -162,8 +162,6 @@ public class SegwitAddressTest {
                     "0014751e76e8199196d454941c45d1b3a323f1433bd6", 0),
             new AddressData("tb1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcccefvpysxf3q0sl5k7", TESTNET,
                     "00201863143c14c5166804bd19203356da136c985678cd4d27a1b8c6329604903262", 0),
-            new AddressData("bc1pw508d6qejxtdg4y5r3zarvary0c5xw7kw508d6qejxtdg4y5r3zarvary0c5xw7kt5nd6y", MAINNET,
-                    "5128751e76e8199196d454941c45d1b3a323f1433bd6751e76e8199196d454941c45d1b3a323f1433bd6", 1),
             new AddressData("BC1SW50QGDZ25J", MAINNET, "6002751e", 16),
             new AddressData("bc1zw508d6qejxtdg4y5r3zarvaryvaxxpcs", MAINNET, "5210751e76e8199196d454941c45d1b3a323", 2),
             new AddressData("tb1qqqqqp399et2xygdj5xreqhjjvcmzhxw4aywxecjdzew6hylgvsesrxh6hy", TESTNET,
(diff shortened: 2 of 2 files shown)


johnzweng:
Note: I just force-pushed a changed commit again, I had overlooked an existing test in `SegwitAddressTest` which failed with my new check. It should work now.

core/src/main/java/org/bitcoinj/core/SegwitAddress.java
```diff
@@ -113,11 +113,28 @@ private SegwitAddress(Network network, byte[] data) throws AddressFormatExceptio
...
+        // or P2SH-wrapped version 1 outputs, remain unencumbered."
+        // In other words: other lengths are not valid Taproot scripts but coins sent there won't be
+        // unspendable, quite the contrary, they will be anyone-can-spend. (Not that easy spendable, because still
+        // not-standard outputs and therefore not relayed, but a willing miner could easily spend them.)
+
+        // Rationale for still restricting length here: creating anyone-can-spend Taproot addresses is probably
+        // not that what callers expect.
+        if (witnessVersion == 1 && witnessProgram.length != WITNESS_PROGRAM_LENGTH_TR)
+            throw new AddressFormatException.InvalidDataLength(
+                    "Invalid length for address version 1: " + witnessProgram.length + ". Resulting address would be anyone-can-spend.");
```
schildbach, line 137: I think I'd remove the second sentence here. To me, it feels too verbose for an exception message.

schildbach (commented):
Thanks for this PR!

I think the intention of BIP 341 to not strongly forbid non-32-byte programs is extensibility. They want to be able to softfork future script types.

But you're right, with bitcoinj being used as a client (rather than a full bitcoin node), we should probably follow policy more closely than consensus – we've done so in other areas as well. This would mean that if a new script type comes up and a consumer of bitcoinj wants to use it, the consumer would be forced to upgrade to a newer bitcoinj (if it already exists). But I guess that's ok.

So I agree with this PR.

I've added one review comment inline, but it's a really minor one. Thanks specifically for the intensive documentation and rationale – it really helps with reviewing!

core/src/main/java/org/bitcoinj/core/SegwitAddress.java
johnzweng: @schildbach I just force-pushed the change and removed the 2nd sentence.

johnzweng:
> Thanks specifically for the intensive documentation and rationale – it really helps with reviewing!

You're welcome! 🙂  I know from a reviewer's perspective how valuable this is.
(Also this makes it easier for other readers to follow the reasoning behind changes.)

schildbach:
Side note: I think segwit v0 used to be extensible too, but they found [a vulnerability](https://github.com/sipa/bech32/issues/51) in the bech32 format. The existing 20 and 32 byte programs were luckily not affected, so the fix was to "close" v0 (restrict it to these 2 usages) and add a new fixed and extensible v1 + bech32m.

johnzweng:
> Side note: I think segwit v0 used to be extensible too, but they found [a vulnerability](https://github.com/sipa/bech32/issues/51) in the bech32 format. The existing 20 and 32 byte programs were luckily not affected. So the fix was to "close" v0 by restricting it to these 2 usages, and add a new fixed and extensible v1 + bech32m.

Interesting learning of Bitcoin history. :) I was aware that bech32m was introduced because of this vulnerability, but didn't know that v0 had been planned to be extensible.

schildbach:
Merged! Let's see if someone complains about future script types…

johnzweng:
> Let's see if someone complains about future script types…

Aaaaand here come: `bc1pfeessrawgf` entered the chat.. 😄

johnzweng:
I am impressed how much love was put into every detail of this specific transaction. 😃

Later on the creator also published a stacker.news article where he reveleaed even more details: https://stacker.news/items/600187 

Mindblowing! 😊

PR #2664, opened 2023-01-30
NativeSecp256k1: remove unused native methods

schildbach:
This should make it clearer which methods we actually need and therefore need to fix.

Note this also removes a Schnorr-related method, which seems odd since Schnorr is likely one of our goals. However it only partially supported  "experimental Schnorr" which has since been removed from libsecp256k1 (and re-added as a module later).

…if we don't remove support for libsecp256k1 entirely, see #2267.


schildbach:
We merged #2267, so this can be closed.

PR #2665, opened 2023-02-01
Remove recached

micheal-swiggs:
@schildbach Was having a browse through the code and noticed that `Message.recached` was unused so decided to remove it.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Message.java
+++ b/core/src/main/java/org/bitcoinj/core/Message.java
@@ -138,10 +138,6 @@ public abstract class Message {
         return payload != null;
     }
 
-    public boolean isRecached() {
-        return recached;
-    }
-
     /**
      * Overrides the message serializer.
      * @param serializer the new serializer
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/core/Message.java
```diff
@@ -138,10 +136,6 @@ public boolean isCached() {
         return payload != null;
     }
 
-    public boolean isRecached() {
```
schildbach, line 141: Maybe we would like to deprecate this for one release cycle, rather than remove it immediately?

schildbach (commented):
Thanks for your contributions!

I merged the `hasMoreBytes()` commit, which seemes independent from this PR. The rest could be squashed and rebased, I guess. I added one review comment.

core/src/main/java/org/bitcoinj/core/Message.java
schildbach: In general, we try to deprecate for one cycle.

I merged the `hasMoreBytes` removal without deprecation because I consider it unlikely that someone has extended the message framework *and* has used this method. I may be wrong, so if someone complains we can re-add it back in deprecated form.

PR #2667, opened 2023-02-01
Removes unused method Peer.isNotFoundMessageSupported

micheal-swiggs:
@schildbach & @msgilligan removes another unused method. This is the only place where `NotFoundMessage.MIN_PROTOCOL_VERSION` is used so perhaps remove that too?


schildbach:
Oh yes, that should probably have gone into ccbd30da8f96a247143514f7e860955904cea291.

schildbach:
Merged.

PR #2668, opened 2023-02-02
Removes redundant cursor statement

micheal-swiggs:
@schildbach Clean clean.


schildbach:
I think it was left as a reminder to increment the cursor if more fields will ever be appended to the structure.

But yes, it's clearly redundant.

schildbach:
Merged.

Issue #2531, opened 2022-07-14
Where to keep `TxConfidenceTable` if not in Context?


schildbach:
`TxConfidenceTable` currently is the point of truth for transaction confidences. It is meant to exist only one time per network. `Transaction.getConfidence()` is only a proxy (with apparently some issues, see #1769).

Currently, `TxConfidenceTable` resides in our `Context` so we have it available whenever we need to. However, this is in the way of making the context optional and used only for testing purposes. Unfortunately, `TxConfidenceTable` is needed in production, there is quite some logic (and usually UI) depending on it.

So the question is: where could `TxConfidenceTable` go?

* Passing in `TxConfidenceTable` references whereever we need them is likely unfeasible, and is probably the reason Context was implemented in the first place. It would affect many methods of the call chain I assume.
* If we keep the table in the Context, but make Context optional at the same time, we would need to auto-create the context if it doesn't exist yet. This would create an empty `TxConfidenceTable`, possibly unexpected and thus introduce hard to detect bugs/inconsistencies. But this might actually be ok, since `TxConfidenceTable` has weak refs anyway, and has a 1000 entry cap. So apps might already have to deal with spurious "losses of confidence". Still it doesn't feel good to me.
* We could make `TxConfidenceTable` a VM-scoped singleton. Since the table is addressed by tx hash, it could actually support multiple networks at the same time (hashes don't overlap).  Still it doesn't feel good to me.
* It could go to an object we already pass around everywhere. `Network` or `NetworkParameters` come to mind, but we want to keep `Network` immutable and get rid of `NetworkParameters` eventually. Is there some omnipresent object left I oversee?
* We could also try to undo the introduction of `TxConfidenceTable` (ages ago) such that the individual transactions keep track of confidence again. But in this case we might get rid of the confidence depth first, see #877 and #1553. Also, we might need to make sure tx instances only exist once per transaction, otherwise we'd have multiple confidences for the same tx. Our quest to make txns immutable might come in handy here?

This ticket is meant for discussing the various options.

msgilligan:
Why not `PeerGroup`? It looks like it may have been there originally:
```
    /** Use "Context.get().getConfidenceTable()" instead */
    @Deprecated @Nullable
    public TxConfidenceTable getMemoryPool() {
        return Context.get().getConfidenceTable();
    }
```
[...]

PR #2670, opened 2023-02-03
Remove deprecated method getMemoryPool

micheal-swiggs:
@schildbach method is deprecated and unused.


schildbach:
I think we should not remove this method until we decided about getting rid of Context for normal (non-test) usage. TxConfidenceTable is one of the current blockers for this endeavor.

schildbach:
See #2531 for discussion about TxConfidenceTable.

msgilligan (approved):
Approved. I think we can merge this. I think we have decided that the "owner" of the `TxConfidenceTable` will not be a `PeerGroup` (though a reference this new owner type/instance may be stored or accessible via a `PeerGroup`.

PR #2674, opened 2023-02-07
Peer: Removes deprecated method toStringServices

micheal-swiggs:
@schildbach removes deprecated method in Peer.


schildbach:
Merged.

PR #2677, opened 2023-02-07
BitcoinURI: Removes deprecated constant `BITCOIN_SCHEME`

micheal-swiggs:
@schildbach removes deprecated constant.


schildbach:
This was deprecated in the current cycle (567d88e13fab8f0267c07ec483f414899b0af7de), so we should wait for a release to remove it.

msgilligan:
Shouldn't this be tagged with a 0.18 milestone, since the current cycle is 0.17?

PR #2678, opened 2023-02-09
Message: deprecate method getHash()

schildbach:
It's not implemented anyways and only used as a generalization of some subclass implementations.


schildbach:
Side note: We could also to remove this without deprecatng first. It was always not implemented, so strictly speaking no one could ever have used it.

msgilligan:
> Side note: We could also to remove this without deprecatng first. It was always not implemented, so strictly speaking no one could ever have used it.

Well, removing it could break compilation, whereas the deprecation will only give a compile-time warning.  (It is possible something is referencing `Message.getHash()`, but not calling it for the subclasses that have the `throws` implementation.)

msgilligan (approved):
Looks good!

schildbach:
Merged.

PR #2681, opened 2023-02-11
Github graalvm.yml workflow: use official setup-graalvm action

msgilligan:
Switch from the DeLaGuardo/setup-graalvm  action to the "official" [graalvm/setup-graalvm](https://github.com/graalvm/setup-graalvm) action.

There are three advantages to using this action:

1. It uses a newer Node.js and eliminates a warning message we are seeing with the other action
2. It combines installing `native-image` with the installation of GraalVM JDK
3. In theory it should be better supported because it is from the GraalVM team

### DIFF
--- a/.github/workflows/graalvm.yml
+++ b/.github/workflows/graalvm.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        java: [ 'java17' ]
+        java: [ '17' ]
         graalvm: [ '22.1.0' ]
         gradle: ['7.4.2']
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/graalvm.yml
```diff
@@ -17,12 +17,12 @@ jobs:
       - name: Git checkout
         uses: actions/checkout@v1
       - name: Set up GraalVM
-        uses: DeLaGuardo/setup-graalvm@4.0
+        uses: graalvm/setup-graalvm@v1
         with:
-          graalvm: ${{ matrix.graalvm }}
-          java: ${{ matrix.java }}
-      - name: Install native-image plugin
-        run: gu install native-image
```
schildbach, line 25: What's this `GITHUB_TOKEN`? Is it necessary, and why was it not necessary before?

schildbach (commented):
Barring the one question, I think this PR can be merged.

.github/workflows/graalvm.yml
msgilligan: > Token for communication with the GitHub API. Please set to ${{ secrets.GITHUB_TOKEN }} (see [templates](https://github.com/bitcoinj/bitcoinj/pull/2681#templates)) to allow the action to authenticate with the GitHub API, which helps to reduce rate limiting issues.

See: https://github.com/graalvm/setup-graalvm#options

I guess the new developers are more concerned about rate-limiting than the previous developer.

msgilligan: The `GITHUB_TOKEN` secret is documented here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret

schildbach:
Merged.

PR #2685, opened 2023-02-11
Wallet/WalletFiles: improve error handling on save

msgilligan:
Catch FileNotFound error earlier and provide more informative exception when it happens.

I discovered this issue running the WalletTool with an invalid directory path to the wallet file. Before this fix the error/exception message is:

    java.io.IOException: No such file or directory

After the fix it is:

    java.io.FileNotFoundException: /Users/sean/bitcoinj (wallet directory not found)


msgilligan:
All three of the Windows test runs failed, so I thought maybe something was broken on Windows, but I looked at the failing tests and they are just a random collection of your typical random failures.

schildbach:
Merged.

Issue #2691, opened 2023-02-17
DefaultRiskAnalysisTest fails when run without a `Context` having been previously created.


msgilligan:
Since JUnit tests are not run in a deterministic order, this error is currently randomly occurring:

```
org.bitcoinj.wallet.DefaultRiskAnalysisTest > nonStandardDust FAILED
[476](https://github.com/bitcoinj/bitcoinj/actions/runs/4207671922/jobs/7302755582#step:4:479)
    java.lang.IllegalStateException: You must construct a Context object before using bitcoinj!
[477](https://github.com/bitcoinj/bitcoinj/actions/runs/4207671922/jobs/7302755582#step:4:480)
        at org.bitcoinj.core.Context.get(Context.java:124)
[478](https://github.com/bitcoinj/bitcoinj/actions/runs/4207671922/jobs/7302755582#step:4:481)
        at org.bitcoinj.core.Transaction.getConfidence(Transaction.java:1624)
[479](https://github.com/bitcoinj/bitcoinj/actions/runs/4207671922/jobs/7302755582#step:4:482)
        at org.bitcoinj.wallet.DefaultRiskAnalysis.analyzeIsFinal(DefaultRiskAnalysis.java:75)
[480](https://github.com/bitcoinj/bitcoinj/actions/runs/4207671922/jobs/7302755582#step:4:483)
        at org.bitcoinj.wallet.DefaultRiskAnalysis.analyze(DefaultRiskAnalysis.java:65)
[481](https://github.com/bitcoinj/bitcoinj/actions/runs/4207671922/jobs/7302755582#step:4:484)
        at org.bitcoinj.wallet.DefaultRiskAnalysisTest.nonStandardDust(DefaultRiskAnalysisTest.java:139)
[482](https://github.com/bitcoinj/bitcoinj/actions/runs/4207671922/jobs/7302755582#step:4:485)
```

See: https://github.com/bitcoinj/bitcoinj/actions/runs/4207671922/jobs/7302755582

The problem can easily be reproduced by running `DefaultRiskAnalysisTest` as a standalone test.

schildbach:
Fixed via e27baa1.

PR #2692, opened 2023-02-17
DefaultRiskAnalysisTest: initialize Context in setup()

msgilligan:
This is a (possible) fix for issue #2691.


msgilligan:
This seems to fix the symptom, but may we might want to look deeper to avoid this problem turning up elsewhere...

schildbach:
I decided to use the Context.propagate() method used in other tests for consistency. See e27baa1.

PR #2693, opened 2023-02-18
Message: add `final Network`, make `params` `final`

msgilligan:
This is an experimental first step towards migrating `Message` and subclasses from `NetworkParameters` to `Network` (and reducing mutability a bit)

* Message can be constructed with either params or network
* params member is now final (but nullable, see: rough edges)
* Network member is final and not nullable
* no-args constructors are deprecated
* Added network() accessor in Peer (to eliminate no-arg constructors)

Rough edges:

* UnknownNetwork is hopefully a temporary solution until all refactoring is finished
* params is set to null when network is UnknownNetwork

Next steps:

* More analysis around the UnknownwNetwork.UNNECESSARY cases
* More analysis around which tests require UNITTESTPARAMS (in those cases we currently must use params not network)
* Make `params` private and provide a `getParams()` accessor
* Consider using an interface to mark which messages have params and/or network
* Keep refactoring towards Network

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/utils/UnknownNetwork.java
+++ b/core/src/main/java/org/bitcoinj/base/utils/UnknownNetwork.java
@@ -1,0 +1,38 @@
+package org.bitcoinj.base.utils;
+
+import org.bitcoinj.base.Coin;
+import org.bitcoinj.base.Monetary;
+import org.bitcoinj.base.Network;
+
+/**
+ * Experimental subclass for use cases where we don't know what the network is
+ */
+public enum UnknownNetwork implements Network {
+    DEPRECATED,          // Unknown network resulting from usage of deprecated API
+    UNNECESSARY;         // Unknown network in subclass where network info is not needed
+
+    @Override
+    public String id() {
+        return "deprecated";
+    }
+
+    @Override
+    public String uriScheme() {
+        return "deprecated";
+    }
+
+    @Override
+    public boolean hasMaxMoney() {
+        return false;
+    }
+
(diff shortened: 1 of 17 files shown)


schildbach:
I remember I tried to get rid of NetworkParams (now: Network) in the Message hierarchy for messages that look the same in all networks. VersionAck is one of those as far as I can tell. However this PR re-adds a constructor with Network – is this the direction we want to go?

msgilligan:
> I remember I tried to get rid of NetworkParams (now: Network) in the Message hierarchy for messages that look the same in all networks. 

That's probably the best approach. I hinted at this in these bullet-points:

>> * More analysis around the `UnknownNetwork.UNNECESSARY` cases
>> * Consider using an interface to mark which messages have params and/or network

> VersionAck is one of those as far as I can tell. However this PR re-adds a constructor with Network – is this the direction we want to go?

This was an experiment. I was trying to (a) make things more immutable and (b) move from `NetworkParameters` to `Network`.

You've convinced me that requiring `Network` (or params) in all constructors and making `Network` final and non-null is the wrong way to go.

We could create something like:
```
public interface NetworkAwareMessage extends Message {
    Network network();
}
```
and then implement that interface on all messages that have a network property.

Or maybe there's no need for the interface and it would just be that some messages have a network and some don't.

schildbach:
> Or maybe there's no need for the interface and it would just be that some messages have a network and some don't.

This. An interface would need some justification I'd say, and personally I can't think of one. (I may be wrong.)

PR #2694, opened 2023-02-19
Bech32Test: Test that Bech32m is invalid Bech32 and vice-versa

msgilligan:
* Use the array of valid Bech32m addresses and make sure they are invalid Bech32
* Use the array of valid Bech32 addresses and make sure they are invalid Bech32m

### DIFF
--- a/core/src/test/java/org/bitcoinj/base/Bech32Test.java
+++ b/core/src/test/java/org/bitcoinj/base/Bech32Test.java
@@ -71,11 +71,17 @@ public class Bech32Test {
     public void invalid_bech32() {
         for (String invalid : INVALID_BECH32)
             invalid(invalid);
+        // Valid BECH32M codes are invalid BECH32
+        for (String invalid : VALID_BECH32M)
+            invalid(invalid);
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/bitcoinj/base/Bech32Test.java
```diff
@@ -71,12 +71,18 @@ private void valid(String valid) {
     public void invalid_bech32() {
         for (String invalid : INVALID_BECH32)
             invalid(invalid);
+        // Valid BECH32M codes are invalid BECH32
```
schildbach, line 74: I think this is a false assumption. It assumes that the decoder would decode only one of the two formats "at a time", but it decodes both at the same time.

msgilligan: This makes sense. I'll make a PR with some JavaDoc for `decode` to make this more clear.

We might want to add an optional encoding parameter to `decode` so it will fail if string is in the wrong encoding. (Similar to what we do with the network parameter for addresses.) This could be used by the unit tests to verify that the distinctions between Bech32 and Bech32m are handled properly.

PR #2695, opened 2023-02-21
Add Network.segwitAddressHrp(), deprecate NetworkParameters.getSegwitAddressHrp()

msgilligan:
This PR makes the following changes:

* Add `String segwitAddressHrp()` method to `Network` interface
* Add implementation of `segwitAddressHrp()` to `BitcoinNetwork` interface
* Deprecate `public String getSegwitAddressHrp()` in `NetworkParameters`
* Add `SegwitAddress.SegwitHrp` Java `enum` defining 3 HRPs for the 4 `BitcoinNetwork` values.
* Migrate `SegwitAddress` from `NetworkParameters` to `Network` (and using `SegwitAddress.SegwitHrp`)
* Simplify `DefaultAddressParser` by using `Network.segwitAddressHrp()`
* Add a simple implementation of `segwitAddressHrp()` to `MockAltNetwork` (i.e. `return "mock";`)

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
+++ b/core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
@@ -101,6 +101,16 @@ public enum BitcoinNetwork implements Network {
         return id;
     }
 
+
+    /**
+     * Return the standard Bech32 {@link org.bitcoinj.base.SegwitAddress.SegwitHrp} (as a {@code String}) for
+     * this network.
+     * @return The HRP as a (lowercase) string.
+     */
+    public String segwitAddressHrp() {
+        return SegwitAddress.SegwitHrp.ofNetwork(this).toString();
+    }
+
     /**
      * The URI scheme for Bitcoin.
      * @see <a href="https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki">BIP 0021</a>
--- a/core/src/main/java/org/bitcoinj/base/Network.java
+++ b/core/src/main/java/org/bitcoinj/base/Network.java
@@ -27,6 +27,12 @@ public interface Network {
     String id();
 
     /**
+     * Human-readable part (HRP) of bech32 encoded segwit addresses for this network.
+     * @return HRP (lowercase)
+     */
+    String segwitAddressHrp();
+
+    /**
      * The URI scheme for this network. See {@link BitcoinNetwork#uriScheme()}.
      * @return The URI scheme for this network
      */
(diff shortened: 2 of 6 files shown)


schildbach:
Maybe too early to comment, but I wonder what does it do to extensibility if the hrp is now a Java enum rather than a string. Perhaps not too relevant, but I though I'd bring it up.

msgilligan:
> Maybe too early to comment...

Not at all. Early feedback can be valuable and that's one of the reasons I posted this as WIP.

> but I wonder what does it do to extensibility if the hrp is now a Java enum rather than a string. Perhaps not too relevant...

It's a good question. I've thought about this a little and here are some points to consider:

1. At the `Network` level HRP is still a `String`.
2. Within the `SegwitAddress` implementation, HRP is an `enum`
3. One of my goals was to get the list of valid HRPs for Segwit in one place (4 lines of code) for easy reference.
4. Note that in the test-support class `MockNetwork` just uses a simple string for its HRP.
5. We could change SegwitAddress to use a set of string constants instead of `enum`s  and this would only affect `SegwitAddress` plus a 1-line change in `BitcoinNetwork`.
6. One advantage to using the `enum` is that the `of` and `find` methods allow throwing and non-throwing validation of Segwit HRP strings.
7. One disadvantage of using the `enum` is that adding another HRP definition could be a breaking change for some code (e.g. switch statements) although that could also be viewed as an advantage (maybe that code should be updated if a new HRP is added.)

There are a couple alternative approaches:

A. Current approach. (i.e. `Network` independent type is `String`, `BitcoinNetwork` and `SegwitAddress` use `SegwitHrp` internally)
B. Go back to `String` as suggested in *5* above.
C. Create a parent interface for `SegwitHrp` (similar to how `Network` is a parent interface for the `BitcoinNetwork` enum)
D. In addition to an interface (as in C), create a custom subclass type for HRPs.

I'm happy about the approach in this PR, but would also be happy with *B* as well. *C* and *D* seem like overkill, but we could investigate them if you like.

msgilligan:
@schildbach I think this commit is a slight improvement: dbbaa20de34670fd3953b0b19f4c3f9c7d9e2588 (make `BitcoinNetwork[]` a property of each HRP)

msgilligan:
@schildbach Even better, we can use an `EnumSet`: b79432e61ff0bc9d9e71384a368d8f9bae12189d

This is ready-for-review now (see comment:: https://github.com/bitcoinj/bitcoinj/pull/2695#issue-1592717573)

I think the `EnumSet` implementation is the best, but the `switch/case` or `BitcoinNetwork[]` implementations are fine with me, too. Let me know which one you want (if any) and I will squash/merge...

msgilligan:
One more thing: added some JavaDoc: 71ddcdeee5f378a4638251ea94fa8666cd93fcba

schildbach:
Squashed & merged.

PR #2699, opened 2023-02-24
LegacyAddressTest/SegwitAddressTest: convert to Network from NetParams

msgilligan:
Finish conversion from NetworkParameters to Network in the tests. This cleans up the code, too.

There are still some references to NetworkParameters in LegacyAddressTest:

1. Tests of deprecated methods that can be removed when those methods are removed
2. Tests that need getP2SHHeader() which hasn't moved to Network yet.


schildbach:
Merged.

PR #2700, opened 2023-02-24
DefaultAddressParser: move from o.b.core to base

msgilligan:
(The remaining dependencies on `core` are on params p2sh header and in deprecated methods)


schildbach:
Merged.

PR #2702, opened 2023-02-24
ECKey,DumpedPrivateKey,EncodedPrivateKey to crypto

msgilligan:
also move associated tests and test resources


schildbach:
Merged.

PR #2703, opened 2023-02-24
Remove DRMWorkaround for AES-256 in older Java

msgilligan:
Android and OpenJDK don't need this and I'm pretty sure that  Oracle JDK 9 and even recent builds of Oracle JDK don't need it either.

Maybe we should verify this before merging this PR or releasing this version.


schildbach:
Are we even using the JCE API any more? We've got BouncyCastle.

core/src/main/java/org/bitcoinj/crypto/BIP38PrivateKey.java
```diff
@@ -152,7 +152,6 @@ private ECKey decryptNoEC(String normalizedPassphrase) {
             byte[] key = Arrays.copyOfRange(derived, 32, 64);
             SecretKeySpec keyspec = new SecretKeySpec(key, "AES");
 
-            DRMWorkaround.maybeDisableExportControls();
             Cipher cipher = Cipher.getInstance("AES/ECB/NoPadding");
```
msgilligan, line 155: @schildbach I'm guessing this line uses JCE (or at least used to)

schildbach:
It is my understanding that if you'd need the workaround but have not applied it,  your code (e.g. unit tests) would fail rather than silently yield wrong results. Under that assumption, I'd tend to merge this.

msgilligan:
> Are we even using the JCE API any more? We've got BouncyCastle.

I think we are. See my comment in `BIP38PrivateKey.java`

msgilligan:
> It is my understanding that if you'd need the workaround but have not applied it, your code (e.g. unit tests) would fail rather than silently yield wrong results. Under that assumption, I'd tend to merge this.

Yeah, I think we should just merge it.

core/src/main/java/org/bitcoinj/crypto/BIP38PrivateKey.java
schildbach: Answering my own question: Grepping for `getInstance("` yields 8 occurences.

schildbach:
According to the exception message in `DRMWorkaround`, the parts in bitcoinj that needed this are: Tor (which we don't have any more) and BIP38 (implemented via `BIP38PrivateKey.java`)

schildbach:
Merged.

PR #2704, opened 2023-02-24
Removes unused parameter `addressToken` in BitcoinURI

micheal-swiggs:


core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
```diff
@@ -186,12 +185,13 @@ public BitcoinURI(@Nonnull Network network, String input) throws BitcoinURIParse
         }
 
         // Attempt to parse the rest of the URI parameters.
-        parseParameters(network, addressToken, nameValuePairTokens);
+        parseParameters(network, nameValuePairTokens);
 
+        String addressToken = addressSplitTokens[0];  // may be empty!
```
msgilligan, line 190: I think moving this line is confusing and also unrelated to removing the unused parameter.  i.e. the code reads much better if we can easily see that the address is before the split (`[0]`) and the other tokens are after the split (`[1]`)

msgilligan (changes requested):

schildbach:
Just a note for the future, to amend a PR you don't need to open a new PR. You can either add commits to your branch and push, or change the existing commits and force-push. In any case, the GitHub UI makes it fairly easy to review the diff.

PR #2705, opened 2023-02-24
BtcFormat: remove unused variable `minus`

micheal-swiggs:


core/src/main/java/org/bitcoinj/utils/BtcFormat.java
```diff
@@ -1555,7 +1555,6 @@ public String pattern() { synchronized(numberFormat) {
         String moneySep = String.valueOf(s.getMonetaryDecimalSeparator());
         String zero = String.valueOf(s.getZeroDigit());
         String boundary = String.valueOf(s.getPatternSeparator());
-        String minus = String.valueOf(s.getMinusSign());
```
msgilligan, line 1558: @schildbach Should we be removing this unused variable or should we be adding code to use it properly?

schildbach: Yes, that's the correct question to ask (-:  I'd love to forward this decision to Adam Mackler, but it seems he isn't around on GitHub.

(A third option would be to deprecate and remove the class (and its descendants) eventually. It's kind of redundant to `MonetaryFormat` and exists mainly because it plugs into the `java.text` API.)

schildbach:
I think I'd like to keep this as a reminder to implement this.

PR #2713, opened 2023-02-27
Create CryptoUtils, move sha256hash160 from Utils

msgilligan:
`Utils. sha256hash160()` is currently used by the `crypto` package (to be module) and requires Bouncy Castle.

Move it to a new class `CryptoUtils` in a new `internal` sub package.


schildbach:
Merged.

PR #2715, opened 2023-02-27
Move FutureUtils to base.internal (and remove Guava dependency)

msgilligan:


schildbach:
I did just that and merged.

PR #2717, opened 2023-02-27
PlatformUtils: new class with os/runtime detection from core.Utils

msgilligan:


schildbach:
I think this is a good idea. The PR seems to be missing the new class though.

msgilligan:
@schildbach Fixed and force-pushed.

schildbach:
Merged.

PR #2720, opened 2023-02-27
LegacyAddress: Replace 2 core imports with paths in JavaDoc

msgilligan:
This helps us keep better track of our progress in migrating away from core.


schildbach:
Merged.

PR #2722, opened 2023-02-27
crypto package: remove use of Guava Stopwatch

msgilligan:
It was used in KeyCrypterScript and MnemonicCode for logging derivation times. This is not really necessary and the best way to remove the Stopwatch dependency is to just remove the timing and logging.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/KeyCrypterScrypt.java
+++ b/core/src/main/java/org/bitcoinj/crypto/KeyCrypterScrypt.java
@@ -17,7 +17,6 @@
 
 package org.bitcoinj.crypto;
 
-import com.google.common.base.Stopwatch;
 import com.google.protobuf.ByteString;
 import org.bitcoinj.wallet.Protos;
 import org.bitcoinj.wallet.Protos.ScryptParameters;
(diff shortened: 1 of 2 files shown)


schildbach:
These are the kind of measurements we're using `Stopwatch` for. I think if we remove this, we can just as well eliminate `Stopwatch` everywhere. Is this what you're aiming for?

msgilligan:
> These are the kind of measurements we're using `Stopwatch` for. I think if we remove this, we can just as well eliminate `Stopwatch` everywhere. Is this what you're aiming for?

I was just trying to eliminate it in `crypto`, but I suppose we could increase the scope and try to eliminate it everywhere...

schildbach:
Ok, I see three possible options:

1. Remove just the Stopwatch usage in the crypto package (as proposed by this PR)
2. Remove all Stopwatch usage, as the ultimate goal is getting rid of Guava
3. Replace all Stopwatch usage by a (simplified) copy of Stopwatch

I suggest the deadline of our decision is the point in time where all other Guava refs have been eliminated from crypto.

msgilligan:
> 3. Replace all Stopwatch usage by a (simplified) copy of Stopwatch

This PR https://github.com/bitcoinj/bitcoinj/pull/1999 was created in 2020 to replace Guava `Stopwatch`. We closed it because at the time we were no where near dropping dependencies on Guava.

schildbach:
See https://github.com/bitcoinj/bitcoinj/pull/2731 for an alternative suggestion to remove usage of Guava Stopwatch.

msgilligan:
Closing in favor of PR #2731.

PR #2723, opened 2023-02-27
MonetaryFormat.satoshisToDecimal: encapsulate decimal conversion

msgilligan:
This is a reasonable standalone refactoring, and is being made in preparation for migration away from a  dependency on Guava's LongMath.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/utils/MonetaryFormat.java
+++ b/core/src/main/java/org/bitcoinj/base/utils/MonetaryFormat.java
@@ -362,16 +362,12 @@ public final class MonetaryFormat {
         checkState(maxDecimals <= smallestUnitExponent,
                 "The maximum possible number of decimals (%s) cannot exceed %s.", maxDecimals, smallestUnitExponent);
 
-        // rounding
+        // convert to decimal
         long satoshis = Math.abs(monetary.getValue());
         int potentialDecimals = smallestUnitExponent - shift;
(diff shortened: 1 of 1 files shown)


schildbach:
This adds a good chunk of code, and I wonder how does it aid to migrate away from "LongMath"?

msgilligan:
> This adds a good chunk of code, and I wonder how does it aid to migrate away from "LongMath"?

If you don't count the comments and the `DecimalNumber` wrapper class for the return value, it doesn't really add any code.  (Note: that the `DecimalNumber` wrapper in the future would be a `record` class)

It aids in migrating away from `LongMath` because it wraps 3 usages of `LongMath` in a single function. So instead of providing replacements for `LongMath.divide` and `LongMath.checkedPow` we can use an alternative method of producing `DecimalNumber`.  Two possibilities for this are using `BigInteger` and/or `BigDecimal` or possibly some methods in the `Coin` class.

Placing this functionality in a dedicated method makes it easier to reason about and makes it more clear which variables are temporary variables.

msgilligan:
With #2817 about to be merged, I think addressing this PR (and any related follow-up) is the last thing we have to do to remove all non-deprecated usages of Guava from `o.b.base`.

schildbach:
I now understand where this PR is going. Merging…

msgilligan:
Merged. (and since then, improved.)

PR #2725, opened 2023-02-28
Coin.COIN_VALUE: replace use of LongMath.pow with long literal

msgilligan:
There's no convenient Java built-in for exponentiation of `long`. And this constant is unlikely to ever change. So we're just going to define it with a `long` literal. 

Fortunately Java 7 added the ability to use `_` to group digits.


schildbach:
Merged.

PR #2726, opened 2023-03-01
ByteUtils: replace Guava BaseEncoding

msgilligan:
* Deprecate `ByteUtils.HEX`
* Replace with `ByteUtils.formatHex` and `ByteUtils.parseHex`
* Direct use of Guava BaseEncoding in tests that use some specific features of `BaseEncoding` that aren't provide by the new methods in `ByteUtils`.


msgilligan:
> I guess you've looked into options that might come with the JDK?

Yes. There is nothing that comes with the JDK that is usable across all versions.  (IIRC, There was something in an XML-related class that was removed from JavaSE)

They (finally!) added `HexFormat` in Java 17: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/HexFormat.html

The class provided in this PR implements a subset of `HexFormat` so should be easy for Java developers to understand and can e replaced with the real HexFormat when/if we move to a Java 17 baseline.

> 
> Maybe a test would be nice. We have tons of (implicit) tests for the positive path, but some failure cases might be worth to consider.

I'll add some basic tests for both positive and negative paths.

>  Theoretically we could also run a mass test with random data and compare the impls…

I thought about this, but I think it is unnecessary.

msgilligan:
@schildbach Tests added and ready for review/merge.

schildbach:
Thanks! Merged.

PR #2731, opened 2023-03-01
core, integration-test: migrate usage of Guava `Stopwatch` to `java.time` API

schildbach:
This replaces Guava Stopwatch usage by usage of `java.time.Instant` and `java.time.Duration`, aided by two new static helpers in `TimeUtils`.

Note: `java.time`requires Android API level 26, which I think we can safely require by now, either in this cycle (0.17) or next.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/internal/TimeUtils.java
+++ b/core/src/main/java/org/bitcoinj/base/internal/TimeUtils.java
@@ -18,6 +18,8 @@ package org.bitcoinj.base.internal;
 
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.Date;
 import java.util.Locale;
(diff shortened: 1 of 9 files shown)


schildbach:
Note that in this PR I decided to try without a separate Stopwatch class, moving the logic from there to the new static helpers. The only bit that nags me is that `Duration` imho has no human friendly `.toString()`, so I had to add millisecond conversions and the string "ms" to each usage in logs.

I thought about wrapping the Duration and implement our own toString(), but unfortunately Duration cannot be extended.

We could perhaps add another helper just for formatting a duration into TimeUtils.

msgilligan (approved):
Looks good to me!

I like the use of the typed, value-based classes `Instant` and `Duration` (for one reason because you don't have to worry about whether a `long` represents seconds, milliseconds, or nanos). In the future it might make sense to migrate more of `TimeUtils` to use them.

PR #2732, opened 2023-03-02
FutureUtils: add failedFuture method

msgilligan:
As we're migrating from ListenableCompletableFuture to standard CompletableFuture, we need a failedFuture() method that returns the standard CompletableFuture object.

TransactionBroadcast already needed this method as it was returning a ListenableCompletableFuture in one case where a standard CompletableFuture would be preferred.

This method can be deprecated when we require Java 9.


schildbach:
Merged.

PR #2733, opened 2023-03-02
Peer: Deprecate ping() and replace with sendPing()

msgilligan:
sendPing:

1. Returns Duration rather than Long
2. Uses CompletableFuture rather than ListenableCompletableFuture

I've also changed the error-handling a little. Instead of throwing a ProtocolException (RuntimeException) if PingPong isn't supported the exception is set in the CompletableFuture. We don't currently have anything catching this exception, so wrapping it in the CompletableFuture seems like a safe move.

### DIFF
--- a/integration-test/src/test/java/org/bitcoinj/core/PeerTest.java
+++ b/integration-test/src/test/java/org/bitcoinj/core/PeerTest.java
@@ -45,6 +45,7 @@ import java.net.InetSocketAddress;
 import java.net.SocketException;
 import java.nio.channels.CancelledKeyException;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


integration-test/src/test/java/org/bitcoinj/core/PeerTest.java
```diff
@@ -514,17 +515,17 @@ public void pingPong() throws Exception {
         inbound(writeTarget, new Pong(pingMsg.getNonce()));
         pingAndWait(writeTarget);
         assertTrue(future.isDone());
-        long elapsed = future.get();
-        assertTrue("" + elapsed, elapsed > 1000);
-        assertEquals(elapsed, peer.getLastPingTime());
-        assertEquals(elapsed, peer.getPingTime());
+        Duration elapsed = future.get();
+        assertTrue("" + elapsed, elapsed.toMillis() > 1000);
```
schildbach, line 519: I propose to change the first argument to

```
elapsed.toMilllis() + " ms"
```

Duration.toString() unfortunately isn't very human readable.

schildbach (commented):
Both commits look good to me.

One comment added.

integration-test/src/test/java/org/bitcoinj/core/PeerTest.java
msgilligan: Fixed and force-pushed.

schildbach:
Merged.

PR #2734, opened 2023-03-02
various `java.time` API migrations

schildbach:
The precision of Bitcoin time is only a second, so we need to truncate any surplus precision to avoid rounding errors in the PoW verification.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerGroup.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerGroup.java
@@ -270,9 +270,10 @@ public class PeerGroup implements TransactionBroadcaster {
     }
 
     // Exponential backoff for peers starts at 1 second and maxes at 10 minutes.
-    private final ExponentialBackoff.Params peerBackoffParams = new ExponentialBackoff.Params(1000, 1.5f, 10 * 60 * 1000);
+    private final ExponentialBackoff.Params peerBackoffParams = new ExponentialBackoff.Params(Duration.ofSeconds(1),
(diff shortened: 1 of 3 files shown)


schildbach:
I added a commit on top, with a similar refactoring for `PeerGroup.peerDiscoveryTimeout`.

core/src/main/java/org/bitcoinj/core/PeerGroup.java
```diff
@@ -139,8 +140,8 @@ public class PeerGroup implements TransactionBroadcaster {
      */
     public static final int DEFAULT_CONNECTIONS = 12;
     private volatile int vMaxPeersToDiscoverCount = 100;
-    private static final long DEFAULT_PEER_DISCOVERY_TIMEOUT_MILLIS = 5000;
-    private volatile long vPeerDiscoveryTimeoutMillis = DEFAULT_PEER_DISCOVERY_TIMEOUT_MILLIS;
+    private static final Duration DEFAULT_PEER_DISCOVERY_TIMEOUT_MILLIS = Duration.ofSeconds(5);
```
msgilligan, line 143: Since this is a private, you could remove the `_MILLIS` from the name.

msgilligan (approved):
Approved. With one minor suggestion (see line comment)

core/src/main/java/org/bitcoinj/core/PeerGroup.java
schildbach: Thanks! This was an oversight – fixed.

schildbach:
Merged.

PR #2735, opened 2023-03-02
more various `java.time` API migrations

schildbach:


core/src/main/java/org/bitcoinj/core/PeerAddress.java
```diff
@@ -54,7 +57,8 @@ public class PeerAddress extends ChildMessage {
     private String hostname;    // Used for (.onion addresses) TORV2, TORV3, null otherwise or if not-yet-parsed
     private int port;
     private BigInteger services;
-    private long time;
+    @Nullable
```
msgilligan, line 60: Can we avoid the @Nullable here, somehow?

core/src/main/java/org/bitcoinj/core/PeerAddress.java
```diff
@@ -351,10 +355,17 @@ public BigInteger getServices() {
         return services;
     }
 
-    public long getTime() {
+    @Nullable
```
msgilligan, line 358: We should definitely avoid this `@Nullable`, perhaps by using an `Optional<Duration>`?

core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
```diff
@@ -96,7 +96,7 @@ public DnsSeedDiscovery(NetworkParameters params, String hostname) {
         }
 
         @Override
-        public List<InetSocketAddress> getPeers(long services, long timeoutValue, TimeUnit timeoutUnit) throws PeerDiscoveryException {
+        public List<InetSocketAddress> getPeers(long services, Duration timeout) throws PeerDiscoveryException {
```
msgilligan, line 99: Should we provide a @Deprecated @Override for the old signature, rather than have it call the super?

msgilligan (changes requested):
Looks good, but I have a few minor changes suggested.

core/src/main/java/org/bitcoinj/core/PeerAddress.java
schildbach: Ok, I changed this to Optional.

core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
schildbach: Are you aware of the old signature in `PeerDiscovery`?

```
    /**
     * @deprecated use {@link #getPeers(long, Duration)}
     */
    @Deprecated
    default List<InetSocketAddress> getPeers(long services, long timeoutValue, TimeUnit timeoutUnit) throws PeerDiscoveryException {
        return getPeers(services, Duration.ofMillis(timeoutUnit.toMillis(timeoutValue)));
    }
```

msgilligan: Sorry, I didn't look closely enough. Yeah, that default method will cause the new method in each subclass to be called.

core/src/main/java/org/bitcoinj/core/PeerAddress.java
```diff
@@ -351,10 +353,16 @@ public BigInteger getServices() {
         return services;
     }
 
-    public long getTime() {
+    public Optional<Instant> getTimeInstant() {
```
msgilligan, line 356: This should have JavaDoc explaining under what circumstances (old peer) the `Instant` is not available.

core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
schildbach: So I assume we can mark this as resolved?

core/src/main/java/org/bitcoinj/core/PeerAddress.java
schildbach: You're right. I've just added that.

msgilligan (approved):

schildbach:
Merged.

PR #2738, opened 2023-03-03
MessageVerifyUtils: cast P2SH address to LegacyAddress

msgilligan:
* Change the private `compareP2SHScriptHashDerivedFromPubKey()` method to take a `LegacyAddress`
* In caller cast `address` (after check for P2SH) to `LegacyAddress`

(This is a minor change, but I was doing some code analysis to see in which cases `getHash()` is  called on `SegwitAddress` and this eliminates one false positive)

### DIFF
--- a/core/src/main/java/org/bitcoinj/utils/MessageVerifyUtils.java
+++ b/core/src/main/java/org/bitcoinj/utils/MessageVerifyUtils.java
@@ -1,6 +1,7 @@
 package org.bitcoinj.utils;
 
 import com.google.common.primitives.Bytes;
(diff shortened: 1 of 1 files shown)


schildbach:
Shouldn't the `comparePubKeyHash()` method be changed as well, for uniformity?

msgilligan:
> Shouldn't the `comparePubKeyHash()` method be changed as well, for uniformity?

No, because it can be passed both Legacy and Segwit v0 addresses.

schildbach:
You're right. Merged.

PR #2740, opened 2023-03-03
ByteUtils, StreamUtils: move to o.b.base.internal

msgilligan:
I don't think these should really be considered part of our external API. Of course, there is nothing to prevent them from being used externally, but putting them in the `internal` sub-package at least helps set expectations that they may not be supported in the future.


schildbach:
I agree about `StreamUtils`, also because it is new and tiny. That refactoring could go into its own commit.

`ByteUtils`  is more difficult I guess. Many of those methods have been around for a long time, and have just been deprecated (the `HEX` encoding, for example). Maybe we keep it around for one cycle to aid migration to 0.17? (But add a warning the class is planned to be removed from public API?)

msgilligan:
Yeah, ByteUtils is more likely to have external usage. However, it doesn't really make a difference to those external users whether it moves from o.b.core to o.b.base.utils or to o.b.base.internal. The `internal` name is purely advisory.

However, once we add Java Module Support (module-info.java), we have the option of not exporting the `internal` sub-package or only exporting it to "friend" modules (such as `crypto` and `core`)

We could actually add a module-info.java file in this release, but of course it will only affect modular applications (for example possibly wallettemplate.)

There is a simple Gradle plugin (there's something for Maven, too, if we ever decide to switch back) that will allow compiling a module-info.jar into a Java 8 bytecode JAR. This is a common practice (and some say best practice) used even by some semi-official JARs (e.g. Jakarta)

My goal is to do that in this release, though I'll create a separate issue and/or pull request for that.

So, with that contextual information, I would say there are three options:

1. Move it directly to `o.b.base.internal`
2. Move it to both `o.b.base.utils` and `o.b.base.internal` and deprecate the class in `utils`
3. Move it to `o.b.b.utils` only (for now)

Note that if we use option 1 and enable modules, only applications that enable modules would lose access.

I'll create a separate PR for StreamUtils and we can decide here what to do about ByteUtils.

msgilligan:
Converted to DRAFT. PR #2748 was created to move only `StreamUtils`.

Once we decide what to do about `ByteUtils` we can rebase, update, or replace this PR.

msgilligan:
PR #2748 was merged and this PR has been rebased.

msgilligan:
> I would say there are three options for moving `ByteUtils ` in the 0.17 release:
> 
> 1. Move it to `o.b.base.internal` (What this PR currently does)
> 2. Move it to `o.b.base.utils` (The current state of `master` as we have already moved it from `core`)
> 3. Move it to both `o.b.base.utils` and `o.b.base.internal` and deprecate the class in `utils`

I vote for (1) and @schildbach if you agree, this PR is ready-to-merge.

schildbach:
Rebased and merged.

PR #2741, opened 2023-03-03
ScriptBuilder: cast address to LegacyAddress before calling getHash()

msgilligan:
In both of these cases we have already checked the ScriptType of the  address and know that it is a LegacyAddress.


schildbach:
Merged.

Issue #2697, opened 2023-02-22
SegwitAddress: Store witnessProgram in 8-bit bytes (not 5-bit bytes)


msgilligan:
Currently the `SegwitAddress` class stores the `witnessProgram` in a partially-converted-to-Bech32 format that is:

1. non-standard
2. not very useful
3. confusing to maintainers (and is currently making the Bech32 class hard to use for general Bech32 encoding/decoding)
4. requires slightly more storage space.

The Bech32 encoding of binary data can be viewed as a two-step process:

1. Convert from an arbitrary byte-array (where each `byte` can contain any value) to a 8/5 larger byte-array where each `byte` only contains values from 0-32 (5-bits.)  This is currently done by the calling: `convertBits(witnessProgram, 0, witnessProgram.length, 8, 5, true)`
2. Convert each 5-byte value to a valid Bech32 ascii character using a string/array containing: `"qpzry9x8gf2tvdw0s3jn54khce6mua7l"`.

The current `SegwitAddress(Network network, int witnessVersion, byte[] witnessProgram)` constructor, performs step1 1 of the Bech32 conversion (using `convertBits`) on the `witnessProgram` and stores the results in the `byte[] bytes` array in `Address`. I propose removing this step and instead just pre-pending the `witnessProgram` to a `byte` representation of `witnessVersion`.

This will require internal conversion changes in the following places:

1. `getWitnessProgram()` - this method currently must convert from the 5-bit format back to the 8-bit format. With the proposed change, this conversion can be removed.
2. `toBech32()` - this method currently performs step 2 of the Bech32 conversion and would be modified to perform step 1 and step 2.
3. There is also a conversion back to 8-bit format in the constructor (for witness length validation) that would be unnecessary with the proposed change.

Note that the proposed change would affect the Java hash of `SegwitAddress` instances, but I don't think we are providing an API-level guarantee that this will be unchanged between versions.

Related PR: https://github.com/bitcoinj/bitcoinj/pull/1840 I think this PR had a good idea and its current status is that it needs rebasing and a new review. #1840 differs slightly from what I am proposing in that it removes the `witnessVersion` from `byte[] bytes` in the `Address` superclass and adds `private final int witnessVersion` to the `SegwitAddress` subclass. My recommendation is that we implement the change from 5-bit to 8-bit representation first and then separately consider removing `witnessVersion` from `byte[] bytes`.

schildbach:
[...]

PR #2742, opened 2023-03-04
Address,SegwitAddress: Deprecate getHash()

msgilligan:
* Deprecate Address.getHash() and SegwitAddress.getHash()
* Use SegwitAddress.getWitnessProgram() in the cases where SegwitAddress.getHash() was previously used.

It seems that both cases where a segwit address "hash" is being used are special-cases and we should not be providing "hash" has a general property of address.  Especially, given that hash does not currently include `witnessVersion`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/Address.java
+++ b/core/src/main/java/org/bitcoinj/base/Address.java
@@ -127,7 +127,9 @@ public abstract class Address implements Comparable<Address> {
      * Get either the public key hash or script hash that is encoded in the address.
      * 
      * @return hash that is encoded in the address
+     * @deprecated Use {@link LegacyAddress#getHash()} or {@link SegwitAddress#getWitnessProgram()} as appropriate
      */
+    @Deprecated
     public abstract byte[] getHash();
 
     /**
--- a/core/src/main/java/org/bitcoinj/base/LegacyAddress.java
+++ b/core/src/main/java/org/bitcoinj/base/LegacyAddress.java
@@ -214,7 +214,6 @@ public class LegacyAddress extends Address {
     }
 
     /** The (big endian) 20 byte hash that is the core of a Bitcoin address. */
-    @Override
     public byte[] getHash() {
         return bytes;
     }
--- a/core/src/main/java/org/bitcoinj/base/SegwitAddress.java
+++ b/core/src/main/java/org/bitcoinj/base/SegwitAddress.java
@@ -221,7 +221,11 @@ public class SegwitAddress extends Address {
         return convertBits(bytes, 1, bytes.length - 1, 5, 8, false);
     }
 
+    /**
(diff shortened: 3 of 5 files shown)


msgilligan:
Related:
* https://github.com/bitcoinj/bitcoinj/pull/1840
* https://github.com/bitcoinj/bitcoinj/issues/2697

schildbach:
It's documented in `Address.getHash()`:

> Get either the public key hash or script hash that is encoded in the address.

So it's expected that it doesn't include the `witnessVersion`. And for that matter, it's not meant as a replacement to `hashCode()` or similar.

But indeed, it's debatable we need this as a general property – probably not.

schildbach:
Perhaps is makes sense to just remove the method from the public API, but keep it for internal use?

msgilligan:
> It's documented in `Address.getHash()`:
> 
> 
> 
> > Get either the public key hash or script hash that is encoded in the address.
> 

I wasn't sure whether this comment had been updated for segwit. I suppose we could explicitly mention the segwit case and also distinguish from `.hashCode()`.

msgilligan:
My main motivation for exploring this was to eliminate the `byte[] data` in the parent `Address` class and to make `Address` into an `interface`.

Both of those goals can be achieved while leaving the `Address.getHash()` method.

So the question becomes: "is getHash a general property of all addresses?"

It seems there are (at least) two use cases for this information:

1. Verifying message signatures
2. Providing a unique binary lookup key for addresses (e.g. in a wallet)

I don't think either of these uses should be considered internal.

I'm not sure how a full Taproot implementation will affect this (if at all)

So it really comes down to: what is the best API for `Address`?

schildbach:
> Both of those goals can be achieved while leaving the Address.getHash() method.

In that case I'd suggest to eliminate this commit from the other PRs so we can go ahead with the easy ones first.

I'll respond to your comment more thoroughly later today (hopefully).

msgilligan:
> > Both of those goals can be achieved while leaving the Address.getHash() method.
> 
> In that case I'd suggest to eliminate this commit from the other PRs so we can go ahead with the easy ones first. 

Yes, I was about to suggest that.

> I'll respond to your comment more thoroughly later today (hopefully).

I'd like to hear your thoughts. However, I don't think there's a big hurry on changing this after we rebase the other PRs.

msgilligan:
Converting to DRAFT. PR #2747 submitted to continue the PR-chain skipping this PR/commit.

msgilligan:
> However based on our discussion I think we should…
> 
> 1. rename it to a more approprate name.
> 2. make clear in the JavaDoc that this is not a hash (hashCode, whatever) of the address. Rather the hash is an element (field) of the address or like I said "the address represents a hash". I'm not sure about the best wording here.

I agree. Let's put this off until we have resolved some other issues. I'll leave this PR open as a reminder.

PR #2750, opened 2023-03-05
Wallet: migrate `fromWatchingKeyB58()` and `fromSpendingKeyB58()` methods to `java.time` API

schildbach:


core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -391,8 +393,17 @@ public static Wallet fromWatchingKey(NetworkParameters params, DeterministicKey
     /**
      * Creates a wallet that tracks payments to and from the HD key hierarchy rooted by the given watching key. The
      * account path is specified. The key is specified in base58 notation and the creation time of the key. If you don't
-     * know the creation time, you can pass {@link DeterministicHierarchy#BIP32_STANDARDISATION_TIME_SECS}.
+     * know the creation time, you can pass empty and it will use {@link DeterministicHierarchy#BIP32_STANDARDISATION_TIME_SECS}.
```
msgilligan, line 396: We should deprecate `DeterministicHierarchy#BIP32_STANDARDISATION_TIME_SECS` and define `public static final Instant DeterministicHierarchy#BIP32_STANDARDISATION_TIME`.

msgilligan: We should also consider defaulting creationTime of Deterministic Keys to DeterministicHierarchy#BIP32_STANDARDISATION_TIME.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -391,8 +393,17 @@ public static Wallet fromWatchingKey(NetworkParameters params, DeterministicKey
...
      */
+    public static Wallet fromWatchingKeyB58(NetworkParameters params, String watchKeyB58, Optional<Instant> creationTime) {
```
msgilligan, line 398: Hopefully we can avoid the `Optional` parameter here.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -391,8 +393,17 @@ public static Wallet fromWatchingKey(NetworkParameters params, DeterministicKey
...
+        if (creationTime.isPresent())
+            return fromWatchingKeyB58(params, watchKeyB58, creationTime.get().getEpochSecond());
```
msgilligan, line 400: Isn't this calling a deprecated method?

msgilligan (changes requested):

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: > We should deprecate `DeterministicHierarchy#BIP32_STANDARDISATION_TIME_SECS` and define `public static final Instant DeterministicHierarchy#BIP32_STANDARDISATION_TIME`.

I agree. I saved this for a future commit.

schildbach: > We should also consider defaulting creationTime of Deterministic Keys to DeterministicHierarchy#BIP32_STANDARDISATION_TIME.

Yes, and we also could default single keys to the invention to Genesis time.

However, these changes are out of scope for this PR.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: Yes, I decided to implement it that way around because it was easier.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: > However, these changes are out of scope for this PR.

I agree, but they might help reduce some of the method-overload boilerplate.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: I generally prefer to implement the deprecated methods with the non-deprecated methods for the following reasons:

1. It makes things clear -- at the level of the code -- what the canonical way of doing things is.
2. It eliminates deprecated warnings in compiles and in the IDE
3. It makes it easier to remove the deprecated method in the future.

I'll address this in my attempt at an alternative commit.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: Honestly if we want to insert defaults we need to plan and think through carefully. And if you think it's a pre-requisite to this PR, we should talk about defaults first. I'll open a separate issue.

schildbach: Done: https://github.com/bitcoinj/bitcoinj/issues/2755

msgilligan:
https://github.com/bitcoinj/bitcoinj/pull/2756 provides an alternate implementation using method overloading rather than passing `Optional<>` method parameters.

schildbach:
Closing in favour of #2756.

(I sense another American vs British spelling in this sentence.)

PR #2752, opened 2023-03-05
Wallet, KeyChainGroup: migrate `keyRotationTime` field to `java.time` API

schildbach:


core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -5294,29 +5274,55 @@ public void setKeyRotationTime(@Nullable Date time) {
...
+        else
+            setKeyRotationTime(Instant.ofEpochMilli(time.getTime()));
+    }
+
     /**
-     * Returns the key rotation time, or null if unconfigured. See {@link #setKeyRotationTime(Date)} for a description
+     * Returns the key rotation time, or null if unconfigured. See {@link #setKeyRotationTime(Instant)} for a description
      * of the field.
      */
+    public Optional<Instant> getKeyRotationTimeInstant() {
```
msgilligan, line 5307: I suppose it would also be possible to use `Instant.MAX` (aka "never") for the case where there is no rotation enabled.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -284,9 +286,9 @@ protected boolean removeEldestEntry(Map.Entry<Sha256Hash, Transaction> eldest) {
     protected volatile WalletFiles vFileManager;
     // Object that is used to send transactions asynchronously when the wallet requires it.
     protected volatile TransactionBroadcaster vTransactionBroadcaster;
-    // UNIX time in seconds. Money controlled by keys created before this time will be automatically respent to a key
+    // Money controlled by keys created before this time will be automatically respent to a key
     // that was created after it. Useful when you believe some keys have been compromised.
-    private volatile long vKeyRotationTimestamp;
+    private volatile Optional<Instant> vKeyRotationTime = Optional.empty();
```
msgilligan, line 291: This could be `Instant.MAX` by default if we want to avoid using `Optional` altogether.

msgilligan (changes requested):
Although I believe that `Optional` is useful (particularly for the return-value case) I also believe that it should be used in moderation.

Some of the methods could use overloads to avoid `Optional` and/or we could use `Instant.MAX` to represent the case of "never rotate")

schildbach:
If we do method overloading, then I fear we'll end up with a lot of conditionals for the passing through of `Optional<Instant/Duration>` values.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: Currently (as a long) we use 0 as the magic constant for "never". I'm a bit surprised by this recommendation to use a magic value. I thought `Optional` is exactly meant for this, or `null`.

schildbach:
Ok, I get it. `Optional` should not be used in parameters.

I investigated and discovered `Optional.ofNullable()` and `.orElse(null)`, providing relatively easy conversions between `Optional` and nullable types.

So I decided to rewrite this (as a commit on top for clarity) with nullable parameters, and a nullable field. It removes a bit of boilerplate.

Does this look better?

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -5346,29 +5326,55 @@ public void setKeyRotationTime(@Nullable Date time) {
...
+    public void setKeyRotationTime(@Nullable Date time) {
+        if (time == null)
+            vKeyRotationTime = null;
+        else
+            setKeyRotationTime(Instant.ofEpochMilli(time.getTime()));
+    }
+
     /**
-     * Returns the key rotation time, or null if unconfigured. See {@link #setKeyRotationTime(Date)} for a description
+     * Returns the key rotation time, or empty if unconfigured. See {@link #setKeyRotationTime(Instant)} for a description
```
msgilligan, line 5356: This is one of the few places where the `null`/`empty` case is documented. I think this could be improved.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -5346,29 +5326,55 @@ public void setKeyRotationTime(@Nullable Date time) {
...
+                TimeUtils.dateTimeFormat(keyRotationTime.toEpochMilli()));
+        vKeyRotationTime = keyRotationTime;
         saveNow();
     }
 
+    /** @deprecated use {@link #setKeyRotationTime(Instant)} */
+    @Deprecated
+    public void setKeyRotationTime(long timeSecs) {
+        if (timeSecs == 0)
+            vKeyRotationTime = null;
```
msgilligan, line 5341: I think setKeyRotationTime() should be called here, like this:
```
setKeyRotationTime(null);
```

Or maybe use an intermediate, nullable value and call setKeyRotationTime() once.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -5346,29 +5326,55 @@ public void setKeyRotationTime(@Nullable Date time) {
...
+        else
+            setKeyRotationTime(Instant.ofEpochSecond(timeSecs));
+    }
+
+    /** @deprecated use {@link #setKeyRotationTime(Instant)} */
+    @Deprecated
+    public void setKeyRotationTime(@Nullable Date time) {
+        if (time == null)
+            vKeyRotationTime = null;
```
msgilligan, line 5350: Same comment as above.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -5346,29 +5326,55 @@ public void setKeyRotationTime(@Nullable Date time) {
      * The given time cannot be in the future.
      * </p>
      */
-    public void setKeyRotationTime(long unixTimeSeconds) {
-        checkArgument(unixTimeSeconds <= TimeUtils.currentTimeSeconds(), "Given time (%s) cannot be in the future.",
-                TimeUtils.dateTimeFormat(unixTimeSeconds * 1000));
-        vKeyRotationTimestamp = unixTimeSeconds;
+    public void setKeyRotationTime(Instant keyRotationTime) {
```
msgilligan, line 5329: `Instant` should probably be annotated as `@Nullable`.

Also, The JavaDoc should be updated as it currently says:

> You can stop key rotation by calling this method again with {@code 0} as the argument.

When updating the JavaDoc it would be good to add an actual `@param` line for the argument.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -286,9 +287,10 @@ protected boolean removeEldestEntry(Map.Entry<Sha256Hash, Transaction> eldest) {
     protected volatile WalletFiles vFileManager;
     // Object that is used to send transactions asynchronously when the wallet requires it.
     protected volatile TransactionBroadcaster vTransactionBroadcaster;
-    // UNIX time in seconds. Money controlled by keys created before this time will be automatically respent to a key
+    // Money controlled by keys created before this time will be automatically respent to a key
     // that was created after it. Useful when you believe some keys have been compromised.
-    private volatile long vKeyRotationTimestamp;
+    @Nullable
```
msgilligan, line 292: I think it would be good to explain here what `null` means.

msgilligan (changes requested):
This looks good. I've got a few minor style change requests and some JavaDoc change requests.

msgilligan:
> Does this look better?

Yes!

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -5346,29 +5326,55 @@ public void setKeyRotationTime(@Nullable Date time) {
...
+    }
+
     /**
-     * Returns the key rotation time, or null if unconfigured. See {@link #setKeyRotationTime(Date)} for a description
+     * Returns the key rotation time, or empty if unconfigured. See {@link #setKeyRotationTime(Instant)} for a description
      * of the field.
      */
+    public Optional<Instant> getKeyRotationTimeInstant() {
+        Optional<Instant> keyRotationTime = Optional.ofNullable(vKeyRotationTime);
+        return keyRotationTime;
```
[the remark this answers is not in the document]
schildbach: Done.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: I added another comment to the  `vKeyRotationTime` field.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: Done.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: Done.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: I added another comment to the vKeyRotationTime field.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: Both done.

PR #2754, opened 2023-03-05
wallettemplate: Upgrade to JavaFX 17.0.6

msgilligan:
17.0.6 is currently the latest version of the current LTS version of JavaFX:

https://gluonhq.com/products/javafx/

Note that Gluon has recently announced that ongoing maintenance builds of JavaFX 17 will be free and publicly available. They have also announced that JavaFX 20 and later will require JDK 17 and later. (So it's time to start thinking about upgrading this project to require JDK 17.)

### DIFF
--- a/wallettemplate/build.gradle
+++ b/wallettemplate/build.gradle
@@ -13,7 +13,7 @@ dependencies {
 }
 
 javafx {
-    version = '17.0.2'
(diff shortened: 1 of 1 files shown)


schildbach:
Luckily JavaFX 17 is LTS until 2026…

Merged.

PR #2756, opened 2023-03-05
Wallet: migrate fromWatchingKeyB58, fromSpendingKeyB58 to java.time API

msgilligan:
This is an alternative implementation to the one in PR #2750 that uses method overloads rather than Optional<Instant> to handle the case of an unknown creation time.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -114,6 +114,7 @@ import java.math.BigInteger;
 import java.math.RoundingMode;
 import java.nio.ByteBuffer;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -389,15 +390,39 @@ public static Wallet fromWatchingKey(NetworkParameters params, DeterministicKey
...
 
+     /**
+     * Creates a wallet that tracks payments to and from the HD key hierarchy rooted by the given watching key. The
+     * account path is specified.
+     * @param params The network
+     * @param watchKeyB58 The key in base58 notation
+     * @param creationTime creation time of the key (if unknown use {@link #fromWatchingKeyB58(NetworkParameters, String)}
+     * @return a new wallet
+     */
+    public static Wallet fromWatchingKeyB58(NetworkParameters params, String watchKeyB58, Instant creationTime) {
```
msgilligan, line 401: Since we've deprecated the old `fromWatchingKeyB58` method, we should probably change the 1st parameter from `NetworkParameters` to `Network` as well.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -413,13 +438,36 @@ public static Wallet fromSpendingKey(NetworkParameters params, DeterministicKey
...
     /**
      * Creates a wallet that tracks payments to and from the HD key hierarchy rooted by the given spending key.
-     * The key is specified in base58 notation and the creation time of the key. If you don't know the creation time,
-     * you can pass {@link DeterministicHierarchy#BIP32_STANDARDISATION_TIME_SECS}.
+     * @param params The network
+     * @param spendingKeyB58 The key in base58 notation
+     * @param creationTime creation time of the key (if unknown use {@link #fromWatchingKeyB58(NetworkParameters, String)}
+     * @return a new wallet
+     */
+    public static Wallet fromSpendingKeyB58(NetworkParameters params, String spendingKeyB58, Instant creationTime) {
```
msgilligan, line 446: Since we've deprecated the old `fromWatchingKeyB58` method, we should probably change the 1st parameter from `NetworkParameters` to `Network` as well.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -389,15 +390,39 @@ public static Wallet fromWatchingKey(NetworkParameters params, DeterministicKey
...
+
+    /**
+     * @deprecated Use {@link #fromWatchingKeyB58(NetworkParameters, String, Instant)} or {@link #fromWatchingKeyB58(NetworkParameters, String)}
+     */
+    @Deprecated
     public static Wallet fromWatchingKeyB58(NetworkParameters params, String watchKeyB58, long creationTimeSeconds) {
-        final DeterministicKey watchKey = DeterministicKey.deserializeB58(null, watchKeyB58, params);
-        watchKey.setCreationTimeSeconds(creationTimeSeconds);
-        return fromWatchingKey(params, watchKey, outputScriptTypeFromB58(params, watchKeyB58));
+        return (creationTimeSeconds == 0)
```
schildbach, line 423: I generally like an if-else statement more for cases like this, also because if there were no return value you could not use the terniary op.

But if course this is only bikeshedding.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -413,13 +438,36 @@ public static Wallet fromSpendingKey(NetworkParameters params, DeterministicKey
...
+
+    /**
+     * @deprecated Use {@link #fromSpendingKeyB58(NetworkParameters, String, Instant)} or {@link #fromSpendingKeyB58(NetworkParameters, String)}
      */
+    @Deprecated
     public static Wallet fromSpendingKeyB58(NetworkParameters params, String spendingKeyB58, long creationTimeSeconds) {
-        final DeterministicKey spendKey = DeterministicKey.deserializeB58(null, spendingKeyB58, params);
-        spendKey.setCreationTimeSeconds(creationTimeSeconds);
-        return fromSpendingKey(params, spendKey, outputScriptTypeFromB58(params, spendingKeyB58));
+        return (creationTimeSeconds == 0)
```
schildbach, line 468: Same here.

schildbach (commented):
I have the feeling this case study was "too easy". I should have pointed you at #2752 I guess…

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: I've become a much bigger fan of the ternary operator as I've been using functional-style Java. Because it is an expression and not a statement.  (I'm not a huge fan of the `boolean ? expr1 : expr2` syntax, but I try to make do and once I saw the current formatting that I generally use I found that fairly readable)

In this case I think it makes the code more reasonable because you see that all the method does it return a value.  (i.e. the whole method is simply `return expression`)

In the case of a void method (i.e. one with side-effects) it is by definition imperative (not functional) code and can not be a single expression.

I'm happy to rewrite this if you insist, though.  (Although long-run it doesn't matter because this method is deprecated for removal)

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: Same response here as above.

msgilligan:
> I have the feeling this case study was "too easy". I should have pointed you at #2752 I guess…

Fair enough. I'll do 2752 now...

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: I trust you on what's best for functional style.

schildbach:
Fixed the missing import and merged.

PR #2757, opened 2023-03-05
DownloadProgressTracker: migrate `progress()` method to `java.time` API

schildbach:
I'm aware the new method invokes the deprecated old method. I did not find an elegant way that works well for both unmigrated code and migrated code. It should go away with the removal of the deprecated method though.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/listeners/DownloadProgressTracker.java
+++ b/core/src/main/java/org/bitcoinj/core/listeners/DownloadProgressTracker.java
@@ -26,6 +26,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
+import java.time.Instant;
 import java.util.Date;
 import java.util.Locale;
 import java.util.concurrent.CompletableFuture;
@@ -68,7 +69,7 @@ public class DownloadProgressTracker implements BlockchainDownloadEventListener 
             caughtUp = true;
             if (lastPercent != 100) {
                 lastPercent = 100;
-                progress(lastPercent, blocksLeft, new Date(block.getTimeSeconds() * 1000));
+                progress(lastPercent, blocksLeft, block.getTimeInstant());
             }
             doneDownload();
             future.complete(peer.getBestHeight());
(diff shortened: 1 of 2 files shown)


msgilligan (approved):
Looks good!  (Update: upon closer look I have a question that should be answered and maybe addressed before merging.)

core/src/main/java/org/bitcoinj/core/listeners/DownloadProgressTracker.java
```diff
@@ -89,11 +90,23 @@ public void onBlocksDownloaded(Peer peer, Block block, @Nullable FilteredBlock f
...
+
+    /**
+     * Called when download progress is made.
+     *
+     * @param pct  the percentage of chain downloaded, estimated
+     * @param time the time of the last block downloaded
+     * @deprecated use {@link #progress(double, int, Instant)}
+     */
+    @Deprecated
+    protected void progress(double pct, int blocksSoFar, Date time) {
```
msgilligan, line 109: Should this do one of the following:
* Call the new method?
* Throw a RuntimeException?
* Be removed entirely to cause a breaking compile error?

msgilligan (changes requested):

core/src/main/java/org/bitcoinj/core/listeners/DownloadProgressTracker.java
schildbach: Well, that's the question. I had the idea that a user either overrides one or the other.

It should be noted that this method is not meant to be invoked by users, it's more like a callback.

schildbach: Of course, we could simply switch the signature in those cases, and not provide backwards compatibility. Is that what you prefer?

msgilligan: I think that's reasonable and probably simpler for everyone.

schildbach: Ok, I removed the deprecated method and its invocation.

PR #2758, opened 2023-03-05
DetermnisticHierarchy/MnemonicCode: BIPXX_STANDARDISATION_TIME as `Duration`

msgilligan:
Deprecate BIPXX_STANDARDISATION_TIME_SECS int constants. Replace with BIPXX_STANDARDISATION_TIME Duration constants.

I did not update all usages of the deprecated constants because we have pending (and competing) PRs that would have conflicts. Once those PRs are finalized/merged we can update the deprecated usages.


schildbach:
Merged.

PR #2759, opened 2023-03-06
Wallet: fix minor JavaDoc error that was breaking JDK 11 build.

msgilligan:


schildbach:
Good catch! Merged.

schildbach:
I wonder why this wasn't showing up in my local builds. I'm using OpenJDK 11 for building all the time.

PR #2761, opened 2023-03-06
WalletTool:  convert to Instant in createWallet()

msgilligan:
This is a minor change that removes use of some deprecated members. It does not change behavior.


schildbach:
I have that one in the pipeline as well, but I think we should do the `creationTime` refactoring first.

msgilligan:
Closing.

Issue #2755, opened 2023-03-05
Defaults for seed/key/keychain creation times?


schildbach:
**Current state**

Currently we represent seed/key/chain creation times as a long (seconds since epoch).

In cases where a creation time is not known, we use `0`.

In cases where a creation time is derived from its parent (key from keychain), the getter hands the query through to its parent and the field value is ignored.

The protobuf representation is complicated. I think in some cases the field is omitted.

**Proposal**

The idea is to always use defaults, and disallow `0` or `null`. Single keys would get Genesis time assigned. Keychains would get BIP32 standardization time assigned (though they could probably also get Genesis time for simplicity – would not differ much these days). Derived keys would get their parents creation time assigned I guess.

Pro: Creation times would always exist, the "unknown" codepaths would go away.

Con: The information that a keys creation time is actually not known would be lost. Atm I'm not sure about the consequences.

**Related**

There is the notion of "earliestKeyCreationTime" (mostly in keychains). We currently use `0` for "unknown" and `Long.MAX` for chains without keys. I think those magic values are picked so that you can always use the a `min` function to determine the earliest time.

schildbach:
While developing #2770 and #2784 it came to my mind that we could possibly merge "creationTime" and "earliestCreationTime" into one.

PR #2762, opened 2023-03-06
WalletTool: add getCreationTime() that returns Instant

msgilligan:
Replace getCreationTimeSeconds() that returned a long # of seconds

Note: That this **changes behavior** somewhat in certain cases. In the case
where a wallet is created without a command-line option to set `date` or
`unixtime`, BIP32_STANDARDISATION_TIME will be used instead of `EPOCH`.

See Issue https://github.com/bitcoinj/bitcoinj/issues/2755


schildbach:
I believe this is now more or less contained in #2784.

schildbach:
I believe this can be closed.

msgilligan:
Yes, it can be closed...

PR #2763, opened 2023-03-06
KeyChainGroupTest: remove use of BIP39_STANDARDISATION_TIME_SECS

msgilligan:
There is one other usage, but it's fixed in PR #2761


schildbach:
I have that one in the pipeline as well, but I think we should do the `creationTime` refactoring first.

msgilligan:
> I have that one in the pipeline as well, but I think we should do the `creationTime` refactoring first.

Okay, I'm going to close this, then.

PR #2765, opened 2023-03-06
TimeUtils: Use AtomicReference<Instance> internally

msgilligan:
While doing some refactoring, I was getting false-positives (failures that should have succeeded) in tests* that were comparing actual vs. expected `Instant` values. The problem is that as currently implemented `mockTime` only has millisecond resolution, but `Instant` has higher resolution.  (These failures occurred in routines that were manipulating the MockClock)

I believe that using `Instant` internally will fix these problems.

The `AtomicReference` was necessary for the "roll" functionality.

I think there is small race condition in rollMockClockMillis(long) between the check for null (which throws IllegalStateException if null) and the `updateAndGet`. I'm not sure if the correct exception will be thrown if `updateAndGet` fails because of the `null`. That's why this is a DRAFT.

* see https://github.com/bitcoinj/bitcoinj/actions/runs/4340826601/jobs/7579742318#step:4:585 for an example of a failure.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/internal/TimeUtils.java
+++ b/core/src/main/java/org/bitcoinj/base/internal/TimeUtils.java
@@ -23,81 +23,86 @@ import java.time.Instant;
 import java.util.Date;
 import java.util.Locale;
 import java.util.TimeZone;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Utilities for time and mock time.
  */
 public class TimeUtils {
     private static final TimeZone UTC = TimeZone.getTimeZone("UTC");
+    // TODO: See if java.time.Clock can help us in our mocking
     /**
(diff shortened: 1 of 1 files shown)


schildbach:
I just published my branch of mock clock modernization: #2771
However it fails the integration tests :crying_cat_face: 

It's not using AtomicReference by now, which you do in this PR.

msgilligan:
> I just published my branch of mock clock modernization: #2771 However it fails the integration tests 😿

You implemented my TODO (`// TODO: See if java.time.Clock can help us in our mocking`)!

> It's not using AtomicReference by now, which you do in this PR.

If you can make it work without that, that's great. I think I did that because I was seeing an IDE warning about non-atomically updating a `volatile`.

msgilligan:
Closing in favor of PR #2771.

PR #2766, opened 2023-03-06
WIP: Refactoring key rotation to use Instant

msgilligan:
This is W.I.P.

* The first commit is an internal change inside `Wallet` to use `Instant` internally for `vKeyRotationTimestamp`.
* The second creates ``getEarliestKeyCreationInstant` and deprecates `getEarliestKeyCreationTime` in  `PeerFilterProvider` and `Wallet`.

The second commit exposes issues that I believe are resolved by (WIP) PR #2765.

I decided to explore the continuing to use special values on the (`Instant`) timeline in an analogous manner to how the current code uses special values on the `long` seconds timeline. Whether these special values are considered "magic numbers" or not is something we can discuss. I also went down the path of refactoring the routines that return key timestamps before refactoring the routines like `freshAddress()` where we were discussion whether or not to use an `Optional`.

I'll keep exploring this and have more to share/discuss later.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -3491,7 +3491,7 @@ public class Wallet extends BaseTaggableObject
 
             // Do the keys.
             builder.append("\nKeys:\n");
-            builder.append("Earliest creation time: ").append(TimeUtils.dateTimeFormat(getEarliestKeyCreationTime() * 1000))
+            builder.append("Earliest creation time: ").append(TimeUtils.dateTimeFormat(getEarliestKeyCreationInstant().toEpochMilli()))
                     .append('\n');
             final Date keyRotationTime = getKeyRotationTime();
             if (keyRotationTime != null)
@@ -3591,14 +3591,14 @@ public class Wallet extends BaseTaggableObject
      * If there are no keys in the wallet, the current time is returned.
      */
     @Override
-    public long getEarliestKeyCreationTime() {
+    public Instant getEarliestKeyCreationInstant() {
         keyChainGroupLock.lock();
         try {
-            long earliestTime = keyChainGroup.getEarliestKeyCreationTime();
+            Instant earliestTime = Instant.ofEpochSecond(keyChainGroup.getEarliestKeyCreationTime());
             for (Script script : watchedScripts)
-                earliestTime = Math.min(script.getCreationTimeSeconds(), earliestTime);
-            if (earliestTime == Long.MAX_VALUE)
-                return TimeUtils.currentTimeSeconds();
+                earliestTime = TimeUtils.minInstant(Instant.ofEpochSecond(script.getCreationTimeSeconds()), earliestTime);
+            if (earliestTime == Instant.MAX)
+                return Instant.now();
             return earliestTime;
         } finally {
             keyChainGroupLock.unlock();
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -288,7 +288,14 @@ public class Wallet extends BaseTaggableObject
     protected volatile TransactionBroadcaster vTransactionBroadcaster;
     // UNIX time in seconds. Money controlled by keys created before this time will be automatically respent to a key
     // that was created after it. Useful when you believe some keys have been compromised.
-    private volatile long vKeyRotationTimestamp;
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -288,7 +288,14 @@ protected boolean removeEldestEntry(Map.Entry<Sha256Hash, Transaction> eldest) {
     protected volatile TransactionBroadcaster vTransactionBroadcaster;
     // UNIX time in seconds. Money controlled by keys created before this time will be automatically respent to a key
     // that was created after it. Useful when you believe some keys have been compromised.
-    private volatile long vKeyRotationTimestamp;
```
schildbach, line 291: Note that in the description you say `MAX`, here you implement `MIN`. `MIN` is closer to the previous `0`, however if the special case isn't handled correctly it could lead into a mass-rotation of wallets. `MAX` somehow protects against that I'd say.

Either way, I don't really like it. `MAX` or `MIN` are magic constants. How is it better than using `null`?

Looking at it at a broader scale, magic constants like those are not always available. `null` or `Optional` can be used with any type (also native types using autoboxing).

schildbach:
Special care must be taken when serializing and deserializing this value to ProtoBuf. Basically, we need to stay compatible to old ProtoBufs "in all eternity".

schildbach:
Another side note about time values within bitcoinj:

We're using two precisions…

- second precision for all P2P protocol messages and some stuff derived from that. I think this is called "Bitcoin time".
- millisecoind precision (or perhaps: undefined) for everything else

I'm not sure of we've got a guarantee that any magic constant will stay the same when being truncated from milliseconds to seconds. This can happen when serializing to ProtoBuf and back, for example. I guess the `Instant.EPOCH` constant would have the best chance to survive, if it represents "0 seconds since epoch".

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: > Note that in the description you say `MAX`, here you implement `MIN`. `MIN` is closer to the previous `0`, however if the special case isn't handled correctly it could lead into a mass-rotation of wallets. `MAX` somehow protects against that I'd say.

Yes, I've been a little bit confused and kept changing my WIP code. There are different use-cases and different meanings it seems for some of the uses of 0.  (e.g. in a key creation time, it means unknown but for key rotation 0 has a different meaning)

> 
> Either way, I don't really like it. `MAX` or `MIN` are magic constants. How is it better than using `null`?

I think `null` is probably fine (and possibly "better") in this use case. But, I do think that sometimes it is better to use 0, MAX, or MIN instead of null -- especially in cases where a "natural" integer comparison will handle the special case without having to worry about NPE.

(As I said elsewhere my opinion has changed on this in recent years as I've been adopting "functional-style" Java. My abhorrence for `null` is now greater than my abhorrence for "magic constants".)

> 
> Looking at it at a broader scale, magic constants like those are not always available. `null` or `Optional` can be used with any type (also native types when using autoboxing).

Yes. And I think (as you've commented elsewhere) in the general case `Optional` is preferred for return types and for parameters it is generally preferred to use either nullable or overloading.

msgilligan: >

msgilligan: > Also I'm specifically interested in how you would refactor the `getKeyRotationTime()` accessor. I assume it would get a `getKeyRotationTimeInstant()` sibling, but would you expose the magic constant here, too? (This is an even more important question that the above, as it touched public API.)

In my work last night, I was unable to satisfactorily answer this question. Part of the problem (IMHO) is that I didn't see any comments that clearly explained what these timestamps meant (I may not have looked carefully enough.) So I don't really have an answer yet.

msgilligan:
> Another side note about time values within bitcoinj:
> 
> We're using two precisions…
> 
> * second precision for all P2P protocol messages and some stuff derived from that. I think this is called "Bitcoin time".
> * millisecond precision (or perhaps: undefined) for everything else
> 
> I'm not sure of we've got a guarantee that any magic constant will stay the same when being truncated from milliseconds to seconds.

These are good points.  I found there were issues that required special-case mapping between Long.MAX_VALUE and Instant.MAX (e.g. if you try to map Long.MAX_VALUE to an `Instant` it is out-of-range.)

msgilligan:
Closing in favor of PR #2752

PR #2767, opened 2023-03-06
TransactionConfidence: migrate `lastBroadcastedAt` field to `java.time` API

schildbach:
Note this also adds a couple of nullable declarations, as that's how the current code works – and I didn't change that.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
@@ -32,6 +32,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Executor;
@@ -84,7 +85,7 @@ public class TransactionConfidence {
     private CopyOnWriteArrayList<PeerAddress> broadcastBy;
     /** The time the transaction was last announced to us. */
     @Nullable
-    private Instant lastBroadcastedAt = null;
+    private Instant lastBroadcastTime = null;
     /** The Transaction that this confidence object is associated with. */
     private final Sha256Hash hash;
     // Lazily created listeners array.
@@ -306,7 +307,7 @@ public class TransactionConfidence {
      * @return true if marked, false if this address was already seen
      */
     public boolean markBroadcastBy(PeerAddress address) {
-        lastBroadcastedAt = TimeUtils.currentTime();
+        lastBroadcastTime = TimeUtils.currentTime();
         if (!broadcastBy.addIfAbsent(address))
             return false;  // Duplicate.
         synchronized (this) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
```diff
@@ -337,23 +339,37 @@ public boolean wasBroadcastBy(PeerAddress address) {
     }
 
     /** Return the time the transaction was last announced to us. */
-    public Date getLastBroadcastedAt() {
+    @Nullable
+    public Instant getLastBroadcastedAtInstant() {
```
msgilligan, line 343: You could name this `getLastBroadcastTime()` if you like.

Also, as this is a return value, using `Optional` here would probably be better.

core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
```diff
@@ -337,23 +339,37 @@ public boolean wasBroadcastBy(PeerAddress address) {
...
+    /** @deprecated use {@link #getLastBroadcastedAtInstant()} */
+    @Deprecated
+    @Nullable
+    public Date getLastBroadcastedAt() {
+        return lastBroadcastedAt != null ? Date.from(lastBroadcastedAt) : null;
+    }
+
     /** Set the time the transaction was last announced to us. */
-    public void setLastBroadcastedAt(Date lastBroadcastedAt) {
+    public void setLastBroadcastedAt(Instant lastBroadcastedAt) {
```
msgilligan, line 355: Maybe name this `setBroadcastTime`?  Also annotate as `@Nullable`?

Actually, could we not allow it to be set to `null`? Maybe it's only initialized to `null` and the only valid use case is "set to an actual time"?

core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
```diff
@@ -337,23 +339,37 @@ public boolean wasBroadcastBy(PeerAddress address) {
...
         this.lastBroadcastedAt = lastBroadcastedAt;
     }
 
+    /** @deprecated use {@link #setLastBroadcastedAt(Instant)} */
+    @Deprecated
```
msgilligan, line 360: Was this every set to `null` before?

core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
```diff
@@ -82,7 +83,8 @@ public TransactionConfidence createConfidence(Sha256Hash hash) {
      */
     private CopyOnWriteArrayList<PeerAddress> broadcastBy;
     /** The time the transaction was last announced to us. */
-    private Date lastBroadcastedAt;
+    @Nullable
+    private Instant lastBroadcastedAt = null;
```
msgilligan, line 87: Since this is private and the type is changing maybe rename it to `lastBroadcastTime` or `lastBroadcastInstant`?

msgilligan (changes requested):
Looks good, but I have a few minor suggestions.

core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
schildbach: Its not really meant to use null (clear the time). I'll add a `checkNotNull()`. I'll also rename to your suggestion.

core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
schildbach: Will do.

core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
schildbach: Good idea!

core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
schildbach: This method is not meant to be called with `null`, if that's what you're asking.

PR #2769, opened 2023-03-06
DumpedPrivateKeyTest: move p2shAddressCreationFromKeys from LegacyAddressTest

msgilligan:
This test has dependencies on o.b.crypto and is arguably a test of DumpedPrivateKey, so let's move it there.

### DIFF
--- a/core/src/test/java/org/bitcoinj/base/LegacyAddressTest.java
+++ b/core/src/test/java/org/bitcoinj/base/LegacyAddressTest.java
@@ -21,11 +21,8 @@ import nl.jqno.equalsverifier.EqualsVerifier;
 import nl.jqno.equalsverifier.Warning;
 import org.bitcoinj.base.exceptions.AddressFormatException;
 import org.bitcoinj.base.utils.ByteUtils;
-import org.bitcoinj.crypto.DumpedPrivateKey;
-import org.bitcoinj.crypto.ECKey;
 import org.bitcoinj.core.NetworkParameters;
 import org.bitcoinj.params.Networks;
-import org.bitcoinj.script.Script;
 import org.bitcoinj.script.ScriptBuilder;
 import org.bitcoinj.script.ScriptPattern;
 import org.bitcoinj.testing.MockAltNetworkParams;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/bitcoinj/base/LegacyAddressTest.java
```diff
@@ -228,23 +225,6 @@ public void p2shAddress() {
...
-        ECKey key2 = DumpedPrivateKey.fromBase58(MAINNET, "5Jb7fCeh1Wtm4yBBg3q3XbT6B525i17kVhy3vMC9AqfR6FH2qGk").getKey();
-        key2 = ECKey.fromPrivate(key2.getPrivKeyBytes());
-        ECKey key3 = DumpedPrivateKey.fromBase58(MAINNET, "5JFjmGo5Fww9p8gvx48qBYDJNAzR9pmH5S389axMtDyPT8ddqmw").getKey();
-        key3 = ECKey.fromPrivate(key3.getPrivKeyBytes());
-
-        List<ECKey> keys = Arrays.asList(key1, key2, key3);
-        Script p2shScript = ScriptBuilder.createP2SHOutputScript(2, keys);
-        LegacyAddress address = LegacyAddress.fromScriptHash(MAINNET,
-                ScriptPattern.extractHashFromP2SH(p2shScript));
-        assertEquals("3N25saC4dT24RphDAwLtD8LUN4E2gZPJke", address.toString());
```
schildbach, line 245: This is the actual test.

schildbach (commented):
I get why you want to move this.

I've marked the region of test code that in my opinion is the acrtual test. And I think the actual test should define what test class it belongs to, not the prerequisites of the test. I think it's in the right class atm.

To solve this, I see these options:

1. We allow _test_ dependencies from base to (in this case) cryto. Note this would not corrupt our goal for consumers of base to not depend on other bitcoinj modules.
2. I think it would be nice to have a way to parse the "dumped private key" format in base, as it's very common just like parsing addresses. I'm aware ECKey is currently rooted in crypto for good, but maybe we could separate ECKey into a part that only represents the key data (which could go to base) and another part that has all the crypto operations.

schildbach:
3. Split the test itself. I'll come up with a proposal for that.

core/src/test/java/org/bitcoinj/base/LegacyAddressTest.java
msgilligan: Agreed. Note that this still has a decency on `o.b.script`, but that's a separate issue.

core/src/test/java/org/bitcoinj/base/LegacyAddressTest.java
```diff
@@ -228,23 +225,6 @@ public void p2shAddress() {
...
-        // import some keys from this example: https://gist.github.com/gavinandresen/3966071
-        ECKey key1 = DumpedPrivateKey.fromBase58(MAINNET, "5JaTXbAUmfPYZFRwrYaALK48fN6sFJp4rHqq2QSXs8ucfpE4yQU").getKey();
-        key1 = ECKey.fromPrivate(key1.getPrivKeyBytes());
-        ECKey key2 = DumpedPrivateKey.fromBase58(MAINNET, "5Jb7fCeh1Wtm4yBBg3q3XbT6B525i17kVhy3vMC9AqfR6FH2qGk").getKey();
-        key2 = ECKey.fromPrivate(key2.getPrivKeyBytes());
-        ECKey key3 = DumpedPrivateKey.fromBase58(MAINNET, "5JFjmGo5Fww9p8gvx48qBYDJNAzR9pmH5S389axMtDyPT8ddqmw").getKey();
-        key3 = ECKey.fromPrivate(key3.getPrivKeyBytes());
-
-        List<ECKey> keys = Arrays.asList(key1, key2, key3);
-        Script p2shScript = ScriptBuilder.createP2SHOutputScript(2, keys);
```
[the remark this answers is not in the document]
msgilligan: Good point.

msgilligan:
> 1. We allow _test_ dependencies from base to (in this case) cryto. Note this would not corrupt our goal for consumers of base to not depend on other bitcoinj modules.

A similar alternative would be to move the tests that require dependencies on multiple modules to `integration-test`. This will also let us use JUnit 5.


> 2. I think it would be nice to have a way to parse the "dumped private key" format in base, as it's very common just like parsing addresses. I'm aware ECKey is currently rooted in crypto for good, but maybe we could separate ECKey into a part that only represents the key data (which could go to base) and another part that has all the crypto operations.

Yes, I've thought of this from time-to-time. It could be a fairly logical complement to `AesKey`.

msgilligan:
Closing in favor of PR #2782.

schildbach:
> A similar alternative would be to move the tests that require dependencies on multiple modules to `integration-test`. This will also let us use JUnit 5.

We'd suddenly find 95% of our "unit tests" in the integration tests package… (-:

msgilligan:
> > A similar alternative would be to move the tests that require dependencies on multiple modules to `integration-test`. This will also let us use JUnit 5.
> 
> We'd suddenly find 95% of our "unit tests" in the integration tests package… (-:

Fair enough. But maybe there is some _subset_ of tests requiring dependencies on ~~multiple modules~~ _modules that should be dependent, not dependencies_ that could be moved to `integration-test`.

PR #2770, opened 2023-03-07
Wallet, KeyChain, KeyChainGroup, PeerFilterProvider: migrate `getEarliestKeyCreationTime()` methods to `java.time` API

schildbach:
I decided that in the case of `earliestKeyCreationTime()`, use of magic constants is to be preferred to `null` or `Optional`, because it's all about comparisons.

For the "at least one creation time is unknown" case I picked `Instant.EPOCH` because it follows the previous `0` and `Instant.MIN` would be at risk of deterioration if millis/nanos were truncated.

For the "have no keys" case I picked `Instant.MAX` because it follows the `Long.MAX_VALUE` of some previous implementations and most documentation. Some documentation said "current time", but I decided against that (current time moves) and updated the documentation accordingly.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
@@ -711,10 +711,10 @@ public class DeterministicKeyChain implements EncryptableKeyChain {
 
     @Override
     public Instant getEarliestKeyCreationTimeInstant() {
-        if (seed != null)
-            return Instant.ofEpochSecond(seed.getCreationTimeSeconds());
-        else
-            return Instant.ofEpochSecond(getWatchingKey().getCreationTimeSeconds());
+        return Instant.ofEpochSecond(seed != null ?
+                seed.getCreationTimeSeconds() :
+                getWatchingKey().getCreationTimeSeconds()
+        );
     }
 
     @Override


schildbach:
This PR is draft because it fails integrations tests. I'm investigating.

msgilligan:
> use of magic constants ... because it's all about comparisons.
> 
> `Instant.EPOCH` because it follows the previous `0` and `Instant.MIN` ... at risk ... if millis/nanos were truncated.
> 
> For the "have no keys" case I picked `Instant.MAX` ... updated the documentation accordingly.

This all sounds good.

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
```diff
@@ -283,14 +284,21 @@ public boolean removeKey(ECKey key) {
...
     @Override
-    public long getEarliestKeyCreationTime() {
+    public Instant getEarliestKeyCreationTimeInstant() {
         lock.lock();
         try {
             return hashToKeys.values().stream()
                     .mapToLong(ECKey::getCreationTimeSeconds)
-                    .min()
-                    .orElse(Long.MAX_VALUE);
+                    .mapToObj(t -> Instant.ofEpochSecond(t))
```
msgilligan, line 299: You'll get an exception here if `t` (time in seconds) is `Long.MAX_VALUE`. You might need a mapping function that maps `Long.MAX_VALUE` to `Instant.MAX`. I can't remember where in the code I ran into this issue, but be careful.

core/src/test/java/org/bitcoinj/store/WalletProtobufSerializerTest.java
```diff
@@ -356,9 +356,9 @@ public void testRoundTripWatchingWallet() throws Exception {
         assertEquals(creationTime.getEpochSecond(), wallet.getWatchingKey().getCreationTimeSeconds());
         assertEquals(creationTime.getEpochSecond(), wallet2.getWatchingKey().getCreationTimeSeconds());
         assertEquals(creationTime.getEpochSecond(), wallet3.getWatchingKey().getCreationTimeSeconds());
-        assertEquals(creationTime.getEpochSecond(), wallet.getEarliestKeyCreationTime());
-        assertEquals(creationTime.getEpochSecond(), wallet2.getEarliestKeyCreationTime());
-        assertEquals(creationTime.getEpochSecond(), wallet3.getEarliestKeyCreationTime());
+        assertEquals(creationTime, wallet.getEarliestKeyCreationTimeInstant());
+        assertEquals(creationTime, wallet2.getEarliestKeyCreationTimeInstant());
+        assertEquals(creationTime, wallet3.getEarliestKeyCreationTimeInstant());
```
msgilligan, line 361: IIRC, this was where I was seeing a failure due to mock clock only having seconds-level resolution.

core/src/main/java/org/bitcoinj/base/internal/TimeUtils.java
```diff
@@ -107,6 +107,13 @@ public static Duration elapsedTime(Instant start) {
         return Duration.between(start, currentTime());
     }
 
+    /**
+     * Determines the earlier of two instants.
+     */
+    public static Instant earlierTime(Instant time1, Instant time2) {
```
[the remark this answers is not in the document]
schildbach: Agree, will change.

core/src/test/java/org/bitcoinj/store/WalletProtobufSerializerTest.java
schildbach: It hadn't been of a problem in this one, but in others. I currently use `.truncatedTo(ChronoUnit.SECONDS)` when that happens, but in this case it isn't necessary (Instant.ofEpochSecond() is always second precision).
 
In future, I plan to propose introducing the notion of a "Bitcoin time" for certain data structures (mainly P2P messages), and maybe checks at certain places that force users to provide instants without millis or nanos.

core/src/test/java/org/bitcoinj/wallet/BasicKeyChainTest.java
```diff
@@ -66,10 +68,10 @@ public void onKeysAdded(List<ECKey> keys2) {
 
     @Test
     public void importKeys() {
-        TimeUtils.setMockClock();
-        long now = TimeUtils.currentTimeSeconds();
+        Instant now = TimeUtils.currentTime().truncatedTo(ChronoUnit.SECONDS);
+        TimeUtils.setMockClock(now.getEpochSecond());
```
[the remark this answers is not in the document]
schildbach: See my previous comment.

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
schildbach: Good point. Would

```
mapToObj(t -> t == Long.MAX_VALUE ? Instant.MAX : Instant.ofEpochSecond(t))
```

work?

schildbach: I just noticed I have to do the Long.MAX_VALUE <> Instant.MAX conversion at other places, too.

msgilligan: > Good point. Would
> 
> ```
> mapToObj(t -> t == Long.MAX_VALUE ? Instant.MAX : Instant.ofEpochSecond(t))
> ```

Yes, that should work. I would write a utility method for this so you can use a method reference.

schildbach: Well, that piece of code will be gone soon, as `getCreationTime()` will migrate to `Instant` too. Coming up in this theatre…

schildbach:
I think I addressed all your comments (except the `.truncateTo()` stuff which I will address in a future PR).

Also, I fixed the integration tests.

core/src/main/java/org/bitcoinj/core/PeerFilterProvider.java
```diff
@@ -25,11 +27,18 @@
...
+     * blocks be downloaded, and thus this should default to {@link Instant#MAX}.
      */
-    long getEarliestKeyCreationTime();
+    Instant getEarliestKeyCreationTimeInstant();
+
+    /** @deprecated use {@link #getEarliestKeyCreationTimeInstant()} */
+    @Deprecated
+    default long getEarliestKeyCreationTime() {
+        Instant earliestKeyCreationTime = getEarliestKeyCreationTimeInstant();
+        return earliestKeyCreationTime.equals(Instant.MAX) ? Long.MAX_VALUE : earliestKeyCreationTime.getEpochSecond();
```
msgilligan, line 40: This is a nit, but I think we should standardize on the following format:
```
return condition
              ? valueWhenTrue
              : valueWhenFalse;
```

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -710,11 +710,11 @@ public int numLeafKeysIssued() {
...
 
     @Override
-    public long getEarliestKeyCreationTime() {
+    public Instant getEarliestKeyCreationTimeInstant() {
         if (seed != null)
-            return seed.getCreationTimeSeconds();
+            return Instant.ofEpochSecond(seed.getCreationTimeSeconds());
         else
-            return getWatchingKey().getCreationTimeSeconds();
+            return Instant.ofEpochSecond(getWatchingKey().getCreationTimeSeconds());
```
msgilligan, line 717: I would recommend the ternary operator here. With or without an intermediate (effectively final) `seconds` variable.

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
```diff
@@ -283,14 +284,21 @@ public boolean removeKey(ECKey key) {
...
     @Override
-    public long getEarliestKeyCreationTime() {
+    public Instant getEarliestKeyCreationTimeInstant() {
         lock.lock();
         try {
             return hashToKeys.values().stream()
                     .mapToLong(ECKey::getCreationTimeSeconds)
-                    .min()
-                    .orElse(Long.MAX_VALUE);
+                    .mapToObj(t -> t == Long.MAX_VALUE ? Instant.MAX : Instant.ofEpochSecond(t))
```
msgilligan, line 299: We should probably create a utility static method for this. Then a method reference could be used here.

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
```diff
@@ -283,14 +284,21 @@ public boolean removeKey(ECKey key) {
...
+                    .min((t1, t2) -> t1.compareTo(t2))
```
msgilligan, line 300: Can a method reference be used here?  (Maybe with a `Comparators` helper?)

core/src/main/java/org/bitcoinj/wallet/KeyChain.java
```diff
@@ -76,11 +77,19 @@ enum KeyPurpose {
...
+     *         {@link Instant#MAX} if no keys in this chain
      */
-    long getEarliestKeyCreationTime();
+    Instant getEarliestKeyCreationTimeInstant();
+
+    /** @deprecated use {@link #getEarliestKeyCreationTimeInstant()} */
+    @Deprecated
+    default long getEarliestKeyCreationTime() {
+        Instant earliestKeyCreationTime = getEarliestKeyCreationTimeInstant();
+        return earliestKeyCreationTime.equals(Instant.MAX) ? Long.MAX_VALUE : earliestKeyCreationTime.getEpochSecond();
```
msgilligan, line 91: Helper method for conversion. Multi-line ternary.

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
```diff
@@ -466,7 +467,7 @@ public List<DeterministicKeyChain> getActiveKeyChains(@Nullable Instant keyRotat
         checkState(supportsDeterministicChains(), "doesn't support deterministic chains");
         List<DeterministicKeyChain> activeChains = new LinkedList<>();
         for (DeterministicKeyChain chain : chains)
-            if (keyRotationTime == null || chain.getEarliestKeyCreationTime() >= keyRotationTime.getEpochSecond())
+            if (keyRotationTime == null || chain.getEarliestKeyCreationTimeInstant().compareTo(keyRotationTime) >= 0)
```
msgilligan, line 470: Use `Instant.isAfter()`

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
```diff
@@ -488,7 +489,7 @@ public final DeterministicKeyChain getActiveKeyChain(ScriptType outputScriptType
         Collections.reverse(chainsReversed);
         for (DeterministicKeyChain chain : chainsReversed)
             if (chain.getOutputScriptType() == outputScriptType
-                    && (keyRotationTime == null || chain.getEarliestKeyCreationTime() >= keyRotationTime.getEpochSecond()))
+                    && (keyRotationTime == null || chain.getEarliestKeyCreationTimeInstant().compareTo(keyRotationTime) >= 0))
```
msgilligan, line 492: Use `Instant.isAfter()`

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
```diff
@@ -844,19 +845,29 @@ public List<ECKey> getImportedKeys() {
...
         if (chains == null)
-            return Long.MAX_VALUE;
+            return Instant.MAX;
         return chains.stream()
-                .mapToLong(DeterministicKeyChain::getEarliestKeyCreationTime)
-                .min()
-                .orElse(Long.MAX_VALUE);
+                .map(DeterministicKeyChain::getEarliestKeyCreationTimeInstant)
+                .min((t1, t2) -> t1.compareTo(t2))
+                .orElse(Instant.MAX);
```
msgilligan, line 870: Ternary operator.

core/src/main/java/org/bitcoinj/core/PeerFilterProvider.java
```diff
@@ -25,11 +27,18 @@
...
     /**
-     * Returns the earliest timestamp (seconds since epoch) for which full/bloom-filtered blocks must be downloaded.
-     * Blocks with timestamps before this time will only have headers downloaded. {@code 0} requires that all
-     * blocks be downloaded, and thus this should default to {@link System#currentTimeMillis()}/1000.
+     * Returns the earliest time for which full/bloom-filtered blocks must be downloaded.
+     * Blocks with timestamps before this time will only have headers downloaded. {@link Instant#EPOCH} requires that all
+     * blocks be downloaded, and thus this should default to {@link Instant#MAX}.
      */
-    long getEarliestKeyCreationTime();
+    Instant getEarliestKeyCreationTimeInstant();
```
msgilligan, line 34: Should this be shortened to `getEarliestKeyCreationInstant()`

msgilligan (changes requested):
Looks good. I have a handful of minor requests/suggestions.

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
schildbach: I assume it could. However I think this line of code reads well, without the indirection of a method reference. Your milage may very, but you're probably well ahead in grokking the functional style.

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
schildbach: See above.

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
msgilligan: I suppose we could use `! i1.isAfter(i2)`...

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
schildbach: Ok, rewritten.

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
schildbach: Ok, I'm using a static method now. Can you review 108b0d7e3eaf42c8bb87d08ca22f4002e698ae7e if that will work as expected?

I'm stoked that such method does not exist in `Objects`. They is one for non-natural orders, but none for the natural order.

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
```diff
@@ -297,7 +298,7 @@ public Instant getEarliestKeyCreationTimeInstant() {
             return hashToKeys.values().stream()
                     .mapToLong(ECKey::getCreationTimeSeconds)
                     .mapToObj(t -> t == Long.MAX_VALUE ? Instant.MAX : Instant.ofEpochSecond(t))
-                    .min((t1, t2) -> t1.compareTo(t2))
+                    .min(TimeUtils::compare)
```
msgilligan, line 301: This should be `Instant:compareTo`.

core/src/main/java/org/bitcoinj/wallet/KeyChain.java
schildbach: I get your point. Since this case, and the other two occurences, will go away with removing the deprecated methods, I actually prefer this little bit of duplicated code. A helper method will be forgotten to remove, unless marked as deprecated to begin with (which will create deprecation warnings).

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
msgilligan: > We should probably create a utility static method for this.

I meant a static utility method for seconds-to-Instant like this:

public Instant secondsToInstantWithMax(long t) {
    return t == Long.MAX_VALUE ? Instant.MAX : Instant.ofEpochSecond(t);
 }
 
 There's probably a better name.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -711,10 +711,10 @@ public int numLeafKeysIssued() {
...
     @Override
     public Instant getEarliestKeyCreationTimeInstant() {
-        if (seed != null)
-            return Instant.ofEpochSecond(seed.getCreationTimeSeconds());
-        else
-            return Instant.ofEpochSecond(getWatchingKey().getCreationTimeSeconds());
+        return Instant.ofEpochSecond(seed != null ?
+                seed.getCreationTimeSeconds() :
+                getWatchingKey().getCreationTimeSeconds()
+        );
```
msgilligan, line 717: I would probably do the white-space differently ( `?` and `:` as first character of line) but that's minor.

msgilligan (changes requested):

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
msgilligan: Github is truncating the code too much and I was unclear. I mean the entire method:

```
return (chains == null)
             ? Instant.MAX;
             : chains.stream()
                 .map(DeterministicKeyChain::getEarliestKeyCreationTimeInstant)
                 .min((t1, t2) -> t1.compareTo(t2))
                 .orElse(Instant.MAX);
```

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
schildbach: Ok, I've reverted the previous attempt and created 7ffa950cefdd28d2cf6a35bece7ef5d64c1ce9c9.

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
msgilligan: and this is probably irrelevant because "the code will go away as soon as getCreationTime() will migrate to Instant"

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
schildbach: Ok thanks for clarifying. This is again a case of easier review vs getting it right in the first place. Since this is a larger refactoring, I'm leaning towards easier review. But I'll put a commit on top so that we don't forget.

msgilligan (changes requested):
I guess there's one more requested  change here. The ternary operator starting at line 865.  (and maybe `?` and `:` on line ~715)

core/src/main/java/org/bitcoinj/wallet/KeyChainGroup.java
schildbach: Done: a0a6a116b09f2235eeebb2482011f1eb9f2c76e2

schildbach:
> I guess there's one more requested change here. The ternary operator starting at line 865. (and maybe `?` and `:` on line ~715)

The first should be adressed in a0a6a116b09f2235eeebb2482011f1eb9f2c76e2.

About the `?` and `:` I'm currently unsure. I personally like them at the end, and feel it's more readable to me, but I can see this is very subjective. Are there any objective reasons for one or the other?

msgilligan (changes requested):

msgilligan (approved):
We're down to bike shedding about where the `?`and `:` go. Let's merge this and I'll make a separate PR with the proposed change.

schildbach:
Merged.

PR #2771, opened 2023-03-07
TimeUtils: migrate mock clock to `java.time` API

schildbach:
* Migrate `setMockClock()` parameters to Instant
* Migrate `rollMockClock()` parameters to Duration
* Rename to `clearMockClock()` from `resetMocking()`
* Use `Clock` internally to provide the time

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/TransactionTest.java
+++ b/core/src/test/java/org/bitcoinj/core/TransactionTest.java
@@ -42,6 +42,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.math.BigInteger;
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.Iterator;
@@ -147,7 +148,7 @@ public class TransactionTest {
     @Test
     public void testEstimatedLockTime_WhenParameterSignifiesBlockHeight() {
         int TEST_LOCK_TIME = 20;
-        Instant now = TimeUtils.currentTime();
+        Instant now = TimeUtils.currentTime().truncatedTo(ChronoUnit.SECONDS);
 
         BlockChain mockBlockChain = createMock(BlockChain.class);
         EasyMock.expect(mockBlockChain.estimateBlockTimeInstant(TEST_LOCK_TIME)).andReturn(now);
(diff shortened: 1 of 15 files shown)


schildbach:
This is draft because again the integration tests fail.

Note I've decided to not deprecate but remove the old methods, because this API is internal and also it should only be used from tests anyway.

msgilligan:
Looks very nice, so far!

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -147,7 +148,7 @@ public void coinbaseScriptSigTooLarge() {
     @Test
     public void testEstimatedLockTime_WhenParameterSignifiesBlockHeight() {
         int TEST_LOCK_TIME = 20;
-        Instant now = TimeUtils.currentTime();
+        Instant now = TimeUtils.currentTime().truncatedTo(ChronoUnit.SECONDS);
```
msgilligan, line 151: If this use-case is common, you might want to consider a `TimeUtils.currentBlockTime()` which would return the time truncated to seconds.

core/src/test/java/org/bitcoinj/wallet/WalletTest.java
```diff
@@ -3030,15 +3031,15 @@ public void fragmentedReKeying() {
...
+        TimeUtils.rollMockClock(Duration.ofDays(1));
         for (int i = 0; i < 800; i++) {
             sendMoneyToWallet(AbstractBlockChain.NewBlockType.BEST_CHAIN, CENT, address);
         }
 
         MockTransactionBroadcaster broadcaster = new MockTransactionBroadcaster(wallet);
 
         Instant compromise = TimeUtils.currentTime().truncatedTo(ChronoUnit.SECONDS);
-        TimeUtils.rollMockClock(86400);
+        TimeUtils.rollMockClock(Duration.ofDays(1));
```
msgilligan, line 3042: above (3041) is another use case of `.truncatedTo(ChronoUnit.SECONDS)`

schildbach:
Back to the integration-test failure. What happens is that the first test in `PeerGroupTest` times out and then all others fail due to bind errors (the socket is already bound). I think this is interesting because it's an error I've seen many times on our CI, but only every so many builds.

I think now that we can reproduce this issue on all builds (even locally) we have the chance to finally squash it!

I suspect this has to do with the increased precision of millis (Date) vs nanos (Instant). Story is in development…

schildbach:
Great news, I squashed the bug. It was a stack overflow due to a method stupidly calling itself…

msgilligan (approved):
Looks great! Let's merge!

PR #2772, opened 2023-03-07
Base58,Bech32: provide public CHARSET strings

msgilligan:


schildbach:
Is this a specific need for somebody? Maybe providing a `Regex` would be more useful?

Regarding the work for output descriptors, I'd rather wait until we work on that.

schildbach:
Also, it feels counterintuitive to me that the "charset" is not the `char[]`.

msgilligan:
> Is this a specific need for somebody?

I think if we view bitcoinj as a reference implementation (our _aspiring_ to be one) it should make available representations of key parts of the BIPs like this. I think there will always be a need for these types of constants. And certainly no harm in providing them.

> Maybe providing a `Regex` would be more useful?

A Regex might be useful, but I don't think you can combine regexes at the level needed for something like output descriptors. (I might be wrong.) So I made it available as a string that I could `String.format()` into a regex.

 
> Regarding the work for output descriptors, I'd rather wait until we work on that.

That's entirely reasonable.

> Also, it feels counterintuitive to me that the "charset" is not the `char[]`.

Yeah, we could change the name(s).

msgilligan:
> > > Is this a specific need for somebody?
> > 
> > 
> > I think if we view bitcoinj as a reference implementation (our _aspiring_ to be one) it should make available representations of key parts of the BIPs like this. I think there will always be a need for these types of constants. And certainly no harm in providing them.
> 
> There is already a near-identical constant consumers can use.
> 
> I'd say we should aim to either provide the alphabet as a `String` or a `char[]`, but not both.
> 
> > > Maybe providing a `Regex` would be more useful?
> > 
> > 
> > A Regex might be useful, but I don't think you can combine regexes at the level needed for something like output descriptors. (I might be wrong.) So I made it available as a string that I could `String.format()` into a regex.
> 
> That constant could be used for forming a regex almost as easily. However, I consider it an antipattern to compose regexes like that, as without proper escaping a "regex part provider" (us, in this case) could inject regexes, similar to SQL injection. I'd rather not advocate for this code pattern.

I'm doing this (in Output Descriptor) for code-readability (and perhaps testing and maintenance) and all the strings and regexes are static constant, so I don't think an injection-attack is much of a worry.  (Note that with GraalVM `native-image` all this static initialization is done by default at compile time.)

> If we had ready-made regexes for bech32 and base58, we could perhaps use them in our parse methods. We could provide better errors ("looks like base58, but checksum isn't right").
> 
> I should add I'm not adamant on this issue. If you're convinced it's a good idea, I'm also ok with merging it.

There's no hurry on this. It was mostly done to support `OutputDescriptor` (I found it useful to use strings that I know had been unit-tested and were the correct ones rather than manually checking a duplicate string)

I'm going to mark this as DRAFT and we can pick it up again later.

PR #2773, opened 2023-03-07
AesKey: replace o.bouncycastle.c.p.KeyParameter

msgilligan:
We were using KeyParameter as a wrapper for a byte array containing an AES Key and passing it around our API. This introduced an unnecessary API dependency on Bouncy Castle. This PR adds an AesKey class for the same purpose to replace it.

This is a breaking change, but one that should be easy to accommodate.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/AesKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/AesKey.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright by the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bitcoinj.crypto;
+
+import org.bouncycastle.crypto.params.KeyParameter;
+
+/**
+ *  Wrapper for a {@code byte[]}  containing an AES Key. This is a replacement for Bouncy Castle's {@link KeyParameter} which
+ *  was used for this purpose in previous versions of <b>bitcoinj</b>. Unfortunately, this created a Gradle _api_ dependency
+ *  on Bouncy Castle when that wasn't strictly necessary.
+ *  <p>
+ *  We have made this change without deprecation because it affected many method signatures and because updating is a trivial change.
+ *  If for some reason you have code that uses the Bouncy Castle {@link KeyParameter} type and need to convert
+ *  to or from {@code AesKey}, you can temporarily use {@link #ofKeyParameter(KeyParameter)} or {@link #toKeyParameter()}
+ */
+public class AesKey {
+    private final byte[] bytes;
+
+    /**
+     * Wrapper for a {@code byte[]} containing an AES Key
+     * @param keyBytes implementation-dependent AES Key bytes
+     */
+    public AesKey(byte[] keyBytes) {
+        // Make defensive copy, so AesKey is effectively immutable
+        this.bytes = new byte[keyBytes.length];
+        System.arraycopy(keyBytes, 0, this.bytes, 0, keyBytes.length);
+    }
+
+    /**
+     * @return The key bytes
+     */
+    public byte[] bytes() {
+        return bytes;
+    }
+
+    /**
+     * Provided to ease migration from {@link KeyParameter}.
+     * @return The key bytes
+     * @deprecated Use {@link #bytes()}
+     */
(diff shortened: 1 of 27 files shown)


msgilligan:
@schildbach If this approach looks good, I'll make another pass and update all the JavaDoc.

core/src/main/java/org/bitcoinj/crypto/AesKey.java
```diff
@@ -0,0 +1,55 @@
...
+ *  to {@code AesKey}, you can use {@link #ofKeyParameter(KeyParameter)} or {@link #toKeyParameter()}
+ *  // TODO: Update JavaDoc
+ */
+public class AesKey {
+    private final byte[]  key;
+
+    public AesKey(
+            byte[]  key)
+    {
+        this.key = key;
```
schildbach, line 38: Note that the replaced class is taking ownership of the key data (copying) here.

Nitpick: the formatting could be improved around this line.

core/src/main/java/org/bitcoinj/crypto/AesKey.java
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bitcoinj.crypto;
+
+import org.bouncycastle.crypto.params.KeyParameter;
+
+/**
+ *  Wrapper for a byte array containing an AES Key. This is a replacement for Bouncy Castle's{@link KeyParameter} which
```
schildbach, line 22: Missing space.

core/src/main/java/org/bitcoinj/crypto/AesKey.java
```diff
@@ -0,0 +1,55 @@
...
+public class AesKey {
+    private final byte[]  key;
+
+    public AesKey(
+            byte[]  key)
+    {
+        this.key = key;
+    }
+
+    public byte[] getKey()
```
schildbach, line 41: Since the class name already has "key in its name, this could be named just "get()".

schildbach:
I just checked with my app and can testify that while this change breaks some code, it looks easy to fix and it indeed gets rid of BouncyCastle imports apart from my own "Crypto" class.

core/src/main/java/org/bitcoinj/crypto/AesKey.java
msgilligan: I wanted to keep the same accessor name as the replaced class. But we don't have to, we could also provide a deprecated `getKey()` method along with one with a new name.

We could also name the accessor `bytes()`.

core/src/main/java/org/bitcoinj/crypto/AesKey.java
msgilligan: > Note that the replaced class is taking ownership of the key data (copying) here.
> 

My mistake. `byte[]` is mutable and should be defensively copied.

msgilligan:
Rebased to `master`

core/src/main/java/org/bitcoinj/crypto/AesKey.java
msgilligan: @schildbach do you agree with a deprecated `getKey()` and a new accessor name of `bytes()`?

core/src/main/java/org/bitcoinj/crypto/AesKey.java
msgilligan: Fixed.

core/src/main/java/org/bitcoinj/crypto/AesKey.java
msgilligan: I'm going to assume that is ok with you and make the change. If you don't like it, let me know in the next review.

msgilligan:
@schildbach All issue addressed, force-pushed, and ready-to-review/merge.

core/src/main/java/org/bitcoinj/crypto/AesKey.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class AesKey {
+    private final byte[] bytes;
+
+    /**
+     * Wrapper for a {@code byte[]} containing an AES Key
+     * @param keyBytes implementation-dependent AES Key bytes
+     */
+    public AesKey(byte[] keyBytes)
+    {
```
schildbach, line 38: opening brace at wrong position? (3 occurences)

core/src/main/java/org/bitcoinj/crypto/AesKey.java
schildbach: Yes, that's fine as well. Actually I had this thought, too.

schildbach (commented):
Looks great!

core/src/main/java/org/bitcoinj/crypto/AesKey.java
msgilligan: I'll fix it.

msgilligan: Fixed.

msgilligan:
@schildbach OK, 3 opening-brace issues fixed.  (I'm gonna blame Bouncy Castle and IntelliJ for that 🤷‍♂️)

schildbach:
Merged!

PR #2780, opened 2023-03-09
tools/wallettool `build.gradle`: use `mainClass` not `main` in `JavaExec` task

msgilligan:
Fix Gradle 8.0 deprecation warning by changing `main` to `mainClass` in the `JavaExec` tasks.

**Note**: this change **breaks Gradle prior to 6.4.x**, so to support both ancient Gradle (currently 4.4) and Gradle 8, we're going to need to add some kind of conditional code to handle `main` vs `mainClass`.

So this will be a **DRAFT** PR for now.

You can see the failure on Github Actions because Ubuntu/JDK 11/Gradle 4.10.3 is in our test matrix: https://github.com/bitcoinj/bitcoinj/actions/runs/4370061279

### DIFF
--- a/tools/build.gradle
+++ b/tools/build.gradle
@@ -21,7 +21,7 @@ compileJava {
 
 task build_checkpoints(type: JavaExec) {
     description = 'Create checkpoint files to use with CheckpointManager.'
-    main = 'org.bitcoinj.tools.BuildCheckpoints'
+    mainClass = 'org.bitcoinj.tools.BuildCheckpoints'
     if (project.hasProperty('appArgs') && appArgs.length() > 0)
         args = Arrays.asList(appArgs.split("\\s+"))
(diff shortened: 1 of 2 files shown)


schildbach:
I'm not sure if its worth adding a conditional just to remove a deprecation warning.

msgilligan:
> I'm not sure if its worth adding a conditional just to remove a deprecation warning.

I agree. I'm just looking ahead to JDK 20.

msgilligan:
Update: Fixing this may not be necessary, I think Gradle decided to not remove this until Gradle 9.

msgilligan:
> Update: Fixing this may not be necessary, I think Gradle decided to not remove this until Gradle 9.

This fix is not necessary for Gradle 8 support. Closing.

PR #2782, opened 2023-03-09
LegacyAddressTest: separate out and move part of p2shAddressCreationFromKeys

schildbach:
This is an alternative proposal to #2769

The goal is to get rid of the `crypto` imports in LegacyAddressTest.

### DIFF
--- a/core/src/test/java/org/bitcoinj/script/ScriptPatternTest.java
+++ b/core/src/test/java/org/bitcoinj/script/ScriptPatternTest.java
@@ -19,12 +19,17 @@
 package org.bitcoinj.script;
 
 import com.google.common.collect.Lists;
+import org.bitcoinj.base.utils.ByteUtils;
+import org.bitcoinj.crypto.DumpedPrivateKey;
 import org.bitcoinj.crypto.ECKey;
 import org.junit.Test;
 
+import java.util.Arrays;
 import java.util.List;
 
+import static org.bitcoinj.base.BitcoinNetwork.MAINNET;
 import static org.bitcoinj.script.ScriptOpCodes.OP_CHECKMULTISIG;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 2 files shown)


schildbach:
A downside of this approach is that the test gets dissociated a bit from the [original test data](https://gist.github.com/gavinandresen/3966071). I'm not sure how important that is, as plain P2SH is legacy tech and nobody is going to work on these tests.

core/src/test/java/org/bitcoinj/script/ScriptPatternTest.java
```diff
@@ -92,4 +97,19 @@ public void testCreateOpReturnScript() {
...
+
+    @Test
+    public void p2shScriptHashFromKeys() {
+        // import some keys from this example: https://gist.github.com/gavinandresen/3966071
+        ECKey key1 = DumpedPrivateKey.fromBase58(MAINNET, "5JaTXbAUmfPYZFRwrYaALK48fN6sFJp4rHqq2QSXs8ucfpE4yQU").getKey();
+        key1 = ECKey.fromPrivate(key1.getPrivKeyBytes());
+        ECKey key2 = DumpedPrivateKey.fromBase58(MAINNET, "5Jb7fCeh1Wtm4yBBg3q3XbT6B525i17kVhy3vMC9AqfR6FH2qGk").getKey();
+        key2 = ECKey.fromPrivate(key2.getPrivKeyBytes());
+        ECKey key3 = DumpedPrivateKey.fromBase58(MAINNET, "5JFjmGo5Fww9p8gvx48qBYDJNAzR9pmH5S389axMtDyPT8ddqmw").getKey();
+        key3 = ECKey.fromPrivate(key3.getPrivKeyBytes());
```
msgilligan, line 109: Why is each key initialized twice?

msgilligan (approved):
Looks good. I have one question about the code, though...

core/src/test/java/org/bitcoinj/script/ScriptPatternTest.java
schildbach: No idea. I didn't look at the actual code (-:

schildbach: We can fix that in a future commit.

schildbach:
Merged.

PR #2784, opened 2023-03-09
EncryptableItem, ECKey, DeterministicKey, DeterministicSeed: migrate `creationTime` field to `java.time` API

schildbach:


schildbach:
Note this PR doesn't use defaults for creation time as discussed in #2755. Rather:

- fields are nullable
- for setters, there are variants to clear creation time
- for getters, creation time is returned as optional
- for constructors except in DeterministicSeed, there are variants without creation time
- constructors within DeterministicSeed accept nullable creation time, because I didn't find a way to make the constructor chaining work (dreaded "this() must be first" limitation)

schildbach:
Here is an excerpt of the problem to make variants with and without creation time work:

```
    public DeterministicSeed(byte[] seed, List<String> mnemonic, Instant creationTime) {
        this(seed, mnemonic);
        this.creationTime = checkNotNull(creationTime);
    }

    public DeterministicSeed(byte[] seed, List<String> mnemonic) {
        this.seed = checkNotNull(seed);
        this.mnemonicCode = checkNotNull(mnemonic);
        this.encryptedMnemonicCode = null;
        this.encryptedSeed = null;
    }

    /** @deprecated use {@link #DeterministicSeed(byte[], List, Instant)} or {@link #DeterministicSeed(byte[], List)} */
    @Deprecated
    public DeterministicSeed(byte[] seed, List<String> mnemonic, long creationTimeSecs) {
        if (creationTimeSecs > 0)
            this(seed, mnemonic, Instant.ofEpochSecond(creationTimeSecs));
        else
            this(seed, mnemonic);
    }
```

In the last deprecated constructor, I need to call one or the other non-deprecated constructors, but calling `this()` isn't allowed in a conditional. Sadly, I think this cannot expressed as a ternary op or similar – or is there some trick I don't know?

msgilligan:
> In the last deprecated constructor, I need to call one or the other non-deprecated constructors, but calling `this()` isn't allowed in a conditional. Sadly, I think this cannot expressed as a ternary op or similar – or is there some trick I don't know?

If you need to share code between constructors and the `this()`-must-be-called-first restriction is getting in your way, the standard way to solve that problem is to use a static method.

In this case I would recommend a `private` constructor that takes a nullable creation time. That will give us a public API that doesn't use nullables and support the deprecated constructor.

schildbach:
It's easier said than done, because without further hacks the private helper has the same signature as the public constructor with time. Guess I'll add fake parameters for now.

schildbach:
The above example now looks like this:

```
    public DeterministicSeed(byte[] seed, List<String> mnemonic, Instant creationTime) {
        this(seed, mnemonic, checkNotNull(creationTime), null);
    }

    public DeterministicSeed(byte[] seed, List<String> mnemonic) {
        this(seed, mnemonic, null, null);
    }

    /** @deprecated use {@link #DeterministicSeed(byte[], List, Instant)} or {@link #DeterministicSeed(byte[], List)} */
    @Deprecated
    public DeterministicSeed(byte[] seed, List<String> mnemonic, long creationTimeSecs) {
        this(seed, mnemonic, creationTimeSecs > 0 ? Instant.ofEpochSecond(creationTimeSecs) : null);
    }

    private DeterministicSeed(byte[] seed, List<String> mnemonic, @Nullable Instant creationTime, Void _NULL_) {
        this.seed = checkNotNull(seed);
        this.mnemonicCode = checkNotNull(mnemonic);
        this.encryptedMnemonicCode = null;
        this.encryptedSeed = null;
        this.creationTime = creationTime;
    }
```

schildbach:
Ok, I decided I'll seriously clean up the constructor mess first. PR coming up…

msgilligan:
> It's easier said than done, because without further hacks the private helper has the same signature as the public constructor with time. Guess I'll add fake parameters for now.

Right, the public constructor and the private constructor would be the same if the only difference is the `@Nullable` annotation (or equivalent.)

Two other possibilities (besides fake parameter for resolving this)
1. Replace the public constructor with a public static factory method. (I think the IDE will allow this as a refactor)
2. Swap the order of the parameters in the private constructor.

schildbach:
Static constructors did the trick! Implemented in #2786.

I rebased on and adapted this PR to #2786, and  I can proudly announce that all old constructors are now properly deprecated (using the annotations) and the new Instant-based constructors are either private or package-private. I'm super happy that this little step inbetween made this refactoring so relatively easy.

schildbach:
I just rebased this one last time, on current master. This is now ready for review!

I'm aware this is a rather chunky changeset. I didn't find any loose ends that would make sense to refactor separately.

core/src/main/java/org/bitcoinj/crypto/ECKey.java
```diff
@@ -152,9 +155,9 @@ public class ECKey implements EncryptableItem {
     @Nullable protected final BigInteger priv;  // A field element.
     protected final LazyECPoint pub;
 
-    // Creation time of the key in seconds since the epoch, or zero if the key was deserialized from a version that did
+    // Creation time of the key, or null if the key was deserialized from a version that did
     // not have this field.
-    protected long creationTimeSeconds;
+    @Nullable protected Instant creationTime = null;
```
msgilligan, line 160: Can this be `final` and constructors with a creation instant parameter be added. Since we're refactoring here can we make things less mutable?  (Perhaps in a separate PR)

core/src/main/java/org/bitcoinj/wallet/DeterministicSeed.java
```diff
@@ -52,39 +55,70 @@ public class DeterministicSeed implements EncryptableItem {
     @Nullable private final List<String> mnemonicCode; // only one of mnemonicCode/encryptedMnemonicCode will be set
     @Nullable private final EncryptedData encryptedMnemonicCode;
     @Nullable private final EncryptedData encryptedSeed;
-    private long creationTimeSeconds;
+    // Creation time of the seed, or null if the seed was deserialized from a version that did not have this field.
+    @Nullable private Instant creationTime = null;
```
msgilligan, line 59: Can this be `final`?

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -1207,13 +1208,13 @@ private byte[] parseAsHexOrBase58(String data) {
...
+    private Optional<Instant> getCreationTime() {
         if (unixtime != null)
-            return unixtime;
+            return Optional.of(Instant.ofEpochSecond(unixtime));
         else if (date != null)
-            return date.getTime() / 1000;
+            return Optional.of(date.toInstant());
         else
-            return 0;
+            return Optional.empty();
```
msgilligan, line 1217:
```
    return (unixtime != null)
                    ? Optional.of(Instant.ofEpochSecond(unixtime))
                    : Optional.ofNullable(date.toInstant());
```

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -1275,18 +1276,18 @@ private void printWallet(@Nullable AesKey aesKey) {
...
+                creationTime.ifPresentOrElse(seed::setCreationTime, seed::clearCreationTime);
+
         }
-        if (creationTime > 0)
-            System.out.println("Setting creation time to: " + TimeUtils.dateTimeFormat(creationTime * 1000));
-        else
-            System.out.println("Clearing creation time.");
+        creationTime.ifPresentOrElse(
+                time -> System.out.println("Setting creation time to: " + TimeUtils.dateTimeFormat(time.toEpochMilli())),
+                () -> System.out.println("Clearing creation time."));
```
msgilligan, line 1290:
```
String message = creationTime
                    .map(t -> "Setting creation time to: " + TimeUtils.dateTimeFormat(t))
                    .orElse("Clearing creation time.");
System.out.println(message);
```

msgilligan (changes requested):
Looks good. I have a few minor idiomatic suggestions.

I'm also wondering if we should try to make some of these classes less mutable (if possible) by making one or more of the nullable timestamp field(s) final based upon constructor parameters.  (Could be a separate PR)

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
msgilligan: or:
```
System.out.println(creationTime.map(t -> "Setting creation time to: " + TimeUtils.dateTimeFormat(t))
                               .orElse("Clearing creation time."));
```

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
msgilligan: or
```
return Optional.ofNullable(unixtime != null
                              ? Instant.ofEpochSecond(unixtime)
                              :  date != null
                                          ? date.toInstant()
                                          : null);
 ```

File: core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java [hunk 5]
```diff
@@ -307,8 +309,13 @@ public DeterministicKey encrypt(KeyCrypter keyCrypter, AesKey aesKey, @Nullable
...
         EncryptedData encryptedPrivateKey = keyCrypter.encrypt(privKeyBytes, aesKey);
         DeterministicKey key = new DeterministicKey(childNumberPath, chainCode, keyCrypter, pub, encryptedPrivateKey, newParent);
-        if (newParent == null)
-            key.setCreationTimeSeconds(getCreationTimeSeconds());
+        if (newParent == null) {
+            Optional<Instant> creationTime = getCreationTime();
+            if (creationTime.isPresent())
+                key.setCreationTime(creationTime.get());
+            else
+                key.clearCreationTime();
```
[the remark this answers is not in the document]
schildbach: Yes, the `ifPresent*` variants are nice!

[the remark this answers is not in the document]
schildbach: Heck, it would even boil down to:

```
getCreationTime().ifPresentOrElse(
         key::setCreationTime,
         key::clearCreationTime
     );
```

(I similar, can't remember exactly.)

I really like this!

core/src/main/java/org/bitcoinj/crypto/ECKey.java
```diff
@@ -1074,22 +1077,40 @@ public DumpedPrivateKey getPrivateKeyEncoded(NetworkParameters params) {
...
-    public void setCreationTimeSeconds(long newCreationTimeSeconds) {
-        if (newCreationTimeSeconds < 0)
-            throw new IllegalArgumentException("Cannot set creation time to negative value: " + newCreationTimeSeconds);
-        creationTimeSeconds = newCreationTimeSeconds;
+    public void setCreationTime(Instant creationTime) {
+        this.creationTime = checkNotNull(creationTime);
+    }
+
+    /**
+     * Clears the creation time of this key. This is mainly used deserialization and cloning. Normally you should not
```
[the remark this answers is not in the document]
schildbach: I said cloning, but in fact it's something like `encrypt()` / `decrypt()`, which effectively clones the entire class but encryprs/decrypts some of its fields.

core/src/main/java/org/bitcoinj/crypto/ECKey.java
schildbach: I'm open for trying this out, but in a different PR.

core/src/main/java/org/bitcoinj/wallet/DeterministicSeed.java
schildbach: I'm open for trying this out, but in a different PR.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
schildbach: I see your smiley (-:

Honestly, without a reason (for example being forced to use only functions in a constructor) I prefer the if-else-else. And in any case, I would change that in a followup commit so that this one stays as small and reviewable as possible.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
schildbach: I see, true functional style… just learned another bit, thanks!

I changed to your second example and pushed a commit on top.

schildbach:
I think the mutability and "how to clone" change is out of scope for this PR. (But I'm all for immutable classes!)

schildbach:
Rebased and merged.

PR #2785, opened 2023-03-09
MnemonicCode: in `toMnemonic()` throw unchecked rather than checked exception

schildbach:
When converting from entropy to mnemonic with `toMnemonic()`, a `MnemonicLengthException` was thrown for entropies with invalid length. This is bad for two reasons:

- it is a checked exception, and the method will not be called with user input
- misleading exception name, the error is about the entropy length not the mnemonic

We now throw a `RuntimeException` instead. The resulting simplification of calling code can be witnessed in `DeterministicSeed`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/MnemonicCode.java
+++ b/core/src/main/java/org/bitcoinj/crypto/MnemonicCode.java
@@ -37,6 +37,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


msgilligan (approved):
This looks good!

Suggestion: The JavaDoc for `@param entropy`  should probably also be updated to specify the minimum length in bytes (or bits) and `@param entropy` should be added to `toMnemonic()`.

schildbach:
Good idea! I added/edited both `@param entropy` as suggested.

schildbach:
Merged.

PR #2786, opened 2023-03-09
DeterministicSeed: introduce static constructors meant for external consumers

schildbach:
* native constructors that were meant to be called by consumers now have static equivalents
* all native constructors are discouraged from being used by a JavaDoc comment
* one constructor that wasn't meant to be used by consumers is switched to private access
* tests and tools that resemble consumers are updated to use the new API

Note: although most native constructors are now deprecated, we're not using annotations because we're still calling them.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
@@ -308,10 +308,10 @@ public class DeterministicKeyChain implements EncryptableKeyChain {
 
             if (random != null)
                 // Default passphrase to "" if not specified
-                return new DeterministicKeyChain(new DeterministicSeed(random, bits, getPassphrase()), null,
+                return new DeterministicKeyChain(DeterministicSeed.fromRandom(random, bits, getPassphrase()), null,
                         outputScriptType, accountPath);
             else if (entropy != null)
-                return new DeterministicKeyChain(new DeterministicSeed(entropy, getPassphrase(), creationTimeSecs),
+                return new DeterministicKeyChain(DeterministicSeed.fromEntropy(entropy, getPassphrase(), creationTimeSecs),
                         null, outputScriptType, accountPath);
             else if (seed != null)
                 return new DeterministicKeyChain(seed, null, outputScriptType, accountPath);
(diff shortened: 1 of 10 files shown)


schildbach:
Note this PR is before the creationTime refactoring – hence the `long` creation times.

msgilligan:
> Note: although most native constructors are now deprecated, we're not using annotations because we're still calling them.

Did you try IntelliJ's constructor-to-static-factory refactor?

schildbach:
> > Note: although most native constructors are now deprecated, we're not using annotations because we're still calling them.
> 
> Did you try IntelliJ's constructor-to-static-factory refactor?

No, why?

msgilligan:
One other note: Since making this refactoring first, you'll probably need to remember that you don't need to keep the factory methods that take `long seconds` since you're going to deprecate those constructors anyways and point to the factory methods.

msgilligan:
> > Did you try IntelliJ's constructor-to-static-factory refactor?
> 
> No, why?

Because you said there were still some uses of the old constructors. The refactoring should have made it easier to convert all those.

msgilligan:
Here's the doc for it: https://www.jetbrains.com/help/idea/replace-constructor-with-factory-method.html

schildbach:
Thoses uses are from within bitcoinj, but from other classes and packages. It will be difficult even for IntelliJ to restrict access to it. Or add annotations, some problem.

schildbach:
Or lets put it this way: static constructors can't actually construct. They need native constructors to do that. You can't make a deprecated native constructor call a static constructor for backwards compatibility, like we do with all of the normal deprecated methods.

schildbach:
I pushed the suggested rename of static constructors on top.

msgilligan:
> Or lets put it this way: static constructors can't actually construct. They need native constructors to do that. You can't make a deprecated native constructor call a static constructor for backwards compatibility, like we do with all of the normal deprecated methods.

Look closely at what the refactoring does.  Then imagine doing the IntelliJ refactoring  as the first step, to get all the external usages calling the static constructor. Then adjust the native constructors as needed manually.

I might be missing something here and I haven't looked to see how many external calls to the remains public constructors remain. I just thought the first step above could easily convert all external usages to the factory methods.

msgilligan (approved):
Approved.

schildbach:
I could have saved some time using the IntelliJ refactoring, but in the end it doesn't add the annotations, which is the only thing "missing" in this PR.

But the good news is: #2784 frees the old native constructors of all references, so it now properly deprecates them. And the new variant (with the nullable Instants) are all private or package-private)!

schildbach:
Merged.

PR #2788, opened 2023-03-10
Wallet: migrate `updateTime` field to `java.time` API

schildbach:


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -90,9 +92,9 @@ public class Transaction extends ChildMessage {
...
     public static final Comparator<Transaction> SORT_TX_BY_UPDATE_TIME = (tx1, tx2) -> {
-        final long time1 = tx1.getUpdateTime().getTime();
-        final long time2 = tx2.getUpdateTime().getTime();
-        final int updateTimeComparison = -(Long.compare(time1, time2));
+        Instant t1 = tx1.getUpdateTimeInstant().orElse(Instant.EPOCH);
+        Instant t2 = tx2.getUpdateTimeInstant().orElse(Instant.EPOCH);
+        final int updateTimeComparison = -(t1.compareTo(t2));
         //If time1==time2, compare by tx hash to make comparator consistent with equals
         return updateTimeComparison != 0 ? updateTimeComparison : tx1.getTxId().compareTo(tx2.getTxId());
     };
```
msgilligan, line 100: This can probably be simplified (later) with `Comparators.Comparing`

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -520,18 +522,40 @@ public boolean isEveryOwnedOutputSpent(TransactionBag transactionBag) {
 
     /**
      * Returns the earliest time at which the transaction was seen (broadcast or included into the chain),
-     * or the epoch if that information isn't available.
+     * or empty if that information isn't available.
      */
+    public Optional<Instant> getUpdateTimeInstant() {
```
msgilligan, line 527: `getUpdateInstant()`?

msgilligan (approved):
One suggestion for the future, one optional suggestion for this commit.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: I actually tried that, but so far failed:

```
    public static final Comparator<Transaction> SORT_TX_BY_UPDATE_TIME__ = Comparator
            .comparing(tx -> tx.getUpdateTimeInstant().orElse(Instant.EPOCH), Comparator.reverseOrder())
            .thenComparing(tx -> tx.getTxid());
```

As soon as I add the ".thenComparing", it doesn't recognize tx as type Transaction and I get "Cannot resolve method 'getUpdateTimeInstant' in 'Object'" on the two getters.

core/src/main/java/org/bitcoinj/core/Transaction.java
msgilligan: I don't think we've been entirely consistent on this, then -- I'm pretty sure we already have cases with just `.getXxxInstant()`.  I think you'll be able to find them well enough with `/get*Instant()` and with the shorter names it's ok if we don't make the second change right away.

Either way is fine with me, though.

PR #2790, opened 2023-03-10
gradle.yml: set log level to info

schildbach:
This is meant to look at verbose build output.

It should not be merged.


schildbach:
After a bit of discussion, we merged this. If it's too much, we can disable it any time.

PR #2792, opened 2023-03-10
Transaction: reimplement `SORT_TX_BY_UPDATE_TIME` and `SORT_TX_BY_HEIGHT` comparators in a functional way

schildbach:


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -87,29 +87,25 @@
...
-        final int height2 = confidence2.getConfidenceType() == ConfidenceType.BUILDING
-                ? confidence2.getAppearedAtChainHeight() : Block.BLOCK_HEIGHT_UNKNOWN;
-        final int heightComparison = -(Integer.compare(height1, height2));
-        //If height1==height2, compare by tx hash to make comparator consistent with equals
-        return heightComparison != 0 ? heightComparison : tx1.getTxId().compareTo(tx2.getTxId());
-    };
+    public static final Comparator<Transaction> SORT_TX_BY_HEIGHT =
+            Comparator.<Transaction, Integer>comparing(
+                            tx -> tx.getConfidence().getConfidenceType() == ConfidenceType.BUILDING ?
+                                    tx.getConfidence().getAppearedAtChainHeight() :
```
schildbach, line 104: On these lines I'm get `getConfidence()` twice. Previously, I was caching it in a local variable. How can I do this with functions?

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -87,29 +87,25 @@
...
-        Instant t2 = tx2.getUpdateTimeInstant().orElse(Instant.EPOCH);
-        final int updateTimeComparison = -(t1.compareTo(t2));
-        //If time1==time2, compare by tx hash to make comparator consistent with equals
-        return updateTimeComparison != 0 ? updateTimeComparison : tx1.getTxId().compareTo(tx2.getTxId());
-    };
+    public static final Comparator<Transaction> SORT_TX_BY_UPDATE_TIME =
+            Comparator.<Transaction, Instant>comparing(
+                            tx -> tx.getUpdateTimeInstant().orElse(Instant.EPOCH),
+                            Comparator.reverseOrder())
+                    .thenComparing(SORT_TX_BY_ID);
```
msgilligan, line 99: I would extract `tx -> tx.getUpdateTimeInstant().orElse(Instant.EPOCH)` to a private, static method. Maybe you can come up with a better name for the private method. I suppose the method could also be added as an instance method if that makes more sense.

```
    public static final Comparator<Transaction> SORT_TX_BY_UPDATE_TIME =
            Comparator.comparing(Transaction::txUpdateTimeOrElseEpoch, Comparator.reverseOrder())
                        .thenComparing(SORT_TX_BY_ID);
                        
    private static Instant txUpdateTimeOrElseEpoch(Transaction tx) {
        return tx.getUpdateTimeInstant().orElse(Instant.EPOCH);
    }
```

msgilligan: Note that the extracted method eliminates the need for `<Transaction, Instant>`

core/src/main/java/org/bitcoinj/core/Transaction.java
msgilligan: See below...

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -87,29 +87,25 @@
...
-        return heightComparison != 0 ? heightComparison : tx1.getTxId().compareTo(tx2.getTxId());
-    };
+    public static final Comparator<Transaction> SORT_TX_BY_HEIGHT =
+            Comparator.<Transaction, Integer>comparing(
+                            tx -> tx.getConfidence().getConfidenceType() == ConfidenceType.BUILDING ?
+                                    tx.getConfidence().getAppearedAtChainHeight() :
+                                    Block.BLOCK_HEIGHT_UNKNOWN,
+                            Comparator.reverseOrder())
+                    .thenComparing(SORT_TX_BY_ID);
+
```
msgilligan, line 108: Here the extracted method helps by:
1.  Caches the `TransactionConfidence` in a local variable
2. Eliminates the need for the `<Transaction, Integer>`
3. Makes the compactor more readable

Again, you might want a better method name. And it also would be possible to add this as an instance method.

```
    public static final Comparator<Transaction> SORT_TX_BY_HEIGHT =
            Comparator.comparing(Transaction::txGetHeight, Comparator.reverseOrder())
                        .thenComparing(SORT_TX_BY_ID);

    private static int txGetHeight(Transaction tx) {
        TransactionConfidence conf = tx.getConfidence();
        return conf.getConfidenceType() == ConfidenceType.BUILDING
                ? conf.getAppearedAtChainHeight()
                : Block.BLOCK_HEIGHT_UNKNOWN;
    }
```

core/src/main/java/org/bitcoinj/core/Transaction.java
msgilligan: Note that you can also use the `return` form Lambda:
```
tx -> {
   var cache = xyz;
   computation;
    return <integer>
}

msgilligan (changes requested):

core/src/main/java/org/bitcoinj/core/Transaction.java
msgilligan: The helper method also lets you write a comment, which I lazily didn't do.

schildbach:
> return form Lambda

Cool – I already forgot this exists!

schildbach:
Thanks for your help!

Merged.

PR #2797, opened 2023-03-12
WIP: Refactor Bech32, support arbitrary byte[] encodings

msgilligan:


msgilligan:
TODO:

* Add more JavaDoc
* Try to come up with better names for classes and methods
* Decide to what extent we want to globally use `ByteArray` (or similar class) as a generic wrapper for `byte[]`
* Consider visibility changes: public, package, private
* Consider issues of backward compatibility more thoroughly (e.g. any deprecations, etc.)
* Consider breaking into smaller commits

schildbach:
Would the Nostr test vectors also work with the current code? In that case, we could merge these tests right away.

core/src/main/java/org/bitcoinj/base/Bech32.java
```diff
@@ -51,15 +54,71 @@ public class Bech32 {
...
+        /**
+         * Wrapper for a {@code byte[]} array.
+         *
+         * @param bytes bytes to be copied
+         */
+        protected Bech32Bytes(byte[] bytes) {
+            super(bytes);
+        }
+
+        private Bech32Bytes(byte first, byte[] rest) {
```
schildbach, line 67: I think the plain Bech32 spec (just the ["Bech" chapter](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#user-content-Bech32) of BIP173), and it doesn't have this "first byte + rest" concept. It's just hrp, separator (irrelevant) and data.

So I think a Bech32Bytes class should probably not have this, at least not without a good reason. Maybe a more specialized "SegwitBech32Bytes" class on top would make sense?

core/src/main/java/org/bitcoinj/base/internal/ByteArray.java
```diff
@@ -0,0 +1,54 @@
+/*
```
schildbach, line 1: Do we already have a second use (besides Bech32Bytes) for this? If not, we could start by folding this into Bech32Bytes – especially if we choose to extract the "first byte + rest" into a SegwitBech32Bytes class (see comment above).

core/src/main/java/org/bitcoinj/base/Bech32.java
schildbach: I just read through the Nostr ["bech32-encoded entities" spec](https://github.com/nostr-protocol/nips/blob/master/19.md).

It seems to not have any mention of "first byte" or "version byte". It also always uses Bech32 (_not_ "m") encoding. Segwit ties the encoding strictly to the version byte: 0 uses Bech32, >0 uses Bech32m.

msgilligan: > … the ["Bech" chapter](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#user-content-Bech32) of BIP173, doesn't have this "first byte + rest" concept.

True, but this is really just an alternate representation of a `byte[]`. So it doesn't really violate in "layering" principles in the spec.

> without a good reason

This is a convenience constructor (that simplifies the code and avoids a byte[] copy) for the Segwit implementation but without using any Segwit types.

Note that it is `private`.

> Maybe a more specialized "SegwitBech32Bytes" or "VersionedBech32Bytes" class on top would make sense?

Personally, I think we have enough classes in this implementation already and feel the convenience constructor is the simpler solution.

Maybe trying to eliminate the copy is unnecessary optimization (though there are applications that might deal with thousands or millions of `Address` and need to serialize them.)

I originally coded it without this method and added it later, so if you insist we could break it into multiple commits and maybe when you see it you'll agree with me. I also consider having the first parameter be a `short` or an `int` instead of a `byte`, but that seems to violate "layering" a little more.

msgilligan: > the Nostr ["bech32-encoded entities" spec](https://github.com/nostr-protocol/nips/blob/master/19.md).

> … seems to not have any mention of "first byte" or "version byte".

That is true. This constructor was added for the refactored `SegwitAddress` implementation.

core/src/main/java/org/bitcoinj/base/internal/ByteArray.java
msgilligan: > Do we already have a second use?

`AesKey` is the one that was on my mind, and we're considering creating a similar wrapper for `ECKey`.

I think there are probably plenty of other cases where having a typed, immutable `byte[]`-wrapper class would be handy. It could also have some utility methods like `formatHex` that are currently in `ByteUtils`.

msgilligan:
> Would the Nostr test vectors also work with the current code? In that case, we could merge these tests right away.

The current code has the methods `encode8to5`, `decode5to8`, and `convertBits` in `SegwitAddress` as `private`. This refactoring includes:

> * Move/refactor utility methods from SegwitAddress to `Bech32`

... which would be the bare minimum for supporting the Nostr test vectors.

> * Add `encodeBytes` and `decodeBytes` to `Bech32` for encoding/decoding arbitrary byte[]

Provides an API for the more general (including Nostr NIP-19) use case. And:

* Add `Bech32.Bech32Bytes` class for wrapping Bech32 5-bit byte arrays.

... provides some type safety for the 5-bit format arrays we were using. (Which would have helped with my confusion in my first attempt to use `Bech32.java` for NIP-19 encoding.)

core/src/main/java/org/bitcoinj/base/Bech32.java
schildbach: Ok, I can live with this one. Bech32 is close enough to segwit that some of its artifacts can spill over.

But there is also its cousin in ByteArray, see my other comment.

msgilligan: Fixed in new commits.

schildbach:
Looks good!

schildbach:
> Consider issues of backward compatibility more thoroughly (e.g. any deprecations, etc.)

I believe Bech32Bytes has been changed in this cycle already. Not sure if it's worth adding deprecated methods. Do you have any in mind?

> Try to come up with better names for classes and methods

Looks good to me.

> Consider visibility changes: public, package, private

I would perhaps investigate if the package-private methods can be changed to private or protected.

> Consider breaking into smaller commits

It's fine as is.

> Add more JavaDoc

Yes, I think the private/package-private methods in Bech32Bytes could use a bit of explanation about what they do.

schildbach:
Not sure if it catches on, but there is a proposal for an encoding very similar, but not identical to Bech32: Codex32. Apparently, it's designed for encoding/decoding with pen & paper.

https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2023-February/021469.html

I'm just mentioning this because it uses some of the building blocks of Bech32, so we could perhaps share code.

msgilligan:
I'm ready to try to finish this one now. First step: PR #3088

msgilligan:
> > Consider issues of backward compatibility more thoroughly (e.g. any deprecations, etc.)
> 
> I believe Bech32Bytes has been changed in this cycle already. Not sure if it's worth adding deprecated methods. Do you have any in mind?

None in particular at this point. Marked as completed.

> 
> > Consider visibility changes: public, package, private
> 
> I would perhaps investigate if the package-private methods can be changed to private or protected.

I made a few methods private, where possible.  I also made `decode5to8()` `public` because I it is needed for the likely common use-case of decoding to `Bech32Data`, checking the encoding and/or HRP and then needed access to the fully decoded data.

> > Add more JavaDoc
> 
> Yes, I think the private/package-private methods in Bech32Bytes could use a bit of explanation about what they do.

Done.

msgilligan:
OK, I rebased and force-pushed to update to the latest master. Then I updated the JavaDoc and adjusted visibility of some methods and did a second force-push.

Ready-for-review.

406712774:
No kiero nada de estos correos señor Gatito por favor así como usted no
Kiere ni verme porque se enferma pues así igualmente yo tampoco kiero Saver
nada de ustedn dígame k tengo k aser para no volver a saber niver nada de
usted nada más dígamelo y yo lo ago contal de no volver a saber nada de
usted tampoco kiero ver su nombre en ningún lado Demi vida

PR #2801, opened 2023-03-13
BuildCheckpoints: migrate to `java.time` API

schildbach:


tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
```diff
@@ -135,17 +138,18 @@ public Integer call() throws Exception {
...
 
-        final long timeAgo = now - (86400 * days);
-        System.out.println("Checkpointing up to " + TimeUtils.dateTimeFormat(timeAgo * 1000));
+        Instant timeAgo = now.minus(days, ChronoUnit.DAYS);
+        System.out.println("Checkpointing up to " + TimeUtils.dateTimeFormat(timeAgo.toEpochMilli()));
 
         chain.addNewBestBlockListener(Threading.SAME_THREAD, block -> {
             int height = block.getHeight();
-            if (height % params.getInterval() == 0 && block.getHeader().getTimeSeconds() <= timeAgo) {
+            if (height % params.getInterval() == 0 && block.getHeader().getTimeInstant().compareTo(timeAgo) <= 0) {
```
msgilligan, line 149: I would use `isBefore()` or `isAfter()` here. Although using `compareTo` and `<=` looks like it's closer to the old code/easier to compare, it's actually different (I briefly thought the `0` value was replacing `timeAgo` with epoch seconds.)

msgilligan (changes requested):
Minor change requested.

tools/src/main/java/org/bitcoinj/tools/BuildCheckpoints.java
schildbach: Ok, I'll rewrite it with `isAfter()` (and the operands swapped).

schildbach:
Merged.

PR #2803, opened 2023-03-13
VersionMessage: migrate `time` field to `java.time` API

schildbach:


core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -388,8 +388,7 @@ public String toString() {
         String servicesStr = Strings.emptyToNull(VersionMessage.toStringServices(vPeerVersionMessage.localServices));
         helper.add("services",
                 vPeerVersionMessage.localServices + (servicesStr != null ? " (" + servicesStr + ")" : ""));
-        long peerTime = vPeerVersionMessage.time * 1000;
-        helper.add("time", String.format(Locale.US, "%tF %tT", peerTime, peerTime));
+        helper.add("time", TimeUtils.dateTimeFormat(vPeerVersionMessage.time .toEpochMilli()));
```
msgilligan, line 391: Looks like there's an unwanted space before `.toEpachMilli()`

msgilligan (changes requested):
One minor whitespace issue that I think you'll want to fix.

core/src/main/java/org/bitcoinj/core/Peer.java
schildbach: Fixed.

schildbach:
Merged.

PR #2805, opened 2023-03-14
base: extinguish unwanted references from main classpath

schildbach:
The idea of this patchset is to check if base has no unwanted dependencies to libraries or bitcoinj modules.

It was created this way:

- Remove all references of non-base packages (imports, fully qualified names)
- Try fix "cannot resolve symbol" compiler errors only by removing deprecated methods or variables

Optional: After the next release, we could merge this.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/Address.java
+++ b/core/src/main/java/org/bitcoinj/base/Address.java
@@ -16,9 +16,6 @@
 
 package org.bitcoinj.base;
 
-import org.bitcoinj.crypto.ECKey;
-import org.bitcoinj.core.NetworkParameters;
-
 import java.util.Comparator;
 
 /**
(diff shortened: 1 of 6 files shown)


msgilligan:
Next task for me after the Bech32 stuff is to go after these remaining uses of `NetworkParameters`.

Note that this patch set does not verify a lack of external dependencies (e.g. Guava) in `base`.

schildbach:
Damn, we still have these usages in `DefaultAddressParserProvider`. I should not have assigned this PR to 0.18, moving to 0.17 even though the merge is planned for 0.18.

The milestone is only meant for resolving the remaining usages that prevent us from cutting the base module in 0.18.

msgilligan:
After PR #3464 is merged, you should be able to rebase this PR and it should pass tests.

msgilligan (approved):
Yay!

schildbach:
Merged.

PR #2815, opened 2023-03-15
WIP: TimeLock: refactor to subtypes

msgilligan:
This is to support pattern-matching in JDK 16 and later and supports testing with `instanceof` on all JDKs (but may require casting after the tests on pre-16 JDKs)

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/LockTime.java
+++ b/core/src/main/java/org/bitcoinj/core/LockTime.java
@@ -20,8 +20,6 @@ import org.bitcoinj.base.internal.TimeUtils;
 
 import java.time.Instant;
 import java.util.Objects;
-import java.util.Optional;
-import java.util.OptionalInt;
 
 /**
  * Wrapper for transaction lock time, specified either as a block height or as a timestamp (in seconds
@@ -54,14 +52,6 @@ public abstract /* sealed */ class LockTime {
         }
     }
 
-    // TODO: Should we have a separate NoLock case, or should it be HeightLock with value ZERO?
-    public static final class NoLock extends LockTime {
-        private static final NoLock NO_LOCK = new NoLock();
-        public NoLock() {
-            super(0);
-        }
-    }
-
     /**
      * Wrap a raw value (as used in the Bitcoin protocol) into a lock time.
      * @param rawValue raw value to be wrapped
@@ -70,9 +60,7 @@ public abstract /* sealed */ class LockTime {
     public static LockTime of(long rawValue) {
         if (rawValue < 0)
             throw new IllegalArgumentException("illegal negative lock time: " + rawValue);
-        return rawValue == 0
-                ? NoLock.NO_LOCK
-                : rawValue < LockTime.THRESHOLD
+        return rawValue < LockTime.THRESHOLD
                     ? new HeightLock(rawValue)
                     : new TimeLock(rawValue);
     }
@@ -106,13 +94,13 @@ public abstract /* sealed */ class LockTime {
      * Construct an unset lock time.
      * @return unset lock time
      */
-    public static NoLock unset() {
-        return NoLock.NO_LOCK;
+    public static LockTime unset() {
+        return LockTime.ofBlockHeight(0);
     }
 
     /**
      * Raw values below this threshold specify a block height, otherwise a timestamp in seconds since epoch.
-     * Consider using {@link #isBlockHeight()} or {@link #isTimestamp()} before using this constant.
+     * Consider using {@code lockTime instance of HeightLock} or {@code lockTime instance of TimeLock} before using this constant.
      */
     public static final long THRESHOLD = 500000000; // Tue Nov  5 00:53:20 1985 UTC
 
@@ -131,46 +119,12 @@ public abstract /* sealed */ class LockTime {
     }
 
     /**
-     * The lock time is considered to be set only if its raw value is greater than zero.
+     * This is equivalent to {@code lockTime instanceof HeightLock || lockTime instanceof TimeLock}.
+     * (The lock time is considered to be set only if its raw value is greater than zero.)
      * @return true if lock time is set
      */
     public boolean isSet() {
-        return !(this instanceof NoLock);
-    }
-
-    /**
-     * Determine if this lock time is specified as a block height. That means its raw value is below {@link #THRESHOLD}.
-     * TODO: this maps NoLock into a BlockLock. I'm not sure this is what we want, but I think some use-cases need this.
-     * @return true if specified as a block height
-     */
-    public boolean isBlockHeight() {
-        return this instanceof HeightLock || this instanceof NoLock;
-    }
-
-    /**
-     * Gets the lock time as a block height.
-     * @return lock time as a block height
-     */
-    public OptionalInt getBlockHeight() {
-        return isBlockHeight() ? OptionalInt.of(((HeightLock)this).blockHeight()) : OptionalInt.empty();
-    }
-
-    /**
-     * Determine if this lock time is specified as a timestamp in seconds since epoch. That means its raw value is
-     * equal or above {@link #THRESHOLD}.
-     * @return true if specified as a timestamp
-     */
-    public boolean isTimestamp() {
-        return this instanceof TimeLock;
-    }
-
-    /**
-     * Gets the lock time as a timestamp.
-     * @return lock time as a timestamp
-     * @throws IllegalStateException if the lock time is not specified as a timestamp
-     */
-    public Optional<Instant> getTimestamp() {
-        return isTimestamp() ? Optional.of(((TimeLock) this).timestamp()) : Optional.empty();
+        return value > 0;
     }
 
     @Override
@@ -187,14 +141,8 @@ public abstract /* sealed */ class LockTime {
 
     @Override
     public String toString() {
-        String result;
-        if (this instanceof NoLock) {
-            result = "no lock";
-        } else if (this instanceof HeightLock) {
-            result =  "block " + ((HeightLock)this).blockHeight();
-        } else {
-            result = TimeUtils.dateTimeFormat(((TimeLock)this).timestamp());
-        }
-        return result;
+        return this instanceof HeightLock
+                ? "block " + ((HeightLock)this).blockHeight()
+                : TimeUtils.dateTimeFormat(((TimeLock)this).timestamp());
     }
 }
--- a/core/src/main/java/org/bitcoinj/core/LockTime.java
+++ b/core/src/main/java/org/bitcoinj/core/LockTime.java
@@ -32,7 +32,6 @@ import java.util.Objects;
 public abstract /* sealed */ class LockTime {
 
     public static final class HeightLock extends LockTime {
-
         public HeightLock(long value) {
             super(value);
         }
@@ -119,8 +118,8 @@ public abstract /* sealed */ class LockTime {
     }
 
     /**
-     * This is equivalent to {@code lockTime instanceof HeightLock && rawValue() > 0 || lockTime instanceof TimeLock}.
-     * (The lock time is considered to be set only if its raw value is greater than zero.)
+     * The lock time is considered to be set only if its raw value is greater than zero.
+     * In other words, it is set if it is either a non-zero block height or a timestamp.
      * @return true if lock time is set
      */
     public boolean isSet() {
--- a/core/src/main/java/org/bitcoinj/core/LockTime.java
+++ b/core/src/main/java/org/bitcoinj/core/LockTime.java
@@ -20,6 +20,8 @@ import org.bitcoinj.base.internal.TimeUtils;
 
 import java.time.Instant;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.OptionalInt;
 
 /**
  * Wrapper for transaction lock time, specified either as a block height or as a timestamp (in seconds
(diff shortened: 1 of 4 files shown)


msgilligan:
Even though we are still supporting JDK 8, it will allow apps using the latest JDKs to do stuff like this:
```
public class TestTimeLock {
    @MethodSource("testParams")
    @ParameterizedTest(name = "Locktime {0} = value {1}")
    void canConstruct(Locktime locktime, long expectedValue) {

        long value = switch (locktime) {
            case TimeLock   t -> t.timestamp().getEpochSecond();
            case HeightLock h -> h.blockHeight();
            case NoLock     n -> 0;
            case null         -> throw new IllegalArgumentException("null");        // This is optional and useful when null-checking is needed.
            default           -> throw new IllegalStateException("can't happen");   // With Sealed Classes this is eliminated
        };

        assertEquals(value, expectedValue);
    }

    private static Stream<Arguments> testParams() {
        return Stream.of(
                // Note: For BIP32 wallets the Network does not affect the path
                Arguments.of(Locktime.of(0), 0),
                Arguments.of(Locktime.of(Locktime.THRESHOLD-1),     499_999_999),
                Arguments.of(Locktime.of(Locktime.THRESHOLD),            500_000_000),
                Arguments.of(Locktime.ofBlockHeight(1),                             1),
                Arguments.of(Locktime.ofBlockHeight(Locktime.THRESHOLD-1), 499_999_999),
                Arguments.of(Locktime.ofTime(Instant.EPOCH.plus(365 * 50, ChronoUnit.DAYS)), 1576800000),
                Arguments.of(Locktime.ofTime(Instant.MAX),            31556889864403199L)
        );
    }
}
```

core/src/main/java/org/bitcoinj/core/LockTime.java
```diff
@@ -29,7 +31,37 @@
...
+        public TimeLock(long value) {
+            super(value);
+        }
+
+        public Instant timestamp() {
+            return Instant.ofEpochSecond(value);
+        }
+    }
+
+    // TODO: Should we have a separate NoLock case, or should it be HeightLock with value ZERO?
```
schildbach, line 57: Since you're basing this PR on a commit with "unset == blockHeight 0", I'd pick the second choice.

core/src/main/java/org/bitcoinj/core/LockTime.java
```diff
@@ -38,40 +70,44 @@ public class LockTime {
     public static LockTime of(long rawValue) {
         if (rawValue < 0)
             throw new IllegalArgumentException("illegal negative lock time: " + rawValue);
-        return new LockTime(rawValue);
+        return rawValue == 0
+                ? NoLock.NO_LOCK
+                : rawValue < LockTime.THRESHOLD
+                    ? new HeightLock(rawValue)
+                    : new TimeLock(rawValue);
```
schildbach, line 77: …it would also get rid of the nesting.

msgilligan:
TODO:

1. Decide whether `NoLock` should be its own type or just `HeightLock` with a height of zero
2. Decide whether to include the `Optional`/`OptionalInt` getters
3. Decide which idiom(s) to use for handling conditionals in various places (e.g. ternary, if/else w/`instanceof`, if/else w/`.isXxx`, etc.)
4. Review naming choices
5. Consider removing the `.isXyz()` methods
6. Any updates to the test
7. JavaDoc

core/src/main/java/org/bitcoinj/core/LockTime.java
```diff
@@ -99,26 +135,24 @@ public long rawValue() {
...
     /**
      * Gets the lock time as a block height.
      * @return lock time as a block height
-     * @throws IllegalStateException if the lock time is not specified as a block height
      */
-    public int blockHeight() {
-        if (!isBlockHeight())
-            throw new IllegalStateException("lock time not specified as a block height");
-        return Math.toIntExact(value);
+    public OptionalInt getBlockHeight() {
```
schildbach, line 154: `blockHeight()`?

core/src/main/java/org/bitcoinj/core/LockTime.java
```diff
@@ -127,18 +161,16 @@ public int blockHeight() {
...
     /**
      * Gets the lock time as a timestamp.
      * @return lock time as a timestamp
      * @throws IllegalStateException if the lock time is not specified as a timestamp
      */
-    public Instant timestamp() {
-        if (!isTimestamp())
-            throw new IllegalStateException("lock time not specified as a timestamp");
-        return Instant.ofEpochSecond(value);
+    public Optional<Instant> getTimestamp() {
```
schildbach, line 172: `timestamp()`?

core/src/main/java/org/bitcoinj/core/LockTime.java
```diff
@@ -155,8 +187,14 @@ public int hashCode() {
...
-                TimeUtils.dateTimeFormat(timestamp());
+        String result;
+        if (this instanceof NoLock) {
+            result = "no lock";
+        } else if (this instanceof HeightLock) {
+            result =  "block " + ((HeightLock)this).blockHeight();
+        } else {
+            result = TimeUtils.dateTimeFormat(((TimeLock)this).timestamp());
+        }
+        return result;
```
schildbach, line 198: Sadly we can't use pattern matching here.

I would use 3 return statements and get rid of 2 lines…

core/src/test/java/org/bitcoinj/core/LockTimeTest.java
```diff
@@ -73,11 +76,32 @@ public void unset() {
 
     @Test(expected = IllegalStateException.class)
     public void blockHeight_mismatch() {
-        LockTime.ofTimestamp(Instant.MAX).blockHeight();
+        LockTime.ofTimestamp(Instant.MAX).getBlockHeight().orElseThrow(IllegalStateException::new);
```
schildbach, line 79: `.orElseThrow()`  requires Java 9 API I think. Maybe we should make a clear statement about the tests in core. I wouldn't mind switching both language features and API officially over to Java 11.

core/src/test/java/org/bitcoinj/core/LockTimeTest.java
```diff
@@ -73,11 +76,32 @@ public void unset() {
...
+        if (timestamp instanceof TimeLock) {
+            assertTrue(((TimeLock) timestamp).timestamp().isAfter(Instant.EPOCH));
+        }
+        if (height instanceof HeightLock) {
+            assertTrue(((HeightLock) height).blockHeight() > 0);
+        }
+        if (unset instanceof NoLock) {
+            assertTrue(unset.rawValue() == 0);
+        }
     }
```
schildbach, line 106: Maybe use 3 separate tests?

schildbach (commented):
Generally this looks okay.

The question remains, should we keep the `is*()` methods (as public)?

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: We don't want this to conflict with the `.blockHeight()` method in the subclass. That's why I added the `get`...

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: Again, we don't want to conflict with the method in the subclass.  So the `get` implies some level of indirection.

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: Fair enough. We should probably have a conversation about the pros and cons of the various idioms we have available pre-pattern-matching and try to standardize a little. I deliberately used several of the alternatives in the PR just for comparison.

core/src/test/java/org/bitcoinj/core/LockTimeTest.java
msgilligan: Java 9 provides `.orElseThrow()` with no arguments. The above is valid Java 8.

core/src/test/java/org/bitcoinj/core/LockTimeTest.java
msgilligan: I could probably write more than three...

msgilligan:
> Generally this looks okay.

So, I'l take that as a "concept ack"?

> The question remains, should we keep the `is*()` methods (as public)?

I guess I'm leaning towards removing them or making them private.

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: But you might be open to three subclasses in a subsequent PR?

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: ... and of course we could remove this altogether.

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: ... and of course we could remove this altogether.

schildbach:
> So, I'l take that as a "concept ack"?

The concept is fine, but the use for the typical JDK 8 / 11 crowd is limited. At first, I misunderstood you to somehow retrofit language features to old Java.

> I guess I'm leaning towards removing them or making them private.

Me too.

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: I'm reducing to two subtypes for this PR.

core/src/main/java/org/bitcoinj/core/LockTime.java
```diff
@@ -131,46 +119,12 @@ public long rawValue() {
     }
 
     /**
-     * The lock time is considered to be set only if its raw value is greater than zero.
+     * This is equivalent to {@code lockTime instanceof HeightLock || lockTime instanceof TimeLock}.
```
schildbach, line 122: Really? I thought "unset" is same as a HeightLock of 0?

core/src/test/java/org/bitcoinj/core/LockTimeTest.java
schildbach: I mean this specifc test can literally be split in 3 tests (now two, after NoLock has been removed).

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: Removed.

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: Removed.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -131,9 +131,8 @@ private static int sortableBlockHeight(Transaction tx) {
     public static final int SERIALIZE_TRANSACTION_NO_WITNESS = 0x40000000;
 
     /**
-     * @deprecated use {@link LockTime#THRESHOLD},
-     *                 {@link LockTime#isBlockHeight()} or
-     *                 {@link LockTime#isTimestamp()}
+     * @deprecated use {@code lockTime instanceof HeightLock} or
```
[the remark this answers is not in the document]
msgilligan: Fixed.

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: Oops. This was an error. Fixed.

core/src/main/java/org/bitcoinj/core/LockTime.java
```diff
@@ -119,7 +119,7 @@ public long rawValue() {
     }
 
     /**
-     * This is equivalent to {@code lockTime instanceof HeightLock || lockTime instanceof TimeLock}.
```
schildbach: I think this expression would now require proper brackets.

However, maybe better use human words:

> The lock time is considered to be set only if its raw value is greater than zero. In other words, it is set if it is either a non-zero block height or a timestamp.

core/src/main/java/org/bitcoinj/core/LockTime.java
```diff
@@ -29,7 +29,29 @@
  * Instances of this class are immutable and should be treated as Java
  * <a href="https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/doc-files/ValueBased.html#Value-basedClasses">value-based</a>.
  */
-public class LockTime {
+public abstract /* sealed */ class LockTime {
+
+    public static final class HeightLock extends LockTime {
+
```
schildbach, line 35: Superfluous empty line.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -843,10 +845,10 @@ public String toString(@Nullable AbstractBlockChain chain, @Nullable CharSequenc
             s.append(indent).append("time locked until ");
             LockTime locktime = lockTime();
             s.append(locktime);
-            if (locktime.isBlockHeight()) {
+            if (locktime instanceof HeightLock) {
                 if (chain != null) {
                     s.append(" (estimated to be reached at ")
-                            .append(TimeUtils.dateTimeFormat(chain.estimateBlockTimeInstant(locktime.blockHeight())))
+                            .append(TimeUtils.dateTimeFormat(chain.estimateBlockTimeInstant(((HeightLock)locktime).blockHeight())))
```
schildbach, line 851: Imho, missing space after the cast.

core/src/main/java/org/bitcoinj/core/LockTime.java
msgilligan: Will fix.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1868,9 +1870,9 @@ public boolean isFinal(int height, long blockTimeSeconds) {
      */
     public Instant estimateLockTimeInstant(AbstractBlockChain chain) {
         LockTime locktime = lockTime();
-        return locktime.isBlockHeight() ?
-                chain.estimateBlockTimeInstant(locktime.blockHeight()) :
-                locktime.timestamp();
+        return locktime instanceof HeightLock ?
+                chain.estimateBlockTimeInstant(((HeightLock) locktime).blockHeight()) :
+                ((TimeLock)locktime).timestamp();
```
schildbach, line 1875: Imho, missing space after the cast.

msgilligan: Fixed.

core/src/main/java/org/bitcoinj/core/Transaction.java
msgilligan: Fixed.

msgilligan:
@schildbach I added JavaDoc and more (and improved) tests. I think this is ready for final review.

schildbach:
Rebased, edited the tests a bit, and merged.

PR #2817, opened 2023-03-16
base: replace Guava preconditions with conditionals

schildbach:
This has the added benefit of exception messages only being evaluated on demand.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/Bech32.java
+++ b/core/src/main/java/org/bitcoinj/base/Bech32.java
@@ -22,7 +22,7 @@ import javax.annotation.Nullable;
 import java.util.Arrays;
 import java.util.Locale;
 
-import static com.google.common.base.Preconditions.checkArgument;
+import static org.bitcoinj.base.internal.Preconditions.checkArgument;
 
 /**
  * <p>Implementation of the Bech32 encoding.</p>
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/bitcoinj/base/Bech32.java
```diff
@@ -128,8 +126,10 @@ public static String encode(final Bech32Data bech32) {
 
     /** Encode a Bech32 string. */
     public static String encode(Encoding encoding, String hrp, final byte[] values) {
-        checkArgument(hrp.length() >= 1, "Human-readable part is too short");
-        checkArgument(hrp.length() <= 83, "Human-readable part is too long");
+        if (hrp.length() < 1)
+            throw new IllegalArgumentException("human-readable part is too short: " + hrp.length());
+        if (hrp.length() > 83)
+            throw new IllegalArgumentException("human-readable part is too long: " + hrp.length());
```
msgilligan, line 132: I think we could make an exception (or new rule) to allow `if (cons) throw exception` on 1-line for argument checking.

core/src/main/java/org/bitcoinj/base/LegacyAddress.java
```diff
@@ -66,11 +64,15 @@ public class LegacyAddress implements Address {
      *            20-byte hash of pubkey or script
      */
     private LegacyAddress(Network network, boolean p2sh, byte[] hash160) throws AddressFormatException {
-        this.network = normalizeNetwork(checkNotNull(network));
-        this.bytes = checkNotNull(hash160);
+        if (network == null)
+            throw new NullPointerException();
```
msgilligan, line 68: This could be `Objects.requireNonNull(network)`

msgilligan: Or even:
```
this.network = normalizeNetwork(Objects.requireNonNull(network));
```

We could also us a static import for `Objects.requireNonNull`.

msgilligan (changes requested):
I definitely think we should use `Objects.requireNonNull` where possible.

I also prefer the brevity of `checkState` and think we should consider creating an equivalent static utility method.

Note that our utility method could take a `Supplier<String>` argument, so invocation would look like:
```
checkArgument(hrp.length() >= 1, () -> "human-readable part is too short: " + hrp.length());
```

schildbach:
This PR should be reviewed and merged first: https://github.com/bitcoinj/bitcoinj/pull/2819

I will then rebase this one and address your comments.

core/src/main/java/org/bitcoinj/base/LegacyAddress.java
schildbach: This has been resolved via #2819.

schildbach:
Rebased, and added Preconditions utility methods.

Ready for another review.

core/src/main/java/org/bitcoinj/base/Bech32.java
schildbach: In the context of this PR, this is no longer necessary.

msgilligan (approved):
Looks great!

schildbach:
Merged.

PR #2818, opened 2023-03-16
SegwitAddressTest: replace Guava ToStringHelper with custom string building

schildbach:


schildbach:
In my experience, most `toString()` are special/custom enough that a generic helper like ToStringHelper doesn't fit well. So I'd lean against coming up with an own utility.

PR #2821, opened 2023-03-17
MonetaryFormat: replace Guava `checkedPow()` by our own `checkedPowOf10()`

schildbach:
Since we only need a handful, we precompute the powers and look it up in a table.

I deliberately kept this in MonetaryFormat (rather than moving it a utils class) because it's very tailored to the usecase.


schildbach:
TODO: Currently, this is throwing an IllegalArgumentException rather than the previous ArithmeticException.

schildbach:
If we decide to make this a generic util, we could precompute the powers with exponents from 0 to 18 (inclusive). That's the highest power of 10 we can store in a long.

JAYP1255:
Nicely put

schildbach:
Closing in favor of #2823.

PR #2822, opened 2023-03-17
Preconditions: add generic `check()` that allows to supply a custom exception

schildbach:
The idea is to use this in #2821 to throw an `ArithmeticException` rather than a `IllegalArgumentException`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/internal/Preconditions.java
+++ b/core/src/main/java/org/bitcoinj/base/internal/Preconditions.java
@@ -25,8 +25,7 @@ public class Preconditions {
      * @throws IllegalArgumentException if {@code expression} is false
      */
     public static void checkArgument(boolean expression) {
-        if (!expression)
-            throw new IllegalArgumentException();
+        check(expression, () -> new IllegalArgumentException());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


schildbach:
This PR is now kind of without direct use, as #2821 will not be merged.

However it might be useful on its own. I recently added tests to this PR.

core/src/main/java/org/bitcoinj/base/internal/Preconditions.java
```diff
@@ -25,8 +25,7 @@ public class Preconditions {
      * @throws IllegalArgumentException if {@code expression} is false
      */
     public static void checkArgument(boolean expression) {
-        if (!expression)
-            throw new IllegalArgumentException();
+        check(expression, () -> new IllegalArgumentException());
```
msgilligan, line 28: I think you can use `IllegalArgumentException::new` here.

msgilligan, line 28: I think you can use `IllegalArgumentException::new` here.

msgilligan (changes requested):
Looks good. One minor suggestion.

core/src/main/java/org/bitcoinj/base/internal/Preconditions.java
schildbach: Thanks, added.

schildbach: Added.

schildbach:
Merged.

PR #2823, opened 2023-03-17
MonetaryFormat: use `BigDecimal` for converting satishos to number and decimals

schildbach:
This gets rid of Guava `LongMath`.

The potential downside of this solution is BigDecimal is expensive.

If we merge this, in a future commit I'd inline the `satoshisToDecimal()` method again, because it's really simple enough and we can get rid of the result wrapper class.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/utils/MonetaryFormat.java
+++ b/core/src/main/java/org/bitcoinj/base/utils/MonetaryFormat.java
@@ -19,6 +19,7 @@ package org.bitcoinj.base.utils;
 import org.bitcoinj.base.Coin;
 import org.bitcoinj.base.Monetary;
 
+import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.text.DecimalFormatSymbols;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


msgilligan (approved):
This looks great!

msgilligan:
> This gets rid of Guava `LongMath`.
> 
> The potential downside of this solution is BigDecimal is expensive.

I suppose this _could_ be an issue if formatting a large number of amounts for output, but even that seems unlikely to me.

> If we merge this, in a future commit I'd inline the `satoshisToDecimal()` method again, because it's really simple enough and we can get rid of the result wrapper class.

I would argue against inlining. I think the method helps readability and also allows for the possibility of unit tests. If you want to eliminate the wrapper class you could return `BigDecimal[]` or `long[]`.   (Or wait until we're JDK 17 and replace it with a record.)

schildbach:
Merged.

schildbach:
> > If we merge this, in a future commit I'd inline the `satoshisToDecimal()` method again, because it's really simple enough and we can get rid of the result wrapper class.
> 
> I would argue against inlining. I think the method helps readability and also allows for the possibility of unit tests. If you want to eliminate the wrapper class you could return `BigDecimal[]` or `long[]`. (Or wait until we're JDK 17 and replace it with a `record`.)

We're not in a hurry with this, so I'm ok with keeping it as is. The main goal of eliminating `LongMath` has been achieved.

However the code following the method could profit from `BigDecimal` usage too, so the conversion would be in the way.

PR #2824, opened 2023-03-17
Preconditions: replace Guava `checkArgument()` and `checkState()` for the entire codebase

schildbach:
This also fixes many precondition messages, and rewrites two cases of `checkPositionIndex()` and `checkElementIndex()`.


msgilligan (approved):
LGTM. We should be aware that at some point we might discover a need to add a `checkArgument` method that takes a simple `String` parameter in some places to eliminate the overhead of creating a lambda.

schildbach:
Thanks! Merged.

PR #2826, opened 2023-03-18
WIP TimeUtils: introduce the notion of Bitcoin time

schildbach:
The idea is to make sure Instants that are transmitted over the P2P network never have milli- or nanoseconds. That way, we cannot loose them on the way.

A `checkBitcoinTime()` method is called at any setters or constructors of P2P messages.

A `currentBitcoinTime()` convenience method is used for truncating millis from the current time.

At present, only these classes were identified:

* `Block`,
* `LockTime` and
* `VersionMessage`

 I did not include these cases on purpose, but perhaps we should (to be discussed):

* instants that loose their millis due to protobuf serialization (wallet, payment protocol)
* some instants are truncated to millis by serialization, so we still have the same problem wth nanoseconds

I also decided against a full type safe solution, e.g. by wrapping an `Instant` in a `BitcoinInstant`. It seemed overkill to me for a relatively harmless problem.


schildbach:
This is draft because I'm primarly interested in comments/discussion about the concept.

core/src/main/java/org/bitcoinj/base/internal/TimeUtils.java
```diff
@@ -70,6 +71,28 @@ public static Instant currentTime() {
...
+     * It has no higher precision than the second.
+     * @param time time to be checked for absence of precision higher than the second
+     * @return the time that was checked
+     */
+    public static Instant checkBitcoinTime(Instant time) {
+        Preconditions.checkArgument(time.getNano() == 0, () ->
+                "not a Bitcoin time: " + DateTimeFormatter.ISO_INSTANT.format(time));
+        return time;
+    }
+
```
msgilligan, line 95: Do we want to "check and throw `RuntimeException`" or do we just want to truncate?

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -215,7 +215,7 @@ public Block(NetworkParameters params, long version, Sha256Hash prevBlockHash, S
         this.version = version;
         this.prevBlockHash = prevBlockHash;
         this.merkleRoot = merkleRoot;
-        this.time = time;
+        this.time = TimeUtils.checkBitcoinTime(time);
```
msgilligan, line 218: throw or truncate?

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -930,7 +930,7 @@ public Date getTime() {
     @VisibleForTesting
     public void setTime(Instant time) {
         unCacheHeader();
-        this.time = time.truncatedTo(ChronoUnit.SECONDS); // convert to Bitcoin time
+        this.time = TimeUtils.checkBitcoinTime(time);
```
msgilligan, line 933: throw or truncate?

core/src/main/java/org/bitcoinj/core/LockTime.java
```diff
@@ -97,7 +97,7 @@ public static HeightLock ofBlockHeight(int blockHeight) {
      * @return wrapped timestamp
      */
     public static TimeLock ofTimestamp(Instant time) {
-        long secs = time.getEpochSecond();
+        long secs = TimeUtils.checkBitcoinTime(time).getEpochSecond();
```
msgilligan, line 100: throw or truncate?

msgilligan (commented):
I'm not sure I understand the implications of checking vs truncating well enough to approve or reject this PR.

core/src/main/java/org/bitcoinj/core/Block.java
msgilligan: Out of scope, but related: we should add `time` to the constructors and deprecate this method.

msgilligan:
Whether we throw or truncate when Bitcoin Time is required for a parameter, the JavaDoc should probably say that Bitcoin time is required/used.

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: Absolute! I believe that's part of the Message hierarchy rework.

schildbach:
The idea is that if such P2P message is created with a user-supplied value, that value will be checked because it needs to be Bitcoin time, with an exception raised if it is not.

To get such a Bitcoin time, you either convert from seconds (e.g. when deserialitzing), or use the current time via `getBitcoinTime()`, or truncate some other time yourself (hopefully only in unit test, if at all).

PR #2828, opened 2023-03-18
WIP: Decouple TxConfidenceTable & Context

micheal-swiggs:
Work in progress...

Having a look at https://github.com/bitcoinj/bitcoinj/pull/1553 and https://github.com/bitcoinj/bitcoinj/issues/2531. Originally my aim was to find a non breaking way to remove `depth` but realised that this can't really be done without addressing where `TxConfidenceTable` should be located. This P.R assumes that there is going to be one `TxConfidenceTable` per `Network`. Also rather than attaching it to an existing object like `Network` or `NetworkParameters` it's just attached as a static member in `TxConfidenceTable`. The intention is that in the future a  `TxConfidenceTable` is created once there is an `AbstractBlockChain` and is set as a Singleton for a `Network` on `TxConfidenceTable`. 

The reason why `TxConfidenceTable` has an indirect dependency on `AbstractBlockChain` is to be able to calculate `depth` by using `AbstractBlockChain.getBestChainHeight()`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -18,6 +18,7 @@
 package org.bitcoinj.core;
 
 import com.google.common.base.Preconditions;
+import org.bitcoinj.base.Network;
 import org.bitcoinj.base.Sha256Hash;
 import org.bitcoinj.core.listeners.NewBestBlockListener;
 import org.bitcoinj.core.listeners.ReorganizeListener;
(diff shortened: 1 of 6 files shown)


msgilligan:
This is good thinking and an overall solid approach. We still need to do something like this.

See PR #3189

msgilligan:
I just created Issue #4093 to track our work towards removal of the `Context` object.

This PR was a helpful experiment and drove our discussion in Issue #2531, but should probably be closed now that we have changed direction slightly and made other related improvements. I'll leave it open a little while longer...

Thanks @micheal-swiggs for you work on this!

PR #2836, opened 2023-03-19
Message: remove `length` arguments from constructors of the hierarchy

schildbach:
There was never a case where `length` wasn't equal to `payload.length`, which is supplied by the same constructors. This also does away with a couple of constructor variants.

Finally a small achievement in the quest to get rid of cruft in the Message hierarchy!


schildbach:
I pushed two fixes on top.

msgilligan (approved):
Nice!

schildbach:
Merged.

PR #2837, opened 2023-03-19
Peer: calculate lastPing & averagePing when Pong is received

msgilligan:
Peer: calculate lastPing & averagePing when Pong is received. This eliminates the need for locking in the getters.

The fields are marked as volatile so that when written, they will be flushed to "main memory".

**Note:** this changes behavior slightly: It will calculate the average time more accurately. Previously if only two pongs have been received the following calculation will be made:

```
var avgPing = (lastPing + previousPing * 19) / 20
```

With this change the calculation will be:
```
var avgPing = (lastPing + previousPing) / 2
```

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -171,6 +171,9 @@ public class Peer extends PeerSocketHandler {
     // Outstanding pings against this peer and how long the last one took to complete.
     private final ReentrantLock pingIntervalsLock = new ReentrantLock();
     @GuardedBy("pingIntervalsLock") private final Deque<Duration> pingIntervals = new ArrayDeque<>(PING_MOVING_AVERAGE_WINDOW);
+    private volatile Duration lastPing = null;  // should only be written while holding pingIntervalsLock
+    private volatile Duration avgPing = null;   // should only be written while holding pingIntervalsLock
+
     private final CopyOnWriteArrayList<PendingPing> pendingPings;
     // Disconnect from a peer that is not responding to Pings
     private static final int PENDING_PINGS_LIMIT = 50;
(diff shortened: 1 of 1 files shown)


msgilligan:
Note that dc2f95c6fed26e45589e941c5d9e3163af2bdd49 changes behavior slightly: It will calculate the average time more accurately. Previously if only two pongs have been received the following calculation will be made:

```
var avgPing = (lastPing + previousPing * 19) / 20
```

With this change the calculation will be:
```
var avgPing = (lastPing + previousPing) / 2
```

schildbach:
I cherry-picked af8596edb447aa48159804e0fa0a5e6599910ac1 right away.

msgilligan:
> I cherry-picked [af8596e](https://github.com/bitcoinj/bitcoinj/commit/af8596edb447aa48159804e0fa0a5e6599910ac1) right away.

Rebased and force-pushed.

schildbach:
Picked 98c3a14a540e13672a83728354595496de919b7c too.

I think the remaining two can be squashed – the second seems a bugfix of the first.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -169,7 +170,10 @@ public GetDataRequest(Sha256Hash hash) {
 
     // Outstanding pings against this peer and how long the last one took to complete.
     private final ReentrantLock pingIntervalsLock = new ReentrantLock();
-    @GuardedBy("pingIntervalsLock") private Duration[] pingIntervals = null;
+    @GuardedBy("pingIntervalsLock") private final Deque<Duration> pingIntervals = new ArrayDeque<>(PING_MOVING_AVERAGE_WINDOW);
+    @GuardedBy("pingIntervalsLock") private volatile Duration lastPing = null;
+    @GuardedBy("pingIntervalsLock") private volatile Duration avgPing = null;
```
schildbach, line 175: Is this `@GuardedBy` actually right? I mean, like you said in the getters it's now unguarded. And it's `volatile`.

msgilligan: Well, it's guarded for writes.

schildbach: But then the read access would be violating of the locking policy declared by `@GuardedBy`, doesn't it?

I think the idea of the annotation is that it can be checked by build tools/IDEs.

schildbach: I wrote the above under the assumption that the two fields are independent of each other.

If it needs to be ensured that `lastPing` and `avgPing` are consistent with each other, we'd need to switch to a different model anyway. But I'd say that would be overkill.

msgilligan:
Renamed due to cherry picks of previous components of this PR.

schildbach:
Merged.

PR #2838, opened 2023-03-19
Message: do away with retaining of original message payload

schildbach:
This removes all caching of original message payload and "retain mode" in message serializers.

There is some effect on tests. Asserts on the cache status have been removed, rendering some tests almost pointless. Rather than removing them, comments have been added to explain the history of these tests. An assert on a large binary blob of a block was dependent on retain mode, likely due to non-optimal encoding. It had to be disabled.

Note: Some classes still cache individual values, like hashes.


schildbach:
I should note that despite the removal of cache optimisations I do not notice any difference in runtime of the unit tests.

msgilligan (approved):
Looks good! Lots of mutability removed!

schildbach:
Merged!

PR #2839, opened 2023-03-20
Peer: WIP/Question: replace ReentrantLock

msgilligan:
Replace `pingIntervalsLock.lock()` with `synchronizedLock(pingIntervals)`.

I'm not sure this is equivalent or if we are losing any functionality by doing this. Note that before changing to the `Deque`,  the `Duration[]` was nullable, so there had to be another object to lock on. Now that the Deque is final, we can use it.


msgilligan:
`ReentrantLock` is more flexible, but I don't think we are using (or need) that flexibility.  (Also, my copy of _Java Concurrency in Practice_ is at the office, so I need to review that to make sure I'm not missing anything.)

schildbach:
I think it would be fine to use it. I'll be awaiting your book study though.

PR #2813, opened 2023-03-15
WIP: Step 0 of (de)serialization logic move from Message to BitcoinSerializer

schildbach:
Introduces a wrapper for payload bytes and a cursor position for reading, and an offset.

* Replaces the payload, cursor and offset fields in Message
* All the "reading from payload" helper methods move from Message to here
* Reduces the amount of constructors in the Message hierarchy: previously there were often the permutation "payload" and "payload, offset"
* Should make it easier to move parse() from Message to BitcoinSerializer

The Payload class is missing many of our quality marks (JavaDocs and tests, just to name a few). For now, this PR is only to test the concept.

Sadly, as with the other PR, it fails unit tests so it's draft for now.

PR #2840, opened 2023-03-20
Payload: wrap payload bytes and offset in message constructors

schildbach:
This reduces arguments of constructors and does away with a couple of constructor variants.

Note: This PR is split out from #2813. Baby steps. At least it doesn't fail the unit tests. That other PR still hints at the vision.


schildbach:
Actually I just implemented the same baby step approach with a `ByteBuffer`: #2841

schildbach:
Closing in favor of #2841.

PR #2843, opened 2023-03-20
ByteUtils: rename methods that write to byte arrays or streams

schildbach:
They now all use the prefix "write". Note the reading methods already use the prefix "read".


schildbach:
A followup PR will add methods for reading from and writing to `ByteBuffer`.

msgilligan (approved):
LGTM

schildbach:
Merged.

PR #2844, opened 2023-03-20
ByteUtils: add reading from and writing to `ByteBuffer`

schildbach:
This gets rid of a lot of manual bit shuffling.

Follow-up to #2846. Support for `ByteBuffer` will be needed if we use it for message (de)serialization.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
+++ b/core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
@@ -22,8 +22,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.math.BigInteger;
+import java.nio.BufferOverflowException;
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.util.Comparator;
 
+import static org.bitcoinj.base.internal.Preconditions.check;
 import static org.bitcoinj.base.internal.Preconditions.checkArgument;
 
 /**
@@ -96,6 +101,17 @@ public class ByteUtils {
     }
 
     /**
+     * Write 2 bytes to the buffer as unsigned 16-bit integer in little endian format.
+     * @param val value to be written
+     * @param buf buffer to be written into
+     * @return the buffer
+     * @throws BufferOverflowException if the value doesn't fit the remaining buffer
+     */
+    public static ByteBuffer writeUint16LE(int val, ByteBuffer buf) throws BufferOverflowException {
+        return buf.order(ByteOrder.LITTLE_ENDIAN).putShort((short) val);
+    }
+
+    /**
      * Write 2 bytes to the byte array (starting at the offset) as unsigned 16-bit integer in little endian format.
      * @param val value to be written
      * @param out buffer to be written into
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -95,78 +100,99 @@ public static BigInteger bytesToBigInteger(byte[] bytes) {
         return new BigInteger(1, bytes);
     }
 
+    /** Write 2 bytes to the buffer as unsigned 16-bit integer in little endian format. */
+    public static void writeUint16LE(int val, ByteBuffer buf) throws BufferOverflowException {
+        buf.order(ByteOrder.LITTLE_ENDIAN).putShort((short) val);
```
msgilligan, line 110: Do we want to verify that `int val` is not going to lose information when cast to `short`?  (The existing methods don't do this, so not strictly necessary.)

schildbach: Maybe we should. I'm not planning to deprecate the variants that take an array, so I'd say yes.

msgilligan: Would it be useful to return the `ByteBuffer` from this method instead of `void`? This might allow some method chaining and/or make conversion to using `ByteBuffer` easier in some use-cases?

schildbach:
Sorry, I just rebased this on the documentation PR #2846 while you were already reviewing. I hope it didn't disturb you too much.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -104,8 +119,43 @@ public static BigInteger bytesToBigInteger(byte[] bytes) {
      *                                        if value doesn't fully fit the buffer
      */
     public static void writeUint16LE(int val, byte[] out, int offset) throws ArrayIndexOutOfBoundsException {
-        out[offset] = (byte) (0xFF & val);
-        out[offset + 1] = (byte) (0xFF & (val >> 8));
+        check(offset >= 0 && offset <= out.length - 2, () ->
+                new ArrayIndexOutOfBoundsException(offset));
+        writeUint16LE(val, ByteBuffer.wrap(out, offset, out.length - offset));
```
msgilligan, line 124: There's an object creation here,  don't know if there are any use cases where this would affect performance.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -293,32 +382,42 @@ public static long readInt64(byte[] bytes, int offset) throws ArrayIndexOutOfBou
...
+        check(offset >= 0 && offset <= bytes.length - 8, () ->
+                new ArrayIndexOutOfBoundsException(offset));
+        return readInt64(ByteBuffer.wrap(bytes, offset, bytes.length - offset));
     }
 
     /**
      * Read 2 bytes from the stream as unsigned 16-bit integer in little endian format.
      * @param is stream to be read from
      */
     public static int readUint16FromStream(InputStream is) {
```
msgilligan, line 416: I suppose this could be renamed to just `readUnit16` (in a separate PR)

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
schildbach: Actually it would be a better idea to check for the exact bounds, depending on if the methods take signed or unsigned values. I'll prepare something, probably as a followup commit.

msgilligan (approved):
LGTM. I made a few comments,  but no objections to merging as is.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
schildbach: > Would it be useful to return the `ByteBuffer` from this method instead of `void`? This might allow some method chaining and/or make conversion to using `ByteBuffer` easier in some use-cases?

Great idea! Implemented.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
schildbach: Tbh I don't see a noticable regression of the unit tests. I'd say we can optimize later, if we run into problems here.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
schildbach: Yes, agree.

schildbach: Renamed in 7303043f37a0a073841a7a368e3fb2a360b1a7b8.

schildbach:
Merged.

PR #2847, opened 2023-03-21
ByteUtils: check range of unsigned values supplied to write helpers

schildbach:
I'll leave this as draft, because it turns out this fails tests that needs to be fixed.


schildbach:
This is a followup PR to #2860.

msgilligan (changes requested):
I (still) believe this is all unnecessary. We should remove the unsigned/signed distinction from the "write helpers". IMO it is not their job to validate whether what they are (de)serializing is signed or unsigned, just to write the bytes.

Maybe there should be some other tests that make sure these int-in-long values have expected/specified values, but we shouldn't crate two sets of serialization methods just to validate them.

schildbach:
I think it is a good idea to have this range check for the methods dedicated to usage with values as per Java Unsigned Integer API. Especially now that we have the proper documentation in place.

So I rebased this and I'm putting this up for another review.

> We should remove the unsigned/signed distinction from the "write helpers".

I think as long as we have code that uses the Unsigned Integer API these helpers are valuable.

msgilligan (approved):
LGTM

schildbach:
Merged.

PR #2849, opened 2023-03-21
SeedPeers: fix conversion of int representation of IPv4 addresses to `InetAddress`

schildbach:
Usage of `writeUint32LE()` is wrong on two counts:

* `InetAddress.getByAddress()` expects big-endian values, and the method above encodes as little-endian.
* Java int is signed, even if the Java hex literal appears unsigned. The above method expects unsigned values.

So we add the previously missing `writeInt32BE()` to ByteUtils and use it.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
+++ b/core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
@@ -201,6 +201,31 @@ public class ByteUtils {
     }
 
     /**
+     * Write 4 bytes to the buffer as signed 32-bit integer in big endian format.
+     * @param val value to be written
+     * @param buf buffer to be written into
+     * @return the buffer
+     * @throws BufferOverflowException if the value doesn't fit the remaining buffer
+     */
+    public static ByteBuffer writeInt32BE(int val, ByteBuffer buf) throws BufferOverflowException {
+        return buf.order(ByteOrder.BIG_ENDIAN).putInt((int) val);
+    }
+
+    /**
+     * Write 4 bytes to the byte array (starting at the offset) as signed 32-bit integer in big endian format.
+     * @param val value to be written
+     * @param out buffer to be written into
+     * @param offset offset into the buffer
+     * @throws ArrayIndexOutOfBoundsException if offset points outside of the buffer, or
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -200,6 +200,31 @@ public static void writeUint32BE(long val, byte[] out, int offset) throws ArrayI
         writeUint32BE(val, ByteBuffer.wrap(out, offset, out.length - offset));
     }
 
+    /**
+     * Write 4 bytes to the buffer as signed 32-bit integer in big endian format.
```
msgilligan, line 204: It doesn't really matter whether it is singed or unsigned. We should remove "signed" from the JavaDoc here.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -200,6 +200,31 @@ public static void writeUint32BE(long val, byte[] out, int offset) throws ArrayI
...
+     * @param buf buffer to be written into
+     * @return the buffer
+     * @throws BufferOverflowException if the value doesn't fit the remaining buffer
+     */
+    public static ByteBuffer writeInt32BE(int val, ByteBuffer buf) throws BufferOverflowException {
+        return buf.order(ByteOrder.BIG_ENDIAN).putInt((int) val);
+    }
+
+    /**
+     * Write 4 bytes to the byte array (starting at the offset) as signed 32-bit integer in big endian format.
```
msgilligan, line 215: Remove "signed".

msgilligan (changes requested):
Looks good, but there are two minor changes to the JavaDoc comments.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
msgilligan: This method should work for both signed and unsigned. It's using `.putInt((int) val)` which simply extracts the lower 32-bits and writes them without modification to the `ByteBuffer`, right?

schildbach:
Forgot to mark this as merged.

PR #2851, opened 2023-03-21
VarInt: don't save original encoded size (Breaking)

msgilligan:
It looks like getOriginalSizeInBytes() is only being used to update the cursor position in various places in the parsing code. If we came up with another mechanism to do that, could we avoid storing the original size in every instance of this class?

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/VarInt.java
+++ b/core/src/main/java/org/bitcoinj/base/VarInt.java
@@ -24,7 +24,6 @@ import org.bitcoinj.base.internal.ByteUtils;
  */
 public class VarInt {
     private final long value;
-    private final int originallyEncodedSize;
 
     /**
      * Constructs a new VarInt with the given unsigned long value.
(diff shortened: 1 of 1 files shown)


msgilligan:
It looks like this only breaks one unit test (and triggers a second failure because somewhere the clock is not reset)

This is just an experiment, but perhaps the correct way to approach this would be to provide an alternate mechanism (perhaps using ByteBuffer) to update the cursor and either remove or deprecated the `getOriginalSizeInBytes()` method.

schildbach:
I'm already working on replacing the cursor by the position inside `ByteBuffer`.

schildbach:
On top of that, I assume there is a need for restoring a VarInt into the exact encoding it was originally parsed in, especially since we removed the concept of "parse retain" lately. I may be wrong on that, but I think for concensus code it's critical.

msgilligan:
> On top of that, I assume there is a need for restoring a VarInt into the exact encoding it was originally parsed in, especially since we removed the concept of "parse retain" lately. I may be wrong on that, but I think for consensus code it's critical.

If we decide we really need the space-savings of eliminating the extra `int`, we could make two subclasses: one for the optimal case (where original and optimal encodings are the same) and one for the suboptimal case where we need to preserver the exact original encoding.

I'm going to close this PR as I think it is mostly a distraction.

PR #2852, opened 2023-03-21
TransactionInput: check `sequence` bounds

schildbach:
It turns out a test is supplying an invalid fake value. This is now fixed.


core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -36,6 +36,9 @@
  * from {@code org.bitcoinj.core.Utils}.
  */
 public class ByteUtils {
+    /** Maximum unsigned value that can be expressed by 32 bits. */
+    public static final long MAX_UNSIGNED_INTEGER = (1L << Integer.SIZE) - 1;
+
```
msgilligan, line 41: This should be changed as I suggested in PR #2850

msgilligan (approved):
One suggestion that I assume will be addressed in PR #2850.

schildbach:
Merged.

PR #2854, opened 2023-03-21
Message: parse the original `ByteBuffer`, rather than a copy of it

schildbach:
This has the advantage of using the read position contained in `ByteBuffer`, so we remove our `offset` and `cursor` fields.


msgilligan (approved):
Looks great!

PR #2860, opened 2023-03-23
BloomFilter: check `nonce` bounds

schildbach:


core/src/main/java/org/bitcoinj/core/BloomFilter.java
```diff
@@ -126,6 +126,8 @@ public BloomFilter(int elements, double falsePositiveRate, long randomNonce, Blo
         // Optimal number of hash functions for a given filter size and element count.
         hashFuncs = (int)(data.length * 8 / (double)elements * log(2));
         hashFuncs = max(1, min(hashFuncs, MAX_HASH_FUNCS));
+        checkArgument(randomNonce >= 0 && randomNonce <= ByteUtils.MAX_UNSIGNED_INTEGER, () ->
+                "nonce out of range: " + randomNonce);
```
msgilligan, line 130: Rather than checking here, we should just use an `int`.

core/src/main/java/org/bitcoinj/net/FilterMerger.java
```diff
@@ -43,7 +44,7 @@
  */
 public class FilterMerger {
     // We use a constant tweak to avoid giving up privacy when we regenerate our filter with new keys
-    private final long bloomFilterTweak = (long) (Math.random() * Long.MAX_VALUE);
+    private final long bloomFilterTweak = Integer.toUnsignedLong(new Random().nextInt());
```
msgilligan, line 47: This can be an `int`.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4992,7 +4993,7 @@ public int getBloomFilterElementCount() {
     public BloomFilter getBloomFilter(double falsePositiveRate) {
         beginBloomFilterCalculation();
         try {
-            return getBloomFilter(getBloomFilterElementCount(), falsePositiveRate, (long) (Math.random() * Long.MAX_VALUE));
+            return getBloomFilter(getBloomFilterElementCount(), falsePositiveRate, Integer.toUnsignedLong(new Random().nextInt()));
```
msgilligan, line 4996: Conversion to unsigned long is not needed here or anywhere else.

msgilligan (changes requested):
I recommend using PR #2862 instead.

core/src/main/java/org/bitcoinj/core/BloomFilter.java
schildbach: Agreed, as this is arbitrary data.

PR #2861, opened 2023-03-23
Ping: require a nonce value

schildbach:
Pings without nonce are not protocol compliant since version 60001. Our minimum version is at 70000.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Ping.java
+++ b/core/src/main/java/org/bitcoinj/core/Ping.java
@@ -22,55 +22,50 @@ import org.bitcoinj.base.internal.ByteUtils;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
+import java.util.Random;
 
 /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/bitcoinj/core/Ping.java
```diff
@@ -22,55 +22,50 @@
...
+
+    /** @deprecated returns true */
+    @Deprecated
     public boolean hasNonce() {
-        return hasNonce;
+        return true;
     }
-    
+
     public long getNonce() {
```
msgilligan, line 69: `.getNonce()` can now be just `.nonce()`

msgilligan (changes requested):
Looks good. 1 minor change requested `.nonce()` -- **correction:** no change needed!

msgilligan (approved):

core/src/main/java/org/bitcoinj/core/Ping.java
msgilligan: This is out-of-scope (I was a little bit confused by the other pending PR) for this PR. We can fix it when we make Ping immutable.

schildbach:
Merged.

PR #2862, opened 2023-03-23
BloomFilter: Use `int` for nonce (nTweak)

msgilligan:
The original implementation of **bitcoinj** was treating the bloom filter `nonce` and `nTweak` field as an unsigned int and storing it in a `long`. However, it is only used to store 32-bits of entropy. Storing as a long uses extra space and can lead to confusion and errors in implementation. This patch switches to an `int`.

Use `new Random().nextInt()` to generate a value.

Notes:

* The wire format is 32-bits.
* The value can be signed or unsigned it just holds randomness
* Should we store a `Random` instance somewhere and re-use it?
* Should we provide deprecated methods that take `long` everywhere?  (I only provided them for `BloomFilter` constructors)
 
We were previously generating a random double, multiplying by Long.MAX_VALUE and then truncating the low 32-bits of the long. Using `.nextInt()` might actually generate MORE randomness since there will be no truncation.

This should be thoroughly reviewed.


core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4992,7 +4993,7 @@ public int getBloomFilterElementCount() {
     public BloomFilter getBloomFilter(double falsePositiveRate) {
         beginBloomFilterCalculation();
         try {
-            return getBloomFilter(getBloomFilterElementCount(), falsePositiveRate, (long) (Math.random() * Long.MAX_VALUE));
+            return getBloomFilter(getBloomFilterElementCount(), falsePositiveRate, new Random().nextInt());
```
msgilligan, line 4996: This is where the randomness for Bloom filters for a wallet is generated.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -289,6 +289,12 @@ public static void writeUint32LE(long val, OutputStream stream) throws IOExcepti
         stream.write(buf);
     }
 
+    public static void writeInt32LE(int val, OutputStream stream) throws IOException {
+        byte[] buf = new byte[4];
+        writeUint32LE(val, ByteBuffer.wrap(buf));
+        stream.write(buf);
+    }
+
```
msgilligan, line 297: This method isn't strictly needed.

core/src/test/java/org/bitcoinj/core/BloomFilterTest.java
```diff
@@ -59,7 +59,7 @@ public void insertSerializeTest() {
     
     @Test
     public void insertSerializeTestWithTweak() {
-        BloomFilter filter = new BloomFilter(3, 0.01, 2147483649L);
```
schildbach, line 62: Why not simply strip the `L`?

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -385,6 +391,10 @@ public static long readUint32(ByteBuffer buf) throws BufferUnderflowException {
         return Integer.toUnsignedLong(buf.order(ByteOrder.LITTLE_ENDIAN).getInt());
     }
 
+    public static int readInt32(ByteBuffer buf) throws BufferUnderflowException {
```
schildbach, line 394: There is a JavaDoc missing here (-:

schildbach (commented):
Concept ack. I was thinking along the same lines yesterday, but wanted to fix the bug first, in a least intrusive way.

schildbach:
> * Should we provide deprecated methods that take `long` everywhere?  (I only provided them for `BloomFilter` constructors)

We're talking about `PeerFilterProvider` and `FilterMerger`. Hard to say – I don't think these APIs are used much externally.

core/src/test/java/org/bitcoinj/core/BloomFilterTest.java
schildbach: Ah, probably the number is too large to fit an int and you didn't want to change the test data.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: This, or maybe FilterMerger line 47.

msgilligan:
> > This should be thoroughly reviewed.
> 
> Do you have a particular aspect in mind? I think the nonce isn't very safety critical. In fact, varying the nonce reduces privacy.

Yeah, I read that it wasn't safety critical, but I don't understand the protocol well enough to understand why.

I also wanted to make sure the difference in `Wallet.java` between `new Random().nextInt()` and `(long) (Math.random() * Long.MAX_VALUE)` is benign or an improvement.

(And yes, I need to update the JavaDoc.)

msgilligan:
> `PeerFilterProvider` and `FilterMerger` … I don't think these APIs are used much externally.

Then I vote for not doing deprecation.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
msgilligan: I can add it if you want to move forward with this PR.

schildbach: Yes, please. I'm curious about the exact wording when it comes to signedness (-:

schildbach:
> I also wanted to make sure the difference in `Wallet.java` between `new Random().nextInt()` and `(long) (Math.random() * Long.MAX_VALUE)` is benign or an improvement.

It's an improvement if the goal is randomness.

The question is if randomness is good. I wonder why we don't use a constant value all the time, e.g. `0`. The privacy issue is that if an observer gets 2 different filters for the same wallet, s/he can filter out the (intended) false positives by applying two filters on the same data. Only if both filters match, it's very likely a "real positive".

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
msgilligan: Updated JavaDoc has been pushed.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -289,6 +289,13 @@ public static void writeUint32LE(long val, OutputStream stream) throws IOExcepti
         stream.write(buf);
     }
 
+    /**
+     * Write a 4-byte integer to the output stream in little endian format.
+     * The integer may be signed or unsigned, its bytes are just serialized as is.
```
schildbach, line 294: It's good that you mention both of these cases. However as a consumer of this method I have almost no idea what "its bytes are just serialized as is" means.

I'd like to know:

* Can I use it to serialize signed integers as `int`? – probably yes, as that's the bread-and-butter case for Java
* Can I use it to serialize unsigned `long` integers? – it really should, if you claim "the integer may be signed", because that's how Java supports unsigned int
* Can I use it to serialize unsigned `int`? – that case should be explained in more detail, it's not really expected

I think the JavaDoc should make the answers to the questions clear. It's my perspective as a consumer who doesn't really want to think about these things.

msgilligan: "The integer may be signed or unsigned, in both cases it is serialized correctly."


> * Can I use it to serialize signed integers as int?

Yes.

> * Can I use it to serialize unsigned long integers?

No, because the parameter type is an `int`.


> * Can I use it to serialize unsigned int?

Yes, is the above text (" in both cases it is serialized correctly") sufficient.

schildbach: The wording "in both cases it is serialized correctly" is at least reassuring (-:

But I'm not happy with the "the integer may be signed or unsigned", as long as I cannot supply unsigned `long` integers. If you claim "the integer may be signed" I'd expect that. To fix that, I'd either add the appropriate overload or be very specifc about it only accepts the unsigned `int` form. I'd prefer the overload, I guess it could call though to the original method with a cast.

(I'm trying to make a case here, such that we can apply the outcome of this discussion to the other write helpers.)

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
schildbach: Sorry, I accidently continued this discussion in the other helper above.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -289,6 +289,19 @@ public static void writeUint32LE(long val, OutputStream stream) throws IOExcepti
...
+     * @param val value to be written
+     * @param stream stream to be written into
+     * @throws IOException if an I/O error occurs
+     */
+    public static void writeInt32LE(int val, OutputStream stream) throws IOException {
```
schildbach, line 299: I think this write helper can be removed, as per our agreement to not have unused helpers methods.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -385,6 +398,17 @@ public static long readUint32(ByteBuffer buf) throws BufferUnderflowException {
         return Integer.toUnsignedLong(buf.order(ByteOrder.LITTLE_ENDIAN).getInt());
     }
 
+    /**
+     * Read a 4-byte integer from the buffer in little endian format.
+     * The integer may be signed or unsigned, the bytes are just read as is.
+     * @param buf buffer to be read from
+     * @return A 4-byte integer (that may be signed or unsigned depending upon the source.)
```
schildbach, line 405: For reading, there is this similar case we have above, just with the difference that we can't use an overload.

So again, the problem I have with the documentation is that it claims "the integer may be unsigned", but then it doesn't return a long for that case.

msgilligan:
@schildbach Ready for final review, but needs a squash before merging.

PR #2865, opened 2023-03-23
Message: get rid of `payload` field

schildbach:
It's now passed into the `parse()` method, and passed along from there.

PR #2866, opened 2023-03-23
Buffers: new utils class for common operations on P2P message `ByteBuffer`

schildbach:
~~This is a followup PR to #2865.~~

I was wrong. It should actually be merged first.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/internal/Buffers.java
+++ b/core/src/main/java/org/bitcoinj/base/internal/Buffers.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright by the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bitcoinj.base.internal;
+
+import org.bitcoinj.base.VarInt;
+
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+
+import static org.bitcoinj.base.internal.Preconditions.check;
+import static org.bitcoinj.base.internal.Preconditions.checkArgument;
+import static org.bitcoinj.base.internal.Preconditions.checkState;
+
+/**
+ * Utility methods for common operations on Bitcoin P2P message buffers.
+ */
+public class Buffers {
+    /**
+     * Read given number of bytes from the buffer.
+     *
+     * @param buf    buffer to read from
+     * @param length number of bytes to read
+     * @return bytes read
+     * @throws BufferUnderflowException if the read value extends beyond the remaining bytes of the buffer
+     */
+    public static byte[] readBytes(ByteBuffer buf, int length) throws BufferUnderflowException {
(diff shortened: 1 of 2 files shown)


schildbach:
One open question for this PR is the package/module choice. If the helpers in this class remain to be used only for serialization purposes, maybe it should go to `core.internal`?.

core/src/main/java/org/bitcoinj/base/internal/Buffers.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * First read a {@link VarInt} from the buffer and use it to determine the number of bytes to read. Then read
+     * that many bytes and interpret it as an UTF-8 encoded string to be returned. This construct is frequently used
+     * by Bitcoin protocols.
+     *
+     * @param buf buffer to read from
+     * @return read string
+     * @throws BufferUnderflowException if the read value extends beyond the remaining bytes of the buffer
+     */
+    public static String readString(ByteBuffer buf) throws BufferUnderflowException {
```
msgilligan, line 68: If `readByteArray` is renamed, this should have a similar name.

core/src/main/java/org/bitcoinj/base/internal/Buffers.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+
+import static org.bitcoinj.base.internal.Preconditions.check;
+import static org.bitcoinj.base.internal.Preconditions.checkArgument;
+import static org.bitcoinj.base.internal.Preconditions.checkState;
+
+public class Buffers {
```
msgilligan, line 29: A class-level JavaDoc comment could also be a good place to explain that their are VarInt length fields in the buffer.

msgilligan: This class could also be named BitcoinBuffers or P2PBuffers since it contains methods specific to that application.

msgilligan (changes requested):
Concept ack. A few suggestions about naming and comments.

msgilligan:
> One open question for this PR is the package/module choice. If the helpers in this class remain to be used only for serialization purposes, maybe it should go to `core.internal`?.

The nice thing about `internal` is that we are free to move it from one place to another without notice.

core/src/main/java/org/bitcoinj/base/internal/Buffers.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * First read a {@link VarInt} from the buffer and use it to determine the number of bytes to be read. Then read
+     * that many bytes into the byte array to be returned. This construct is frequently used by Bitcoin protocols.
+     *
+     * @param buf buffer to read from
+     * @return read bytes
+     * @throws BufferUnderflowException if the read value extends beyond the remaining bytes of the buffer
+     */
+    public static byte[] readByteArray(ByteBuffer buf) throws BufferUnderflowException {
```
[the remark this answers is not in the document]
schildbach: Renamed to `readLengthPrefixedBytes()`.

core/src/main/java/org/bitcoinj/base/internal/Buffers.java
msgilligan: > I've just added a class-level JavaDoc, but explaining VarInt there seems a bit odd. After all, VarInt is only used by two methods.

No need to explain VarInt there, I guess, but I would say: "common operations on Bitcoin P2P message buffers" or something like that.

> I'd rather make its intended audience clear by putting it next to BitcoinSerializer and the Message hierarchy.

That makes sense.

schildbach: > common operations on Bitcoin P2P message buffers

Added this wording.

msgilligan (approved):
LGTM. I'm assuming this in preparation for some next steps...

schildbach:
Merged.

PR #2867, opened 2023-03-23
ByteUtils: add `readUint64()` helper to read unsigned 64-bit values

schildbach:
This is a followup PR to #2866.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/internal/Buffers.java
+++ b/core/src/main/java/org/bitcoinj/base/internal/Buffers.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright by the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


schildbach:
I should note I've added the method to `ByteUtils` because all the other readers are there. If we decide to move them to somewhere else first, or structure them a different way, I'm happy to rebase and adapt this PR.

msgilligan (changes requested):
I would use an (unsigned) `long` here. If you look at all the use cases they're converting to a `long` (at some point) anyway.  "Services" is stored in `PeerAddress` as a `BigInteger` but is converted to a long in many places. This value doesn't seem to be used arithmetically, so I think it is simpler to use a `long` for it.

schildbach:
> I would use an (unsigned) `long` here. If you look at all the use cases they're converting to a `long` (at some point) anyway. "Services" is stored in `PeerAddress` as a `BigInteger` but is converted to a long in many places. This value doesn't seem to be used arithmetically, so I think it is simpler to use a `long` for it.

I get what you're saying, but how do you know this assumption is true now and in future? I really don't want to encourage consumers to store unsigned long in a Java long. If they want to do that, they can always use `.longValue()`.

The service bits case is a valid point. Maybe we should create a custom class for that, or use something like a bit-field class – not aware of the available options in the Java API atm. Such a class would then offer read and write helpers that don't take the detour via `BigInteger`.

schildbach:
I will close this in favor of #2874, #2871 and #2872. If those are merged, the helper is rendered unnecessary.

schildbach:
Closing.

PR #2868, opened 2023-03-23
Coin: add reading from and writing to `ByteBuffer`

schildbach:
Includes a test.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/Coin.java
+++ b/core/src/main/java/org/bitcoinj/base/Coin.java
@@ -19,6 +19,10 @@ package org.bitcoinj.base;
 import org.bitcoinj.base.utils.MonetaryFormat;
 
 import java.math.BigDecimal;
+import java.nio.BufferOverflowException;
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 
 import static org.bitcoinj.base.internal.Preconditions.checkArgument;
 
@@ -88,6 +92,9 @@ public final class Coin implements Monetary, Comparable<Coin> {
      */
     public static final Coin NEGATIVE_SATOSHI = Coin.valueOf(-1);
 
+    /** Number of bytes to store this amount. */
+    public static final int BYTES = 8;
+
     /**
      * The number of satoshis of this monetary value.
      */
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/bitcoinj/core/FeeFilterMessage.java
```diff
@@ -47,7 +47,7 @@ protected void bitcoinSerializeToStream(OutputStream stream) throws IOException
 
     @Override
     protected void parse() throws ProtocolException {
-        feeRate = Coin.ofSat(readUint64().longValue());
+        feeRate = Coin.read(payload);
```
msgilligan, line 50: I would go with: `Coin.ofSat(readInt64())` and not add the `ByteBuffer` stuff to `Coin`.

msgilligan (changes requested):
Personally, I would skip this altogether.

schildbach:
I converted this to draf for now and will try to rework this a bit. I rebased on current master.

As a first change, the read and write helpers are now using ByteBuffer directly. Since we're dealing with signed long, it's no big deal.

schildbach:
I rebased this, and replaced the `asBuffer()` helper with `serialize()` like with the other data classes.

core/src/main/java/org/bitcoinj/base/Coin.java
```diff
@@ -354,6 +372,27 @@ public BigDecimal toBtc() {
...
+    /**
+     * Allocates a byte array and serializes the amount.
+     *
+     * @return serialized amount
+     */
+    public byte[] serialize() {
+        ByteBuffer buf = ByteBuffer.allocate(BYTES);
+        return write(buf).array();
+    }
+
```
msgilligan, line 395: I still think that `serialize()` should not be using an intermediate `ByteBuffer` here, but I suppose you could call that "premature optimization", so I'm going to go ahead an approve the PR anyway.

msgilligan (approved):
Approved, with one suggestion that we can consider later.

core/src/main/java/org/bitcoinj/base/Coin.java
```diff
@@ -354,6 +369,18 @@ public BigDecimal toBtc() {
...
+     *
+     * @param buf buffer to write into
+     * @return the buffer
+     * @throws BufferOverflowException if the value doesn't fit the remaining buffer
+     */
+    public ByteBuffer write(ByteBuffer buf) throws BufferOverflowException {
+        ByteUtils.writeInt64LE(this.value, buf);
+        return buf;
+    }
+
```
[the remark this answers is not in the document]
schildbach: We'll remove the unused helpers before release.

core/src/main/java/org/bitcoinj/base/Coin.java
schildbach: I don't see a way to prevent this, if we want to use `ByteBuffer.putLong(value)` and *not* do our own bitshuffling.

schildbach:
Merged.

PR #2869, opened 2023-03-23
TransactionOutput: disallow negative values

schildbach:
Draft because it fails unit tests.


schildbach:
I'm setting the milestone to 0.17 because I think we should fix this before relase.

schildbach:
Seems we have to allow `-1` values for the time being. Ready for review.

msgilligan (approved):
LGTM

PR #2871, opened 2023-03-23
FeeFilterMessage: disallow negative fee rate

schildbach:
Fee rates are signed "int64_t", according to BIP-133. The spec doesn't mention, but it goes without saying that negative fee rates make no sense.


schildbach:
Merged.

PR #2872, opened 2023-03-24
Services: wrapper for node services bitfield

schildbach:
This was a holy mess of BigInteger, long, int and VarInt…

I'm mainly looking for concept ack. No JavaDoc has been written so far.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -389,9 +389,8 @@ public class Peer extends PeerSocketHandler {
         helper.addValue(getAddress());
         helper.add("version", vPeerVersionMessage.clientVersion);
         helper.add("subVer", vPeerVersionMessage.subVer);
-        String servicesStr = Strings.emptyToNull(VersionMessage.toStringServices(vPeerVersionMessage.localServices));
-        helper.add("services",
-                vPeerVersionMessage.localServices + (servicesStr != null ? " (" + servicesStr + ")" : ""));
+        if (vPeerVersionMessage.localServices.hasAny())
+            helper.add("services", vPeerVersionMessage.localServices.toString());
         helper.add("time", TimeUtils.dateTimeFormat(vPeerVersionMessage.time));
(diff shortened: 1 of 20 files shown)


msgilligan (commented):
Concept ACK. We might want to try defining an `enum` for service flags and using an `EnumSet` to represent a collection of them.

schildbach:
> We might want to try defining an `enum` for service flags and using an `EnumSet` to represent a collection of them.

I generally agree. I decided to not change it in this PR, because the main purpose was to wrap what we already have, and get stuff out of the message class.

If we later use the EnumSet approach, I assume this class would stay and we'd only replace the `long` by `EnumSet`? And replace the constants.

schildbach:
In particular, I'm interested if you're ok with the static `read()` constructor. It's using a very natural format, except the little-endian. I'm asking becaus we had this controversy of `Sha256Hash` vs `Coin` and the context of added helpers for reading and writing.

If you're ok with the `read()`, the next question I have is about the `write()`. It currently writes to an output stream, unlike `Sha256Hash` which writes into a buffer (which is then written to output streams by consumers). Would you prefer me to use `ByteBuffer` as well? Or better return a `byte[]` array?

schildbach:
I rebased this and added tests and documentation.

I decided to remove the `write(OutputStream)` and provide `write(ByteBuffer)` instead, for consistency with other wrappers. There is also a `buffer()` method that returns a `ByteBuffer` guaranteed to be backed by an array, so it can be easily used to write into streams and the like.

The "bits as enum" suggestion is a very good one, but will follow up in a future PR.

Ready for final review!

msgilligan:
> If we later use the EnumSet approach, I assume this class would stay and we'd only replace the `long` by `EnumSet`?

We might be able to use the `EnumSet` directly (instead of the wrapper `Service` class)  It would be `EnumSet<EnumType>` so would be fairly idiomatic and would save a wrapper class. But we could also do it in two additional steps:

1. Replace the `long` with `EnumSet`.
2. Remove the wrapper.

msgilligan (approved):
Looks good. I would definitely like to (at least) explore `EnumSet` in a future PR.

PR #2873, opened 2023-03-24
Sha256Hash: direct use of `ByteBuffer` in hashCode()

schildbach:
A hash code is just an arbitrary number. It doesn't care about signedness and endianess.


msgilligan (approved):

schildbach:
Merged.

PR #2875, opened 2023-03-24
Message: allow `parse()` to throw `BufferUnderflowException`

schildbach:
The exception is now handled by the one caller to `parse()`, rather than by various helpers.


msgilligan (approved):
LGTM

schildbach:
Was merged a bit ago.

PR #2877, opened 2023-03-24
NetworkParameters: Change PacketMagic to an `int`

msgilligan:
It's just a binary value and does not have signed-edness.

Replace `writeUint32BE` methods in `ByteUtils` with `writeInt32BE` methods

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
+++ b/core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
@@ -236,6 +236,18 @@ public class ByteUtils {
     }
 
     /**
+     * Write a 32-bit integer to a given buffer in big-endian format.
+     *
+     * @param val value to be written
+     * @param buf buffer to be written into
+     * @return the buffer
+     * @throws BufferOverflowException if the value doesn't fit the remaining buffer
+     */
+    public static ByteBuffer writeInt32BE(int val, ByteBuffer buf) throws BufferOverflowException {
+        return buf.order(ByteOrder.BIG_ENDIAN).putInt((int) val);
+    }
+
+    /**
      * Write a 32-bit integer to a given byte array in big-endian format, starting at a given offset.
      * <p>
      * The value is expected as an unsigned {@code long} as per the Java Unsigned Integer API.
@@ -249,6 +261,19 @@ public class ByteUtils {
     public static void writeInt32BE(long val, byte[] out, int offset) throws ArrayIndexOutOfBoundsException {
         check(offset >= 0 && offset <= out.length - 4, () ->
                 new ArrayIndexOutOfBoundsException(offset));
+        writeInt32BE(val, ByteBuffer.wrap(out, offset, out.length - offset));
+    }
+
+    /**
+     * Write a 32-bit integer to a given byte array in big-endian format, starting at a given offset.
+     *
+     * @param val    value to be written
+     * @param out    buffer to be written into
+     * @param offset offset into the buffer
+     * @throws ArrayIndexOutOfBoundsException if offset points outside of the buffer, or
+     *                                        if the value doesn't fit the remaining buffer
+     */
+    public static void writeInt32BE(int val, byte[] out, int offset) throws ArrayIndexOutOfBoundsException {
         writeInt32BE(val, ByteBuffer.wrap(out, offset, out.length - offset));
     }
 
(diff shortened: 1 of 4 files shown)


schildbach:
Will work on this on Tuesday.

schildbach:
`BitcoinSerializer` and `BlockFileLoader`  do some crazy bitshifting with the packet magic. I'm not sure if that stuff still works on int.

Just an idea, but perhaps we can simplify those usecases first?

schildbach:
In BlockFileLoader, we could perhaps read an entire 32-bit integer in the right order and then compare it to the packet magic in one comparison?

msgilligan:
> `BitcoinSerializer` and `BlockFileLoader` do some crazy bitshifting with the packet magic. I'm not sure if that stuff still works on int.

They are all unsigned shifts (`>>>`) of 24-bits or left and they are all masking with `0xff` to extract the low-order byte, so they work the same on `int` as on `long`.

All unit tests are passing.

> Just an idea, but perhaps we can simplify those usecases first?

What do you have in mind for simplification?  (I've got one related PR coming...)

msgilligan:
> In BlockFileLoader, we could perhaps read an entire 32-bit integer in the right order and then compare it to the packet magic in one comparison?

We can't do that because the packetMagic is matched on a byte-by-byte basis. (I suppose we could read a 32-bit value compare and then reverse the buffer pointer by 3 bytes, but since we're matching byte-by-byte we should probably read byte-by-byte)

schildbach:
Can you rebase this on current master? For me it fails a number of tests.

schildbach:
Other than that, I'm fine to merge.

msgilligan:
> Can you rebase this on current master? For me it fails a number of tests.

I rebased and saw the test failures. `packetMagic` was the last use case of writeInt32BE(long, ...) and was failing the sign
check in those methods. So I replaced them with writeInt32BE(int, ...) equivalents.

It's now passing all tests and has been force-pushed.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -221,34 +221,26 @@ public static void writeInt32LE(long val, byte[] out, int offset) throws ArrayIn
...
-     * The value is expected as an unsigned {@code long} as per the Java Unsigned Integer API.
      *
      * @param val    value to be written
      * @param out    buffer to be written into
      * @param offset offset into the buffer
      * @throws ArrayIndexOutOfBoundsException if offset points outside of the buffer, or
      *                                        if the value doesn't fit the remaining buffer
      */
-    public static void writeInt32BE(long val, byte[] out, int offset) throws ArrayIndexOutOfBoundsException {
-        check(offset >= 0 && offset <= out.length - 4, () ->
```
schildbach, line 250: Please don't remove these checks and comments.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -346,14 +338,12 @@ public static void writeInt32LE(long val, OutputStream stream) throws IOExceptio
 
     /**
      * Write a 32-bit integer to a given output stream in big-endian format.
-     * <p>
```
schildbach, line 349: Please don't remove this comment.

msgilligan:
> I think we have methods for writing signed ints already.

Not for the `writeInt32BE(int...)` case. I have updated the PR by adding those methods and leaving the `long` methods in place.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -221,34 +221,26 @@ public static void writeInt32LE(long val, byte[] out, int offset) throws ArrayIn
...
-     * The value is expected as an unsigned {@code long} as per the Java Unsigned Integer API.
      *
      * @param val value to be written
      * @param buf buffer to be written into
      * @return the buffer
      * @throws BufferOverflowException if the value doesn't fit the remaining buffer
      */
-    public static ByteBuffer writeInt32BE(long val, ByteBuffer buf) throws BufferOverflowException {
-        checkArgument(val >= 0 && val <= MAX_UNSIGNED_INTEGER, () ->
```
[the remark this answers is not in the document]
msgilligan: Added back and force-pushed.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
msgilligan: Added back and force-pushed.

core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
msgilligan: ded back and force-pushed.

schildbach:
Merged.

PR #2878, opened 2023-03-24
ByteUtils write helper cleanup

schildbach:
This proposal has two commits:

* Rename all `writeUInt*()` helpers to `writeInt*()` – This creates overloads in some cases, but I think this is by design.
* Rework JavaDoc of all `writeInt*()` helpers – I tried to make really clear what kind of input value is accepted and specifically linked to helpers taking values as per the Java Unsigned Integer API.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
+++ b/core/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
@@ -104,7 +104,10 @@ public class ByteUtils {
     }
 
     /**
-     * Write 2 bytes to the buffer as unsigned 16-bit integer in little endian format.
+     * Write a 16-bit integer to a given buffer in little-endian format.
+     * <p>
+     * The value is expected as an unsigned {@code int} as per the Java Unsigned Integer API.
+     *
      * @param val value to be written
      * @param buf buffer to be written into
      * @return the buffer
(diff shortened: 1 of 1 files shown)


msgilligan:
* Rename all writeUInt*() helpers to writeInt*()  -- 👍👍  (two thumbs up!)
* Rework JavaDoc -- I would word the JavaDoc a little differently

For the case "value is expected as an unsigned" what is really happening is that the upper half of the bits is being dropped. It doesn't really matter whether it is signed or unsigned. If truncating the value is not what is expected, the user will have a problem. 

I think what you mean by "value is expected as an unsigned" is "value is expected as an unsigned n-bit value stored in a 2n-bit container".

"The value is expected as a signed or unsigned" is fine, but IMO redundant.

I think what you mean by "If you've got an unsigned xxx as per the Java Unsigned Integer API, use yyy" is "if you've stored an unsigned n-bit value in a 2n-bit primitive, you'll need to use yyy".

In other words, the issue in these _write_ routines is truncation, not signed-redness.

msgilligan (changes requested):
See my non-review comment above. 

My recommendation is to either update (or remove, this is an internal API) the comments on the "2n" overloads and to deprecate those methods. Instead of calling `xyz32(long val, ...)`, clients would just use  `xyz32((int) val, ...)` with an explicit cast. This makes it very clear to the user what is happening and allows us to have about 1/2 as many methods.

msgilligan (approved):
I'm changing my review to APPROVE. I don't think discussing the comments should hold this  up.

schildbach:
Merged.

PR #2879, opened 2023-03-24
Network: migrate `legacyAddressHeader()` and `legacyP2SHHeader()` from `NetworkParameters`

schildbach:


msgilligan (commented):
This looks good.

However, it doesn't deprecate the equivalent information in `NetworkParameters`. I was imagining that by adding this to `BitcoinNetwork` we could deprecate and replace the various header parameters in `NetworkParameters`.  There are 6 total.

Because there are 6, I would have created a single class that holds all of them, maybe `HeaderParameters` and had one instance for each network type.  (This would require 6 parameters to the enum constructor.)

But this does what we need to decouple `base` and is simpler. So either is fine with me.

schildbach:
I pushed deprecations for the two accessors – that's equivalent to what we did with hrps.

I don't know what you mean by "6 headers". In addition to the two, there is a third that is used for encoded private keys. But we haven't touched ECKey etc so far. What are the remaining three?

msgilligan:
> I don't know what you mean by "6 headers". 

Actually, I miscounted. I think it is 7:

```
    protected int addressHeader;
    protected int p2shHeader;
    protected int dumpedPrivateKeyHeader;
    protected int bip32HeaderP2PKHpub;
    protected int bip32HeaderP2PKHpriv;
    protected int bip32HeaderP2WPKHpub;
    protected int bip32HeaderP2WPKHpriv;
```

schildbach:
Ah, BIP32, I forgot. I think these headers should go to a class close to the BIP32 implementation (so currently in the wallet "module").

schildbach:
Shall we merge this, even if the other headers are not yet migrated?

msgilligan (approved):
Let's do it!

schildbach:
Merged.

PR #2881, opened 2023-03-26
GuiUtils: replace Guava Throwables.getRootCause() with private helper method

msgilligan:
Private helper walks the chain and avoids loops.


schildbach:
Merged,

PR #2882, opened 2023-03-26
WalletPasswordController: remove use of Guava Longs

msgilligan:
Use ByteBuffer in private methods for long to byte[] conversion.


schildbach:
Merged.

PR #2886, opened 2023-03-26
Transaction: Replace Guava Lists.reversed with JDK equivalent

msgilligan:


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1731,12 +1730,14 @@ public void checkCoinBaseHeight(final int height)
...
     public Sha256Hash findWitnessCommitment() {
         checkState(isCoinBase());
-        for (TransactionOutput out : Lists.reverse(outputs)) {
-            Script scriptPubKey = out.getScriptPubKey();
-            if (ScriptPattern.isWitnessCommitment(scriptPubKey))
-                return ScriptPattern.extractWitnessCommitmentHash(scriptPubKey);
-        }
-        return null;
+        List<TransactionOutput> reversed = new ArrayList<>(outputs);
+        Collections.reverse(reversed);
```
schildbach, line 1734: It may not matter much, but isn't there a performance difference here? It reverses the array physically, rather than just iterating backwards as before.

Does the stream API have a "reverse stream"?

schildbach: You should be able to use `return Lists.reverse(outputs).stream()…`.

msgilligan: Yes, there is probably a slight performance difference.  However, there are only a small number of outputs in a transaction so copying and reversing the list will still be fast. Also `findWitnessCommitment()` is only used for testing.

> Does the stream API have a "reverseStream()"?

No. You can sort with a reverse comparator, but that won't work in this case.

> You should be able to use return Lists.reverse(outputs).stream()….

`Lists` is a Guava class.

schildbach:
Merged.

PR #2890, opened 2023-03-26
ForwardingService: rename forwardingCoinSelector parameter

msgilligan:
... to parentTxId (which is more clear and correct)


schildbach:
Merged.

PR #2894, opened 2023-03-26
WalletTool: rename `t` to `tx` in send and don't unnecessarily re-initialize it.

msgilligan:
Two commits:

1. Make `t` effectively final by not unnecessarily reinitializing it.
2. Rename `t` to `tx`.


schildbach:
Merged.

PR #2896, opened 2023-03-26
ForwardingService: make Closable not AutoCloseable

msgilligan:
Closeable implements AutoCloseable but narrows the exception thrown from `Exception` to `IOException`. ForwardingService.close() currently doesn't throw any exceptions.


schildbach:
There is a comment later in the class about `AutoClosable`. I guess that should be adapted as well.

msgilligan:
> There is a comment later in the class about `AutoClosable`. I guess that should be adapted as well.

I removed that comment (because it could be either `Closeable` or `AutoClosable` and we don't really need to document standard Java 7+ behavior in our comment.)

schildbach:
Merged.

PR #2898, opened 2023-03-26
Similar cases of method overloads with nullable parameter (3 commits)

schildbach:
Changes the signature of the new methods so that the compiler can distinguish between the two even in the `null` case, without cast.


msgilligan (approved):
Looks great!

schildbach:
Merged.

PR #2900, opened 2023-03-26
FullBlockTestGenerator: Write the size as a single Int64, rather than two Int32s.

msgilligan:
Write the size as a single Int64, rather than two Int32s.


schildbach:
Merged.

PR #2901, opened 2023-03-26
VarInt: Use ByteBuffer directly in encoding/decoding instead of ByteUtils

msgilligan:


schildbach (commented):
I think this PR should also replace the read helpers.

msgilligan:
> I think this PR should also replace the read helpers.

Agreed. I fixed the read helpers (including the deprecated constructor) and now there is no dependency on `ByteUtils`.

schildbach:
Great way to remove the code duplication in the deprecated constructor! I thought about how to do this, but did not find a way…

I picked this as a separate refactoring.

PR #2902, opened 2023-03-26
TransactionBroadcast: add awaitRelayed(), deprecate future()

msgilligan:
* Add awaitRelayed() method that returns a CF that completes when relaying is confirmed
* Deprecate future() and reimplement it using awaitRelayed()


core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java
```diff
@@ -102,9 +103,11 @@ public ListenableCompletableFuture<Transaction> future() {
 
     /**
      * @return future that completes when some number of remote peers has rebroadcast the transaction
+     * @deprecated Use {@link #awaitRelayed()} (and maybe {@link CompletableFuture#thenApply(Function)})
      */
+    @Deprecated
     public ListenableCompletableFuture<Transaction> future() {
```
schildbach, line 109: This method is deprecated but still referenced.

schildbach (changes requested):
Looks good, but one comment added.

PR #2903, opened 2023-03-26
Wallet.SendResult: deprecate public fields, replace with accessors

msgilligan:
* Deprecate public fields in `SendResult`
* Make available equivalent accessors
* Update all usages

This prepares the way for replacing `SendResult` with `TransactionBroadcast` in the future.


schildbach:
Merged.

PR #2906, opened 2023-03-27
VarInt: simplify calculation `originallyEncodedSize`

schildbach:
It now uses code that speaks for itself, rather than relying on comments.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/VarInt.java
+++ b/core/src/main/java/org/bitcoinj/base/VarInt.java
@@ -63,20 +63,19 @@ public class VarInt {
     public static VarInt read(ByteBuffer buf) throws BufferUnderflowException {
         buf.order(ByteOrder.LITTLE_ENDIAN);
         int first = Byte.toUnsignedInt(buf.get());
+        int originallyEncodedSize = Byte.BYTES;
         long value;
-        int originallyEncodedSize;
         if (first < 253) {
             value = first;
-            originallyEncodedSize = 1; // 1 data byte (8 bits)
         } else if (first == 253) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/base/VarInt.java
```diff
@@ -63,20 +63,19 @@ public static VarInt ofBytes(byte[] buf, int offset) throws ArrayIndexOutOfBound
     public static VarInt read(ByteBuffer buf) throws BufferUnderflowException {
         buf.order(ByteOrder.LITTLE_ENDIAN);
         int first = Byte.toUnsignedInt(buf.get());
+        int originallyEncodedSize = Byte.BYTES;
```
msgilligan, line 66: To avoid mutating `originallyEncodedSize`, I would remove this line and add a `Byte.BYTES + ` to the three lines inside the conditionals.

msgilligan (changes requested):
Looks good, but I have one minor request.

core/src/main/java/org/bitcoinj/base/VarInt.java
msgilligan: Actually, the constants are used again down in `sizeOf()`. Maybe we should just define them as private constants in the file, e.g.
```
private static int VARINT_BYTE_SIZE = Byte.BYTES;
private static int VARINT_SHORT_SIZE = Byte.BYTES + SHORT.BYTES;
```
etc.

schildbach:
Closing in favor of https://github.com/bitcoinj/bitcoinj/pull/2910.

PR #2907, opened 2023-03-27
VarInt: implement `equals()` and `hashCode()`

schildbach:
`originallyEncodedSize` is not considered on purpose, because it is only relevant for serialization.


msgilligan (approved):
LGTM

PR #2914, opened 2023-03-27
ForwardingService: improve arg validation

msgilligan:
* System.exit(1) rather than throw
* Check max number of arguments


schildbach:
Just created the same (-:

PR #2915, opened 2023-03-27
ForwardingService: Combine the `main` and `forward` methods

msgilligan:
(This is partly in preparation for a forthcoming change that will simplify the main method.)

### DIFF
--- a/integration-test/src/test/java/org/bitcoinj/examples/ForwardingServiceTest.java
+++ b/integration-test/src/test/java/org/bitcoinj/examples/ForwardingServiceTest.java
@@ -43,10 +43,10 @@ public class ForwardingServiceTest {
     }
 
     @Test
-    public void startViaStaticForwardAndImmediatelyInterrupt(@TempDir File tempDir) {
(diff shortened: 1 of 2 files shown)


integration-test/src/test/java/org/bitcoinj/examples/ForwardingServiceTest.java
```diff
@@ -46,7 +46,7 @@ void setupTest() {
     public void startViaStaticForwardAndImmediatelyInterrupt(@TempDir File tempDir) {
```
schildbach, line 46: The test name reads like it now needs to be changed, too?

schildbach (changes requested):
Looks good!

integration-test/src/test/java/org/bitcoinj/examples/ForwardingServiceTest.java
msgilligan: Fixed and force-pushed.

schildbach:
Merged.

PR #2916, opened 2023-03-28
ListenableCompletionStage: deprecate addListener

msgilligan:
When people upgrade to 0.17 this deprecation will make it more clear that they need to switch from this method to `thenRunAsync`

Reminder: we did not deprecate ListenableCompletionStage or ListenableCompletableFuture because they are still used in many places in the bitcoinj API, but this method is not used anywhere in bitcoinj.


schildbach:
Merged.

PR #2917, opened 2023-03-28
PeerGroup: check required services on discovered addresses via `addr/addrv2`

schildbach:
If the discovered peer has the required services, mark it with a slightly higher priority than peers discovered via `PeerDiscovery`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerGroup.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerGroup.java
@@ -319,8 +319,10 @@ public class PeerGroup implements TransactionBroadcaster {
             Collections.shuffle(addresses);
             int numAdded = 0;
             for (PeerAddress address : addresses) {
-                // Add to inactive pool.
-                boolean added = addInactive(address, 0);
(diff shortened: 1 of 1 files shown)


msgilligan:
This is hard for me to review because neither the `onAddr()` method or the `AddressMessage` has any JavaDoc.

msgilligan (commented):
I don't really understand what this change means without more JavaDoc.

schildbach:
Ok, I added the JavaDocs you requested. Then I rebased this.

msgilligan (approved):
LGTM

schildbach:
Priority is just an int. It is a way to priorize peers over others in trying to connect to them.

The idea of this PR is that, for the peers we get via P2P messages we already know the services they support, and do ignore them if they're useless for us. In contrast, the peers we get via PeerDiscovery we usually need to connect in order to find out if they support our services. This creates unnecessary churn, as many peers (especially on testnet) are connected and quickly dropped.

Also, peers from DNS are often hardcoded (in the operators domain config) and thus are more likely to be stale.

schildbach:
Merged.

PR #2922, opened 2023-03-29
BlockFileLoader: Don't use NetworkParameters internally

msgilligan:
* Remove the private `params` field
* Save `packetMagic` and `serializer` instead
* Constructors are unchanged but could change in the future


schildbach:
Merged.

PR #2923, opened 2023-03-29
EmptyMessage: remove `params` and `payload` from constructors in hierarchy

schildbach:
Empty messages don't need any of these.

There is a commit prefixed:
ProtocolException, VerificationException: allow construction without message


msgilligan (approved):
LGTM

PR #2925, opened 2023-03-29
TransactionOutPoint: define parent as TransactionInput in constructor

msgilligan:
It might be possible to do this for more of the messages with parents and removing the Message.parent field and replacing with:

abstract @Nullable Message getParent()

may allow us to narrow the return type in subclasses.


schildbach:
I think this is now obsolete?

msgilligan:
> I think this is now obsolete?

Yes, it was made obsolete by d02853b1f0ace5347c04c82c047a3cea438cdd5e.

PR #2926, opened 2023-03-29
ChildMessage: Store typed parent in subclass

msgilligan:
* Replace parent field in ChildMessage with protected, abstract getParent()
* Remove parent parameter from ChildMessage constructor
* Use correct Message subtype in subclass constructors
* Override getParent() in subtypes and narrow return type
* Replace getParentTransaction with getParent

TODO:
1. Eliminate redundancy in setParent
2. Consider removing unCache and perhaps ChildMessage itself is no longer needed.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/ChildMessage.java
+++ b/core/src/main/java/org/bitcoinj/core/ChildMessage.java
@@ -27,8 +27,6 @@ import java.nio.ByteBuffer;
  */
 public abstract class ChildMessage extends Message {
 
-    @Nullable protected Message parent;
-
     public ChildMessage(NetworkParameters params) {
         super(params);
     }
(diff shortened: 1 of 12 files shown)


schildbach:
I'm currently still with `params` and `serializer`, but from a first glance this seems to duplicate the logic from `ChildMessage`.

I think a better approach would be to identify current subclasses of `ChildMessage` that acually don't need the services of `ChildMessage`. Those should be derived from `Message` instead.

schildbach:
Also, like I said, I think we need an idea where to go with `parent` in future, in an all-immutable world.

I think the less we depend on `parent` the better.

msgilligan:
Closing in favor of PR #2932, PR #2933, PR #2935

PR #2935, opened 2023-03-30
TransactionInput, TransactionOutput: make direct subclass of Message

msgilligan:
Since only these 2 classes need to support the concept of a "Parent", it will be simpler to just add the code to handle that to both of them and remove the `ChildMessage` altogether. This also allows parent to be consistent typed as `Transaction`


schildbach (commented):
Looks good!

Would it perhaps be a good idea to rename the field + setters to `parentTransaction`?

msgilligan:
> Would it perhaps be a good idea to rename the field + setters to `parentTransaction`?

I thought about that. I think I'd prefer to replace `getParentTransaction` with `getParent` in a future PR.

PR #2938, opened 2023-03-30
PeerAddress: remove `params` from constructor

schildbach:
To make this possible, a port has to be specified in all cases.


msgilligan (approved):
LGTM

PR #2941, opened 2023-03-30
VersionMessage: require extended version handshake messages

schildbach:
Versions without the extended fields (e.g. user-agent, block height) are not protocol compliant since version 106. Our minimum version is at 70000.


msgilligan (approved):
LGTM

schildbach:
Merged.

PR #2942, opened 2023-03-30
NetworkParameters: fix protocol version for BIP37 bloom filters

schildbach:
According to BIP111, the protocol version must be greater than 70000. Since we use "greater or equals", 70001 is the correct number.


schildbach:
This currently fails unit tests – investigating.

msgilligan (approved):
LGTM

schildbach:
Merged.

PR #2945, opened 2023-03-30
TransactionOutput: remove `params` from constructors

schildbach:
To make this possible, a check for maximum value is removed.

I think this check is a bit esotheric anyway. The practical limit will of course be input values of a transaction. If we want such a check for a single output, I guess a (future) builder for `TransactionOutput` would be a better place.

On the plus side, a lot of passing through of `NetworkParameters` is unnecessary.


msgilligan (approved):
LGTM

schildbach:
Merged.

PR #2946, opened 2023-03-30
Message hierarchy: convert `NetworkParameters` to `Network `

schildbach:
* It's a huge changeset. We should try to remove more `NetworkParameters` first.
* Currently, it doesn't work. There is no equivalent to `NetworkParameters.UNITTEST` in the `Network` world. Probably we need to get rid of `UNITTEST` first…


schildbach:
Closing this, as there is essentially nothing to convert.

PR #2948, opened 2023-03-30
WalletTool: use `net` over `params` where possible

msgilligan:
(Note that `params` is initialized from `net`)


schildbach:
Merged.

PR #2949, opened 2023-03-31
UTXOProvider: replace getParams() with network()

msgilligan:
Since this only affects FullPrunedBlockStore, this change doesn't need deprecation.


schildbach:
Merged.

PR #2951, opened 2023-03-31
MemoryBlockStore: construct with genesis block not NetParams

msgilligan:
Since MemoryBlockStore is mostly used for testing, we probably don't need a deprecated constructor with NetworkParameters.


schildbach:
Merged.

PR #2953, opened 2023-03-31
SPVBlockStore: Construct from genesis Block

msgilligan:
* Change constructor to take a genesis Block
* Internally store the genesisBlock
* ~~Get serializer from genesis Block~~

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
@@ -16,9 +16,7 @@
 
 package org.bitcoinj.store;
 
-import org.bitcoinj.base.internal.ByteUtils;
 import org.bitcoinj.core.Block;
-import org.bitcoinj.core.NetworkParameters;
 import org.bitcoinj.core.ProtocolException;
 import org.bitcoinj.base.Sha256Hash;
 import org.bitcoinj.core.StoredBlock;
@@ -65,7 +63,7 @@ public class SPVBlockStore implements BlockStore {
     static final byte[] HEADER_MAGIC_V2 = "SPV2".getBytes(StandardCharsets.US_ASCII);
 
     private volatile @Nullable MappedByteBuffer buffer;
-    protected final NetworkParameters params;
+    private final Block genesisHeader;
 
     // The entire ring-buffer is mmapped and accessing it should be as fast as accessing regular memory once it's
     // faulted in. Unfortunately, in theory practice and theory are the same. In practice they aren't.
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -111,9 +113,10 @@ public SPVBlockStore(NetworkParameters params, File file) throws BlockStoreExcep
      * @param grow wether or not to migrate an existing block store of different capacity
      * @throws BlockStoreException if something goes wrong
      */
-    public SPVBlockStore(NetworkParameters params, File file, int capacity, boolean grow) throws BlockStoreException {
+    public SPVBlockStore(Block genesisBlock, File file, int capacity, boolean grow) throws BlockStoreException {
         Objects.requireNonNull(file);
-        this.params = Objects.requireNonNull(params);
+        this.genesisHeader = Objects.requireNonNull(genesisBlock).cloneAsHeader();
+        this.serializer = genesisHeader.getParams().getDefaultSerializer(); // TODO: Better way of getting serializer?
```
schildbach, line 119: I think this line makes this PR a net loss. We want to get rid of both params and serializer in `Block` (genesisBlock).

schildbach (commented):
I think we should look at this again once the Message hierarchy refactorings are finished.

msgilligan:
> I think we should look at this again once the Message hierarchy refactorings are finished.

OK. I'll mark it as DRAFT.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
msgilligan: I have rebased on latest `master` and serializer is no longer in `SPVBlockStore`.

msgilligan:
@schildbach I just rebased this, but I actually think we should not pass in genesis blocks, but `Network` instead of `NetworkParameters`.

schildbach:
> but I actually think we should not pass in genesis blocks, but `Network` instead of `NetworkParameters`.

Why that change of mind?

schildbach (commented):
I like this draft.

msgilligan:
> > but I actually think we should not pass in genesis blocks, but `Network` instead of `NetworkParameters`.
> 
> Why that change of mind?

To simplify the constructor API. The caller should just specify the network, and shouldn't be responsible for finding and providing a genesis block or header. In the future we may be changing the types for blocks/headers, so it would be better if `SPVBlockStore` gets the blocks/headers itself based upon the passed in network.

I'm thinking we could add a `Block.getGenesis(Network)` method to return the genesis block for a given network.

We should also keep the existing `NetworkParameters` constructors and mark them as `@Deprecated`.

msgilligan:
I propose closing this and instead merging: 

* PR #4261

PR #2954, opened 2023-03-31
TransactionInput: remove helper `duplicateDetached()`

schildbach:
It is only used from a test, and that test works just as well by just copying the reference.


msgilligan (approved):
LGTM

PR #2957, opened 2023-03-31
WalletTool.send exception handling cleanup

msgilligan:
This is broken into three separate commits that can be merged as a sequence.


schildbach:
Merged.

PR #2958, opened 2023-03-31
TransactionInput, TransactionOutput: make `setParent()` protected

schildbach:
Rather than using this method, we should prefer passing the parent transaction via constructor.


msgilligan (approved):
Good catch!

schildbach:
Merged.

PR #2959, opened 2023-03-31
Transaction: make `allowWitness()` static and decide on `protocolVersion` only

schildbach:
It doesn't need an entire `MessageSerializer`.


msgilligan (approved):
LGTM

PR #2961, opened 2023-03-31
Transaction: make `network` in `toString()` an optional parameter

schildbach:
If it isn't provided, standard output scripts cannot be converted to addresses.


msgilligan (approved):
LGTM

schildbach:
Merged.

PR #2962, opened 2023-04-01
TransactionTest: don't create global transaction object for each test

schildbach:
It makes tests more self-contained, and only about a third of the tests use the global transaction.


core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -77,34 +77,36 @@ public class TransactionTest {
     private static final NetworkParameters TESTNET = TestNet3Params.get();
     private static final Address ADDRESS = new ECKey().toAddress(ScriptType.P2PKH, BitcoinNetwork.TESTNET);
 
-    private Transaction tx;
-
     @Before
     public void setUp() {
-        tx = FakeTxBuilder.createFakeTx(TESTNET);
         Context.propagate(new Context());
+        Transaction tx = FakeTxBuilder.createFakeTx(TESTNET);
```
msgilligan, line 83: This looks like it is unneeded.

msgilligan (approved):
LGTM. I think one line can be deleted.

PR #2963, opened 2023-04-01
Block: log warning if `solve()` runs a bit long

schildbach:


schildbach:
This is now ready for review.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -445,13 +446,20 @@ public String toString() {
      */
     @VisibleForTesting
     public void solve() {
+        Duration fairlyLong = Duration.ofSeconds(5);
```
msgilligan, line 449: A better name might be `warningThreshold`.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -445,13 +446,20 @@ public String toString() {
...
+        Stopwatch watch = Stopwatch.start();
         while (true) {
             try {
                 // Is our proof of work valid yet?
                 if (checkProofOfWork(false))
                     return;
                 // No, so increment the nonce and try again.
                 setNonce(getNonce() + 1);
+
+                if (watch.isRunning() && watch.elapsed().compareTo(fairlyLong) > 0) {
```
msgilligan, line 459: Can the `.isRunning()` check be removed?

msgilligan (approved):
Looks good. Minor questions/suggestions.

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: No, because then it would print thousands of messages until the block is solved.

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: Ok, I'll change it. Though sometimes I like a bit colloquial names.

schildbach:
Merged.

PR #2965, opened 2023-04-01
Transaction: move `isMature()` to `Wallet.isTransactionMature()`

schildbach:
That method is only needed in `Wallet`, so it's good to have it there. On top of that, Wallet will be able to provide `NetworkParameters` for a long time.


msgilligan (approved):
LGTM

PR #2966, opened 2023-04-01
Transaction, Block: make `verify*()` static helper methods

schildbach:
Two commits, one each for Transaction and Block.

I think it's fair that these must become independant, so that the data classes need not keep the parameters for the validation. We can move the validation to separate classes later if needed.


msgilligan (approved):
Looks good. Longer term I feel like there's probably a better place to put these than in `Wallet` which is already huge.

schildbach:
No worries, I wouldn't move them to Wallet. Merged!

PR #2967, opened 2023-04-01
Stopwatch: introduce tool for measuring time mainly for log messages

schildbach:
Guava Stopwatch is back with a vengeance…

The helpers from `TimeUtils` are affected by the mock clock, which is not desired for debug output.


msgilligan (approved):
Looks good. Fixes issues and also shortens the log message arguments nicely.

schildbach:
Merged.

PR #2968, opened 2023-04-01
DefaultCoinSelector, AllowUnconfirmedCoinSelector: remove singleton pattern

schildbach:
Instantiating these classes is cheap, as they're stateless. And we only do that once per `Wallet` instance.


msgilligan (approved):
Nice!

schildbach:
Merged.

PR #2970, opened 2023-04-01
Transaction: remove `params` from constructors

schildbach:
This is a crazy big changeset, but it removes a lot of params…


schildbach:
Merged. And I'll continue with `Block`…

PR #2973, opened 2023-04-03
VersionMessage: deprecate `isPingPongSupported()` and `ProtocolVersion.PONG`

schildbach:
Our minimum version is at 70000 already.


msgilligan (approved):
LGTM

PR #2974, opened 2023-04-03
VersionMessage: inline helpers `isWitnessSupported()`, `hasBlockChain()` and `hasLimitedBlockChain()`

schildbach:
Our API is now fluent enough that we don't need them any more.

We cannot provide deprecations because they'd need network parameters.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -522,21 +522,22 @@ public class Peer extends PeerSocketHandler {
         // mode nodes because we can't download the data from them we need to find/verify transactions. Some bogus
         // implementations claim to have a block chain in their services field but then report a height of zero, filter
         // them out here.
-        if (!peerVersionMessage.hasLimitedBlockChain() ||
+        Services services = peerVersionMessage.getServices();
+        if (!(services.has(Services.NODE_NETWORK) || services.has(Services.NODE_NETWORK_LIMITED)) ||
                 (!params.allowEmptyPeerChain() && peerVersionMessage.bestHeight == 0)) {
             // Shut down the channel gracefully.
             log.info("{}: Peer does not have at least a recent part of the block chain.", this);
             close();
             return;
         }
-        if (!peerVersionMessage.localServices.has(requiredServices)) {
+        if (!services.has(requiredServices)) {
             log.info("{}: Peer doesn't support these required services: {}", this,
                     Services.of(requiredServices & ~peerVersionMessage.localServices.bits()).toString());
             // Shut down the channel gracefully.
             close();
             return;
         }
-        if (peerVersionMessage.localServices.has(Services.NODE_BITCOIN_CASH)) {
+        if (services.has(Services.NODE_BITCOIN_CASH)) {
             log.info("{}: Peer follows an incompatible block chain.", this);
             // Shut down the channel gracefully.
             close();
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/bitcoinj/core/VersionMessage.java
```diff
@@ -113,6 +113,15 @@ public VersionMessage(NetworkParameters params, int newBestHeight) {
         relayTxesBeforeFilter = true;
     }
 
+    /**
+     * Get the service bitfield that represents the node services being provided.
+     *
+     * @return service bitfield
+     */
+    public Services getServices() {
```
msgilligan, line 121: Perhaps this should be `.services()`?  If not now, then perhaps after `localServices` can be made final.

core/src/main/java/org/bitcoinj/core/VersionMessage.java
```diff
@@ -113,6 +113,15 @@ public VersionMessage(NetworkParameters params, int newBestHeight) {
...
+        return localServices;
+    }
+
```
msgilligan, line 124: For convenience, consider adding:
```
public boolean has(long feature) {
   return localServices.has(feature);
}   
```

msgilligan (changes requested):
I think I see an unintended behavior change and have a few minor suggestions as well.

core/src/main/java/org/bitcoinj/core/VersionMessage.java
schildbach: At the moment, it's mutable. So maybe we continue accessing the field for now and not add the getter until we tackle the immutability?

core/src/main/java/org/bitcoinj/core/VersionMessage.java
schildbach: I don't understand. That's a recursive loop.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -522,21 +522,22 @@ private void processVersionMessage(VersionMessage peerVersionMessage) throws Pro
         // mode nodes because we can't download the data from them we need to find/verify transactions. Some bogus
         // implementations claim to have a block chain in their services field but then report a height of zero, filter
         // them out here.
-        if (!peerVersionMessage.hasLimitedBlockChain() ||
+        Services services = peerVersionMessage.getServices();
+        if (!(services.has(Services.NODE_NETWORK) || services.has(Services.NODE_NETWORK_LIMITED)) ||
```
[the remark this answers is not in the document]
schildbach: "limited blockchain" means either "limited blockchain" or "full blockchain" – hence the "or".

However, I just pushed a commit on top that adds a helper for cases like this.

core/src/main/java/org/bitcoinj/core/VersionMessage.java
msgilligan: Add `VersionMessage.has()` as a shortcut for `VersionMessage.getServices().has()`...

msgilligan (approved):
Looks good as is, but I do have two suggestions that could be implemented now or later:

* `VersionMessage.services()` instead of `VersionMessage.getServices()`
* `VersionMessage.has(serviceFlags)`

schildbach:
I renamed the getter, but kept your other suggestion for a future PR.

Merged.

PR #2975, opened 2023-04-03
TransactionOutPoint: divorce from `Message`

schildbach:

* A static constructor `read()` replaces the native constructor that deserialized from a payload.
* A `serialize()` helper replaces `bitcoinSerialize()`.
* A `BYTES` constant replaces `getMessageSize()`.

This comes with a test.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
@@ -29,8 +29,8 @@ import org.bitcoinj.wallet.KeyBag;
 import org.bitcoinj.wallet.RedeemData;
 
 import javax.annotation.Nullable;
-import java.io.IOException;
-import java.io.OutputStream;
+import java.nio.Buffer;
+import java.nio.BufferOverflowException;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.util.Objects;
@@ -43,9 +43,8 @@ import static org.bitcoinj.base.internal.Preconditions.checkState;
  * 
  * <p>Instances of this class are not safe for use by multiple threads.</p>
  */
-public class TransactionOutPoint extends Message {
-
-    static final int MESSAGE_LENGTH = 36;
+public class TransactionOutPoint {
+    public static final int BYTES = 36;
 
     /** Special outpoint that normally marks a coinbase input. It's also used as a test dummy. */
     public static final TransactionOutPoint UNCONNECTED =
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -43,9 +43,8 @@
  * 
  * <p>Instances of this class are not safe for use by multiple threads.</p>
  */
-public class TransactionOutPoint extends Message {
-
-    static final int MESSAGE_LENGTH = 36;
+public class TransactionOutPoint {
+    public static final int BYTES = 36;
```
msgilligan, line 47: Perhaps this would be better named `LENGTH` or `SIZE` or `BYTE_LENGTH`?

core/src/test/java/org/bitcoinj/core/TransactionOutPointTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private Iterator<TransactionOutPoint> randomOutPoints() {
+        Random random = new Random();
+        return Stream.generate(() -> {
+            byte[] randomBytes = new byte[Sha256Hash.LENGTH];
+            random.nextBytes(randomBytes);
+            return new TransactionOutPoint(Integer.toUnsignedLong(random.nextInt()), Sha256Hash.wrap(randomBytes));
+        }).limit(10).iterator();
+    }
+}
```
msgilligan, line 56: Nice!

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -85,28 +97,40 @@ public TransactionOutPoint(TransactionOutput connectedOutput) {
...
+    /**
+     * Allocates a buffer and writes this transaction outpoint into it. The buffer is backed by a byte array and it's
+     * already rewound for reading.
+     *
+     * @return buffer containing the transaction outpoint
+     */
+    public ByteBuffer buffer() {
+        ByteBuffer buf = write(ByteBuffer.allocate(BYTES));
+        ((Buffer) buf).rewind();
+        return buf;
```
msgilligan, line 121: Maybe we should just have a `bytes()` method that returns an array. It seems every usage is `.buffer().array()` so it might be better to skip the intermediate `ByteBuffer`?

msgilligan (changes requested):
Looks good! Another little milestone. Two minor requests.

core/src/test/java/org/bitcoinj/core/TransactionOutPointTest.java
schildbach: I've used that before, see ServicesTest (-:

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
schildbach: I see what you mean, the ideal form isn't found yet. (There is precident in `Services`.)

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
msgilligan: OK, It _is_ consistent with our other code and the JDK. I'll get used to it.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
schildbach: Ok, it's now `byte[] serialize()`.

schildbach:
Merged.

PR #2976, opened 2023-04-03
Block: move difficulty target range check to `AbstractBlockChain.add()`

schildbach:
It doesn't make much sense in a getter. It should be checked when adding a block to the chain, like all the other checks.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -456,10 +456,7 @@ public class Block extends Message {
      * is thrown.
      */
     public BigInteger getDifficultyTargetAsInteger() throws VerificationException {
-        BigInteger target = ByteUtils.decodeCompactBits(difficultyTarget);
-        if (target.signum() <= 0 || target.compareTo(params.maxTarget) > 0)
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -456,10 +456,7 @@ public void solve() {
      * is thrown.
      */
     public BigInteger getDifficultyTargetAsInteger() throws VerificationException {
-        BigInteger target = ByteUtils.decodeCompactBits(difficultyTarget);
-        if (target.signum() <= 0 || target.compareTo(params.maxTarget) > 0)
-            throw new VerificationException("Difficulty target is bad: " + target.toString());
-        return target;
+        return ByteUtils.decodeCompactBits(difficultyTarget);
```
msgilligan, line 459: JavaDoc needs update.

msgilligan (changes requested):
Looks good but needs JavaDoc update.

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: Ah, indeed. Good catch!

PR #2979, opened 2023-04-03
VersionMessage: remove `params` from constructor

schildbach:


schildbach:
I should add that there is one constructor that takes NetworkParameters, but they are discarded in the constructor already, as soon as the necessary values are extracted.

msgilligan (approved):
Looks good!

schildbach:
Merged.

PR #2983, opened 2023-04-03
TransactionOutPoint: make `hash` and `index` final/immutable

msgilligan:
* make `hash` and `index` `final`
* Remove deprecated `setHash()` and `setIndex()`
* Fix generation of  `b58`  in `FullBlockTestGenerator` that relies on `setIndex()`

This depends on PR #2975

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
@@ -29,7 +29,6 @@ import org.bitcoinj.wallet.KeyBag;
 import org.bitcoinj.wallet.RedeemData;
 
 import javax.annotation.Nullable;
-import java.nio.Buffer;
 import java.nio.BufferOverflowException;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
@@ -51,9 +50,9 @@ public class TransactionOutPoint {
             new TransactionOutPoint(ByteUtils.MAX_UNSIGNED_INTEGER, Sha256Hash.ZERO_HASH);
 
     /** Hash of the transaction to which we refer. */
-    private Sha256Hash hash;
+    private final Sha256Hash hash;
     /** Which output of that transaction we are talking about. */
-    private long index;
+    private final long index;
 
     // This is not part of bitcoin serialization. It points to the connected transaction.
     Transaction fromTx;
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -226,30 +226,10 @@ public Sha256Hash getHash() {
...
-    /**
-     * @param hash new hash
-     * @deprecated Don't mutate this class -- create a new instance instead.
-     */
-    @Deprecated
-    void setHash(Sha256Hash hash) {
-        this.hash = hash;
-    }
-
     public long getIndex() {
```
schildbach, line 229: I think we should at the same time change the accessor names (probably providing deprecations).

core/src/test/java/org/bitcoinj/core/FullBlockTestGenerator.java
```diff
@@ -1131,7 +1131,8 @@ public boolean add(Rule element) {
         {
             Transaction tx = new Transaction(params);
             tx.addOutput(new TransactionOutput(tx, ZERO, new byte[] {}));
-            b58.getSpendableOutput().outpoint.setIndex(42);
+            TransactionOutPoint outPoint = b58.getSpendableOutput().outpoint;
+            b58.getSpendableOutput().outpoint = new TransactionOutPoint(42, outPoint.getHash());;
```
schildbach, line 1135: Have you checked if the other fields need copying?

msgilligan: Well, I ran the tests and they worked. I'm going to change this line to produce a "connected" outPoint by using a different constructor -- one that _should_ result in all the same fields.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
msgilligan: Fixed and force-pushed.

core/src/test/java/org/bitcoinj/core/FullBlockTestGenerator.java
msgilligan: Changes force-pushed.

schildbach:
Merged.

PR #2985, opened 2023-04-03
TransactionInput: divorce from `Message`

schildbach:

* A static constructur `read()` replaces the native constructor that deserialized from a payload.
* A `serialize()` helper replaces `bitcoinSerialize()`.

This comes with a test.


msgilligan (approved):
LGTM

schildbach:
Merged.

PR #2988, opened 2023-04-03
TransactionOutPoint: refactor to interface w/nested `Connected` class

msgilligan:
* Wrap `TransactionOutPoint` in a parent interface
* The implementing class is `TransactionOutPoint.Connected` and has no behavior changes.

The next step is to add an "unconnected" implementation and begin to distinguish the two with `instanceof` (or possibly an `isConnected()` method.)

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
@@ -39,222 +39,236 @@ import static org.bitcoinj.base.internal.Preconditions.checkState;
 
 /**
  * <p>This message is a reference or pointer to an output of a different transaction.</p>
- * 
+ *
  * <p>Instances of this class are not safe for use by multiple threads.</p>
  */
-public class TransactionOutPoint {
-    public static final int BYTES = 36;
-
-    /** Special outpoint that normally marks a coinbase input. It's also used as a test dummy. */
-    public static final TransactionOutPoint UNCONNECTED =
-            new TransactionOutPoint(ByteUtils.MAX_UNSIGNED_INTEGER, Sha256Hash.ZERO_HASH);
-
-    /** Hash of the transaction to which we refer. */
-    private final Sha256Hash hash;
-    /** Which output of that transaction we are talking about. */
-    private final long index;
-
-    // This is not part of bitcoin serialization. It points to the connected transaction.
-    Transaction fromTx;
-
-    // The connected output.
-    TransactionOutput connectedOutput;
-
-    /**
-     * Deserialize this transaction outpoint from a given payload.
-     *
-     * @param payload payload to deserialize from
-     * @return read transaction outpoint
-     * @throws BufferUnderflowException if the read message extends beyond the remaining bytes of the payload
-     */
-    public static TransactionOutPoint read(ByteBuffer payload) throws BufferUnderflowException, ProtocolException {
-        Sha256Hash hash = Sha256Hash.read(payload);
-        long index = ByteUtils.readUint32(payload);
-        return new TransactionOutPoint(index, hash);
-    }
-
-    public TransactionOutPoint(long index, Transaction fromTx) {
-        super();
-        checkArgument(index >= 0 && index <= ByteUtils.MAX_UNSIGNED_INTEGER, () ->
-                "index out of range: " + index);
-        this.index = index;
-        this.hash = fromTx.getTxId();
-        this.fromTx = fromTx;
-    }
-
-    public TransactionOutPoint(long index, Sha256Hash hash) {
-        super();
-        checkArgument(index >= 0 && index <= ByteUtils.MAX_UNSIGNED_INTEGER, () ->
-                "index out of range: " + index);
-        this.index = index;
-        this.hash = hash;
-    }
-
-    public TransactionOutPoint(TransactionOutput connectedOutput) {
-        this(connectedOutput.getIndex(), connectedOutput.getParentTransactionHash());
-        this.connectedOutput = connectedOutput;
-    }
-
-    /**
-     * Write this transaction outpoint into the given buffer.
-     *
-     * @param buf buffer to write into
-     * @return the buffer
-     * @throws BufferOverflowException if the outpoint doesn't fit the remaining buffer
-     */
-    public ByteBuffer write(ByteBuffer buf) throws BufferOverflowException {
-        buf.put(hash.serialize());
-        ByteUtils.writeInt32LE(index, buf);
-        return buf;
-    }
-
-    /**
-     * Allocates a byte array and writes this transaction outpoint into it.
-     *
-     * @return byte array containing the transaction outpoint
-     */
-    public byte[] serialize() {
-        return write(ByteBuffer.allocate(BYTES)).array();
-    }
-
-    /** @deprecated use {@link #serialize()} */
-    @Deprecated
-    public byte[] bitcoinSerialize() {
-        return serialize();
-    }
-
-    /** @deprecated use {@link #BYTES} */
-    @Deprecated
-    public int getMessageSize() {
-        return BYTES;
-    }
-
-    /**
-     * An outpoint is a part of a transaction input that points to the output of another transaction. If we have both
-     * sides in memory, and they have been linked together, this returns a pointer to the connected output, or null
-     * if there is no such connection.
-     */
-    @Nullable
-    public TransactionOutput getConnectedOutput() {
-        if (fromTx != null) {
-            return fromTx.getOutputs().get((int) index);
-        } else if (connectedOutput != null) {
-            return connectedOutput;
-        }
-        return null;
-    }
-
-    /**
-     * Returns the pubkey script from the connected output.
-     * @throws java.lang.NullPointerException if there is no connected output.
-     */
-    public byte[] getConnectedPubKeyScript() {
-        byte[] result = Objects.requireNonNull(getConnectedOutput()).getScriptBytes();
-        checkState(result.length > 0);
-        return result;
-    }
-
-    /**
-     * Returns the ECKey identified in the connected output, for either P2PKH, P2WPKH or P2PK scripts.
-     * For P2SH scripts you can use {@link #getConnectedRedeemData(KeyBag)} and then get the
-     * key from RedeemData.
-     * If the script form cannot be understood, throws ScriptException.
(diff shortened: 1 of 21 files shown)


core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -52,58 +57,101 @@ public interface TransactionOutPoint {
...
+     *
+     * @param payload payload to deserialize from
+     * @return read transaction outpoint (currently returns as {@link TransactionOutPoint.Connected} for compatibility)
+     * @throws BufferUnderflowException if the read message extends beyond the remaining bytes of the payload
+     */
+    static Connected read(ByteBuffer payload) throws BufferUnderflowException, ProtocolException {
+        Sha256Hash hash = Sha256Hash.read(payload);
+        long index = ByteUtils.readUint32(payload);
+        return new Connected(index, hash);
+    }
```
schildbach: Nitpick: I would put read before write/serialize.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -52,58 +57,101 @@ public interface TransactionOutPoint {
...
-                    "index out of range: " + index);
-            this.index = index;
-            this.hash = hash;
+        /**
+         * @deprecated Use {@link #hash()}
+         */
+        @Deprecated
+        public Sha256Hash getHash() {
+            return hash();
         }
```
schildbach: Shouldn't this method go to the interface, as default?

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -52,58 +57,101 @@ public interface TransactionOutPoint {
...
-        public Connected(TransactionOutput connectedOutput) {
-            this(connectedOutput.getIndex(), connectedOutput.getParentTransactionHash());
-            this.connectedOutput = connectedOutput;
+        /**
+         * @deprecated Use {@link #index()}
+         */
+        @Deprecated
+        public long getIndex() {
+            return index();
+        }
```
schildbach, line 143: Same here.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -139,6 +189,36 @@ public byte[] bitcoinSerialize() {
         public int getMessageSize() {
             return BYTES;
         }
+    }
+
+    class Connected extends Unconnected implements TransactionOutPoint  {
+
+        /** Special outpoint that normally marks a coinbase input. It's also used as a test dummy. */
+        public static final Connected UNCONNECTED =
+                new Connected(ByteUtils.MAX_UNSIGNED_INTEGER, Sha256Hash.ZERO_HASH);
```
schildbach, line 199: Why is this in `Connected` rather than in `Unconnected`?

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -139,6 +189,36 @@ public byte[] bitcoinSerialize() {
...
+        public Connected(long index, Transaction fromTx) {
+            super(fromTx.getTxId(), index);
+            this.fromTx = fromTx;
+        }
+
+        public Connected(long index, Sha256Hash hash) {
+            super(hash, index);
+            checkArgument(index >= 0 && index <= ByteUtils.MAX_UNSIGNED_INTEGER, () ->
+                    "index out of range: " + index);
+        }
```
schildbach: Why do these exist? Seems like they only create an unconnected output.

Perhaps we should add an assert to the `Connected` constructors: at least one of `fromTx` and `connectedOutput` must be set. Or even `fromTx` always, I'm not sure. But both `null` should be disallowed at this point, otherwise we wouldn't have the `Unconnected` subclass.

schildbach (commented):
Generally I like the idea of `Connected` extends `Unconnected`.

I wonder if having an interface for `TransactionOutPoint` is perhaps overkill – will we ever have other impls?

I noticed that in your first commit you moved everything over to `Unconnected` and in the second commit moved a lot back into the superclass (interface). This creates some churn. I would have tried to really extract one class from the other, e.g. noticing that `BYTES` and the deprecated methods will probably always stay in the superclass.

But ok, as long as the end result is clean, and the intermediate steps don't break builds, it's fine I guess…

Sometimes I think if we should add a test to each `Test` class that simply calls all the deprecated methods. That way we'd

* have them tested (I had a deprecated method that long threw NPE, but I never notced because we're not testing them)
* have a notice that a signature changed (by looking at the diff)

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -89,19 +90,19 @@ static Connected read(ByteBuffer payload) throws BufferUnderflowException, Proto
 
     class Unconnected implements TransactionOutPoint {
         /** Hash of the transaction to which we refer. */
-        private final Sha256Hash hash;
+        @Nonnull private final Sha256Hash hash;
         /** Which output of that transaction we are talking about. */
         private final long index;
 
-        public Unconnected(Sha256Hash hash, long index) {
+        public Unconnected(@Nonnull Sha256Hash hash, long index) {
```
schildbach, line 97: Why are using this `@Nonnull` lately? I thought everything without `@Nullable` is required, at least for our refactored classes?

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -102,159 +66,238 @@ public TransactionOutPoint(TransactionOutput connectedOutput) {
...
-        return index() == other.index() && hash().equals(other.hash());
+        /**
+         * Allocates a byte array and writes this transaction outpoint into it.
+         *
+         * @return byte array containing the transaction outpoint
+         */
+        @Override
+        public byte[] serialize() {
+            return write(ByteBuffer.allocate(BYTES)).array();
+        }
```
schildbach, line 180: I would put this into the interface, as a default method.

msgilligan:
> Generally I like the idea of `Connected` extends `Unconnected`.
> 
> I wonder if having an interface for `TransactionOutPoint` is perhaps overkill – will we ever have other impls?

It might be overkill, but it is slightly more flexible. We should probably make the container an abstract base class and make private constructors for `Connected` and `Unconnected` to emulate "sealed class" behavior. I was planning on doing that in a future commit once the constructors and static methods are stabilized.

> I noticed that in your first commit you moved everything over to `Unconnected` and in the second commit moved a lot back into the superclass (interface). This creates some churn. I would have tried to really extract one class from the other, e.g. noticing that `BYTES` and the deprecated methods will probably always stay in the superclass.

I tried to do it with refactoring and this was the best I could come up with and still get a clean history.  (I had one or two failed attempts, I've never done this exact type of refactoring before)


> Sometimes I think if we should add a test to each `Test` class that simply calls all the deprecated methods. That way we'd
> 
> * have them tested (I had a deprecated method that long threw NPE, but I never notced because we're not testing them)
> * have a notice that a signature changed (by looking at the diff)

Good point.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
msgilligan: > Why are we using this `@Nonnull` lately?

I guess I"m doing it temporarily as I"m working to document the current behavior and/or start declaring that `null` will no longer be accepted as a parameter.

>  I thought everything without `@Nullable` is required, at least for our refactored classes?

Yes, that is how this class should look when it's finished. Unfortunately nullability is currently hard for me to figure out.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
msgilligan: This is temporary. At present `Connected` class can be either "connected" or "unconnected" based on mutable fields in the class. There is more conversion that needs to be done in the other classes before it can fully use the new semantics.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
msgilligan: Because currently the `Connected` class is a "transformer" it can "connect" an "unconnect".  So I've only updated code to use `Unconnected` when those instances never need to transform. Further steps of refactoring will be needed to handle connection and unconnection properly.

msgilligan:
> > I wonder if having an interface for `TransactionOutPoint` is perhaps overkill – will we ever have other impls?
> 
> …  We should probably make the container an abstract base class and make private constructors for `Connected` and `Unconnected` to emulate "sealed class" behavior. I was planning on doing that in a future commit once the constructors and static methods are stabilized.

Oh yeah, I was tired when I wrote the above. I forgot to mention, I want to be able to use `instanceof` to distinguish between connected and unconnected. This means `Connected` can't be a subclass of `Unconnected`. And upon further reflection, that means the parent class should probably be abstract and contain the two fields.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
schildbach: The mutability would be taken care of via #2997 – we can merge that either before or after your PR, I'm happy to rebase if needed.

schildbach: But going back to my above concern: I think even in this temporary state, the `UNCONNECTED` instance should go either into the `Unconnected` subclass or simply stay at the `TransactionOutPoint` interface level.

Leonardbrett:
My first investment with Charlotte give me profits of over $7,000 and ever since then she has never failed to deliver and I can even say she's the must sincere broker I have ever known'' communicate with her to be a successful investor. On WhatsApp:+𝟭𝟰𝟳𝟱𝟮𝟰𝟮𝟲𝟯𝟯𝟴

msgilligan:
@schildbach This is very obsolete. We have made a lot of related progress in the last year. `TransactionOutPoint` is now fully-immutable and after a few more PRs are merged I think we will be ready to separate the basic, txid-and-index-only `TransactionOutPoint` from the connected version. I don't think there is anything in this PR that is worth keeping, but let's keep it open as a reference until we have finalized its replacement.

PR #2989, opened 2023-04-04
Transaction{Input,Output}.equals: compare parent txId not object reference

msgilligan:


schildbach:
Merged.

msgilligan:
> I know I merged this, but now my stomach is aching a bit. Those equals() methods are highly customized to the way our business logic expects equality,

We should probably define this explicitly and document it (and possibly modify after review)

> and less conformant to how you'd expect equality for value-based classes.

Well the behavior in this PR is not really standard Java behavior. It's only comparing based on parent txId.  (Previously it was comparing based on parent object identity.)

> Especially if you look at TransactionOutput: its equality depends on where it is being used within its parent (index).

The problem I had was a test failing because two identical transactions were in different objects and the comparison was using `==`.  There are several choices for how we implement this:

1. Use parent object ID (before this PR)
2. Use parent TxID (this PR, which uses the default `parent.equals(other)`)
3. Use all fields of parent
4. Use parent TxID and index in the parent (and perhaps parent "completeness")
5. Use all parent fields. 

This will also get easier when we can distinguish between connected and unconnected transactions and between transactions being constructed (builder) and signed immutable transactions.

schildbach:
No, I think the change is a step in the right direction and it's low risk. I just wanted to raise awareness that this equals method in TransactionOutput is a can of worms and we need to be careful.

PR #2990, opened 2023-04-04
TransactionOutPoint: store connectedOutput when storing fromTx

msgilligan:
This requires related changes:

* Clear connectedOutput when clearing fromTx
* Change comparison in TransactionInputTest to reflect new behavior

Depends on PR #2737

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
@@ -80,6 +80,7 @@ public class TransactionOutPoint {
         this.index = index;
         this.hash = fromTx.getTxId();
         this.fromTx = fromTx;
+        this.connectedOutput = Objects.requireNonNull(fromTx.getOutput(index));
     }
 
     public TransactionOutPoint(long index, Sha256Hash hash) {
--- a/core/src/main/java/org/bitcoinj/core/TransactionInput.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionInput.java
@@ -435,6 +435,7 @@ public class TransactionInput {
             // The outpoint is connected using a "standard" wallet, disconnect it.
             connectedOutput = outpoint.fromTx.getOutput((int) outpoint.index());
             outpoint.fromTx = null;
+            outpoint.connectedOutput = null;
         } else if (outpoint.connectedOutput != null) {
             // The outpoint is connected using a UTXO based wallet, disconnect it.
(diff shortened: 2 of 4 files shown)


msgilligan:
I added an additional commit that detects the invalid index a few lines earlier.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -80,6 +80,7 @@ public TransactionOutPoint(long index, Transaction fromTx) {
         this.index = index;
         this.hash = fromTx.getTxId();
         this.fromTx = fromTx;
+        this.connectedOutput = Objects.requireNonNull(fromTx.getOutput(index));
```
schildbach, line 83: This should make line 96 unnecessary (or maybe converted to an assert rather than an assignment).

schildbach (commented):
c0fa8f1be3e19e64c31e026493b32dbd80ae7893 seems like a good idea. If we find out why the test fails, I've got only minor comments.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
schildbach: Also, the first if branch in `getConnectedOutput()` is potentially unnecessary after this change?

`if (fromTx != null) `

schildbach:
I had to investigate into this territory in my quest to eliminate `getParentTransaction()` from input and output.

#2997 makes `connectedOutput` and `fromTx` and introduces methods for the cases where those need mutation. It doesn't fail any tests.

What do you think, should I rebase mine on yours?

schildbach:
Thinking a bit into the future, I'm not sure if storing `connectedOutput` makes sense any more. If all goes well, a transaction output will only be the touple `value` and `script`. No index, no parent transaction. For everything that involves how/where a `TransactionOutput` is being used, you'd need to go through `Ttransaction` anyway.

On the other hand, and I think this may be responsible for the test failures, for the business logic currently it seems to be a difference if `fromTx` is set or if `connectedOutput` is set. The latter case seems to be used for UTXO-based wallets exclusively. We need to work that out I guess.

Leonardbrett:
Hello everyone I'm earning from this great platform I'm earn $4872 after every 7th days from investment of $500 This woman offers fast and legit investment. Connect with Charlotte on 🅦🅗🅐🅣🅢🅐🅐🅟+𝟭𝟰𝟳𝟱𝟮𝟰𝟮𝟲𝟯𝟯𝟴

PR #2993, opened 2023-04-04
TransactionWitness: add read and write helpers

schildbach:
* A static constructur `read()` deserializes from a payload.
* A `serialize()` helper gets the serialized bytes.

This comes with a test.


schildbach:
I pushed two commits on top:

* make immutable, except for the byte arrays themselves
* require non-null pushes

msgilligan (approved):
LTGM

schildbach:
Merged.

PR #2994, opened 2023-04-05
Script: migrate two native constructors into a `parse()` static constructor AND make immutable

schildbach:
This is a prerequisite for further refactortings (extracting an interface).


schildbach:
I pushed three more commits on top:

* migrate a native constructor that wraps chunks to `of()` static constructor
* make field `chunks` immutable
* make field `program` immutable

There is more to come.

schildbach:
Better to review #2995 first.

schildbach:
Rebased, and ready for review.

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -229,38 +229,91 @@ public enum VerifyFlag {
...
     /**
-     * Construct a Script that copies and wraps the programBytes array. The array is parsed and checked for syntactic
-     * validity.
-     * @param programBytes Array of program bytes from a transaction.
-     * @param creationTime creation time of the script
+     * To run a script, first we parse it which breaks it up into chunks representing pushes of data or logical
+     * opcodes. Then we can run the parsed chunks.
      */
-    public Script(byte[] programBytes, Instant creationTime) throws ScriptException {
+    private static void parse(byte[] program, List<ScriptChunk> chunks) throws ScriptException {
```
msgilligan, line 289: This method should be renamed `parseChunks` (it already has a different return type than the other `static` `parse()` methods) and should allocate the `List<ScriptChunk>` internally and return it.

msgilligan (changes requested):
Looks good, but I think the `static void parse()` method should be renamed `parseChunks() and return a `List<ScriptChunks>`

core/src/main/java/org/bitcoinj/script/Script.java
schildbach: I renamed the method to `parseIntoChunks()`. I could not change how the argument handling works, because two of the three usages use this method for partial parsing (in case of exception). I know this is bad, but it's out of scope of this PR to change this.

msgilligan (approved):
Let's merge!

schildbach:
Merged.

PR #2996, opened 2023-04-07
Transaction: lockTime() return LockTime.unset() not null

msgilligan:


schildbach:
I wonder how did the null sneak into your instance? The field itself is non-null. I just checked and all constructors initialize it somewhere in their path.

msgilligan:
> I wonder how did the null sneak into your instance? The field itself is non-null. I just checked and all constructors initialize it somewhere in their path.

I was running unit tests on consensusj using the `0.17-alpha1` release. Let me see if I can reproduce it.  (I added a workaround so, I'll remove it.)

msgilligan:
The `null` value for `lockTime` was coming from a mock `Transaction`  that didn't set that field. 🤦‍♂️

Closing.

PR #2997, opened 2023-04-08
TransactionOutPoint: make fields `connectedOutput` and `fromTx` immutable

schildbach:
I'm not sure if we want to keep these fields in future, but at least they're immutable now.

It introduces methods for the cases where the fields need mutation. In this case, the entire `TransactionOutPoint` is reconstructed.


schildbach:
There is a second commit on top:

* TransactionOutPoint: don't use accessors in `equals()` and `hashCode()`

schildbach:
Marking this as draft for now, because I don't want to sneak past #2988 and #2990.

I think this PR could be a worthwhile preparation for #2988 though, because it makes the different states of "connectedness" and its usages more clear.

msgilligan (approved):
LGTM

schildbach:
Merged.

PR #3001, opened 2023-04-08
Move NetworkParameters.ProtocolVersion to standalone class in o.b.core

msgilligan:
Consists of 3 commits:

1. `NetworkParameters`: deprecate `getProtocolVersionNum()` and use `ProtocolVersion.<instance>.getProtocolVersion()` instead
2. rename `getBitcoinProtocolVersion()` to `intValue()`
3. Move  `NetworkParameters.ProtocolVersion` to standalone class in `o.b.core`

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+++ b/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
@@ -550,26 +550,11 @@ public abstract class NetworkParameters {
         return verifyFlags;
     }
 
-    public abstract int getProtocolVersionNum(final ProtocolVersion version);
-
-    public static enum ProtocolVersion {
-        MINIMUM(70000),
-        @Deprecated
-        PONG(60001),
-        BLOOM_FILTER(70001), // BIP37
-        BLOOM_FILTER_BIP111(70011), // BIP111
-        WITNESS_VERSION(70012),
-        FEEFILTER(70013), // BIP133
-        CURRENT(70013);
-
-        private final int bitcoinProtocol;
-
-        ProtocolVersion(final int bitcoinProtocol) {
-            this.bitcoinProtocol = bitcoinProtocol;
-        }
-
-        public int getBitcoinProtocolVersion() {
-            return bitcoinProtocol;
-        }
+    /**
+     * @deprecated  use {@link ProtocolVersion#intValue()}
+     */
+    @Deprecated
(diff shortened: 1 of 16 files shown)


core/src/main/java/org/bitcoinj/params/BitcoinNetworkParams.java
```diff
@@ -247,7 +247,11 @@ public MonetaryFormat getMonetaryFormat() {
...
     }
 
+    /**
+     * @deprecated  use {@link ProtocolVersion#getBitcoinProtocolVersion()}
+     */
     @Override
+    @Deprecated
     public int getProtocolVersionNum(final ProtocolVersion version) {
         return version.getBitcoinProtocolVersion();
     }
```
schildbach: This deprecated method could be in `NetworkParameters`, and from its 2 subclasses removed.

A `deprecatedMembers()` test could be added. See 4e75444644c820f797bcef92f8cf0a66f59d4b1c as an example.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
```diff
@@ -572,6 +572,17 @@ public static enum ProtocolVersion {
...
+            return bitcoinProtocol;
+        }
+
+        /**
+         * @deprecated Use {@link #intValue()}
+         */
+        @Deprecated
         public int getBitcoinProtocolVersion() {
             return bitcoinProtocol;
         }
```
schildbach: Again, the old method could get a `deprecatedMembers()` test.

schildbach (commented):
In general, looks good. Some comments added.

schildbach:
Fwiw, the version indirection via `NetworkParameters` was meant as an extension point for altcoins. But I think we should get rid of it anyway.

core/src/main/java/org/bitcoinj/params/BitcoinNetworkParams.java
msgilligan: Fixed and force-pushed.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
msgilligan: Fixed and force-pushed.

schildbach:
Merged.

PR #3004, opened 2023-04-08
Transaction: remove caches for (witness) transaction ID

schildbach:
I don't see any regression in unit test runtime, so it seems fine.

On the plus side, this removes a fair bit of complexity from `Transaction`, and usages of `TransactionInput.parent` and `TransactionOutput.parent`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
+++ b/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
@@ -231,7 +231,7 @@ public class BitcoinSerializer extends MessageSerializer {
         } else if (command.equals("getheaders")) {
             return new GetHeadersMessage(payload);
         } else if (command.equals("tx")) {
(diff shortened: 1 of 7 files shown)


Leonardbrett:
My first investment with Charlotte give me profits of over $7,000 and ever since then she has never failed to deliver and I can even say she's the must sincere broker I have ever known'' communicate with her to be a successful investor. On WhatsApp:+𝟭𝟰𝟳𝟱𝟮𝟰𝟮𝟲𝟯𝟯𝟴

msgilligan (approved):
LGTM.

msgilligan:
> I don't see any regression in unit test runtime, so it seems fine.

Although there may be real-world use-cases that aren't reflected in the unit tests. But we can worry about performance once we've refactored to a simpler, less-mutable architecture.

schildbach:
Merged.

PR #3005, opened 2023-04-09
ScriptExecution: move execution and verification helpers to here from `Script`

schildbach:


msgilligan (approved):
This looks like an improvement, but I would suggest renaming it to `ScriptExecutor` and making the `public static` methods instance methods. That will allow overriding, mocking, etc.

schildbach:
Hmm, it would mean that all classes using `ScriptExecutor` keep a mutable reference somewhere, with maybe a setter to override.

The class is just a place to move the static methods from `Script` to, more like a graveyard than a well thought out interface or something.

schildbach:
Rebased.

PR #3008, opened 2023-04-09
Wallet: constructor use Network (w/deprecation)

msgilligan:


schildbach:
If I understand this correctly, this changes only one constructor. Do you plan to migrate all the static constructors too?

msgilligan:
> If I understand this correctly, this changes only one constructor.

Yes, this changes the only constructor (and related external factory) and internal usage of `params`/`network`

> Do you plan to migrate all the static constructors too?

Yes. See https://github.com/bitcoinj/bitcoinj/pull/3009

schildbach:
Merged.

PR #3010, opened 2023-04-10
Transaction.coinbase: make complete transactions

msgilligan:
* Transaction.coinbase() single factory method that makes a complete tx
* Block: use this method to make genesis blocks
* FakeTxBuilder: use this method to make fake coinbase transactions

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -228,8 +228,7 @@ public class Block extends Message {
 
     public static Block createGenesis() {
         Block genesisBlock = new Block(BLOCK_VERSION_GENESIS);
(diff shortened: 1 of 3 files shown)


schildbach:
In my mind, this method is part of our future transaction builder. And I'd like Builder classes stay flexible and not be limited to single usecases, in this case coinbases with only out output.

msgilligan:
> In my mind, this method is part of our future transaction builder. And I'd like Builder classes stay flexible and not be limited to single usecases, in this case coinbases with only out output.

We can always add a method that takes a list of outputs (when needed or when we create the actual builder implementation). In the meantime this removes unneeded mutation and helps us identify (see PR #3011) all our use-cases that create Coinbase transactions.

msgilligan:
Rebased and force-pushed.

PR #3012, opened 2023-04-10
Wallet: in `signTransaction()` don't care about the witness being empty

schildbach:
It does not matter for the data being signed.

This removes `Script.createEmptyWitness()` as it's unnecessary as well.


msgilligan (approved):
LGTM

PR #3015, opened 2023-04-10
VersionMessage: remove field `fromAddr`

schildbach:
The field is unused by the Bitcoin protocol. Bitcoin Core fills it up with zeros. From now on, we'll do the same.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/VersionMessage.java
+++ b/core/src/main/java/org/bitcoinj/core/VersionMessage.java
@@ -71,10 +71,6 @@ public class VersionMessage extends Message {
      */
     public PeerAddress receivingAddr;
     /**
-     * The network address of the node emitting this message. Not used.
-     */
-    public PeerAddress fromAddr;
-    /**
      * User-Agent as defined in <a href="https://github.com/bitcoin/bips/blob/master/bip-0014.mediawiki">BIP 14</a>.
      * Bitcoin Core sets it to something like "/Satoshi:0.9.1/".
      */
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/core/VersionMessage.java
```diff
@@ -152,7 +146,7 @@ protected void parse(ByteBuffer payload) throws BufferUnderflowException, Protoc
         localServices = Services.read(payload);
         time = Instant.ofEpochSecond(ByteUtils.readInt64(payload));
         receivingAddr = new PeerAddress(payload, new DummySerializer(0));
-        fromAddr = new PeerAddress(payload, new DummySerializer(0));
+        Buffers.skipBytes(payload, 26); // addr_from
```
msgilligan, line 149: Should we have a constant defined for this `26` value?

core/src/main/java/org/bitcoinj/core/VersionMessage.java
```diff
@@ -174,7 +168,7 @@ public void bitcoinSerializeToStream(OutputStream buf) throws IOException {
         buf.write(localServices.serialize());
         ByteUtils.writeInt64LE(time.getEpochSecond(), buf);
         receivingAddr.bitcoinSerializeToStream(buf);
-        fromAddr.bitcoinSerializeToStream(buf);
+        buf.write(new byte[26]); // addr_from
```
msgilligan, line 171: Here it is again. We should have a constant.

msgilligan (approved):
LGTM, but there should be a constant defined for `26`.

schildbach:
I added the constant and merged.

PR #3016, opened 2023-04-10
VersionMessage: for field `receivingAddress` don't use `PeerAddress`    

schildbach:
Rather, use `InetSocketAddress` and `Services`. This makes `time` in `PeerAddress`
a required field.

Note that currently this field isn't even properly used by bitcoinj. I hope we can fix that soon.

This is a followup PR to #3015.


msgilligan (approved):
LGTM

schildbach:
Fixed the test errors and merged.

PR #3017, opened 2023-04-11
TransactionSigners: use foreach loops for processing inputs

msgilligan:


schildbach:
I think we should avoid using `In/Out.getIndex()`. It needs the parent, which I'm trying to migrate away from currently.

Even if at some point we've migrated `getIndex()` to `Transaction.getInput/OutputIndex()`, it's costly and will have the problem that output equalness is hard to get right.

schildbach:
We could of course still use the foreach syntax, but I'd advocate for counting the index ourselves.

Is there a Stream that counts element indexes?

schildbach:
> Yes (sort of.) I'll try making a PR using streams and we can look at it.

:+1:

msgilligan:
Converting to DRAFT as a reminder to explore using streams or other alternative to `for(;;)`

PR #3019, opened 2023-04-11
Buffers.readLengthPrefixedBytes: handle prefix too big

msgilligan:
This assumes that `Buffers.readLengthPrefixedBytes` will only be used for parsing data that conforms to the max size for Bitcoin P2P messages. If that is unreasonable assumptions we could alternatively make sure that the `VarInt` does not exceed the remaining size of the `ByteBuffer`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/internal/Buffers.java
+++ b/core/src/main/java/org/bitcoinj/base/internal/Buffers.java
@@ -17,6 +17,8 @@
 package org.bitcoinj.base.internal;
 
 import org.bitcoinj.base.VarInt;
+import org.bitcoinj.core.Message;
+import org.bitcoinj.core.ProtocolException;
 
 import java.nio.BufferOverflowException;
 import java.nio.BufferUnderflowException;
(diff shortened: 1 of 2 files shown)


schildbach:
I actually removed these checks, because they are not necessary with ByteBuffer.

core/src/main/java/org/bitcoinj/base/internal/Buffers.java
```diff
@@ -54,10 +56,12 @@ public static byte[] readBytes(ByteBuffer buf, int length) throws BufferUnderflo
...
      * @return read bytes
      * @throws BufferUnderflowException if the read value extends beyond the remaining bytes of the buffer
+     * @throws ProtocolException if the length prefix is greater than the maximum allowed message size
      */
-    public static byte[] readLengthPrefixedBytes(ByteBuffer buf) throws BufferUnderflowException {
-        int length = VarInt.read(buf).intValue();
-        return readBytes(buf, length);
+    public static byte[] readLengthPrefixedBytes(ByteBuffer buf) throws BufferUnderflowException, ProtocolException {
+        long length = VarInt.read(buf).longValue();
+        check(length < Message.MAX_SIZE, () -> new ProtocolException("length prefix too long"));
```
schildbach, line 63: I think we should check against `Integer.MAX_VALUE` here. This method should not know about Message. And if this is used from message parsing, the payload cannot exceed `Message.MAX_SIZE` anyway.

We could perhaps make the `Integer.MAX_VALUE` check a method of `VarInt`, somehting like `boolean fitsInt()`. Then there would be no need to convert to long and then back to int.

schildbach: Also I think we should continue to throw `BufferunderflowException`, not `ProtocolException`. Because in fact it's exactly that.

schildbach:
While investigating into this, I found that `VarInt.intValue()` was actually not implemented correctly. See #3024.

Then I implemented the guards as I proposed them in the chat: #3025

schildbach:
I think this has been taken care of via a22a446eb1cd5855e8a3376be6e7603753cae7a4.

msgilligan:
Yes, this has been fixed by the above.

PR #3020, opened 2023-04-11
PartialMerkleTree: divorce from `Message`

schildbach:

* Static constructor `read()` replaces the native constructor that deserialized from a payload.
* `write()` helper replaces `bitcoinSerializeToStream()`.
* `serialize()` and `getMessageSize()` helpers replace `bitcoinSerialize()`.


msgilligan (approved):
LGTM

schildbach:
Merged.

PR #3023, opened 2023-04-13
Transaction: remove the input and output re-reading

schildbach:
We needed this to correctly determine the message length, but that requirement is gone.


msgilligan (approved):
LGTM

PR #3024, opened 2023-04-13
VarInt: fix long values accepted as an int value

schildbach:
`-1` and `Integer.MIN_VALUE` (and all values inbetween) are in fact very large VarInts that can only be expressed as a long. So their test data have been moved from `VarIntTest.integerTestVectors` to `longTestVectors`.

`intValue()` is changed to guard against such cases, and a helper `fitsInt()` is introduced.

JavaDocs are added to `intValue()` and `longValue()` to clarify the expected value ranges.


msgilligan (approved):
I'm not crazy about the name `fitsInt()`, but I can't think of anything obviously  better...

schildbach:
Merged.

PR #3025, opened 2023-04-13
Buffers, Messages: in various `parse()` methods, throw `BufferUnderflowException` if read length doesn't fit an int

schildbach:
(followup PR to #3024) 

Previously, `ArithmeticException` was thrown which is unlikely to be handled by consumers.
    
`BufferUnderflowException` is the correct exception to throw, under the assumption that a giant length can't be followed by a collection of equally giant length.


msgilligan (approved):
LGTM

schildbach:
Merged.

PR #3027, opened 2023-04-13
BitcoinNetwork remove ID_UNITTESTNET

msgilligan:
It is no longer needed and will only be a source of confusion. (The UNITTEST network parameters has had the testnet ID string for a while now)

I realized this could be removed when updating the JavaDoc for PR #3026, so to avoid merge conflicts this PR depends on that one.


schildbach:
Merged.

PR #3028, opened 2023-04-13
WalletAppKit: add awaitTerminated() in close(), add basic unit test

msgilligan:
`.close()` is supposed to release resources, so we should wait for termination before returning.


schildbach:
Merged.

PR #3030, opened 2023-04-15
FeeFilterMessage.bitcionSerializeToStream: don't call super() 

msgilligan:
super() (incorrectly) logs an error

Depends on PR #3029

PR #3031, opened 2023-04-15
BaseMessage: make bitcoinSerializeToStream() abstract

msgilligan:
Rather than log an error at runtime, the compiler should fail with an error when subclasses of `BaseMessage` don't implement `bitcoinSerializeToStream()`.

Depends on PR #3030


schildbach:
This method will be replaced soon, but yes it's an improvement. Merged.

PR #2888, opened 2023-03-26
ForwardingService - migrate two methods to Wallet

msgilligan:
* waitForConfirmation - as waitForConfirmations + convenience method with 1 param
* sendTransaction

PR #3032, opened 2023-04-15
WIP: Wallet.sendTransaction(Transaction)

msgilligan:
TODO:
1. Make sure calling commitTx on an externally-provided transaction works properly and perhaps verifies the tx is from this wallet.
2. Consider broader refactoring/cleanup of transaction sending
3. Write tests

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -4353,6 +4353,38 @@ public class Wallet extends BaseTaggableObject
     }
 
     /**
+     * Initiate sending of a signed (should-be-immutable) transaction in a {@link Transaction}. It is expected that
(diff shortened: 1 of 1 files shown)


schildbach:
Fwiw, I currently use a combination of

* `Wallet.sendCoinsOffline()` (which combines `completeTx()` and `commitTx()`) and
* `PeerGroup.broadcastTransaction()`

to accomplish roughly the same goal. These two calls are handled by different modules of my app though, so I'm not sure if I could use the new method without issues. It would be a good incentive to modernize my sending code though.

Just to clarify, doesn't the broadcast send to roughly half of connected peers, rather than just one (as stated in your JavaDoc)?

Yes, tests would be great.

msgilligan:
Closing in favor of https://github.com/bitcoinj/bitcoinj/pull/2888 which was merged.

PR #3034, opened 2023-04-16
RejectMessage: make immutable

schildbach:


msgilligan (changes requested):
Rather than renaming the fields to match the (deprecated) getters, I would rename the accessors to (mostly) match the existing fields:
* `.message()`
* `.code()`
* `.reason()`
* `.rejectedHash()`

Once this is merged, I'll create a PR to eliminate the `@Nullable` field by using the ADT pattern.

schildbach:
Actually I tried to move the field names a bit closer to the actual spec, which imho has not very well though out names:

https://github.com/bitcoin/bips/blob/master/bip-0061.mediawiki

schildbach:
I think just `message` is misleading. At quick glance, it can be mistaken as a textual description of the reject code – similar to Java exception messages. We could rename it to "response-to-msg" (as per the spec), but that seems inwieldy. Which is why I've stuck to the `rejectedMessage` from the getter.

msgilligan:
> `message` is misleading … can be mistaken as a textual description of the reject code – similar to Java exception messages.

Good point.

> `rejectedMessage`

Yes, that is probably the best choice, then.

Maybe:

* `.rejectedMessage()`
* `.code()`
* `.reason()`
* `.rejectedHash()`

?

schildbach:
Ok, I put a commit on top:

* rename rejectCode → code
* rename rejectReason → reason

schildbach:
Are you ok with `rejectedObjectHash` (rather than `rejectedHash`)? Makes it more clear what hash we're talking about.

(I can imagine if you do the subclass refactoring this will be renamed to `rejectedTransactionHash` and `rejectedBlockHash` anyway.)

msgilligan:
> Are you ok with `rejectedObjectHash` 

What about `rejectedMessageHash()`?

schildbach:
Ok, renamed to `rejectedMessageHash` and merged.

msgilligan:
@schildbach points out that these messages are mostly gone, see: https://github.com/bitcoin/bitcoin/pull/15437

Issue #2887, opened 2023-03-26
BitcoinURI: Discover (via Service Loader mechanism) the correct Network from the URI string


schildbach:
There is a TODO in BitcoinURI, which actually breaks my app for any non-mainnet.

```
// TODO: Discover (via Service Loader mechanism) the correct Network from the URI string
```

schildbach:
For the records, my 0.16-based code uses `new BitcoinURI(null, uri);` with uri coming from a QR code, link or similar. The idea is it should let the class detect the network.

Two problems:

1. When used with `null` like above, the compile now cannot decide between the two overloads and crashes (!).
2. When I cast `(NetworkParameters) null` or use the suggested method from the deprecation notice, that one doesn't detect the network but has mainnet hardcoded.

schildbach:
https://github.com/bitcoinj/bitcoinj/pull/2898 will help with case 1 above.

msgilligan:
PR #3041 will help with solving problem 2 above.

schildbach:
I believe both of the above have been fixed.

PR #3041, opened 2023-04-17
BitcoinURI: migrate parsing to private constructor

msgilligan:
Using the private constructor allows us to enforce @Nonnull for Network, allow @Nullable for the deprecated NetworkParameters and allow the parsing code to be restructured to find the network from the input string when appropriate.

This is a step towards a solution to PR #2887.  Note that the `TODO` referenced in that issue is now split into two `TODO` s -- one for looking up the network from the input string and the other for implementing a service provider to support alt-networks.

### DIFF
--- a/core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
+++ b/core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
@@ -22,6 +22,7 @@ import org.bitcoinj.base.exceptions.AddressFormatException;
 import org.bitcoinj.base.Coin;
 import org.bitcoinj.base.Address;
 import org.bitcoinj.base.DefaultAddressParser;
+import org.bitcoinj.base.internal.Preconditions;
 import org.bitcoinj.core.NetworkParameters;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
```diff
@@ -147,6 +149,11 @@ public BitcoinURI(String input, @Nonnull Network network) throws BitcoinURIParse
             throw new BitcoinURIParseException("Bad URI syntax", e);
         }
 
+        // TODO: Determine actual network by parsing the address field of the input URI
+        // TODO: Discover (via Service Loader mechanism) alt-Networks from the scheme/prefix in the uri string
+        Network network = autoNetwork ? BitcoinNetwork.MAINNET : nullableNetwork;
```
schildbach, line 154: Isn't that still hardcoded to mainnet?

The previous behaviour was that it would parse either mainnet or testnet URIs (and I hoped in future also signet), and the app presents a specific warning to the user if there is a mismatch between network (which happens frequently!).

We need basically the same behaviour as `AddressParser.parseAddressAnyNetwork()`.

schildbach: So I think I it needs two conditional branches for line 200: one for if the network is given, and one for if it isn't.

msgilligan: > Isn't that still hardcoded to mainnet?

Yes. It doesn't change behavior from the current `master`. This PR just restructures constructors for the change you are looking for.

msgilligan:
This will need an update after PR #3047 is merged.

PR #3043, opened 2023-04-18
BitcoinURI: rework exception handling for amount field

schildbach:
* The check against `maxMoney` is removed. There BIP21 spec does not restrict the amount towards positive infinity. The new maximum is now dictated by `Coin.parseCoin()`: `Long.MAX_VALUE` satoshis
* The check for negative amounts now throws `OptionalFieldValidationException` directly, rather than re-throwing via `ArithmeticException`.
* Exception messages now use string concatenation like everywhere else.
* Tests around the new max amount are added/changed.

### DIFF
--- a/core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
+++ b/core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
@@ -224,15 +224,11 @@ public class BitcoinURI {
                 // Decode the amount (contains an optional decimal component to 8dp).
                 try {
                     Coin amount = Coin.parseCoin(valueToken);
-                    if (network.exceedsMaxMoney(amount))
-                        throw new BitcoinURIParseException("Max number of coins exceeded");
                     if (amount.signum() < 0)
-                        throw new ArithmeticException("Negative coins specified");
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
```diff
@@ -224,15 +224,11 @@ private void parseParameters(Network network, String[] nameValuePairTokens) thro
                 // Decode the amount (contains an optional decimal component to 8dp).
                 try {
                     Coin amount = Coin.parseCoin(valueToken);
-                    if (network.exceedsMaxMoney(amount))
```
msgilligan, line 227: removal of this line allows us to remove the `Network` parameter from `parseParameters()`

msgilligan (approved):
LGTM

core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
schildbach: Yes, we could.

schildbach:
Merged.

PR #3047, opened 2023-04-19
BitcoinURI: migrate to static from native constructors

schildbach:
While reviewing #3041 I figured we should migrate the new (for this release cycle) constructors to static constructors, because the signatures get in each others way all the time. After that, I'd probably rework #3041 to not use the `autoNetwork` parameter – it feels redundant to me.


msgilligan (approved):
LGTM

msgilligan:
>  I'd probably rework #3041 to not use the `autoNetwork` parameter – it feels redundant to me.

It looks redundant, but it was needed because we couldn't have two constructors with the same parameters: one with a nullable `Network` and the other with a non-nullable `Network`. By moving to static constructors we no longer have this conflict.

schildbach:
Merged.

PR #3049, opened 2023-04-19
MarriedKeyChain: remove married key chains

schildbach:
This is a proposal to remove "married key chains".

The current implementation is based on plain P2SH. I assume as of today one would prefer an implementation based on taproot or at least segwit (P2WSH).

Also removes the `marry` action from `WalletTool`.


schildbach:
@devrandom Afair you've been using married keychains. Do you still need this?

schildbach:
Note the `isFollowing` and `sigsRequiredToSpend` fields in the `DeterministicKey` message of the wallet protobuf are preserved for now, because they might come in handy for a future implementation and in general it's hard to remove fields from protobufs.

schildbach:
@oscarguindzberg I think you have been using this too?

schildbach:
@troggy Are you still using married key chains in any project?

troggy:
@schildbach no, not using

schildbach:
As of today, it looks like we'll remove `MarriedKeyChain`.

As described in the comment above, the "marriage-related" fields in the wallet protobuf are preserved after merging this. So theoretically it would be possible open married wallets in old apps after they have accidently been openend (and saved) with a bitcoinj version that has this PR merged.

Is this desired behaviour? Or should simply remove the fields, and mark their protobuf ID as not to be re-used? This would destroy the information from the "marriage-related" fields.

There would also be the option to keep the fields and throw an exception when loading a wallet with `isFollowing == true` or `sigsRequiredToSpend != 1`.

msgilligan (approved):
LGTM. Do we want to provide deprecated methods that have the `isMarried` parameter that throw if it is set?  (That should make it easier to migrate apps to the new release...)

schildbach:
> LGTM. Do we want to provide deprecated methods that have the `isMarried` parameter that throw if it is set? (That should make it easier to migrate apps to the new release...)

I did something like this last time and you prefered the compiler to throw and not the runtime…

msgilligan:
> > LGTM. Do we want to provide deprecated methods that have the `isMarried` parameter that throw if it is set? (That should make it easier to migrate apps to the new release...)
> 
> I did something like this last time and you prefered the compiler to throw and not the runtime…

I'm not always consistent 🤣, but if I remember correctly in the previous case the runtime throw would always happen. In this case the runtime throw will only happen in the case a married wallet is involved and won't happen in the normal case.

msgilligan:
> So theoretically it would be possible open married wallets in old apps after they have accidentally been opened (and saved) with a bitcoinj version that has this PR merged.

This seems like the safer choice. Not having any experience with married wallets I would generally lean towards the safer choice,.

schildbach:
Merged.

schildbach:
(I added some deprecated methods back before merging)

PR #3050, opened 2023-04-19
gitlab-ci.yml: create template job

msgilligan:
* Add YAML anchor &debian-job-configuration
* Add template job named .debian-test
* Modify existing bullseye-jdk11 and bookworm-jdk17 jobs to extend it

This PR has two advantages:

1. It removes the global before_script/after_script settings that can interfere if we add other jobs
2. It allow common-code between the two debian jobs to be more easily maintained.

Having to use both YAML anchors and a template job to do this is not as elegant as I would like. We could just duplicate the `before_script` and `after_script` in each job.

(Update: see PR #3051 which simply puts the `before_script` and `after_script` in each job.)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,25 +1,29 @@
-before_script:
-  - apt-get update
+.debian-job-configuration: &debian-job-configuration
+  before_script:
+    - apt-get update
+  after_script:
+    - gradle --version
+  script:
+    - apt-get -y install $JDK_PKG gradle
+    - gradle build --stacktrace
 
-after_script:
-  - gradle --version
+.debian-test:
+  stage: test
+  variables:
+    JDK_PKG: openjdk-11-jdk-headless
+  artifacts:
+    name: bitcoinj-$CI_JOB_NAME-$CI_COMMIT_REF_SLUG-$CI_COMMIT_SHORT_SHA
+    paths:
+      - core/build/libs/*.jar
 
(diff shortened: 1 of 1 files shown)


schildbach:
The `.gitlab-ci.yml` file is meant as a simple reference build. For this reason, I just recently removed any indirection like build matrix, variables etc. I dislike the idea of it becoming bloat-ish like our GitHub build.

What options of separation do we have? Can you create another `.gitlab-ci.yml`? Could it live on a separate branch? In a different module, in a different project? Could the entire static analysis perhaps be an external service?

schildbach:
Thanks for the explaination! I don't want to stay in the way of progress. However I also don't like too much of experimentation on the master branch. Ideally, concepts are already tested and agreed on when merged to master.

Perhaps we can separate as follows:

1. `.gitlab-ci-yml` is only contains includes, and those includes all live in a new directory `.gitlab`.
2. Inside `.gitlab`, we've got our existing reference builds, hopefully untouched, and
3. any reports.

What do you think of that plan?

msgilligan:
> What do you think of that plan?

I think it's reasonable. But I'm also learning about some GitLab techniques for managing these things ("Auto DevOps") that are very similar and seem very well designed. I'm experimenting with them on some separate projects so once I understand them better I'll propose them here.

In the meantime, I think #3053 can be merged because it pretty much follows the above rules.

schildbach:
Can be closed?

PR #3051, opened 2023-04-20
gitlab-ci.yml: move before/after script to each job

msgilligan:
This will allow us to add other jobs that don't/can't run these scripts.

This is an alternative to PR #3050.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,13 +1,11 @@
-before_script:
-  - apt-get update
-
-after_script:
-  - gradle --version
-
 bullseye-jdk11:
   image: debian:bullseye-slim
-  script:
+  before_script:
+    - apt-get update
     - apt-get -y install openjdk-11-jdk-headless gradle
+  after_script:
+    - gradle --version
+  script:
     - gradle build --stacktrace
   artifacts:
     name: bitcoinj-$CI_JOB_NAME-$CI_COMMIT_REF_SLUG-$CI_COMMIT_SHORT_SHA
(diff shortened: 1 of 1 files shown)


schildbach:
This change seems reasonable, even on its own.

.gitlab-ci.yml
```diff
@@ -1,11 +1,9 @@
...
-  - gradle --version
-
 bullseye-jdk11:
   image: debian:bullseye-slim
+  before_script:
+    - apt-get update
+  after_script:
+    - gradle --version
   script:
     - apt-get -y install openjdk-11-jdk-headless gradle
```
schildbach, line 8: Since `before_script` isn't shared any more, we should move the `apt-get install` line to `before_script`.

.gitlab-ci.yml
```diff
@@ -16,6 +14,10 @@ bullseye-jdk11:
 
 bookworm-jdk17:
   image: debian:bookworm-slim
+  before_script:
+    - apt-get update
+  after_script:
+    - gradle --version
   script:
     - apt-get -y install openjdk-17-jdk-headless gradle
```
schildbach, line 22: Same here.

.gitlab-ci.yml
schildbach: And after that, we since the `script` will have become the same, we could think about sharing that instead. I would like the restructure on the basis that it would better demonstrate that our product can be built with a simple `gradle build`, no matter what environment.

msgilligan: > Since `before_script` isn't shared any more, we should move the `apt-get install` line to `before_script`.

Fixed and force-pushed.

msgilligan: > And after that, since the `script` will have become the same, we could think about sharing that instead. I would like the restructure on the basis that it would better demonstrate that our product can be built with a simple `gradle build`, no matter what environment.

I assume you mean to do this in a separate PR.

.gitlab-ci.yml
msgilligan: Fixed and force-pushed.

schildbach:
Merged.

Issue #3022, opened 2023-04-13
AddressParser: add provider lookup methods


msgilligan:
The `AddressParser` interface should have at least one provider look up method (that work with JDK 8 providers and in the future will work with JDK 9+ providers)  Some possibilities:

1. `AddressParser.get()`: gets a "default" parser that is configured to parse all known/configured variants
2. `AddressParser.for(Network network)`: gets a validating parser that will only successfully parse addresses for the given network.
3. `NetworkAddressParser.get()` returns a parser than validate for all known networks and for a specified network (see suggestions in Issue #3021)

schildbach:
I think this can be deferred to 0.18.

msgilligan:
> I think this can be deferred to 0.18.

Agreed. Note that dead2c74ab962215aed6f5ecd52f0678fd02ff65 added `getDefault()` and `getDefault(Network)` which could be updated to use `ServiceLocator` to find default implementations.

Issue #3021, opened 2023-04-13
Refine AddressParser interface


msgilligan:
The `AddressParser` interface should probably change from:
```
public interface AddressParser {
    Address parseAddressAnyNetwork(String addressString) throws AddressFormatException;
    Address parseAddress(String addressString, Network network) throws AddressFormatException, AddressFormatException.WrongNetwork;

    @FunctionalInterface
    interface Strict {
        Address parseAddress(String addressString) throws AddressFormatException;
    }
}
```
to something more like:
```
 @FunctionalInterface
public interface AddressParser {
    Address parseAddress(String addressString) throws AddressFormatException;

    interface NetworkAddressParser extends AddressParser {
        Address parseAddress(String addressString) throws AddressFormatException;
        Address parseAddressForNetwork(String addressString, Network network) throws AddressFormatException, AddressFormatException.WrongNetwork;
    }
}
```

This addresses the following issue with the current implementation:
1. There are many use-cases where the `@FunctionalInterface` is useful and in this context it could be "strict" or "non-strict" (i.e. may or may not check the Network.
2. "AnyNetwork" is a poor naming choice. It's "all networks known by the configuration of this parser"
3. A specialized exception should be thrown for the case of the network not matching the provided value (this means the @FunctionalInterface can be used in either way)

schildbach:
I still don't understand the implications of functional interfaces and these specific changes. But if you think it's an improvement why not give it a try?

msgilligan:
PR #3103 Uses a slightly different implementation than suggested above to address issues (1) and (2) above, but does not address issue (3).

msgilligan:
PR #3101 was closed in favor of PR #3129, which has been merged.

schildbach:
Is this issue still open for (3) above, or can it be closed because it has mostly been resolved?

schildbach:
If not closed, I tend to defer this to 0.18 too.

msgilligan:
Let's leave it open for at least another few days and I'll probably defer it to 0.18, but also open a new smaller issue for 0.17 and maybe create a PR to match. I also want to make a PR for validating already parsed addresses against networks as we recently discussed.

schildbach:
My understanding is you created the "smaller PR" as #3384 and it was merged. So can we defer this to 0.18?

schildbach:
I'm removing this as a blocker for 0.17, so that we can go ahead and branch off for beta/rc/release.

PR #3052, opened 2023-04-20
BitcoinURI: restore ability to parse URIs from any network

schildbach:
This includes a test.

### DIFF
--- a/core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
+++ b/core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
@@ -24,7 +24,6 @@ import org.bitcoinj.base.Address;
 import org.bitcoinj.base.DefaultAddressParser;
 import org.bitcoinj.core.NetworkParameters;
 
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
@@ -108,7 +107,7 @@ public class BitcoinURI {
      */
     public static BitcoinURI of(String uri) throws BitcoinURIParseException {
         // TODO: Discover (via Service Loader mechanism) the correct Network from the URI string
-        return new BitcoinURI(uri, BitcoinNetwork.MAINNET);
+        return new BitcoinURI(uri, null);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


schildbach:
Note while this removes the `MAINNET` hardcode, it re-introduces a `BITCOIN_SCHEME` hardcode (similar to what we have in 0.16). I expect this can be resolved once we discover schemes e.g. via service loader.

core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
```diff
@@ -118,7 +117,7 @@ public static BitcoinURI of(String uri) throws BitcoinURIParseException {
      * @param network The network the URI is from
      * @throws BitcoinURIParseException If the input fails Bitcoin URI syntax and semantic checks.
      */
-    public static BitcoinURI of(String uri, @Nonnull Network network) throws BitcoinURIParseException {
+    public static BitcoinURI of(String uri, Network network) throws BitcoinURIParseException {
```
msgilligan, line 120: This should stay `@Nonnull`, I think. Rather than pass `null` the other constructor should be used.

msgilligan: But, maybe you're removing the `@Nonnull` because it is our default?

core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
```diff
@@ -188,12 +188,15 @@ private BitcoinURI(String input, @Nonnull Network network) throws BitcoinURIPars
...
+        parseParameters(nameValuePairTokens);
 
         if (!addressToken.isEmpty()) {
             // Attempt to parse the addressToken as a Bitcoin address for this network
             try {
-                Address address = new DefaultAddressParser().parseAddress(addressToken, network);
+                DefaultAddressParser addressParser = new DefaultAddressParser();
+                Address address = network != null ?
+                        addressParser.parseAddress(addressToken, network) :
+                        addressParser.parseAddressAnyNetwork(addressToken);
```
msgilligan, line 199: This code should be able to be simplified in the future once Issue #3021 and Issue #3022 are implemented.

msgilligan (approved):
Looks good!

core/src/main/java/org/bitcoinj/uri/BitcoinURI.java
schildbach: Yes, but I'll add it back because it's out of scope of this PR.

However I've forgotten a `Objects.requireNotNull()` on that parameter which I'll add before merging.

Issue #2063, opened 2020-10-16
wallet.completeTx() adds the same input twice!


pedr0-fr:
See the code below to replicate the issue

```java
import org.apache.commons.codec.binary.Hex;

public class DoubleInputIssue {

    public static void main(String[] args) throws Exception{
        Path walletPath = Path.of(System.getProperty("user.dir"), "data", "wallet.dat");
        File walletFile = new File(walletPath.toString());
        Wallet wallet = Wallet.loadFromFile(walletFile);

        byte[] model_tx = Hex.decodeHex("0100000001df9b45525d3657c483c2444e3b4e68ad9e7d74d5671bca42498cee49c965249a000000006b483045022100ac2054c509a77a4013449ee2ebd095f451a4ce72cf86687363f76d4820d92dd40220486b752486b8a1631d46b6b54e7c14a3e04cd57ea94b3f6847fe57bcf161854601210295f39e3ab984408746401986c25051cbf993b4535684288e832c6ef890626ac5ffffffff01e8030000000000001976a9143226dda88fda8d9dc88f70636ac5b26d3409aecb88ac00000000");

        SendRequest sr = SendRequest.forTx(new Transaction(TestNet3Params.get(), model_tx));
        sr.setFeePerVkb(Coin.ofSat(10_000));
        wallet.completeTx(sr);
        System.out.println(Hex.encodeHexString(sr.tx.bitcoinSerialize()));
    }
}

```

The code above prints the following transaction:

```json
[...]

PR #3055, opened 2023-04-21
Fixes duplicate inputs in `completeTx`

micheal-swiggs:
Proposed fix for https://github.com/bitcoinj/bitcoinj/issues/2063 . If the input is already in the transaction then it is not added again as an input to the transaction. This fix also checks the wallet for the value of an unconnected input and counts that towards `SendRequest.value`. This changes the current behaviour where the value of unconnected inputs is ignored and simply added to the transaction's fee which seems wasteful.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -139,6 +139,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
 
 import static org.bitcoinj.base.internal.Preconditions.checkArgument;
 import static org.bitcoinj.base.internal.Preconditions.checkState;
@@ -4577,6 +4578,9 @@ public class Wallet extends BaseTaggableObject
             // we don't have the keys for.
             List<TransactionOutput> candidates = calculateAllSpendCandidates(true, req.missingSigsMode == MissingSigsMode.THROW);
 
+            // Remove candidates already used in the SendRequest
+            candidates = removeExistingInputs(candidates, req);
+
             CoinSelection bestCoinSelection;
             TransactionOutput bestChangeOutput = null;
             List<Coin> updatedOutputValues = null;
@@ -4646,6 +4650,24 @@ public class Wallet extends BaseTaggableObject
         } finally {
             lock.unlock();
         }
+    }
+
+    /**
+     * Removes candidates that are already included as inputs
+     * in the SendRequest. This is to avoid using the same
+     * input twice in a Transaction.
+     */
+    private static List<TransactionOutput> removeExistingInputs(List<TransactionOutput> candidates, SendRequest sendRequest) {
+        List<TransactionInput> txInputs = sendRequest.tx.getInputs();
+        return candidates.stream().filter(candidate -> {
+
+            boolean candidateNotExistingInput = !txInputs.stream().anyMatch(txInput -> {
+                TransactionOutPoint outpoint = txInput.getOutpoint();
+                return Objects.equals(candidate.getParentTransactionHash(), outpoint.hash()) &&
+                        candidate.getIndex() == outpoint.index();
+            });
+            return candidateNotExistingInput;
+        }).collect(Collectors.toList());
     }
 
     /**
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -4550,13 +4550,42 @@ public class Wallet extends BaseTaggableObject
             log.info("Completing send tx with {} outputs totalling {} and a fee of {}/vkB", req.tx.getOutputs().size(),
                     value.toFriendlyString(), req.feePerKb.toFriendlyString());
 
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/bitcoinj/wallet/WalletTest.java
```diff
@@ -2737,6 +2739,108 @@ public void higherThanDefaultFee() throws InsufficientMoneyException {
...
+    @Test
+    public void GIVEN_sendrequest_with_input_THEN_dont_reuse_same_input() throws Exception {
+
+        /** ARRANGE */
+        StoredBlock block = new StoredBlock(makeSolvedTestBlock(blockStore, OTHER_ADDRESS), BigInteger.ONE, 1);
+        Transaction tx1 = createFakeTx(TESTNET.network(), COIN, myAddress);
+        wallet.receiveFromBlock(tx1, block, AbstractBlockChain.NewBlockType.BEST_CHAIN, 0);
+        Transaction tx2 = createFakeTx(TESTNET.network(), COIN, myAddress);
+        wallet.receiveFromBlock(tx2, block, AbstractBlockChain.NewBlockType.BEST_CHAIN, 0);
+
```
msgilligan, line 2818: Why do you create two transactions here and test them both below? Can this test be simplified to just use a single transaction?

core/src/test/java/org/bitcoinj/wallet/WalletTest.java
```diff
@@ -2737,6 +2739,108 @@ public void higherThanDefaultFee() throws InsufficientMoneyException {
...
+        /** ACT */
+        wallet.completeTx(request1);
+        wallet.completeTx(request2);
+
+        /** ASSERT */
+        Function<List<TransactionInput>, Set<String>> uniqueInputs = txInputs -> txInputs.stream().map(txInput ->{
+            TransactionOutPoint outpoint = txInput.getOutpoint();
+            String result = outpoint.hash().toString() + outpoint.index();
+            return result;
+        }).collect(Collectors.toSet());
```
msgilligan, line 2836: I think this would be simpler and more readable as a helper method. I also think the type of the `Set` could be `Set<TransactionOutPoint>` and rely on `TransactionOutPoint::equals`. Something like:
```
Set<TransactionOutPoint> uniqueOutPoints(List<TransactionInput>) {
...
}
```

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4648,6 +4675,24 @@ public void completeTx(SendRequest req) throws InsufficientMoneyException, BadWa
...
+     * input twice in a Transaction.
+     */
+    private static List<TransactionOutput> removeExistingInputs(List<TransactionOutput> candidates, SendRequest sendRequest) {
+        List<TransactionInput> txInputs = sendRequest.tx.getInputs();
+        return candidates.stream().filter(candidate -> {
+
+            boolean candidateNotExistingInput = !txInputs.stream().anyMatch(txInput -> {
+                TransactionOutPoint outpoint = txInput.getOutpoint();
+                return Objects.equals(candidate.getParentTransactionHash(), outpoint.hash()) &&
+                        candidate.getIndex() == outpoint.index();
```
msgilligan, line 4690: This could be simplified to:
```
return candidate.getOutPointFor.equals(outpoint);
```

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4648,6 +4675,24 @@ public void completeTx(SendRequest req) throws InsufficientMoneyException, BadWa
...
+            });
+            return candidateNotExistingInput;
```
msgilligan, line 4692: I would recommend moving this filter lambda to a private helper method for readability.

msgilligan (changes requested):
I am really happy to see this work. Thank you!

I would like to ask that you split this PR into multiple PRs for ease of review, testing, etc. The first PR should correspond to the first two commits of this one: the test for duplicate inputs and the fix for duplicate inputs.

I have reviewed the code for those two commits and given some line-by-line comments/suggestions.

I also think that (separately) we should clean up/modernize the code in `Wallet::completeTx` as IMO it could be made simpler and easier to read.

The updating inputValues with data from the wallets utxo list and related tests should then be a separate PR.

Thanks again for this work, I know it was difficult!

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4549,13 +4550,42 @@ public void completeTx(SendRequest req) throws InsufficientMoneyException, BadWa
...
+                        }
+                    }
+                } else {
+                    inputValue = input.getConnectedOutput().getValue();
+                }
+
+                if (inputValue != null){
+                    totalInput = totalInput.add(inputValue);
+                } else {
+                    // TODO: 2023-04-06 is this even a good idea just throwing away coins like this?
```
msgilligan, line 4585: Good point: See WIP PR #3070

core/src/test/java/org/bitcoinj/wallet/WalletTest.java
msgilligan: I have determined that it can be simplified and will submit a new PR with a simplified test.

msgilligan:
PR #3085 is a new PR that is based on this PR. Please review.

msgilligan:
Closing in favor of PR #3085 which has been merged.

PR #3057, opened 2023-04-22
core/build.gradle: conditionally use withSourcesJar/withJavadocJar

msgilligan:
`withSourcesJar()` and `withJavadocJar()` are the cleanest way to do this with recent Gradle versions, but require conditionals because they won't work on Gradle versions earlier than Gradle 6.0.

`classifier` is removed in Gradle 8.x so this PR or an equivalent PR (see PR #3058) that also replaces `classifier` is needed to compile with Gradle 8.x (and Java 20)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -97,19 +97,29 @@ jar {
     }
 }
 
-task javadocJar(type: Jar, dependsOn: javadoc) {
-    classifier = 'javadoc'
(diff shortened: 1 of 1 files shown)


schildbach:
#3058 is much less intrusive, so I'd prefer that one if possible,

msgilligan:
Closing in favor of PR #3058.

PR #2919, opened 2023-03-29
BitcoinSerializer: Use Network not NetworkParameters

msgilligan:
* Use private Network field instead of NetworkParameters
* New constructors that take Network (deprecated old)
* private packetMagic field (requires one-shot use of NetworkParameters to init)
* Deprecate (unused) getParameters() method

After merge TODOs:
1. Maybe figure out how to get packetMagic without using NetworkParameters (add to Network?)
2. equals() and hashCode() can be removed (PR #3062)

PR #3062, opened 2023-05-01
BitcoinSerializer: remove equals() and hashCode()

msgilligan:
These methods are not needed.

(Child of PR #2919)


schildbach:
Merged out of order.

PR #3065, opened 2023-05-01
UnitTestParams: don't `solve()` the genesis block for unit tests

schildbach:
It won't really solve it anyway, because the difficulty check is disabled.


msgilligan (approved):
LGTM

PR #3070, opened 2023-05-04
Wallet.completeTransaction: throw if inputs contain unconnected outputs (WIP)

msgilligan:
Rather than (if not segwit) simply add those unconnected outputs to the miner's fee, throw an IllegalArgumentException.

In the case of segwit, the transaction can't be properly signed without knowing the  input values.

TODO:
1. Consider adding a checked exception for this new error case
2. Rename some of the variables in the unit test


msgilligan:
> I'm fine with throwing an exception. Adding inputs with unknown value to the fee seems way too scary. And as you said, with segwit/taproot we can't continue the behaviour anyway.

I think we might want to throw a checked exception in this case. This would be an API breaking change, but since we are changing behavior it might be a good idea to explicitly let people know about the new behavior.

schildbach:
I would have thrown a new subclass of `CompletionException` but that exception is a `RuntimeException`.

msgilligan:
> might want to throw a checked exception in this case.

The existing exceptions (e.g. `DustySendRequested`, `MultipleOpReturnRequested`) thrown by `completeTx` extend `o.b.w.Wallet.CompletionException` which extends `RuntimeException`, I take back my suggestion that we replace `IllegalArgument` with a _checked_ exception, but we should probably create a new subclass of `o.b.w.Wallet.CompletionException` for this case.

schildbach:
And whatever exception we throw, there are quite some higher-level methods in `Wallet` that will need to throw that exception too. (Unless handled of course.)

msgilligan:
This should be rebased, updated and considered for the v0.18 release.

schildbach:
Feel free to rebase any time.

msgilligan:
> Feel free to rebase any time.

Rebased and the title and main comment above were updated. The second checkbox in the TODO should maybe be added as a second commit. Some variables were not renamed for readability of the diff (e.g. `request3` was moved to another method without a rename and `request4` in the original method should be renamed to `request3`)

PR #3071, opened 2023-05-04
Wallet.commitTx: Always enforce OP_RETURN limit

msgilligan:
Without this fix (and code cleanup) the OP_RETURN limit is not enforced when ensureMinRequiredFee is false or emptyWallet is true.


schildbach:
Looks good, but is there a unit test we could add? Ideally it would demonstrate the issue (fail) before this patch is applied, and would succeed after it is applied.

msgilligan:
> Looks good, but is there a unit test we could add?

I thought about that. Yes, I'll update this PR with a unit test soon.

msgilligan:
> Looks good, but is there a unit test we could add? Ideally it would demonstrate the issue (fail) before this patch is applied, and would succeed after it is applied.

Tests added by parameterizing `twoOpReturnsPerTransactionTest()`. Without the fix, 3 out of the 4 times this test is run will fail.

Force-pushed and ready-for-re-review.

schildbach:
Merged.

PR #3072, opened 2023-05-04
Wallet.commitTx: replace mutated value with effectively final valueNeeded

msgilligan:


core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4536,11 +4536,9 @@ public void completeTx(SendRequest req) throws InsufficientMoneyException, BadWa
         try {
             checkArgument(!req.completed, () ->
                     "given SendRequest has already been completed");
-            // Calculate the amount of value we need to import.
-            Coin value = req.tx.getOutputSum();
```
schildbach, line 4540: I wonder why did you inline this variable? It has 2 uses below.

We could rename it to `outputSum` perhaps?

schildbach (commented):
LGTM, one comment added.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: I decided the simplification of eliminating the `value` variable entirely was worth the minimal overhead of performing an average of 2-3 additions twice. Especially since the first usage is only for the purposes of logging.

If we created an `outputSum` it's only purpose would be a negligible performance gain and when reading line 4551, you'd want to find where `outputSum` is calculated.

The above was my thinking anyway. If you'd like me to change it, I will.

schildbach: No, it's fine. Thanks for the explanation.

schildbach:
Merged.

PR #3074, opened 2023-05-04
Use TransactionOutput.serialize() not bitcoinSerialize()

msgilligan:
Fix the last two usages of this deprecated method.


schildbach:
Merged.

PR #3077, opened 2023-05-06
Wallet: Use List interface where possible

msgilligan:
Some local/private/protected declarations were using LinkedList, ArrayList when they should be using the generic List.


schildbach:
Merged.

PR #3079, opened 2023-05-08
Block: migrate genesis block creation to `createGenesis()` from `NetworkParameters`

schildbach:
This fails unit tests, so it's draft for now.


msgilligan:
It's probably failing unit tests because it's no longer building the special UNITEST block as `network` for the UNITEST `NetworkParameters` is `REGTEST`.

msgilligan:
Concept ack, though.  (Although this could also come after PR #3135 is merged.)

PR #3083, opened 2023-05-09
Wallet.completeTx: use getInputSum(), separately warn about null values

msgilligan:
Note that this changes behavior slightly. We are now warning on inputs with `null` value
and skipping inputs with no value in the sum. Previously we were checking for non-`null`
`getConnectedOutput()`. I reviewed the code and whenever `getConnectedOutput()` is non-`null`,
`getValue()` should also be non-`null`. There might be cases where we have a value and
no connected output, but in those cases I think we should use the value because for signing
it is all we need.

(When we refactor `Transaction` we should have simpler state within the transaction object)

(Update: no longer a child commit, rebased to `master`)


schildbach:
Merged.

PR #3084, opened 2023-05-09
Wallet.completeTx: use forEach to add inputs to req.tx

msgilligan:
(This is in preparation for a forthcoming PR, but is also a reasonable standalone change)


schildbach:
Merged.

PR #3088, opened 2023-05-20
ByteArray: immutable byte array wrapper, use for AesKey

msgilligan:
Will also [tentatively] be used by Bech32 in forthcoming commits.


schildbach:
Merged.

Issue #3092, opened 2023-05-23
WalletTool: replace `CountDownLatch` in `wait()` with a `CompletableFuture<String>`


msgilligan:
Replacing the `CountDownLatch` in `wait()` with a `CompletableFuture<String>` that contains what triggered the wait completion (and having `wait()` return the CF instead of `void`) should simplify the code, decouple the output of the status message from computing them, enable composition and reuse.

msgilligan:
This is implemented in PR #3096.

PR #3093, opened 2023-05-23
WalletTool: move onChange() to a Lambda

msgilligan:
It is a method that is only used in one place and moving this should enable
further simplification (especially replacing the CountDownLatch with a CompletableFuture see PR #3096)

See also Issue #3092

### DIFF
--- a/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
+++ b/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
@@ -947,10 +947,20 @@ public class WalletTool implements Callable<Integer> {
                     latch.countDown();
                     break;
                 }
-                wallet.addCoinsReceivedEventListener((w, t, p, n) -> onChange(latch));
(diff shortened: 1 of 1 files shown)


msgilligan:
@schildbach Rebased fixing conflicts.

schildbach:
I'm a bit reluctant with this because on its own I see no benefit. It introduces the "complexity" of the `Runnable` though.

msgilligan:
> I'm a bit reluctant with this because on its own I see no benefit. It introduces the "complexity" of the `Runnable` though.

The net benefit to this PR is that the `onChange` method is removed and it is made more clear that it's really just common code for the 4 listeners. `Runnable` is pretty standard stuff. Also, by placing the `onChange` code next to the rest of the `BALANCE` case code, it makes it more obvious what is happening and invites further simplification.

But, yeah, I see your point it's not a huge benefit in-and-of-itself.

But take a look at the next step PR #3096.

schildbach:
Merged.

PR #3096, opened 2023-05-23
WalletTool: convert wait() to return CompletableFuture

msgilligan:
This change:
1. Shortens the code by 5 lines while adding many comments
2. Separates output to `System.out` from the async wait handling
3. Separates wallet/peerGroup setup and starting from async wait handling
4. Passes wait condition (along with wait-type enum) as a parameter

If we (separately) decide that the `saveWallet()` call in the `BALANCE`
listener is not needed, the `BALANCE` case can be further simplified.

Behavior should be unchanged with the exception that the balance will
be output in the case that the balance "already meets the given condition".

PR #3093 is a parent.


schildbach:
I extracted d65aefd93e0303ac76db593a2cd7eaf3ff188e06 from ef63187.

msgilligan:
Rebased, tested (see PR #3098 ) and ready for review.

schildbach:
Merged.

PR #3098, opened 2023-05-24
MultiplexingDiscovery: REGTEST fix

msgilligan:
checkArgument should not throw with an empty seeds list *if* running on REGTEST.

I found this bug running `WalletTool` (see the @Disabled test that I added.)

The test is disabled because it requires a REGTEST node that is advanced 1 block (typically via `generateblocktoaddres`) during the course of the test.


schildbach:
Merged.

PR #3099, opened 2023-05-25
Threading: Use BlockingQueue interface internally

msgilligan:
This makes it clear that we're not using specific features of the LinkedBlockingQueue implementation class.


schildbach:
Merged.

PR #3101, opened 2023-05-25
ProtoBuf: move generated files to dedicated package: `org.bitcoinj.protobuf`

msgilligan:
This PR does the following:

* Modify the `java_package` setting in our two `.proto` source files
* Move 2 existing generated-but-checked-in Protos.java files to dedicated packages (i.e. packages without hand-coded "artisanal" files)

This is to be more compatibile with ProtoBuf tooling which sometimes likes to delete the generated files and:

> can't distinguish between "old generated code" and "your artisan source files."

One next step would be to move be to move the `org.bitcoinj.protobuf` package to a separate source set, the other would be to merge this PR and then rebase and merge PR #3059.


schildbach:
Just to make sure, is this change 100% compatible with existing serialized data (e.g. wallets)?

msgilligan:
@schildbach asks:
> Just to make sure, is this change 100% compatible with existing serialized data (e.g. wallets)?

Protobuf is a cross-language serialization protocol and the documentation (See https://protobuf.dev/programming-guides/proto2/) for both `java_outer_classname` and `java_outer_classname` says:

>  If not generating Java code, this option has no effect.

We also have an integration test that reads an existing wallet that is checked in as a resource and it passes after  this change. (Although it does not explicitly test every field)

schildbach:
I added the new packages to `.gitignore` and merged.

PR #3103, opened 2023-05-26
AddressParser: simplify by replacing Strict with Simple

msgilligan:
These changes simplify the interface and change the names to more closely match the use-cases we've been seeing internally and externally with alpha1.

parseAddressAnyNetwork() was both verbose and misleading. AddressParser.Strict was being used (for example in the JavaFX wallet) more as a context-specific simplification of the parent interface than as a "strict" implementation.

parseAddressAnyNetwork() is deprecated as is the Strict sub-interface. They should probably be removed in or before 0.17 final.

This addresses part of Issue #3021, though in a slightly different way than the code proposed in the issue. (The changes proposed in the issue make migration/deprecation harder)

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/AddressParser.java
+++ b/core/src/main/java/org/bitcoinj/base/AddressParser.java
@@ -23,12 +23,12 @@ import org.bitcoinj.base.exceptions.AddressFormatException;
  */
 public interface AddressParser {
     /**
-     * Parse an address that could be for any network
+     * Parse an address for any known/configured network
      * @param addressString string representation of address
      * @return A validated address object
      * @throws AddressFormatException invalid address string
      */
-    Address parseAddressAnyNetwork(String addressString) throws AddressFormatException;
+    Address parseAddress(String addressString) throws AddressFormatException;
 
     /**
      * Parse an address and validate for specified network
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/bitcoinj/base/AddressParser.java
```diff
@@ -39,6 +39,44 @@ public interface AddressParser {
...
     Address parseAddress(String addressString, Network network) throws AddressFormatException;
 
+    /**
+     * Parse an address for any known/configured network
+     * @param addressString string representation of address
+     * @return A validated address object
+     * @throws AddressFormatException invalid address string
+     * @deprecated Use {@link #parseAddress(String)}
+     */
+    @Deprecated
```
schildbach, line 49: Do we need this deprecation? The `AddressParser` class was introduced in the current cycle, right?

core/src/main/java/org/bitcoinj/base/AddressParser.java
```diff
@@ -39,6 +39,44 @@ public interface AddressParser {
...
+
+    /**
+     * Functional interface for address parsing. It takes a single parameter, like {@link AddressParser#parseAddress(String)}
+     * but its behavior is context-specific. This interface may be
+     * implemented by creating a partial application of ({@link AddressParser#parseAddress(String, Network)} providing
+     * a fixed value for {@link Network}. Or it may behave more like {@link #parseAddress(String)} with a context-specific
+     * list of networks.
+     */
+    @FunctionalInterface
+    interface Simple {
```
schildbach, line 62: I don't get the chosen prefix "Simple". What's more simple about "a context-specific way"?

core/src/main/java/org/bitcoinj/base/AddressParser.java
```diff
@@ -39,6 +39,44 @@ public interface AddressParser {
...
+    }
+
+    /**
+     * Functional interface for strict parsing. It takes a single parameter, like {@link AddressParser#parseAddress(String)}
+     * but is used in a context where a specific {@link Network} has been specified. This interface may be
+     * implemented by creating a partial application of ({@link AddressParser#parseAddress(String, Network)} providing
+     * a fixed value for {@link Network}.
+     * @deprecated You probably want to use {@link AddressParser.Simple}, though the semantics of <q>strict</q> no longer apply.
+     */
+    @Deprecated
```
schildbach, line 79: Again, do we need the deprecation?

schildbach (commented):
See comments.

core/src/main/java/org/bitcoinj/base/AddressParser.java
msgilligan: I think so, because we released the `-alpha1` binary. We can remove the deprecation after the next alpha release.

core/src/main/java/org/bitcoinj/base/AddressParser.java
msgilligan: It's "simple" because there's only one method with one parameter. The parent has two methods and a total of three parameters. I'm open to another name, though. Any suggestions?

core/src/main/java/org/bitcoinj/base/AddressParser.java
msgilligan: See above.

msgilligan:
Rebased and force-pushed to resolve merge conflict.

core/src/main/java/org/bitcoinj/base/AddressParser.java
msgilligan: I added a comment on the `@Deprecation` line noting that the deprecation should be removed _before_ 0.17 final.

core/src/main/java/org/bitcoinj/base/AddressParser.java
msgilligan: I added a comment on the `@Deprecation` line noting that the deprecation should be removed _before_ 0.17 final.

core/src/main/java/org/bitcoinj/base/AddressParser.java
msgilligan: @schildbach Does this make sense? Is there a better name?

msgilligan:
Actually, I've been thinking about this some more and now I want to put this on hold. Marking as "DRAFT" for now...

msgilligan:
See PR #3129 which is an improved approach for simplification. So simple that we don't have to name anything "Simple" 🤣

msgilligan:
Closing in favor of PR #3129

PR #3105, opened 2023-05-28
Upgrade to Bouncy Castle 1.73

msgilligan:
I'm not sure if we can start using the `bcprov-jdk18on` variant. (We are currently using the `bcprov-jdk15to18` variant.) So this PR just increments the version number from `1.72` to `1.73`.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -10,7 +10,7 @@ plugins {
 version = '0.17-SNAPSHOT'
 
 dependencies {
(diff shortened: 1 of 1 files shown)


schildbach:
Merged.

Our current Android 8.0 requirement is close to JDK 1.8, but afair there are still some APIs missing. I think it's safer to rely on the JDK 1.5-variant of BC for as long as it exists.

PR #3108, opened 2023-06-07
WalletTool: Set peer address for REG_TEST

msgilligan:
On RegTest we need to add the localhost peer address to the PeerGroup and set max connections to 1, to avoid calling peer discovery.


msgilligan:
This is the merged commit: https://github.com/bitcoinj/bitcoinj/commit/4a45c3b4f4b8341fb5ce83facd704103658527f9

PR #3110, opened 2023-06-13
WalletTool: after `send` action, print the entire transaction

schildbach:
Currently, only the transaction hash is printed.

Alternatively, we could add a switch to select txid, tx or nothing. Feels a bit like overkill though.


msgilligan:
I think we should probably print both the hex transaction and the transaction ID. In most cases, I imagine the first thing the user/developer will do is check a blockchain viewer to see if the transaction has been relayed and they'll need the transaction ID to do that.

msgilligan (changes requested):
I think we should (always) print both txid and transaction hex -- see my comment: https://github.com/bitcoinj/bitcoinj/pull/3110#issuecomment-1591931438

msgilligan (approved):
Looks good!

schildbach:
Merged.

PR #3120, opened 2023-07-05
TransactionOutput: fix regression with calculating getMinNonDustValue() for the multisig case

msgilligan:
This fix was applied on the 0.16 branch by a5fa4004047851180a4602914d7732bd33c10d35, but is not on `master` and is not in `bitcoinj-0.17-alpha1`.


schildbach:
Merged.

PR #3124, opened 2023-07-15
DnsDiscovery, MultiplexingDiscovery: use Network in constructors

msgilligan:


schildbach:
I wonder if this is the way to go. According to the comment, Network(Params) is used only for the (default) port number. Maybe we can just pass the default port number instead and thus get rid of Network(Params) entirely?

Afaicr, we're not planning to make port numbers a property of Network either?

msgilligan:
So there are (at least) three options for eliminating NetworkParams:

1. Use `Network` instead and internally use `NetworkParameters.of()` to get the port number (this PR)
2. Use `Network` instead and add the port number to NetworkParameters
3. Use port number and push off finding the port number onto the client.

Personally, I think 3 complicates the AI for end users and would find either 1 or 2 acceptable.

I also have been thinking about creating some kind of factory/builder for making a set of client objects so that typical clients don't have to worry about constructing chain, chainstore, peer group, etc. But that's out-of-scope for this PR.

If we go with 2, we could merge this PR as is and then make a separate PR to add port to `Network` to simplify the internals of this class.

What approach do you prefer?

msgilligan:
Looking at this more closely and refreshing my memory, `NetworkParameters` is also used for `.getDnsSeeds()`. 

The purpose of this PR is to remove some of the remaining places in the API where applications need to use `NetworkParameters` when setting up wallets and/or peer groups and I think it is a step in that direction.

schildbach:
Merged.

PR #3126, opened 2023-07-15
GitHub graalvm.yml workflow: use latest GraalVM

msgilligan:
As of June 13, 2023, GraalVM versioning has changed and now matches the versioning system of the JDK.

The latest release of GraalVM for JDK 17 is now 17.0.7.


schildbach:
Merged.

PR #3129, opened 2023-07-17
Simpler AddressParsing by making AddressParser a @FunctionalInterface

msgilligan:
The main idea here is that we made `AddressParser` the `@FunctionalInterface` and a nested class is used to define an `AddressParserProvider` (Instead of the other way around.)

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/AddressParser.java
+++ b/core/src/main/java/org/bitcoinj/base/AddressParser.java
@@ -17,42 +17,93 @@
 package org.bitcoinj.base;
 
 import org.bitcoinj.base.exceptions.AddressFormatException;
+import org.bitcoinj.core.NetworkParameters;
+
+import javax.annotation.Nullable;
 
 /**
- * Interface for parsing and validating address strings.
+ * Functional interface for address parsing. It takes a single parameter, but its behavior is context-specific. For example if the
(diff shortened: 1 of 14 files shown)


schildbach:
Yes, it seems to me this is the better approach.

msgilligan:
@schildbach I renamed `DefaultAddressParser` to `DefaultAddressParserProvider` and made it package-private. I also add some missing JavaDoc and force-pushed. It's now ready for final review and merging.

schildbach:
Merged.

core/src/main/java/org/bitcoinj/base/AddressParser.java
```diff
@@ -17,42 +17,93 @@
...
+     * @return A parser that will throw for strings that are not valid for network.
+     */
+    @Deprecated
+    static AddressParser getLegacy(@Nullable NetworkParameters params) {
+        AddressParser.AddressParserProvider provider = DefaultAddressParserProvider.fromNetworks();
+        return (params == null)
+                ? provider.forKnownNetworks()
+                : provider.forNetwork(params.network());
     }
+
```
Rjsoon38, line 108: Would you place a different type of command here if you needed to? Such as if you needed to make changes or additions to a different type of command?

PR #3130, opened 2023-07-20
Enable Gradle Build Scans, separate settings-gradle4.gradle file

msgilligan:
Build changes:

* Default Gradle Build requires Gradle 7 or later
* Legacy Gradle build uses `-c settings-gradle4.gradle`
* Default Gradle build includes Build Scan capability
* Always Publish Build Scans if `CI` environment variable is set.

CI changes:

* Remove special case Gradle 4.10.3 build from GitHub Actions
* Add `-c settings-gradle4.gradle` option to GitLab/Debian builds

Build Scans will help us to troubleshoot our intermittent CI failures.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -12,12 +12,6 @@ jobs:
         java: ['11', '17', '19']
         distribution: ['temurin']
         gradle: ['7.6.2']
-        include:
-          # Special case to test something close to Debian config with the oldest supported (standard) Gradle on Ubuntu
-          - os: ubuntu-latest
-            java: '11'
-            distribution: 'temurin'
-            gradle: '4.10.3'
       fail-fast: false
     name: JAVA ${{ matrix.distribution }} ${{ matrix.java }} OS ${{ matrix.os }} Gradle ${{ matrix.gradle }}
     steps:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,7 +4,7 @@ bullseye-jdk11:
     - apt-get update
     - apt-get -y install openjdk-11-jdk-headless gradle
   script:
-    - gradle build --stacktrace
+    - gradle -c settings-gradle4.gradle build --stacktrace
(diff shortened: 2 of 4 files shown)


bitcoinfinancier (approved):

schildbach:
Taking a first glance at this, I think this PR should be separated into 2 PRs or commits:

- separation of settings into two files, replacing (some) conditionals
- introduction of build scans

My gut feeling about the settings separation is that the default one should be for Gradle 4, and then we've got a settings-gradle7.gradle for modern Gradle. The reference build still uses Debian/Ubuntu Gradle 4.4.1.

My gut feeling about build scans is that it's mostly proprietary…

msgilligan:
> I think this PR should be separated into 2 PRs

Agreed. See PR #3213 

> the default one should be for Gradle 4, and then we've got a settings-gradle7.gradle for modern Gradle. The reference build still uses Debian/Ubuntu Gradle 4.4.1.

Since the reference build is automated on GitLab and is done less frequently (I believe) I think it is more friendly to other contributors (at least me 😁) if the default works with recent Gradle. So PR #3213 keeps recent Gradle as the default, but if you insist, I will change it.

> My gut feeling about build scans is that it's mostly proprietary…

True, but for us:

* They are "free beer"
* Are very helpful in quickly reviewing build failures on CI
* Are easily removed/disabled.

msgilligan:
Today I discovered that it it now possible to "inject" Gradle Enterprise support into our build without changes the `settings.gradle` file:

https://github.com/gradle/gradle-build-action/blob/v2.8.0/README.md#gradle-enterprise-plugin-injection

So maybe this is a better approach for us.

schildbach:
Interesting. Yes, I'd prefer the approach of configuring this on the GitHub actions.

msgilligan:
> Interesting. Yes, I'd prefer the approach of configuring this on the GitHub actions.

Closing this in favor of a future PR implementing the approach suggested in issue #3238.

PR #3135, opened 2023-07-30
Block: new constructors to reduce mutability

msgilligan:
* New constructors and static factory methods
* Make `version` final
* Remove `public` from some @VisibleForTesting setters
* Add @Nullable to `transactions` parameter in public constructor
* Update NetworkParameters classes to use new static factory methods
* Update 2 tests to use new constructors

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -51,6 +51,7 @@ import java.util.Date;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
 
 import static org.bitcoinj.base.Coin.FIFTY_COINS;
 import static org.bitcoinj.base.Sha256Hash.hashTwice;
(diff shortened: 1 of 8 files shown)


msgilligan:
@schildbach Marked as DRAFT because of question of implementation type `ArrayList` or `LinkedList` for `Transaction` list. I think `ArrayList` is probably better.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -178,13 +179,31 @@ private static List<Transaction> readTransactions(ByteBuffer payload) throws Buf
...
-        time = TimeUtils.currentTime().truncatedTo(ChronoUnit.SECONDS); // convert to Bitcoin time
-        prevBlockHash = Sha256Hash.ZERO_HASH;
+        this(setVersion,
+                TimeUtils.currentTime().truncatedTo(ChronoUnit.SECONDS), // convert to Bitcoin time)
+                0x1d07fff8L,
+                0,
+                Collections.emptyList());
+    }
+
+    // For unit-test genesis blocks
```
schildbach, line 193: Should this be `@VisibleForTesting` then?

msgilligan:
@schildbach rebased and force-pushed.

core/src/main/java/org/bitcoinj/core/Block.java
msgilligan: Yes. I'll update it.

schildbach:
I picked some more changes from this PR.

btw. also see #3079

schildbach:
Merged.

PR #3136, opened 2023-07-31
Block: simplify initialization of `genesisTxScriptPubKeyBytes`

msgilligan:
* Use ScriptBuilder rather than static initializer and ByteArrayOutputStream


schildbach:
Great simplification!

Merged.

PR #3137, opened 2023-08-02
ForwardingService: Print exception in listener

msgilligan:
The `coinForwardingListener` was swallowing some exceptions. This change will print those exceptions to the console.


schildbach:
Merged.

PR #3138, opened 2023-08-02
ForwardingService: factor "forward" into its own method

msgilligan:
Move the code that creates the forwarded transaction, signs, sends,
and waits for peer confirmation into a standalone method.

The `forward` function is a relatively straightforward asynchronous function
that returns a `CompletableFuture` that either completes normally or exceptionally.

This is IMO easier to read and theoretically "more reusable".

This is a child of PR #3137


schildbach:
Merged.

PR #3141, opened 2023-08-06
TransactionInput: simplify constructors

msgilligan:
* Don't use setParent() in constructors
* Don't call super() now that Object is the superclass
* Use private constructor via this() in constructor for UNSIGNED input

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionInput.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionInput.java
@@ -131,25 +131,24 @@ public class TransactionInput {
     private TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBytes,
                             TransactionOutPoint outpoint, long sequence, @Nullable Coin value) {
         checkArgument(value == null || value.signum() >= 0, () -> "value out of range: " + value);
+        parent = parentTransaction;
         this.scriptBytes = scriptBytes;
         this.outpoint = outpoint;
         this.sequence = sequence;
         this.value = value;
-        setParent(parentTransaction);
     }
 
     /**
      * Creates an UNSIGNED input that links to the given output
      */
     TransactionInput(Transaction parentTransaction, TransactionOutput output) {
-        super();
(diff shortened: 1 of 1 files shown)


schildbach:
I extracted the if-else to ternary change into its own commit.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -131,28 +131,24 @@ public TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBy
     private TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBytes,
                             TransactionOutPoint outpoint, long sequence, @Nullable Coin value) {
         checkArgument(value == null || value.signum() >= 0, () -> "value out of range: " + value);
+        parent = parentTransaction;
         this.scriptBytes = scriptBytes;
         this.outpoint = outpoint;
         this.sequence = sequence;
         this.value = value;
-        setParent(parentTransaction);
```
schildbach, line 138: Not sure about this one. As long as `setParent` exists, we should probably use it. We could however try to get rid of `setParent` entirely, and ideally `parent` as well at some point.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -131,28 +131,24 @@ public TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBy
...
-        if(output.getParentTransaction() != null ) {
-            outpoint = new TransactionOutPoint(outputIndex, output.getParentTransaction());
-        } else {
-            outpoint = new TransactionOutPoint(output);
-        }
-        scriptBytes = EMPTY_ARRAY;
-        sequence = NO_SEQUENCE;
-        setParent(parentTransaction);
-        this.value = output.getValue();
+        this(parentTransaction,
```
schildbach, line 145: Ack using `this()`.

schildbach:
Merged in multiple commits.

PR #3142, opened 2023-08-06
Wallet: refactor estimateVirtualBytesForSigning

msgilligan:
* Use functional-style 
* Use `List<TransactionOutput>` rather than `CoinSelector`
* Factor out method that calculates bytes for a single scriptPubKey.

This is two separate commits (and was originally a child of PR #3140)


schildbach:
I believe everything is merged now.

PR #3145, opened 2023-08-08
Peer: extract buildBlockLocator() to private method

msgilligan:
This has two sequential commits that can be merged together.


schildbach:
I believe everything is merged.

PR #3152, opened 2023-08-08
Block: use ArrayList consistently for transactions field

msgilligan:
This is a child of and follow-up to PR #3135


schildbach:
Merged out-of-order.

schildbach:
I only wonder if the field itself should be of type `ArrayList` too. It's `@VisibleForTesting` though.

msgilligan:
> I only wonder if the field itself should be of type `ArrayList` too. It's `@VisibleForTesting` though.

Well it should be `final` and hopefully we'll make it `final` soon.

PR #3156, opened 2023-08-08
TransactionOutput: simplify isDust

msgilligan:
Replace `if` with boolean logic and revise related comment.


schildbach:
Merged.

PR #3157, opened 2023-08-08
remove use of Guava @VisibleForTesting on all package-private members

msgilligan:
Both methods it is used on are already package-private and one is deprecated.


schildbach:
If we replace `@VisibleForTesting` with just a comment, imho we should do that for the entire codebase at once.

msgilligan:
@schildbach I implemented (2) above ("Replace all usages of  @VisibleForTesting on package-private items with a comment"), rebased, force-pushed, and updated the title.

schildbach:
Yes, that's a good start. Merged.

PR #3158, opened 2023-08-09
Wallet/InsufficientMoneyException: add more information

msgilligan:
Add a new constructor that provides more information about available funds, total outputs, and fee in addition to amount missing. Use it in `Wallet.calculatFee()` to provide more information when the exception occurs.

When assisting a beginning bitcoinj user, I saw that this could be helpful.


schildbach:
Merged.

PR #3160, opened 2023-08-09
Wallet.CouldNotAdjustDownwards: add more diagnostic info

msgilligan:
Add output.value and calculated nonDustValue to the exceptions message string. Provide those values to the constructor in `calculateFee`

Add a constructor to parent class that takes a message String.


schildbach:
Merged.

PR #3086, opened 2023-05-15
MainNetParams: make `getAddrSeeds()` to return `List<InetSocketAddress>`

schildbach:
The `addrSeeds` array is now generated using the tools from Bitcoin Core `contrib/seeds` (see the README there).

I'm a bit unhappy about the code duplication atm. Sadly it turned out the `/generate-seeds.py` script generates *yet another* BIP155 variant. Maybe I should do another pass.

PR #3161, opened 2023-08-09
TorUtils: new utility for Tor/Onion addresses

schildbach:
Move Tor/Onion related code there.

This should help with code duplication in #3086.


msgilligan (approved):
This is an improvement by moving the Tor encoding/decoding into methods.

I've been thinking that we might want to make subclasses/subinterfaces of PeerAddress for each of the types, but this change doesn't preclude that.

I suppose we could also write test for the new TorUtils class.

schildbach:
Good idea about the tests, I'll add roundtrip tests tomorrow.

schildbach:
I added tests and merged this.

PR #3163, opened 2023-08-09
Wallet: Declare throws CompletionException where appropriate

msgilligan:
Adjust the signature of these methods to match the actual behavior and the JavaDoc. (Note: BadWalletEncryptionKeyException is  subclass of CompletionException which is a RuntimeException)

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/Wallet.java
+++ b/core/src/main/java/org/bitcoinj/wallet/Wallet.java
@@ -4320,7 +4320,7 @@ public class Wallet extends BaseTaggableObject
      * @throws BadWalletEncryptionKeyException if the supplied {@link SendRequest#aesKey} is wrong.
      */
     public Transaction sendCoinsOffline(SendRequest request)
-            throws InsufficientMoneyException, BadWalletEncryptionKeyException {
+            throws InsufficientMoneyException, CompletionException {
         lock.lock();
         try {
             completeTx(request);
@@ -4359,7 +4359,7 @@ public class Wallet extends BaseTaggableObject
      * @throws BadWalletEncryptionKeyException if the supplied {@link SendRequest#aesKey} is wrong.
      */
     public SendResult sendCoins(TransactionBroadcaster broadcaster, Address to, Coin value)
-            throws InsufficientMoneyException, BadWalletEncryptionKeyException {
+            throws InsufficientMoneyException, CompletionException {
         SendRequest request = SendRequest.to(to, value);
         return sendCoins(broadcaster, request);
     }
@@ -4387,7 +4387,7 @@ public class Wallet extends BaseTaggableObject
      * @throws BadWalletEncryptionKeyException if the supplied {@link SendRequest#aesKey} is wrong.
      */
     public SendResult sendCoins(TransactionBroadcaster broadcaster, SendRequest request)
-            throws InsufficientMoneyException, BadWalletEncryptionKeyException {
+            throws InsufficientMoneyException, CompletionException {
         // Should not be locked here, as we're going to call into the broadcaster and that might want to hold its
         // own lock. sendCoinsOffline handles everything that needs to be locked.
         checkState(!lock.isHeldByCurrentThread());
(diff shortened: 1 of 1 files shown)


msgilligan:
See related Issue #3167

schildbach:
I'm a bit puzzled, the JavaDocs explicitly state `BadWalletEncryptionKeyException` and not `CompletionException`.

msgilligan:
`createSend` for example looks like this:

```
     * @throws InsufficientMoneyException if the request could not be completed due to not enough balance.
     * @throws DustySendRequested if the resultant transaction would violate the dust rules.
     * @throws CouldNotAdjustDownwards if emptying the wallet was requested and the output can't be shrunk for fees without violating a protocol rule.
     * @throws ExceededMaxTransactionSize if the resultant transaction is too big for Bitcoin to process.
     * @throws MultipleOpReturnRequested if there is more than one OP_RETURN output for the resultant transaction.
     * @throws BadWalletEncryptionKeyException if the supplied {@link SendRequest#aesKey} is wrong.
````

I think you need to show more context.

schildbach:
Yes, it has `@throws BadWalletEncryptionKeyException` and not `@throws `CompletionException`. But this PR description states

> to match the actual behavior **and the JavaDoc**

msgilligan:
> Yes, it has `@throws BadWalletEncryptionKeyException` and not `@throws `CompletionException`. But this PR description states
> 
> > to match the actual behavior **and the JavaDoc**

Well, all those declared exceptions are sub-classes of `CompletionException` so by "matching" I didn't mean throwing every exception listed in the JavaDoc but throwing the superclass that contains them all. I guess that's a little confusing, but I think the throws should be the superclass and the JavaDoc should continue to explain what each subclass means.

schildbach:
> What do you think we should do?

What actual problem is this PR addressing?

msgilligan:
> > What do you think we should do?
> 
> What actual problem is this PR addressing?

That the JavaDoc lists (unchecked) exceptions that are not declared. I actually think that some of these exceptions (by traditional Java conventions, e.g. Effective Java) should be checked exceptions because they are exceptions that the wallet software would need to handle, e.g. `CouldNotAdjustDownwards`.

We don't need to do anything. But in helping a bitcoinj beginner last weekend (he was trying to modify the `ForwardingService` example to split the incoming transaction in two and forward to two addresses) I realized that dealing with these exceptions is tricky.  So I was wanting to just document things more clearly. 

Certainly you'll agree there is an inconsistency that `sendCoins` uses a `throws` keyword to declare `BadWalletEncryptionKeyException` which is also listed in the JavaDoc, but  `DustySendRequested`, `CouldNotAdjustDownwards` and others are mentioned in the JavaDoc but not listed after the `throws` keyword?

msgilligan:
Related Stack Overflow answers:

* https://stackoverflow.com/a/8082576
* https://stackoverflow.com/a/11116886

The [Documenting Exceptions with @throws Tag](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#throwstag) section of an Oracle's online JavaDoc article is interesting:

> Document the following exceptions with the @throws tag:

> All checked exceptions. 
> (These must be declared in the throws clause.)
> Those unchecked exceptions that the caller might reasonably want to catch. 
> (It is considered poor programming practice to include unchecked exceptions in the throws clause.) 
> Documenting these in the @throws tag is up to the judgment of the API designer, as described below.

> Documenting Unchecked Exceptions

> It is generally desirable to document the unchecked exceptions that a method can throw: this allows (but does not require) the caller to handle these exceptions. For example, it allows the caller to "translate" an implementation-dependent unchecked exception to some other exception that is more appropriate to the caller's exported abstraction.

So by this recommendation we should remove the `BadWalletEncryptionKeyException` in the throws clause...

msgilligan:
Alternatively, we deprecate these methods and add knew ones with a different error-handling mechanism (perhaps using checked exceptions and/or `CompletableFuture`)

schildbach:
> Certainly you'll agree there is an inconsistency that `sendCoins` uses a `throws` keyword to declare `BadWalletEncryptionKeyException` which is also listed in the JavaDoc, but `DustySendRequested`, `CouldNotAdjustDownwards` and others are mentioned in the JavaDoc but not listed after the `throws` keyword?

Yes. One naive fix would be to list them in the `@throws` clause.

schildbach:
> (It is considered poor programming practice to include unchecked exceptions in the throws clause.)

I'd say this is more of a question in what domain those exceptions are. The ones we're talking about are clearly in the problem domain of the `sendCoins()` methods, so I think it's fine to declare them. Very generic ones like `IllegalStateException` or `OutOfMemoryError` would probably be not ok.

schildbach:
Given the negliable risk of API breakage, I went ahead and merged this. And I applied the change to two more `send*` methods I believe you missed.

Maybe we should try to simplify the edge cases with sending. I believe some cases have been removed over the years by changes in fee/mempool policy.

Theschorpioen (approved):

PR #3164, opened 2023-08-09
Use Transaction.messageSize() instead of serialize().length

msgilligan:
(Except in asserts in tests)


schildbach:
Oh, I just coded this up myself.

PR #3171, opened 2023-08-12
BlockImporter: remove unsupported "Disk" option from usage

msgilligan:
"Disk" is listed in USAGE but will result in an "Unknown store" message. Remove it from USAGE.


schildbach:
Merged.

PR #3174, opened 2023-08-12
BlockFileLoader: effectively final sizeBytes, dataBytes

msgilligan:
When reading blocks from the file stream use two effectively final `byte[]` locals named `sizeBytes` and `dataBytes` rather than reuse a single `byte[]` local. This also makes the code more readable by using more meaningful variable names.


msgilligan:
I converted this to DRAFT so it can be rebased and merged after #3180 .

schildbach:
I believe this was merged.

PR #3175, opened 2023-08-12
Remove deprecated AbstractTimeoutHandler

msgilligan:
This was first deprecated in 0.17-alpha1 (i.e. it has is not deprecated in the 0.16.x releases) so we might want to wait until after 0.17 final is released before merging this PR.

However, this is fairly internal to the implementation so we might be able to remove it in the next 0.17 alpha release.


msgilligan:
I've asked for a review to consider this for the 0.17 release, but if we want to play it safe we can postpone to the 0.18 release.

schildbach:
Is there reason for hurry to remove this? To me, it looks we can remove this early in the 0.18 cycle as well.

msgilligan:
> Is there reason for hurry to remove this? To me, it looks we can remove this early in the 0.18 cycle as well.

No reason other than the cleanup will help with code readability & additional refactoring. But I was hoping that since it's practically an internal class we might be able to make an exception to our general process and remove it now.

schildbach:
I think unless we find a more pressing reason to remove it effectively without a deprecation cycle, we should keep it for until after release.

I'll mark this as 0.18 for now.

msgilligan:
@schildbach We can merge this now.

schildbach:
Rebased.

PR #3177, opened 2023-08-13
PeerGroup, FilterMerger: Deprecate setting FP rate (after construction)

msgilligan:
In the bitcoinj code itself, false positive rate is never changed after constructing a PeerGroup or FilterMerger.

* Deprecate methods for setting Bloom Filter FP rate in both methods
* Add constructor params to PeerGroup so they can be set at construction time, if non-default values are needed.


schildbach:
Merged.

PR #3178, opened 2023-08-13
BloomFilter: use streams in more places internally

msgilligan:


schildbach:
Is this draft for a reason?

PR #3179, opened 2023-08-13
FilterMerger: WIP less mutable, more functional

msgilligan:


schildbach:
Why is this WIP?

msgilligan:
> Why is this WIP?

I think it is WIP because I wanted PR #3177 merged first and then I was going to make some additional changes in this PR.

PR #3180, opened 2023-08-13
BlockFileLoader: correct Iterable implementation

msgilligan:
To be Iterable, iterator() must be able to be called more than once. This PR moves the `Iterable` implementation to an inner class that is created when `iterator()` is called.

This is a breaking change, user will no longer be able to use BlockFileLoader directly as an Iterator and must call .iterator() first. See the change in BitcoindComparisonTool.


schildbach:
Thanks for the tests!

Merged.

PR #3182, opened 2023-08-13
BlockFileLoader: refactor to use single-file Block iterator internally 

msgilligan:
* Replace iterator that iterates all blocks in a file list with iterator that iterates all blocks in a single file
* Rewrite stream() to use flatMap() on a stream of blocks in a file (internally  using BlockFileIterator)
* Rewrite iterator() to use stream()

This is a step towards a standalone capability to iterate/stream the blocks in a single .dat file. It also is a step towards higher-performance implementations that can use multiple threads.

This is a child of PR #3181.


schildbach:
I believe I merged all of this.

PR #3185, opened 2023-08-14
Add (Ignored) integration test that reads bitcoind blocks

msgilligan:
This is ignored because:
a) It requires a bitcoind installation with a blockchain dir
b) It currently fails while reading blk00000.dat


msgilligan:
@schildbach This fails when run against `master` or against the tip of my recent 3 PRs changing `BlockFileLoader`

msgilligan:
Yes, there's a a fix in PR #3186 and the test is duplicated there too. I'm going to mark this a draft and once #3186 is merged, I think we can close this.

schildbach:
Merged.

PR #3187, opened 2023-08-14
DnsDiscovery: remove Linux gethostbyname bug workaround from 2015

msgilligan:
This is the commit that introduced the work-around: fe2aff49ae1ab23af9bed01aa131062f47a2e4e7

Note that we could probably remove the override of `createExecutor()` in `DnsDiscovery` but it currently uses different naming and a different `ThreadyFactory`.  I'm not sure at what point this diverged but can probably be simplified.


schildbach:
Merged.

PR #3189, opened 2023-08-16
WIP: Comment out use of TxConfidenceTable in Block.readTransactions()

msgilligan:
This change speeds up an integration test that I am writing by a factor of 15x (from 30 minutes to 2 minutes)  When I am deserializing `Block` s on different threads it seems the contention for the `TxConfidenceTable` lock is preventing effective concurrency.

Removing this line doesn't affect the unit tests, but I don't know what the best solution is for removing this. I don't think that `Block` or `Transaction` should have dependencies on the `TxConfidenceTable`, though.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -173,7 +173,9 @@ public class Block extends BaseMessage {
         for (int i = 0; i < numTransactions; i++) {
             Transaction tx = Transaction.read(payload);
             // Label the transaction as coming from the P2P network, so code that cares where we first saw it knows.
-            tx.getConfidence().setSource(TransactionConfidence.Source.NETWORK);
+            // TODO: Either remove this or add equivalent functionality in a way that doesn't
+            // require a dependency on TxConfidenceTable or slow multi-threaded apps
+            // tx.getConfidence().setSource(TransactionConfidence.Source.NETWORK);
             transactions.add(tx);
         }
         return transactions;


msgilligan:
It also might make sense to try to re-implement `TxConfidenceTable` using `ConcurrentHashMap` (if possible.)

schildbach:
I can say that Bitcoin Wallet makes use of the `source` field, and I believe also `DefaultRiskAnalyer` does.

msgilligan:
`Block.read()` is also called in `StoredBlock.deserializeCompact()` and I don't think setting `TransactionConfidence.Source.NETWORK` is correct in that case.

I think we need to remove setting source in `Block.read`, but make sure the value is set correctly in other places for the various use cases.

msgilligan:
I would definitely like to see a clean solution for this issue in the 0.17 release.

schildbach:
> I think we need to remove setting source in `Block.read`, but make sure the value is set correctly in other places for the various use cases.

Sounds reasonable, and I think this might go into a commit preceding this one.

schildbach:
I just went through the call stack.

Ignoring all the tests (which according to you don't care at the moment), setting the source should be applied to:

* `BlockFileLoader.stream()`
* `Peer.processBlock()`
* `Peer.processTransaction()`

I think that should cover all the cases that were covered by the line removed in this PR.

Would you like to adapt this PR or should I prepare an alternative one?

msgilligan:
> setting the source should be applied to:
> 
> * `BlockFileLoader.stream()`
> * `Peer.processBlock()`
> * `Peer.processTransaction()`
> 
> Would you like to adapt this PR or should I prepare an alternative one?

I'll adapt it!

schildbach:
However, we should probably not set the source if it's already set before. In particular, if we know that we've created a txn ourselves, we should never loose that information.

schildbach:
See #3292 for an added precondition to `TransactionConfidence.setSource()`.

msgilligan:
See PR #3301 for an attempt to implement this properly on the current codebase.

msgilligan:
Closing in favor of PR #3301 which has been merged.

PR #3194, opened 2023-08-21
Wallet: add getConfidence(tx) method and use it in waitForConfirmations

msgilligan:


core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4497,7 +4497,12 @@ public CompletableFuture<TransactionConfidence> waitForConfirmation(Transaction
      * @return a future for an object that contains transaction confidence information
      */
     public CompletableFuture<TransactionConfidence> waitForConfirmations(Transaction tx, int requiredConfirmations) {
-        return tx.getConfidence().getDepthFuture(requiredConfirmations);
+
+        return getConfidence(tx).getDepthFuture(requiredConfirmations);
+    }
+
+    private TransactionConfidence getConfidence(Transaction tx) {
+        return tx.getConfidence(Context.get().getConfidenceTable());
```
schildbach, line 4505: Why is the `Context.get()` added here?

schildbach (commented):
This PR does a but moire than just extract the method, and I'd like to understand the big picture first.

msgilligan:
> This PR does a but moire than just extract the method, and I'd like to understand the big picture first.

The big picture is I'm trying to do the following:

1. Reduce direct usage of the `Transaction` to find the `TransactionConfidence` with an eye towards removing and deprecating the methods in `Transaction` to do this.
2. Have methods in the `Wallet` use `wallet.getConfidence(tx)` to get transaction confidence
3. Have classes outside the `Wallet` use `waitForConfirmations` where possible.
4. Work towards the goal of removing `TxConfidenceTable` from `Context`

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: The other `getConfidence()` methods in `Transaction` eventually call the one that takes a `TxConfidenceTable`, so I wanted to call that one directly. And also prepare for not using `Transaction` to get the `TransactionConfidence`.

schildbach: I think this rationale could go into the commit comment.

msgilligan: Done and force-pushed.

PR #3197, opened 2023-08-22
Block.readTransactions: make more functional with IntStream.range

msgilligan:
We could probably create an unmodifiable list here, but for now let's use ArrayList for backward compatibility.

This is a child of PR #3189 and needs a decision about how to remove the use of `TransactionConfidence` in the `readTransactions()` method before it can be merged.


msgilligan:
Since PR #3301 replaces PR #3189 and we now know it is compatible to return an unmodifiable list, I'm closing this PR in favor of PR #3321.

PR #3201, opened 2023-08-22
TransactionConfidence: store overriding txId (not Transaction ref) (minor BREAKING)

msgilligan:
This is a breaking change as `getOverridingTransaction()` has been removed with a replacement of `getOverridingTxId`. All use cases in this repository only need the ID and there are probably very few (if any) external apps that are accessing `TransactionConfidence` directly _and_ using the `getOverridingTransaction()`

Note that ProtoBuf serialization only stores the TxID, so this change does not affect serialization/deserialization.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
@@ -145,7 +145,7 @@ public class TransactionConfidence {
     private ConfidenceType confidenceType = ConfidenceType.UNKNOWN;
     private int appearedAtChainHeight = -1;
     // The transaction that double spent this one, if any.
-    private Transaction overridingTransaction;
+    private Sha256Hash overridingTxId;
 
     /**
      * Information about where the transaction was first seen (network, sent direct from peer, created by ourselves).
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
```diff
@@ -443,28 +443,43 @@ public void clearBroadcastBy() {
...
      * Called when the transaction becomes newly dead, that is, we learn that one of its inputs has already been spent
      * in such a way that the double-spending transaction takes precedence over this one. It will not become valid now
      * unless there is a re-org. Automatically sets the confidence type to DEAD. The overriding transaction may not
      * directly double spend this one, but could also have double spent a dependency of this tx.
+     * @deprecated Use {@link #getOverridingTxId()}
      */
+    @Deprecated
     public synchronized void setOverridingTransaction(@Nullable Transaction overridingTransaction) {
-        this.overridingTransaction = overridingTransaction;
+        Objects.requireNonNull(overridingTransaction.getTxId());
```
schildbach, line 470: Shouldn't the `requireNonNull()` be applied on the method argument, rather than the txId?

schildbach (commented):
Looks good, but one minor comment.

core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
msgilligan: Oops. Will fix...

msgilligan:
@schildbach rebased and ready-to-review/merge.

schildbach:
Merged.

PR #3202, opened 2023-08-23
ListMessage & subclasses: make immutable (after deprecations removed)

msgilligan:
Make `ListMessage` and its subclasses "almost" immutable. When the deprecated  addItem, removeItem, etc. methods are removed and the constructors are changed to create an ummodifiableList, they will be immutable.


schildbach:
Merged.

PR #3208, opened 2023-08-25
Wallet: Use getConfidence(tx) not tx.getConfidence()

msgilligan:
This is a child of PR #3194

PR #3209, opened 2023-08-26
TxConfidenceTable: add getConfidence(), reduce visibility of 2 methods in Transaction

msgilligan:
* TxConfidenceTable: add getConfidence(Transaction)
* Wallet: Use TxConfidenceTable.getConfidence(Transaction)
* Transaction: Reduce visibility of Transaction.getConfidence(TxConfidenceTable) and Transaction.getConfidence(Context) to package-private.

This is a minor breaking change to anything relying on the methods in Transaction that now have package visibility, but that should be very rare.

This is a child of PR #3208.


schildbach:
Merged (before the rebase)

PR #3211, opened 2023-09-01
Peer: Improve use of generics for GetDataRequest

msgilligan:
This gets rid of several warnings including two that had been suppressed.


schildbach:
Merged.

PR #3212, opened 2023-09-01
Peer: simplify getAddr()

msgilligan:
Simplify getAddr() by using a 1-shot CompletableFuture for sending and receiving the request. This CHANGES BEHAVIOR, but should be OK for most/all existing use-cases. The previous implementation would query the server each time this method is called. The only use-case in bitcoinj itself is in `PeerMonitor` which only queries once upon connection to a new Peer.


msgilligan:
Marked as draft because this implementation doesn't allow retry after a failure.

PR #3215, opened 2023-09-02
build.gradle: generate protobuf classes into the build directory

schildbach:
This is extracted from #3059, but also removes the old directory from `.gitignore`.


msgilligan (approved):
Looks good! (I should have included removing the line in `.gitignore`)

PR #3216, opened 2023-09-02
Upgrade to EasyMock 5.2.0

msgilligan:
This is needed to build successfully with Java 21.


schildbach:
It appears at two places – can we update both?

msgilligan:
> It appears at two places – can we update both?

Fixed and force-pushed.

schildbach:
Merged.

PR #3218, opened 2023-09-05
gitlab-ci.yml: add trixie build for jdk 21

msgilligan:
@schildbach I'm not sure if you want to add trixie builds for both 17 & 21. If you want I can resubmit with just 21.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,6 +26,20 @@ bookworm-jdk17:
     paths:
       - core/build/libs/*.jar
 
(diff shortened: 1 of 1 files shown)


schildbach:
I expect JDK 17 will be removed from trixie once JDK 21 is final. So I'd prefer to just add the trixie / JDK 21 combination.

msgilligan:
> So I'd prefer to just add the trixie / JDK 21 combination.

Fixed and force-pushed.

schildbach:
Merged.

PR #3223, opened 2023-09-06
KeyTimeCoinSelector: refactor for clarity and in preparation to migrate to streams()

msgilligan:
There are three commits in this refactoring:

1. Explicitly check that parent is nonNull in `isConfirmed()`
2. Use `wallet.getConfidence(parent)` in `isConfirmed()` and make Wallet.getConfidence() package-private.
3. Extract `isKeyBeforeCutoff()` method


schildbach:
Merged.

PR #3226, opened 2023-09-06
CoinSelector: mark as FunctionalInterface

msgilligan:


schildbach:
I think it would be good to have a little bit of rationale in its commit message. What will the functional interface do in future, or how will it help?

msgilligan:
> I think it would be good to have a little bit of rationale in its commit message. What will the functional interface do in future, or how will it help?

I'll add more to the commit message and force-push.

The `@FunctionalInterface` annotation doesn't change behavior of `CoinSelector` in any way, it just declares our intention that it be used as a functional interface and will generate compiler errors if someone changes the code so that it is no longer a SAM type. (This is similar to the `@Override` annotation.)

msgilligan:
@schildbach commit message updated and force-pushed.

schildbach:
Thanks! Merged.

PR #3227, opened 2023-09-06
FilteringCoinSelector: use stream in select

msgilligan:
Note: this also prevents mutation of the candidate List which isn't
a problem as used within bitcoinj, but could be a problem if used
externally.

This is accidentally a child of PR #3226 (I marked as DRAFT and will rebase after that one is merged.)


schildbach:
Merged.

PR #3228, opened 2023-09-06
FilteringCoinSelector: make immutable

msgilligan:
This is an API-breaking change because it requires a List of TransactionOutPoint to be passed to the constructor and remove the `excludesOutputsSpentBy(tx)` method.

`Wallet` is updated to use the new constructor.


schildbach:
Merged.

PR #3230, opened 2023-09-06
Transaction: add a constructor as deprecated

schildbach:
Problem: At the moment, we don't have a replacement for the removed `new Transaction(NetworkParameters)`.

Workaround: Add back `new Transaction(NetworkParameters)` or a *previously not existing* `new Transaction()` as deprecated, so there is at least something apps can use.

Context: My code constructs an empty transaction and then adds inputs and outputs as needed. So it's basically using it as a builder.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -323,7 +323,7 @@ public class Transaction extends BaseMessage {
         vLockTime = LockTime.unset();
     }
 
-    /** @deprecated */
+    /** @deprecated use {@link #Transaction()} */
     @Deprecated
     public Transaction(NetworkParameters params) {
         this();
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -323,6 +323,12 @@ public class Transaction extends BaseMessage {
         vLockTime = LockTime.unset();
     }
 
+    /** @deprecated */
+    @Deprecated
+    public Transaction(NetworkParameters params) {
+        this();
+    }
+
     /**
      * Returns the transaction id as you see them in block explorers. It is used as a reference by transaction inputs
      * via outpoints.


schildbach:
I'l not sure if an constructing an empty transaction will continue to be a valid usecase in our "immutable tx" future. If it is, we could add a static `Transaction.empty()` or similar.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -323,6 +323,12 @@ public Transaction() {
         vLockTime = LockTime.unset();
     }
 
+    /** @deprecated */
```
msgilligan, line 326: This should specify the supported alternative.

msgilligan (changes requested):

msgilligan:
> I'l not sure if an constructing an empty transaction will continue to be a valid usecase in our "immutable tx" future. If it is, we could add a static `Transaction.empty()` or similar.

I spent some time thinking about this recently and have a specific proposal that I will soon share. It splits the current `Transaction` into two classes with a common abstract parent: one that is "complete" and signed and has immutable properties for everything in the raw transaction while retaining mutable fields for confirmation status, etc. and another that has mutable fields that support the legacy (to be deprecated, probably) builder behavior. Maybe we should try to implement that in this release since we are making major API changes to Transaction already.

schildbach:
Let's discuss about the names once we've discussed about the plan. For now, `MutableTransaction` and `ImmutableTransaction` would do.

schildbach:
I think we should priorize this issue by some amount. Currently, at least for Bitcoin Wallet, it's a blocker for 0.17.

@msgilligan Are you still up with presenting your API split?

schildbach:
I'm inclined to merge this now and find a better solution later, just to get along with a next alpha.

core/src/main/java/org/bitcoinj/core/Transaction.java
msgilligan: Well, actually there is no need to pass the _almost-deprecated_ `NetworkParameters`, so the `@deprecated` JavaDoc tag should probably suggest calling the no-args constructor. That may be deprecated/replaced in the future, but if we suggest the no-args constructor as an alternative we can at least remove this constructor in 0.18.

msgilligan (changes requested):
We should suggest using the no-args constructor rather than the deprecated single-arg `NetworkParameter` constructor.

msgilligan:
> I'm inclined to merge this now and find a better solution later, just to get along with a next alpha.

I agree. But we should at least suggest people migrate from `NetworkParameters` to the no-arg constructor.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: I agree. I added this in a commit on top.

schildbach:
I removed the draft state. While I still would prefer a better, more thorough solution, I propose to squash and merge this as is – just to get the next alpha rolling.

msgilligan (approved):
Looks good. Let's merge it!

schildbach:
Merged.

PR #3233, opened 2023-09-06
WalletTool: use CoinSelector.fromPredicate and related refactoring

msgilligan:
This is split into 4 commits, the first being the most significant (using `CoinSelector.fromPredicate`) and the others providing some further simplification/cleanup of surrounding code.


schildbach:
Merged.

PR #3234, opened 2023-09-06
DefaultCoinSelector: extract compareByDepth comparator

msgilligan:
This is actually split into two commits, the first is the extract method. The second updates the deprecated `sortOutputs` method to use `List.sort` instead of `Collections.sort` as suggested by the IDE (and the `@deprecation` comment.)


schildbach:
Merged.

PR #3237, opened 2023-09-06
WalletApplication: use switch expression in suffixFromNetwork()

msgilligan:
This is two commits:

1. Convert to switch expression
2. Remove unreachable `default` case.


schildbach:
Merged.

Issue #3239, opened 2023-09-07
NPE when trying to create a wallet with wallet-tool


schildbach:
On current master, I get this:

```
/bitcoinj/wallettool (master)$ ./build/install/wallet-tool/bin/wallet-tool create --wallet=test.wallet
java.lang.NullPointerException
	at org.bitcoinj.wallet.Wallet.saveToFile(Wallet.java:1708)
	at org.bitcoinj.wallet.Wallet.saveToFile(Wallet.java:1761)
	at org.bitcoinj.wallettool.WalletTool.createWallet(WalletTool.java:1107)
	at org.bitcoinj.wallettool.WalletTool.call(WalletTool.java:375)
	at org.bitcoinj.wallettool.WalletTool.call(WalletTool.java:111)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2041)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2453)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2415)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2273)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2417)
	at picocli.CommandLine.execute(CommandLine.java:2170)
	at org.bitcoinj.wallettool.WalletTool.main(WalletTool.java:336)
```

When commenting out the check below, it works.

```
        if (!destParentFile.exists()) {
            throw new FileNotFoundException(destParentFile.getPath() + " (wallet directory not found)");
        }
```

And it leaves a tempfile lying around.

A path like this

```
./build/install/wallet-tool/bin/wallet-tool create --wallet=./test.wallet
```

works as expected.

schildbach:
Quick check against 0.16: This is a regression.

msgilligan:
Fixed in PR #3240

PR #3240, opened 2023-09-07
Wallet: use getAbsoluteFile() when checking parent directory exists

msgilligan:
This fixes Issue #3239 that shows up in WalletTool.


schildbach:
Thanks for the fix! Merged.

PR #3242, opened 2023-09-07
EasingInterpolator: use switch expression

msgilligan:
This has two commits:

1. Migrate from switch statement to switch expression
2. Remove the redundant `default` case


schildbach:
Merged.

PR #3246, opened 2023-09-09
Use Duration in 2 calls to sleepUninterruptibly

msgilligan:
Reasons for doing this:
1) We are standardizing on using Duration where possible.
2) Duration values can be a single named constant if desired
3) When we replace this use of Guava, we'll want to use Duration

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java
@@ -29,6 +29,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -37,7 +38,6 @@ import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
-import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 
 import static org.bitcoinj.base.internal.Preconditions.checkState;
@@ -272,7 +272,7 @@ public class TransactionBroadcast {
 
     private static Runnable dropPeerAfterBroadcastHandler(Peer peer) {
         return () ->  {
-            Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
+            Uninterruptibles.sleepUninterruptibly(Duration.ofSeconds(1));
             peer.close();
         };
     }
--- a/wallettemplate/src/main/java/org/bitcoinj/walletfx/utils/KeyDerivationTasks.java
+++ b/wallettemplate/src/main/java/org/bitcoinj/walletfx/utils/KeyDerivationTasks.java
@@ -26,7 +26,6 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.*;
 import java.time.Duration;
-import java.util.concurrent.TimeUnit;
 
 import static org.bitcoinj.walletfx.utils.GuiUtils.checkGuiThread;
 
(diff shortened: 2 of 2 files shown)


msgilligan:
The compile is failing when run with JDK 11 and Gradle 4.10.3, but works when run with the same JDK 11 and Gradle 7 or Gradle 8.3.

I think Guava is using some Gradle magic or MR-JARs or something like that.  This results in some cases where the 1-arg version of `Uninterruptibles.sleepUninterruptibly()` is present and sometimes it is not-present. Yuck!

Note that if I use the `-jre` version of Guava it works even with Gradle 4.10.3.

Is `Duration` not available/used in the `-android` version of Guava? Is our Android 8.0+ sufficient where we can use the `-jre` version of Guava?

schildbach:
For Guava, Android means API level 15 (ICS), which is pretty much equivalent to JDK 7.

Maybe we should rewrite those two calls to

```
try { `TimeUnit.SECONDS.sleep(xxx); } catch(InterruptedException e) { throw new RuntimeException(e); }
```

optionally via a utils method.

msgilligan:
Given our Android requirement can we switch to the `-jre` variant of Guava?

msgilligan:
I viewed this PR as a precursor to providing our own replacement method, but yeah we could skip this step and go straight to a replacement method.

But at this point I'm more concerned about the build behavior being different with different versions of Gradle.

schildbach:
> Given our Android requirement can we switch to the `-jre` variant of Guava?

According to Google if you use this on Android you should use the `-android` variant. Anything else we'd be on our own and we'd need to try (and maintain).

schildbach:
FTR, here is the error from the log:

```
/home/runner/work/bitcoinj/bitcoinj/core/src/main/java/org/bitcoinj/core/TransactionBroadcast.java:275: error: method sleepUninterruptibly in class Uninterruptibles cannot be applied to given types;
            Uninterruptibles.sleepUninterruptibly(Duration.ofSeconds(1));
                            ^
  required: long,TimeUnit
  found: Duration
  reason: actual and formal argument lists differ in length
```

msgilligan:
> > Given our Android requirement can we switch to the `-jre` variant of Guava?
> 
> According to Google if you use Guava on Android you should use the `-android` variant. Anything else we'd be on our own and we'd need to try (and maintain).

I don't think Google are very clear about whether you an use the `-jre` variant if you are using an Android that supports JDK 8. I have opened Issue #3263 for further research and discussion of this issue.
 
> I think since our goal is to get rid of Guava we should invest our time into eliminating it rather than working out tweaks.

I generally agree with that sentiment, however migrating to the `-jre` variant might help us with this goal.

For example in this PR I was trying to migrate to using `Duration` in `sleepUninterruptibly` as a first step towards replacing the Guava method with our own implementation.  (That way I could write the replacement method to take `Duration` and the refactoring would be in two smaller, logical steps. Sticking with the `-android` version either requires me to write the two-argument version in one commit and then change to `Duration` in a second or make a single commit that changes the API and replaces the Guava implementation with our own.)

I'm not sure how many similar cases we'll stumble upon in our migration away from Guava, but it's possible that uses the JDK 8 flavor of Guava might help us migrate away from Guava.

msgilligan:
I opened Issue #3262 about the build problem we are having.  (An issue which also might be solved by migrating to the `-jre` version of Guava.)

msgilligan:
> Maybe we should rewrite those two calls to
> 
> ```
> try { `TimeUnit.SECONDS.sleep(xxx); } catch(InterruptedException e) { throw new RuntimeException(e); }
> ```

I think we should write our own "sleep uninterruptibly" method that behaves as the Guava method behaves. It ignores the interrupt exception but allows it to be handled elsewhere in the same thread.

I'll leave this issue as "open" as I reminder that I should write a "utils" implementation of that method. I'll first write it as compatible with what `-android` Guava provides (2 arguments) and then refactor to a single `Duration` argument version.

msgilligan:
Closing in favor of:
* PR #3293 
* PR #3294

PR #3248, opened 2023-09-12
PBKDF2SHA512: set hLen to correct value of 64 (and make static/private/final)

msgilligan:
This is split into two commits:

1. Set correct value of 64
2. make static/private/final

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/PBKDF2SHA512.java
+++ b/core/src/main/java/org/bitcoinj/crypto/PBKDF2SHA512.java
@@ -39,12 +39,12 @@ import java.nio.charset.StandardCharsets;
  * @see <a href="https://cryptofreek.org/2012/11/29/pbkdf2-pure-java-implementation/">PBKDF2 – Pure Java Implementation by Cryptofreek</a>
  */
 public class PBKDF2SHA512 {
+    // Length of HMAC result
+    static private final int hLen = 64;
     public static byte[] derive(String P, String S, int c, int dkLen) {
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
 
         try {
-            int hLen = 64;          // Length of HMAC result
-
             if (dkLen > ((Math.pow(2, 32)) - 1) * hLen) {
                 throw new IllegalArgumentException("derived key too long");
             } else {
--- a/core/src/main/java/org/bitcoinj/crypto/PBKDF2SHA512.java
+++ b/core/src/main/java/org/bitcoinj/crypto/PBKDF2SHA512.java
@@ -43,7 +43,7 @@ public class PBKDF2SHA512 {
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
 
         try {
-            int hLen = 20;
+            int hLen = 64;          // Length of HMAC result
 
             if (dkLen > ((Math.pow(2, 32)) - 1) * hLen) {
                 throw new IllegalArgumentException("derived key too long");


schildbach:
What makes you believe that 64 is the correct value, and 20 is wrong? E.g. https://github.com/zcash/kotlin-bip39/blob/3779fe9a9c8691d737a4b11a893995117b02fdd5/bip39-lib/src/jvmMain/kotlin/cash/z/ecc/android/crypto/Pbkdf2Sha512.kt#L40 has 20 too.

schildbach:
The linked CryptoFreak impl has 20 too: https://cryptofreek.org/2012/11/29/pbkdf2-pure-java-implementation/

msgilligan:
> What makes you believe that 64 is the correct value, and 20 is wrong? E.g. https://github.com/zcash/kotlin-bip39/blob/3779fe9a9c8691d737a4b11a893995117b02fdd5/bip39-lib/src/jvmMain/kotlin/cash/z/ecc/android/crypto/Pbkdf2Sha512.kt#L40 has 20 too.

The other implementations copied our code and our bug right along with it.

The original code (cryptofreak) was using SHA1 which produces a 160-bit hash (20 * 8) but Ken modified it to use SHA512-based HMAC which produces a 64-bit hash (64 * 8)  This value is supposed to indicate the length of a single hash produced by the `F` function.  Since, in our main use case `dkLen` = `hLen` and the `System.arraycopy` after the loop copies `dkLen` bytes it wasn't a problem.

schildbach:
From the RFC:

> hLen                              denotes the length in octets of the                   pseudorandom function output

and also

> The length of the derived key is bounded by the length of the hash function output, which is […] 20 octets for SHA-1.

However the algo has been modified to use SHA-512.

schildbach:
Related stackexchange: https://crypto.stackexchange.com/questions/7930/pbkdf2-when-dklen-and-hlen-have-the-same-size

schildbach:
@ksedgwic since you're the original author of the class, I'd like to hear your thought. Do you agree the value of `hLen` should be 64 (because of SHA-512), not the current 20 (used for SHA-1 according to the RFC)?

msgilligan:
I suppose we could change the API to only support length = 64 and a single iteration...

ksedgwic:
@schildbach I'm afraid I don't have any recollection of the old motivation.  When I research now I find this answer which seems rational to me:
https://security.stackexchange.com/a/53917

My interpretation of that reasoning n this use case is:
- It's ok to only use 20 bytes if that's all you need/want
- It's ok to use up to the the native hash's output (64 bytes)

Will changing to 64 bytes alter the mapping from entropy to seed as compared to https://iancoleman.io/bip39/ ?

Is the current code compatible w/ iancoleman?

msgilligan:
> Will changing to 64 bytes alter the mapping from entropy to seed as compared to https://iancoleman.io/bip39/ ?

No. In fact, changing `hLen` to 64 will not change the output at all as `hLen` is only used for two purposes:

1. Check for `IllegalArgumentException("derived key too long")` - which can never fail (for positive int values of `hLen` and `dkLen`)
2. For calculating `l` which is the number of times to call `F`.

Since the results of each call to `F` are appended to a `ByteArrayOutputStream` with `baos.write(T)` and the final result is built by taking the first `dkLen` bytes of the resulting `byte[]`, the only practical effect changing `hLen` has is to eliminate unnecessary calls to `F`. In other words, the incorrect value of `hLen` is causing extra calls to `F` which are just thrown away.

msgilligan:
FYI: This can be implemented using built-in `java.crypto`. See PR #3249 which passes all tests. I am not sure if that version will work on Android. Perhaps when this was first implemented this functionality wasn't available on Android, but is now?

schildbach:
@ksedgwic Thanks for your assessment.

I decided to merge this, as the fix seems logical to me. And our freshly added test vectors give confidence of no regression.

schildbach:
Note this doesn't mean #3249 and #3251 are out of question.

PR #3249, opened 2023-09-13
PBKDF2SHA512: use javax.crypto

msgilligan:
This passes all tests. I don't know if it will work on Android.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/PBKDF2SHA512.java
+++ b/core/src/main/java/org/bitcoinj/crypto/PBKDF2SHA512.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012 Cole Barnes [cryptofreek{at}gmail{dot}com]
+ * Copyright by the original author or authors.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


schildbach:
I don't think this is a good idea, because the loose coupling of the javax.crypto API simply makes the outcome somewhat unpredictable.

Maybe we can import Conscrypt and use the provider directly?

msgilligan:
> I don't think this is a good idea, because the loose coupling of the javax.crypto API simply makes the outcome somewhat unpredictable.

I think we probably can rely on it. By my research `PBKDF2WithHmacSHA512` is bundled with Java 8 and later, [is provided by Android](https://developer.android.com/reference/javax/crypto/SecretKeyFactory) API 26+ (our current requirement), and is also provided in the Bouncy Castle version we are using.

This class was added to the codebase 10 years ago when we couldn't rely on its presence in Java or Android (and probably Bouncy Castle as well, but I don't know when it was added to BC.) I seems to me it would be better to use a standard implementation than to provide our own.

> Maybe we can import Conscrypt and use the provider directly?

I'm not sure what you mean by this. Are you referencing the [Conscrypt library](https://github.com/google/conscrypt)? Do you mean add a JAR to the dependencies or just `import` something that is already in our classpath?

msgilligan:
PR #3251, via this commit: https://github.com/bitcoinj/bitcoinj/pull/3251/commits/9250c5bc8b6ed56a030fe4a2f581f071efaa12f6 demonstrates explicitly using `"BC"` to provide `PBKDF2WithHmacSHA512`.

schildbach:
> > Maybe we can import Conscrypt and use the provider directly?
> 
> I'm not sure what you mean by this. Are you referencing the [Conscrypt library](https://github.com/google/conscrypt)? Do you mean add a JAR to the dependencies or just `import` something that is already in our classpath?

Add a JAR if necessary, and yes use import + constructor.

But actually I think our current code does well, and has lots of test coverage etc. Let's close this for now. We can re-open if there general requirements change.

PR #3251, opened 2023-09-14
Experiment: Use BouncyCastle for PBKDF2WithHmacSHA512

msgilligan:
This is an experiment to test and demonstrate that our Bouncy Castle dependency provides `PBKDF2WithHmacSHA512` in all our tested CI configurations.


schildbach:
I think up to now, we're using BC only directly, not via javax.crypto. Maybe we can do the same for PBKDF2WithHmacSHA512?

msgilligan:
We're using some JCE hash functions directly already so I don't see this is a chance in philosophy. And I'd rather keep this function independent of the external library.

But if you really think it's better to use the direct BC API, I still think that is better than using this copypasta implementation.

schildbach:
I was thinking of maybe using the `PBEPBKDF2.PBKDF2withSHA512` class directly, but didn't understand how to use it exactly. BC is missing JavaDocs unfortunately.

The problem with the "loose coupling" provider mechanism is you never know what code you'll be using and chances are that the code path remains untested (by our unit tests).

msgilligan:
Why can't we use the "SunJCE" provider directly?  (Although maybe we might have to use a different provider on Android?)

If we specify provider(s) don't we know which implementation we're getting?

msgilligan:
See Issue #3912 -- we should probably migrate to using the JDK implementation. I don't think we want to increase our usage of Bouncy Castle in this case.

PR #3253, opened 2023-09-14
PBKDF2SHA512: Remove redundant check of max dkLen 

msgilligan:
This PR consists of 5 small refactoring that improve the error-checking and readability of the  `derive` method.

1. Move parameter check of `dkLen` out of try/catch
2. Don't allow negative count or `dkLen`
3. Remove redundant check of exceeded max `dkLen` value
4. Declare/catch specific exceptions thrown by `F` and caught in main `try`
5. Calculate `l` (loop count) with integer math

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/PBKDF2SHA512.java
+++ b/core/src/main/java/org/bitcoinj/crypto/PBKDF2SHA512.java
@@ -50,9 +50,6 @@ public class PBKDF2SHA512 {
         Preconditions.checkArgument(dkLen > 0, () -> "derived key length must be greater than zero");
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
 
-        if (dkLen > ((Math.pow(2, 32)) - 1) * H_LEN) {
-            throw new IllegalArgumentException("derived key too long");
-        }
         try {
             int l = (dkLen + H_LEN - 1) / H_LEN;    // Divide by H_LEN with rounding up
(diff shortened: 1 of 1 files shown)


schildbach:
I've merged everything except 3. Even if the check is redundant, I think it's good to have – at least as a comment. Reasons:

- Tells reviewers we've implemented the algo described in the RFC
- Helps other who copy our code to languages where the check is not redundant

schildbach:
Feel free to rebase this PR so it becomes clear it's now only about 3.

msgilligan:
Yes, I'll rebase.

I might also make a competing PR that makes the max len check a comment or maybe implement it with integer math. (I really try to avoid floating point math whenever possible.)

(Of course, the compiler may figure out the result is always false and optimize it out.)

msgilligan:
Rebased, force-pushed, and title is update to reflect that this PR is now only item 3 (see above.)

msgilligan:
Note: There are now three PRs with three different ways of simplifying/removing this calculation:

1. https://github.com/bitcoinj/bitcoinj/pull/3253 (remove with no comment)
2. https://github.com/bitcoinj/bitcoinj/pull/3254 (replace calculation with equivalent using `long` value)
3. https://github.com/bitcoinj/bitcoinj/pull/3255 (remove with comment explaining why check is unnecessary)

At most one of these should be merged. This PR is for option 1.

schildbach:
Closing in favor of #3255.

PR #3256, opened 2023-09-16
Wallet: replace Interables.concat with Stream.concat in getWatchedOutputs

msgilligan:
This simplifies the code and removes a Guava dependency.

It is done in two steps:

1. replace nested for loops with streams
2. Replaces Interables.concat with Stream.concat


schildbach:
Merged.

Issue #3262, opened 2023-09-16
Builds with recent versions of Gradle use wrong variant of Guava (`-jre` not `-android`)


msgilligan:
For some reason our builds with more recent Gradle versions are compiling against the `-jre` version of Guava, not the `-android` version. This means usage of methods only found in the `-jre` version won't result in a build failure with those versions of Gradle. Fortunately, they will be detected by the Gradle 4.10.3 build in our current GitHub Actions CI configuration.

I'm not sure why Gradle is doing this, but I am noticing that these newer Gradle versions are pulling a `-jre` parent pom. The fix may be to tell Gradle to lock the `-android` version.

See PR #3246 for an example of where the use of methods only available in the `-jre` version are not detected by builds with Gradle 8.3.

msgilligan:
[This Stack Overflow](https://stackoverflow.com/questions/65646026/truth-extensions-causing-rest-of-project-to-downgrade-to-guava-android) provides a hint:

> So you can actually fix this by depending on any -jre version from 30.0-jre up: 30.0-jre is considered "newer" than 30.0-android because of alphabetical order. Fun!

The OP in the Stack Overflow is having the opposite issue of us: Getting the Android version when the JRE version is needed, but the root cause of the problem is the same.

msgilligan:
The "Selecting the appropriate flavor" section of the Release notes for Guava v32.1.0 explains the issue and provides workarounds. (See https://github.com/google/guava/releases/tag/v32.1.0#user-content-selecting-flavor)

msgilligan:
PR #3264 implements the recommendation from the Guava 32.1.0 release notes.

schildbach:
Fixed via #3264

PR #3264, opened 2023-09-16
core/build.gradle: Use constraint to enforce Android Guava

msgilligan:
This is a solution to Issue #3262 

I tested this against PR #3246 which was only failing with Gradle 4.10.3 but should have been failing for _all_ builds in the GitHub CI matrix. With this change the build properly fails when JRE-only Guava methods are used.


schildbach:
Merged.

PR #3271, opened 2023-09-17
build.gradle: drop the Eclipse plugin

schildbach:
We assume that these days the plugin isn't needed to set up bitcoinj within Eclipse.

If you feel this is not the case, and/or if you're actually using Eclipse and would like to continued support for it, please comment here.

I'll leave this open for a while deliberately. There is no need to hurry.


schildbach:
Merged.

PR #3272, opened 2023-09-18
.gitlab-ci.yml: Add :bitcoinj-wallettool:installDist Gradle targets

msgilligan:
Add it to all three build jobs.

The full build of wallet-tool should be part of regular CI.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,7 +4,7 @@ bullseye-jdk11:
     - apt-get update
     - apt-get -y install openjdk-11-jdk-headless gradle
   script:
-    - gradle build --stacktrace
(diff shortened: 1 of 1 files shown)


schildbach:
I'm not sure if our CI builds need this. Actually I planned to propose to "degrade" of those builds to just the `test` task and not publish any artifacts from those builds at all.

In future, we've got our reference build for that and we should only publish those artifacts.

msgilligan:
I think every build should make sure every output can be built and potentially be a reference build. Ideally we could even run tests on the built wallet-tool.

We don't have to publish it as an artifact I suppose, but I don't see any harm in doing that.

schildbach:
Merged.

PR #3273, opened 2023-09-18
.gitlab-ci.yml: Add sha256sum

msgilligan:
Add sha256sum for bitcoinj-core:build and :bitcoinj-wallettool:installDist
Gradle outputs

This is a child of PR #3272

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,6 +8,8 @@ reference:
     - update-alternatives --set iptables /usr/sbin/iptables-legacy
   script:
     - buildah build --file build.Containerfile --output build .
+  after_script:
+    - sha256sum build/core/* build/wallettool/bin/*  build/wallettool/lib/*
   artifacts:
     name: bitcoinj-$CI_JOB_NAME-$CI_COMMIT_REF_SLUG-$CI_COMMIT_SHORT_SHA
     paths:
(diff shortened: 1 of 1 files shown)


schildbach:
I'm pretty sure the checksums will be different on different JDK versions.

msgilligan:
That's fine.

.gitlab-ci.yml
```diff
@@ -5,6 +5,7 @@ bullseye-jdk11:
     - apt-get -y install openjdk-11-jdk-headless gradle
   script:
     - gradle build :bitcoinj-wallettool:installDist --stacktrace
+    - sha256sum core/build/libs/*.jar wallettool/build/install/wallet-tool/bin/*  wallettool/build/install/wallet-tool/lib/*.jar
```
schildbach, line 8: These commands could perhaps go to the after_script sections?

schildbach (commented):
Ok, I think if we add this it should include the job for the reference build as well. Could you rebase this?

schildbach:
Also, we could consider of creating a (published) `SHA256SUMS` file rather than printing into a log. That way, the file could be GPG-signed by us at some point.

msgilligan:
> Ok, I think if we add this it should include the job for the reference build as well. Could you rebase this?

I will rebase!

.gitlab-ci.yml
msgilligan: I'll try that.

.gitlab-ci.yml
```diff
@@ -8,6 +8,8 @@ reference:
     - update-alternatives --set iptables /usr/sbin/iptables-legacy
   script:
     - buildah build --file build.Containerfile --output build .
+  after_script:
+    - sha256sum build/core/* build/wallettool/bin/*  build/wallettool/lib/*
```
schildbach, line 12: If we want to recursively shasum the entire build directory, we could use

`find build/ -exec sha256sum {} +`

msgilligan: I would do that but because we want to compare that line with the lines in the non-reference builds, I think it's better to enumerate the paths.

PR #3274, opened 2023-09-18
core/build.gradle: set JAR dirMode and fileMode for reproducible builds

msgilligan:


schildbach:
I haven't checked but this will most likely remove the `x` bit from the executables.

msgilligan:
are their executables inside a JAR?

schildbach:
> are there executables inside a JAR?

Yes, we've got `wallet-tool` and `wallet-tool.bat`.

msgilligan:
> > are there executables inside a JAR?
> 
> 
> 
> Yes, we've got `wallet-tool` and `wallet-tool.bat`.

But they're not in a JAR, right?

schildbach:
Ok, that's true.

schildbach:
I folded this PR into #3270.

schildbach:
Re-opening this, because I will merge this ahead of #3270 once we've confirmed that #3270 produces the checksums for you.

msgilligan:
> Re-opening this, because I will merge this ahead of #3270 once we've confirmed that #3270 produces the checksums for you.

I agree this should be reviewed/merged independently.

schildbach:
That could be archieved with

```
tasks.withType(Jar) {
    …
}
```

schildbach:
We still might want the other attributes unchanged.

msgilligan:
> We still might want the other attributes applied to all archives.

I pushed an update to change the `withType` to `Jar` before seeing this message. Splitting into two closures is probably a good idea. We should also probably put it in the root `build.gradle` so it applies to all subprojects.

msgilligan:
@schildbach rebased to add dirMode and fileMode options to separate `withType(Jar)` closure.

schildbach:
Merged.

PR #3275, opened 2023-09-19
WIP: Reference build with reusable build container

msgilligan:
This commit splits the reference build into two steps:

1. Build a reference build container with `reference-container-build.sh`
2. Use the container to create a bitcoinj build with `bitcoinj-reference-build.sh` 

`reference-container.ContainerFile` is the ContainerFile for the reusable build container.

### DIFF
--- a/bitcoinj-reference-build.sh
+++ b/bitcoinj-reference-build.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# Run reference-container-build.sh before using this script
+# The output will be produced in the same location as a locally hosted build
+docker run --rm --name bitcoinj-reference-build -v .:/project -it \
+    bitcoinj/reference-build:latest \
+    /usr/bin/gradle --no-build-cache --no-daemon --no-parallel --project-dir /project/ \
+        :bitcoinj-core:build :bitcoinj-wallettool:installDist
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,28 @@
+version: "3.5"
+
+#
+# Docker Compose File for reproducibly building bitcoinj
+#
+# To build use:
+#     docker compose run --rm build
+#
+# This will create a local image `named bitcoinj-build-tools-image` (if it does
+# not already exist) using `reference-container.ContainerFile`. This is a debian slim image
+# with openjdk and gradle installed via `apt-get`.
+# The `build` "service" will mount this directory in the container as "/project" and run
+# the Gradle build inside the container, writing artifacts to the same locations on your
+# disk as if you had run `gradle` on the host.
+#
+services:
+    build:
+        image: bitcoinj-build-tools-image 
+        depends_on:
+            - build-image-builder
+        volumes:
+            - .:/project
+        command: /usr/bin/gradle --no-build-cache --no-daemon --no-parallel --project-dir /project/ clean :bitcoinj-core:build :bitcoinj-wallettool:installDist
+    build-image-builder:
+        build:
+            context: .
+            dockerfile: reference-container.ContainerFile
+        image: bitcoinj-build-tools-image             # this service builds the image
--- a/reference-container-build.sh
+++ b/reference-container-build.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+#
+# This script builds a Container image that can be used to run a reference build of bitcoinj.
+# Run this script to build the container and then use bitcoinj-reference-build.sh to
+# make the reference build of bitcoinj.
+# When building the continer `apt-get update` and `apt-get install` are run once and can
+# then be reused for multiple builds of bitcoinj.
+#
+docker build --file reference-container.ContainerFile -t bitcoinj/reference-build .
+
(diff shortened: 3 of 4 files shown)


core/build.gradle
```diff
@@ -55,6 +55,8 @@ protobuf {
 tasks.withType(AbstractArchiveTask) {
     preserveFileTimestamps = false
     reproducibleFileOrder = true
+    dirMode = 0755
```
schildbach, line 58: This is interesting – it could perhaps replace the `chmod` I added earlier?

schildbach: However, it will likely remove the `x` bit from the executables – the `X` syntax keeps those.

schildbach:
What problem does this PR solve?

core/build.gradle
msgilligan: I don't think we have any of executables inside JARs, right?

msgilligan:
1. It builds a reusable container image that has all the apt-get commands completed. So you can now do repeated local builds that run very quickly.

2. Local reproducible builds are run with <docker/podman> run with the local repository mounted as a volume. (This is how many GitLab build tools run, e.g. SAST scans)

msgilligan:
If we wanted we could publish this image to our GitLab container build and speed up our GitLab jobs.

core/build.gradle
msgilligan: > This is interesting – it could perhaps replace the chmod I added earlier?

Yes. In my testing it works for that purpose.

msgilligan:
@schildbach This approach also allows us to snapshot the version of container (complete with the versions of the `apt` packages) used to make the build and to possibly reproduce it later.

schildbach:
One goal I have with my containerized approach is that validators don't need to trust us on doing bad things to their system, as long as they trust their container tech to properly contain our build.

This is something I frequently criticize with other projects I build/validate, because I have to edit away their Makefile or shellscript wrappers, or entirely understand their possible side effects. In my eyes, this is a major hurdle for getting projects validated etc.

The Docker way is basically to run "docker build .", and that's it. (Or, you could use docker-compose, or Docker Desktop, or Podman, or buildah, or maybe in future buildx.)

On top of that, the script approach from this PR makes the build process environment dependent again, as the `SHASUM_CMD` variable and `/bin/bash` declaration demonstrate.

The goal of snapshotting a specfic arrangement of Debian packages we could also archive by splitting out a `setup-stage` from the `build-stage` and invoking `docker build --target setup-stage -t <tag>`. But as said elsewhere I don't expect we'll run into this problem at all.

msgilligan:
@schildbach rebased now that PR #3274 is merged. I also removed the shasum command from the second script.

msgilligan:
This approach also works with `podman machine` (tested on macOS)  (When using podman machine you must configure the mount path relative to the root of the VM, e.g. `-v /Users/name/path/bitcoinj:/project`)

msgilligan:
@schildbach The shell scripts can be replaced with a `docker-compose.yml` file, which I assume could also be used with Podman.  See 790f2ae4884611ed06113fe41c24a331d20be1ac

schildbach:
`docker-compose` feels like the wrong tool for building, it's meant to orchestrate services. Our build is not a service however.

Volumes, similar thing. They're meant to persist the state of services across restarts/migrations. (I know there are read-only bind mounts.)

The "Docker way" of building something is `docker build` and using the build context for transforming input to output. It's a bit like pipes when using the shell.

msgilligan:
> `docker-compose run` feels like the wrong tool for building, it's meant to orchestrate services. Our build is not a service however.

Yeah, this seems to be a slightly unusual way of using `docker-compose`, but it works.

> 
> Volumes, similar thing. They're meant to persist the state of services across restarts/migrations. (I know there are read-only bind mounts.)

Volumes are used for build pipelines all the time. GitLab uses this extensively. It's a standard way to run a containerized command-line tool on a local directory. That is what we are doing.

> The "Docker way" of building something is `docker build` and using the build context for transforming input to output. It's a bit like pipes when using the shell.

I see you are pretty set on your current approach and I can live with it if you insist upon using it. But it doesn't work with podman on macOS (i.e. remote mode) and it doesn't make our reference build environment as a named container that can be used for other purposes. This solution does both of those things.

msgilligan:
> The goal of snapshotting a specfic arrangement of Debian packages we could also archive by splitting out a `setup-stage` from the `build-stage` and invoking `docker build --target setup-stage -t <tag>`.

I want to do this. Is this something we can add to PR #3270 ?

> But as said elsewhere I don't expect we'll run into this problem at all.

I'm not sure what you mean by this.

schildbach:
> I want to do this. Is this something we can add to PR #3270 ?

Done, via 477b223e05f0b6243dcf34b10061cee9f8dd3a38.

schildbach:
> I'm not sure what you mean by this.

I meant the problem of a moving Debian (due to security patches) destabilizing our reproducibility.

msgilligan:
Closing in favor of PR #3270.

PR #3277, opened 2023-09-19
MnemonicCode: fix comment about iteration count

### LINKED ISSUE #3276: BIP-39 iteration count
The below is a part of source code that converts mnenomic back to a seedphrase:

https://github.com/bitcoinj/bitcoinj/blob/56658e44de78156bad2a9cd6427d8c60300b7f1a/core/src/main/java/org/bitcoinj/crypto/MnemonicCode.java#L142

Per [BIP-39, the iteration count should be 2048](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed).

What's the reason behind choosing 4096 over the spec?
All I'm aware is that it makes the generated hash more secure, but wouldn't that generate a different hash than the applications following BIP-39 standard?

schildbach:
Fixes #3276.


msgilligan (approved):
Approved!

schildbach:
Merged.

PR #3278, opened 2023-09-20
graalvm.yml: Add GraalVM CE 21 to the Matrix

msgilligan:
This consists of 2 commits:

1. Remove unneeded declaration of native-image component (This is not needed with GraalVM 17 and will cause an error with GraalVM 21)
2. Add GraalVM 21 to the Matrix


schildbach:
Merged.

PR #3279, opened 2023-09-21
build.Containerfile: update reference build to use JDK 17

schildbach:
Also updates the container to Debian Bookworm, because we need its specific Gradle version.

Note this change makes the JavaDoc build artifact reproducible.


msgilligan (approved):
Approved!

schildbach:
Merged.

PR #3280, opened 2023-09-21
gradle.yml: use JDK `21` in the matrix

msgilligan:
Upgrade from `21-ea` to `21`.

~~This is a DRAFT PR because `21` isn't available on GitHub yet and the build is failing. When `21` is available we can re-run tests on this PR and then merge.~~


schildbach:
Merged.

PR #3282, opened 2023-09-21
.gitlab-ci.yml: remove publishing of artifacts for CI builds

schildbach:
We have our reference build now and all published artifacts go through that one.

Example of an entire build (all jobs):

https://gitlab.com/bitcoinj/bitcoinj/-/pipelines/1012760479/builds

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -22,10 +22,6 @@ bullseye-jdk11:
     - gradle build --stacktrace
   after_script:
     - gradle --version
-  artifacts:
-    name: bitcoinj-$CI_JOB_NAME-$CI_COMMIT_REF_SLUG-$CI_COMMIT_SHORT_SHA
-    paths:
-      - core/build/libs/*.jar
 
 bookworm-jdk17:
   image: debian:bookworm-slim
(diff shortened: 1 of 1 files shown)


msgilligan (approved):
Reluctantly approved. I don't see any harm in making these binary artifacts available and I sometimes find them useful when I'm troubleshooting.

schildbach:
Can you describe a situation where you downloaded one of these artifacts and found this useful?

(I don't want to take away something that is useful, and can't be substituted by a slightly different procedure.)

msgilligan:
> Can you describe a situation where you downloaded one of these artifacts and found this useful?

Just this week while debugging reference builds -- both the implementation on the (now merged) PR and alternate implementations as well. And I was debugging with Bookworm/JDK 17 while the reference build PR was still on Bullseye/JDK 11. So this situation could arise again when we want to start migrating to Trixie/JDK 21.

Note that I can make a sha256sum-matching build on macOS and as a Mac (or Windows) user I might want to verify that my build (perhaps a JDK 21-based build) generates the same thing as under Trixie on GitLab (or under macOS/Windows on GitHub)

schildbach:
I decided to merge this one for now. We do have checksums for the builds now, so checking for bit identity is possible. For more introspection at the moment I recommend building either locally or via an own GitLab repository.

PR #3283, opened 2023-09-21
Gradle build: support "reproducible" and "modern" variants

msgilligan:
Simplify Gradle feature testing and reduce the number of versions of Gradle that the build supports.

* support Gradle 4.4 - 4.11 and Gradle 7.5+
* simplify feature testing to use minModernGradleVersion

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -81,9 +81,9 @@ test {
     }
 }
 
-def gradleVersionToolchains = GradleVersion.version("6.7")
+def minModernGradleVersion = GradleVersion.version("7.4")
 
-if (GradleVersion.current().compareTo(gradleVersionToolchains) > 0) {
+if (GradleVersion.current().compareTo(minModernGradleVersion) > 0) {
     // If the Gradle Java Toolchains feature is available, run tests on older JDKs
     System.err.println "Adding 'testOnJdk8' task, because ${GradleVersion.current()}"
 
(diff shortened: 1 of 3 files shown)


schildbach:
In general, I agree this seems like the a sane way forward, given that our Gradle 4.4 requirement isn't going to budge within the next months.

However, I'm not sure if we should drop all the carefuilly tested version checks from the build.gradles. We could just keep them and use "minModernGradleVersion" checks (7.4+) on new stuff that might come around.

Also, I'm sure this PR requires some README changes?

msgilligan:
> given that our Gradle 4.4 requirement isn't going to budge within the next months.

Next years, right? There won't be any movement beyond Gradle 4.4.x until Trixie is released, right? And I haven't seen any new versions on Trixie or Sid, yet.

> carefully tested version checks from the build.gradles.

Well they were carefully researched and carefully tested when they were committed. But we haven't been regularly testing with "intermediate" Gradle versions as I discovered while testing PR #3264.

> We could just keep them and use "minModernGradleVersion" checks (7.4+) on new stuff that might come around.

I suppose we could keep them as long as we add code to _disallow_ these "intermediate" Gradle versions that we are not testing. We could then either use "minModernGradleVersion" for new stuff or add new checks on a feature-by-feature basis that use the appropriate version for that feature.

To me the most important part of this PR is disallowing the "intermediate" versions since they are not tested. (And I believe they currently do not work.)

> Also, I'm sure this PR requires some README changes?

Yes. Let's agree on what we want to do and I will rebase and add README changes.

schildbach:
I think given that we're close to a 0.17 release we should postpone this to shortly after branching off for release.

msgilligan:
Closing in favor of PR #3449

PR #3289, opened 2023-09-28
GitHub Actions: Use build-scan-agree.gradle init script

msgilligan:
Add a Gradle initialization script to non-interactively agree to the Gradle Build Scan terms of service. Provided by Stefan Wolf via GitHub Issues.

Works with Gradle 4.4 - Gradle 8.4 (at least)

See:

https://github.com/gradle/gradle/issues/26316#issuecomment-1739245349

`--scan --init-script build-scan-agree.gradle` added to gradle.yml and graalvm.yml GitHub workflows.

(If we decide to add to GitLab CI that can be in a separate PR)

### DIFF
--- a/.github/workflows/graalvm.yml
+++ b/.github/workflows/graalvm.yml
@@ -27,7 +27,7 @@ jobs:
         uses: gradle/gradle-build-action@v2.8.0
         with:
           gradle-version: ${{ matrix.gradle }}
-          arguments: test nativeCompile --info --stacktrace
+          arguments: test nativeCompile --init-script build-scan-agree.gradle --scan --info --stacktrace
       - name: Upload wallet-tool as artifact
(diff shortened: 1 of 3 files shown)


schildbach:
I wonder _who_ exactly is agreeing to the TOS if we merge this?

Did someone already agree because of this script being run (in context of this PR)?

What happens if someone branches off after this PR has been merged and the script runs without s/he really being aware of it?

msgilligan:
I suppose we could add an GitHub (repository-level, probably) environment variable named something like `EXTRA_CI_GRADLE_OPTS` (which would be empty string if not defined) 

See: https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository

This way it would only run in repositories where this variable is set.

schildbach:
Merged.

PR #3292, opened 2023-10-01
TransactionConfidence: check for mutation of an already set source

schildbach:
This uncovers a couple of cases where a tx confidence source, that was already set to `SELF`, is being reset to `NETWORK`.

In particular, I think the reorgs should be fixed.


msgilligan (approved):
Looks good!

schildbach:
Merged.

PR #3293, opened 2023-10-01
KeyDerivationTasks: use Thread.sleep instead of Guava Uninterrruptables, define interval with a Duration

msgilligan:
There are two commits:

1. define sleep interval with a `Duration` constant
2. Use `Thread.sleep` directly rather than `Uninterruptables.sleepUninterrruptibly()`

We don't really need to sleep uninterrruptibly here. The code is only displaying a progress indicator and the `call()` method is already throwing `Exception`.


msgilligan:
Temporarily converting to DRAFT. I need to remove some import statements and force-push...

schildbach:
I think the first commit will fail, because the method doesn't exist in guava-android?

schildbach:
I'll reorder the commits.

schildbach:
Or maybe I'm wrong.

msgilligan:
> I think the first commit will fail, because the method doesn't exist in guava-android?

The first commit still calls the guava-android method signature.

schildbach:
Merged.

PR #3296, opened 2023-10-01
TransactionBroadcastTest: Use AtomicLong over Guava AtomicDouble

msgilligan:
Use AtomicLong with some wrapping and unwrapping instead.


schildbach:
Merged.

PR #3297, opened 2023-10-01
TestWithPeerGroup: remove unneeded use of ListeningScheduledExecutorService

msgilligan:
There is no longer a need to decorate ScheduledThreadPoolExecutor to support ListeningScheduledExecutorService since PeerGroup no longer uses it.


schildbach:
Merged.

PR #3300, opened 2023-10-01
FakeTxBuilder.BlockPair: make fields final

msgilligan:
They are only set in one place, so let's make them final and initialized in the constructor.


schildbach:
Merged.

PR #3303, opened 2023-10-03
TransactionBroadcast: allow interrupt in dropPeerAfterBroadcastHandler

msgilligan:
If an `InterruptedException` occurs while sleeping in `dropPeerAfterBroadcastHandler`, close the peer immediately with `peer.close()` instead of delaying the exception to preserve a full 1-second delay.

PeerGroups can be shut down, and they should shut down as quickly as possible.

This is in response to comments from @schildbach on PR #3294 suggesting the sleep here should be interruptible.


schildbach:
Merged.

Issue #3299, opened 2023-10-01
Rewrite `WalletAppKit` to not extend Guava `Service`


msgilligan:
This is perhaps our most significant Guava dependency in the `bitcoinj-core` API.

A rewrite that doesn't subclass Guava `Service` and is a little more asynchronous and `CompletableFuture`-friendly is something I would like to see in the 0.18 release.

msgilligan:
Actually, I've been thinking about this further and I'm now leaning towards the following approach:

1. [optional] In the 0.17 release move `WalletAppKit`to it's own `org.bitcoinj.walletappkit` package
2. In the 0.18 release move `WalletAppKit` to its own JAR/module
3. Provide a newly designed alternative in `core` or possibly another module.

Step 1 is optional because we could leave it in the `org.bitcoinj.kits` package and avoid users having to change their import declaration. But long-term it might be nice for the package name and module name to match.

The reasoning behind this approach is that `WalletAppKit` has been the recommended way to build wallets for a long time and it might be better for users if it remains available for use (and for sending PRs) for some number of years.

This would also give us the freedom to build an alternative from a clean slate and allow users to migrate at their own pace.

msgilligan:
PR #3316 implements step (1) above and is proposed to target milestone 0.17.

schildbach:
I think we should not rename the package `kits` to `walletappkit`, at least not for now. Packages are thought of as collections of classes, in this case various "kits". A future better impl could go to this package too.

If we manage to provide a Guava-free re-implementation for 0.18, we could possibly do without step 2 (separate submodule) too. But I'm not opposed to separating `WalletAppKit` out, it seems like a sensible approach.

msgilligan:
> I think we should not rename the package `kits` to `walletappkit`, at least not for now. 

OK, I'll leave this open but move to the 0.18 milestone.

ItoroD:
Hi @msgilligan do we still plan to create a guava-free re-implementation of walletappkit? 
Do you mind sharing why we want to take out the guava dependency

msgilligan:
> Hi @msgilligan do we still plan to create a guava-free re-implementation of walletappkit?

Yes. I would like to see this happen!

> Do you mind sharing why we want to take out the guava dependency

* In general, we are trying to remove dependencies wherever we can
* When mixing bitcoinj with other libraries you can get Guava version conflicts
* `WalletAppKit` extending `AbstractIdleService` IMO makes the code harder to read and debug
[...]

PR #3316, opened 2023-10-15
Rename `kits` package to `walletappkit`

msgilligan:
This is in preparation for a proposed move to a standalone module/JAR.

See: 
https://github.com/bitcoinj/bitcoinj/issues/3299#issuecomment-1745436911

### DIFF
--- a/core/src/main/java/org/bitcoinj/walletappkit/WalletAppKit.java
+++ b/core/src/main/java/org/bitcoinj/walletappkit/WalletAppKit.java
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.bitcoinj.kits;
+package org.bitcoinj.walletappkit;
 
 import com.google.common.io.Closeables;
 import com.google.common.util.concurrent.AbstractIdleService;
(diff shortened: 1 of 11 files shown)


schildbach:
Even if we decide to make this package its own submodule, I'm not sure if the rename makes sense. Packages typically contain more of something than just one. I'd perhaps keep the name and still (maybe) extract a submodule in future?

msgilligan:
> ... I'm not sure if the rename makes sense. Packages typically contain more of something than just one.

Generally they do contain more than one item/class. But this package only contains one class and will probably stay that way. I think it was originally named `kits` (by Mike) because he expected there to be more added in the future.

So there are (at least) two reasons for renaming:

1. The current name is misleading as it implies more than one. (A minor issue, I realize)
2. If we extract  a submodule (which I strongly feel we should) it would be good for the package name to match the module name.

I agree this change doesn't make much sense if we decide _not_ to extract a submodule. So we need to decide if we want to extract a submodule _first_. Even if we decide to extract a submodule this change isn't strictly necessary but is still IMO a good idea (though I feel less strongly about the rename than about the extraction.)

msgilligan:
Based upon https://github.com/bitcoinj/bitcoinj/issues/3299#issuecomment-1813118236 I'm deferring this for consideration in the 0.18 release.

PR #3321, opened 2023-10-21
Block: Use stream in readTransactions

msgilligan:
* Use IntStream to build the list
* Return an unmodifiable list

This is a child of PR #3301.

(This is a replacement for PR #3197 which I just closed.)


schildbach:
Merged.

PR #3327, opened 2023-11-18
build.gradle: update protobuf-javalite to 3.25.1

msgilligan:


schildbach:
Perhaps also update the protoc dep at the same time?

Also, I wonder what's our current blocker to update 0.17 to the latest protobuf version? I admit I lost track of the reason to stay with 3.22.x.

msgilligan:
> also update the protoc dep

Fixed and force-pushed.

> what's our current blocker to update 0.17 to the latest protobuf version? I admit I lost track of the reason to stay with 3.22.x.

I don't remember if there ever was a reason. At one point we decided that we didn't want to use the latest Gradle Protobuf Plugin because we wanted to minimize the differences from the Debian build. Perhaps that's what you're thinking of?

schildbach:
> I don't remember if there ever was a reason. At one point we decided that we didn't want to use the latest Gradle Protobuf Plugin because we wanted to minimize the differences from the Debian build. Perhaps that's what you're thinking of?

Yes, exactly. Thanks!

msgilligan:
> > I don't remember if there ever was a reason. … we didn't want to use the latest Gradle Protobuf Plugin because we wanted to minimize the differences from the Debian build. Perhaps that's what you're thinking of?
> 
> Yes, exactly. Thanks!

This PR does _not_ change the version of the Gradle Protobuf Plugin (it's set separately in the root `build.gradle`) and I don't think there ever was a reason for not updating the protobuf compile/runtime JAR to the latest version which is what this PR does.

So we should be able to merge this PR. Right?

schildbach:
Should we update this to 3.25.3 before we merge, or even update to 4.26.1?

msgilligan:
My vote would be for the latest versions unless there is some reason otherwise.

schildbach:
Ok, I opened #3381 to update to latest.

schildbach:
Closing in favor of #3381.

PR #3334, opened 2023-12-15
WIP: Add flake.nix for a Nix build

msgilligan:
You should now be able to use:

* `nix develop` -- to get a development shell
* `nix run .#wallettool` to build and run the wallet-tool
* `nix run .#wallettemplate` to build and run the wallettemplate

see README_NIX.adoc for details.

Known issues:

1. The builds are currently using the "escape hatch" method and should be using something like `gradle2nix`.
2. There should be a separate target to install `bitcoinj-core` and possibly for some of the other module and dependency JARs.
3. As a result of the above two issues there is a conflict if you install both applications. It can be resolved with the `--priority` option see the note below.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,106 @@
+{
+  "nodes": {
+    "devshell": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1700815693,
+        "narHash": "sha256-JtKZEQUzosrCwDsLgm+g6aqbP1aseUl1334OShEAS3s=",
+        "owner": "numtide",
+        "repo": "devshell",
+        "rev": "7ad1c417c87e98e56dcef7ecd0e0a2f2e5669d51",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "devshell",
+        "type": "github"
+      }
+    },
+    "flake-parts": {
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -0,0 +1,162 @@
...
+    };
+  };
+
+  outputs = inputs @ {
+    flake-parts,
+    devshell,
+    gitignore,
+    ...
+  }:
+    flake-parts.lib.mkFlake {inherit inputs;} {
```
schildbach, line 33: Sure this isn't one indentation too many? It doesn't cleanly unwind if you look at the end of the file.

flake.lock
```diff
@@ -0,0 +1,106 @@
+{
```
schildbach, line 1: Is putting a "lock" file under version control intentional?

flake.nix
msgilligan: I'm not sure -- I'm still learning the Nix language and any coding style conventions that come with it.

I based the code on this example: https://github.com/brianmcgee/nix-gradle-sample/blob/main/flake.nix and have not changed the indentation. Note that there is one other place in the code with a `}:` and both are indented.

flake.lock
msgilligan: It is intentional. The point is to capture the SHA hashes of the dependencies for reproducibility purposes.

msgilligan:
This has been rebased and also updated to use `nixpkgs-unstable` and the `jdk21` package.

PR #3340, opened 2023-12-18
build.gradle: update Guava to 33.0.0-android

schildbach:


msgilligan (approved):
Looks good!

> Added an Automatic-Module-Name to failureaccess, Guava's one strong runtime dependency. 

That's good news for our someday Java Module conversion.

PR #3343, opened 2024-01-10
Two new service bits

schildbach:
Two commits:

* Services: add NODE_P2P_V2 service bit
* Services: add NODE_COMPACT_FILTERS service bit

The goal is to name these bits in `toString()`, which might be shown in node explorers etc.


schildbach:
Merged.

PR #3346, opened 2024-01-31
gradle.yml, graalvm.yml: gradle/actions/setup-gradle@v3

msgilligan:
Upgrade to gradle/actions/setup-gradle@v3, which claims to be backward-compatible with v2 (with the exception of requiring node 20)

The `arguments` parameter is now
deprecated (it is recommended to "setup" in one step and to use the standard "run" action in a separate step), see:

See: https://github.com/gradle/gradle-build-action/issues/996

However, for this PR we'll just upgrade to the latest version. We'll deal with the deprecation in a separate PR.


schildbach:
Merged.

PR #3351, opened 2024-03-10
LazyECPoint: make private `curve` member static

msgilligan:
We do not need to specify the curve for each instance as bitcoinj uses SECP256K1 curve exclusively.

This change reduces per instance memory usage, simplifies the API, and reduces public API dependence on Bouncy Castle.

One two-arg constructor is deprecated and replaced with a single-arg constructor that no longer requires the curve parameter.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
+++ b/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
@@ -25,20 +25,21 @@ import java.math.BigInteger;
 import java.util.Arrays;
 import java.util.Objects;
 
+import static org.bitcoinj.base.internal.Preconditions.checkArgument;
+
 /**
- * A wrapper around ECPoint that delays decoding of the point for as long as possible. This is useful because point
+ * A wrapper around a SECP256K1 ECPoint that delays decoding of the point for as long as possible. This is useful because point
  * encode/decode in Bouncy Castle is quite slow especially on Dalvik, as it often involves decompression/recompression.
  */
 public class LazyECPoint {
-    // If curve is set, bits is also set. If curve is unset, point is set and bits is unset. Point can be set along
-    // with curve and bits when the cached form has been accessed and thus must have been converted.
+    private static final ECCurve curve = ECKey.CURVE.getCurve();
 
-    private final ECCurve curve;
+    // bits will be null if LazyECPoint is constructed from an (already decoded) point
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
```diff
@@ -33,7 +33,7 @@ public class LazyECPoint {
     // If curve is set, bits is also set. If curve is unset, point is set and bits is unset. Point can be set along
```
schildbach, line 33: I see a couple of now outdated comments here and there, about the curve.

core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
```diff
@@ -42,15 +42,27 @@ public class LazyECPoint {
...
      * Construct a LazyECPoint from a public key. Due to the delayed decoding of the point the validation of the
      * public key is delayed too, e.g. until a getter is called.
      *
      * @param curve a curve the point is on
      * @param bits  public key bytes
+     * @deprecated Use {@link LazyECPoint#LazyECPoint(byte[])}
      */
+    @Deprecated
     public LazyECPoint(ECCurve curve, byte[] bits) {
-        this.curve = curve;
```
schildbach, line 53: I wonder if this can be dangerous for users who specify a different curve than the Bitcoin curve. Not that I expect anyone has ever used this constructor. I would still guard this with an exception.

schildbach (commented):
Almost approved. See inline comments.

core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
msgilligan: Good idea. I highly doubt anyone is using that with a different curve, though.

core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
msgilligan: Resolved and force-pushed.

core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
msgilligan: Resolved and force-pushed.

schildbach:
Merged.

PR #3353, opened 2024-03-11
ECKey: make `CURVE` package private, provide accessor

msgilligan:
* Make the static field ECKey.CURVE (which is a Bouncy Castle type) package private and provide the ecDomainParameters() accessor for those who really need it.

* Use the accessor in `DefaultRiskAnalysisTest`

* We will not update code within the `o.b.crytpo` package to use the accessor, since we intend to migrate all/most of those usages to the `java.security` types.


schildbach:
Merged.

PR #3354, opened 2024-03-11
WIP: Migration for Bouncy Castle types to `java.security` types.

msgilligan:


schildbach:
Looks fine in general. Can we progress this to the point where we can delete the `BC_CURVE` field?

msgilligan:
> Looks fine in general. Can we progress this to the point where we can delete the `BC_CURVE` field?

I think we can, but we might have to create a secp256k1-jdk wrapper for it, so we can get things like point types in the correct format for _either_ secp256k1-ffm or secp256k1-bouncy. I think this will allow us to minimize format conversion on point types.

msgilligan:
Most of this has already been implemented differently/better. I'm not ready to close it quite yet, but almost...

PR #3355, opened 2024-03-11
ECKey: Make HALF_CURVE_ORDER package private

msgilligan:
This isn't a Bouncy Castle type, but it isn't currently being used publicly and can and should be made package private.

If anyone needs the constant, they can still get it via `ECDomainParameters` (BC) or `ECParameterSpec` (`java.security`)


schildbach:
Merged.

PR #3356, opened 2024-03-15
modular-architecture.md: Update for secp256k1-jdk

msgilligan:
This just updates the proposed 0.19 release to use `secp256k1-jdk` instead of a more generically defined crypto provider API.

We may want to consider doing this in the 0.18 release or moving some ECKey definitions into `o.b.base` but this change does not address either of those proposals.


schildbach (approved):

schildbach:
Merged.

PR #3360, opened 2024-04-06
updated link and version to access jitpack dependency for 0.16.2

ItoroD:
I noticed the version for the jitpack dependency was still pointing to 0.15 

- I have updated the link and the version number

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -92,7 +92,7 @@ Now you are ready to https://bitcoinj.github.io/getting-started[follow the tutor
 
 Building apps with official releases of *bitcoinj* is covered in the https://bitcoinj.github.io/getting-started[tutorial].
 
-If you want to develop or test your app with a https://jitpack.io[Jitpack]-powered build of the latest `master` or `release-0.15` branch of *bitcoinj* follow the dynamically-generated instructions for that branch by following the correct link.
+If you want to develop or test your app with a https://jitpack.io[Jitpack]-powered build of the latest `master` or `release-0.16` branch of *bitcoinj* follow the dynamically-generated instructions for that branch by following the correct link.
 
 * https://jitpack.io/#bitcoinj/bitcoinj/master-SNAPSHOT[master] branch
-* https://jitpack.io/#bitcoinj/bitcoinj/release-0.15-SNAPSHOT[release-0.15] branch
+* https://jitpack.io/#bitcoinj/bitcoinj/release-0.16-SNAPSHOT[release-0.16] branch


README.adoc
```diff
@@ -92,7 +92,7 @@ Now you are ready to https://bitcoinj.github.io/getting-started[follow the tutor
 
 Building apps with official releases of *bitcoinj* is covered in the https://bitcoinj.github.io/getting-started[tutorial].
 
-If you want to develop or test your app with a https://jitpack.io[Jitpack]-powered build of the latest `master` or `release-0.15` branch of *bitcoinj* follow the dynamically-generated instructions for that branch by following the correct link.
+If you want to develop or test your app with a https://jitpack.io[Jitpack]-powered build of the latest `master` or `v0.16.2` branch of *bitcoinj* follow the dynamically-generated instructions for that branch by following the correct link.
```
msgilligan, line 95: You should use the `release-0.16` _branch_, not the `v0.16.2` _tag_ because that branch will always have the latest code from that branch. There is little value in using Jitpack to reference the 0.16.2 tag because that will always be the same as the 0.16.2 official release.

At the current time the branch and the tag are the same, but if we fix a bug on that branch or start work on a 0.16.3 release that branch and the tag will differ.

README.adoc
```diff
@@ -92,7 +92,7 @@ Now you are ready to https://bitcoinj.github.io/getting-started[follow the tutor
...
 
 * https://jitpack.io/#bitcoinj/bitcoinj/master-SNAPSHOT[master] branch
-* https://jitpack.io/#bitcoinj/bitcoinj/release-0.15-SNAPSHOT[release-0.15] branch
+* https://jitpack.io/#bitcoinj/bitcoinj/v0.16.2[v0.16.2] branch
```
msgilligan, line 98: This should also be `release-0.16` see above.

msgilligan (changes requested):
Thanks for the PR. It should be updated to use the _branch_ and not the _tag_, though.

README.adoc
ItoroD: Makes sense. But I do not see a build in jitpack for release-0.16. Also tried pulling into my project but the artifact is not found as expected. I do see release-0.16-SNAPSHOT and that is getting pulled.

msgilligan: Right. It needs to have `-SNAPSHOT` for jitpack to work correctly.

msgilligan (approved):
Looks great! Thank you!

schildbach:
Merged.

PR #3365, opened 2024-04-21
README: Use double-dash to fix the label "bitcoinj-users" in the Matrix badge

msgilligan:
We were incorrectly (and mostly unnoticed) displaying a space where we should be displaying a single-dash.

See: https://shields.io/badges


schildbach:
Merged.

PR #3368, opened 2024-04-23
ForwardingService: let main() throw InterruptedException

msgilligan:
This simplifies main() and increases readability. And it is fine to terminate the process if the thread is interrupted.


schildbach:
Merged.

PR #3371, opened 2024-04-23
ForwardingService: print network, not network.id()

msgilligan:
Both the code and the output is simpler this way.


schildbach:
Merged.

PR #3373, opened 2024-04-23
ForwardingService: Move CoinSelector inline in forward(), remove unnecessary null-check in forwardingCoinSelector

msgilligan:
Two commits:

* Remove unnecessary null-check in forwardingCoinSelector
* Move CoinSelector to be inline in forward()


schildbach:
Merged.

PR #3372, opened 2024-04-23
BitcoinNetwork.strings() lists all networks as strings, use it to create usage message in ForwardingService

msgilligan:
This PR contains two commits:

* Add `strings()` to `BitcoinNetwork` (also internal stream() method to avoid duplication)
* Use it in `ForwardingService`

The `strings()` method should also make it easier to add a network parameter to other examples.

PR #3375, opened 2024-04-23
ForwardingService: Use BitcoinNetwork.strings() to build USAGE string

msgilligan:
Use BitcoinNetwork.strings() for the list of valid BitcoinNetwork values included in the USAGE string.

This is a follow-up to PR #3372.


schildbach:
Merged.

PR #3378, opened 2024-04-26
WIP: ForwardingService refactoring for JDK 11 with record-like `Config` class.

msgilligan:
This is WIP for an improved JDK 11 `ForwardingService`


core/src/main/java/org/bitcoinj/base/BitcoinNetwork.java
```diff
@@ -16,6 +16,9 @@
 
 package org.bitcoinj.base;
 
+import org.bitcoinj.base.exceptions.AddressFormatException;
+import org.checkerframework.checker.units.qual.A;
+
```
msgilligan, line 21: These shouldn't be here.

integration-test/build.gradle
```diff
@@ -18,13 +18,14 @@ dependencies {
     testRuntimeOnly "org.junit.vintage:junit-vintage-engine:5.10.0"
 }
 
-sourceCompatibility = 11
+sourceCompatibility = 17
```
msgilligan, line 21: This can go back to 11.

integration-test/build.gradle
```diff
@@ -18,13 +18,14 @@ dependencies {
...
 compileJava.options.encoding = 'UTF-8'
 compileTestJava.options.encoding = 'UTF-8'
 javadoc.options.encoding = 'UTF-8'
 
 compileJava {
     options.compilerArgs << '-Xlint:deprecation'
+    options.compilerArgs.addAll(['--release', '17'])
```
msgilligan, line 28: Same here.

integration-test/src/test/java/org/bitcoinj/examples/ForwardingServiceTest.java
```diff
@@ -36,18 +36,24 @@
...
     static final BitcoinNetwork network = BitcoinNetwork.TESTNET;
     static final Address forwardingAddress = new ECKey().toAddress(ScriptType.P2WPKH, network);
-    static final String[] args = new String[] { forwardingAddress.toString(), network.toString() };
+    ForwardingService.Config config;
 
     @BeforeEach
-    void setupTest() {
+    void setupTest(@TempDir File tempDir) {
         Context.propagate(new Context());
+        config = new ForwardingService.Config(network, forwardingAddress);
```
msgilligan, line 44: Wow. This change was done automatically by IntelliJ, I think.

settings.gradle
```diff
@@ -5,8 +5,10 @@ import org.gradle.api.GradleScriptException
 def minGradleVersion = GradleVersion.version("4.4")
 // Minimum Gradle version for JUnit5
 def minJunit5GradleVersion = GradleVersion.version("4.6")
-// Minimum Gradle version for builds of JavaFX 17 module using JDK 17
+// Minimum Gradle version for builds of JavaFX module using JDK 17
 def minFxGradleVersion = GradleVersion.version("7.3")
+// Minimum Gradle version for builds of Java 17 module
+def minJava17GradleVersion = GradleVersion.version("7.0")
```
msgilligan, line 11: All changes in this file can be reverted.

msgilligan (commented):
I can't review my own PR, but I found some things that need changing.

examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
```diff
@@ -120,40 +120,37 @@ void coinForwardingListener(Wallet wallet, Transaction incomingTx, Coin prevBala
...
                     System.out.printf("Transaction %s is signed and is being delivered to %s...\n", broadcast.transaction().getTxId(), wallet.network());
                     return broadcast.awaitRelayed(); // Wait until peers report they have seen the transaction
-                });
+            })
+            .whenComplete((broadcast, throwable) -> {
+                if (broadcast != null) {
+                    System.out.printf("Sent %s onwards and acknowledged by peers, via transaction %s\n",
+                            broadcast.transaction().getOutputSum().toFriendlyString(),
+                            broadcast.transaction().getTxId());
+                } else {
```
Itorouk: Maybe we should check if throwable is null. Instead of just an else before we print throwable

```
if (throwable != null) {
                System.out.println("Exception occurred: "  + throwable);
 }
```

ItoroD: Maybe we want to check if throwable is null before we print. (using if instead of else)

```
  if (throwable != null) {
                System.out.println("Exception occurred: "  + throwable);
            }
```

ItoroD: on second thought, you can ignore my comment. When I did this in kotlin I got a null pointer exception. I think this was down to me not making the throwable parameter nullable like so

`.whenComplete { broadcast: TransactionBroadcast?, throwable: Throwable? ->`

PR #3345, opened 2024-01-31
gradle.yml: add `macos-14` (aarch64) to the build Matrix.

msgilligan:
`macOS-14` will run on a macOS "Apple Silicon"/M1 Macintosh.

PR #3379, opened 2024-04-26
gradle.yml, graalvm.yml: explicitly use macos-12

msgilligan:
GitHub has bumped macos-latest to be macos-14 (and ARM) so our tests are currently broken. This PR is a short-term fix to restore the status quo and makes the minimum change to fix the broken build.

See https://github.com/bitcoinj/bitcoinj/pull/3345 for more discussion.

### DIFF
--- a/.github/workflows/graalvm.yml
+++ b/.github/workflows/graalvm.yml
@@ -8,7 +8,7 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


schildbach:
Merged.

Should we perhaps do this for `ubuntu-latest` and `windows-latest`, too?

msgilligan:
> Should we perhaps do this for `ubuntu-latest` and `windows-latest`, too?

Yes. I'll make a PR to do it.

PR #3380, opened 2024-04-27
gradle.yml, graalvm.yml: explicitly use ubuntu-22.04, windows-2022

msgilligan:
This makes explicit which versions of which OSs are used in CI. After this is merged, we will need to manually update to newer OS versions.


schildbach:
Merged.

PR #3382, opened 2024-04-27
gradle.yml: Use ubuntu-22.04 for gradle 4.10.3 special case

msgilligan:
I missed this entry in today's earlier (already merged) PR.


schildbach:
Merged.

PR #3383, opened 2024-04-27
gradle.yml, graalvm.yml: fix Gradle Action deprecation warning

msgilligan:
Remove `arguments` parameter from `setup-gradle` and add add a basic `run` step with the actual `gradle` command.

This fixes a deprecation warning.


schildbach:
Merged.

PR #3386, opened 2024-04-28
examples: remove JavaScript examples

msgilligan:
These are old, unsupported, and almost certainly do not work as is.

We separately plan on replacing the JavaScript example on the documentation site with a Kotlin example.


schildbach:
Merged.

PR #3387, opened 2024-04-28
examples: remove forwarding.py

msgilligan:
This example is 8 years old, unsupported, and almost certainly doesn't work.


schildbach:
Merged.

PR #3392, opened 2024-04-29
ForwardingService: Use ternary operator to set `network`

msgilligan:


examples/src/main/java/org/bitcoinj/examples/ForwardingService.java
```diff
@@ -81,14 +81,11 @@ public static void main(String[] args) throws InterruptedException {
...
-            network.checkAddress(forwardingAddress);
-        } else {
-            // Else network not-specified, extract network from address
-            network = (BitcoinNetwork) forwardingAddress.network();
-        }
+        // If network not provided in args[1], extract network from forwardingAddress
+        network = (args.length >= 2)
+                ? BitcoinNetwork.fromString(args[1]).orElseThrow()
+                : (BitcoinNetwork) forwardingAddress.network();
+        network.checkAddress(forwardingAddress);    // Make sure forwardingAddress is valid for network
```
schildbach, line 88: I think it would be cleaner to just run `checkAddress` if the network is specified. Otherwise we're just checking our own logic.

msgilligan: It doesn't hurt to check it twice (in the case of 1 argument) and I think the overall simplification of the code is worth it.

msgilligan: OK, I see your point. I also think this issue becomes moot after further refactoring.

msgilligan:
Closing after feedback from Andreas.

PR #3396, opened 2024-05-03
Change Key.Type enum from ENCRYPTED_SCRYPT_AES to ENCRYPTED

oliveraemmer:
The protobuf Key.Type enum ENRCRYPTED_SCRYPT_AES can be simplified to ENCRYPTED since the encryption type can already be defined with Wallet.EncryptionType.

This is a separate PR for a small change from a different PR: https://github.com/bitcoinj/bitcoinj/pull/3362 

Mentioned in this comment: https://github.com/bitcoinj/bitcoinj/pull/3362#discussion_r1565746921

### DIFF
--- a/core/src/main/proto/wallet.proto
+++ b/core/src/main/proto/wallet.proto
@@ -86,11 +86,18 @@ message DeterministicKey {
  */
 message Key {
   enum Type {
+    option allow_alias = true;
     /** Unencrypted - Original bitcoin secp256k1 curve */
     ORIGINAL = 1;
 
-    /** Encrypted with Scrypt and AES - Original bitcoin secp256k1 curve */
+    /**
+    * Encrypted - Original bitcoin secp256k1 curve
+    * Details of the encryption are declared via Wallet.EncryptionType enum
+    */
     ENCRYPTED = 2;
+
+    /** Encrypted with Scrypt and AES - Original bitcoin secp256k1 curve */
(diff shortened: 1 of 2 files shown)


schildbach:
I think this change is reasonable. And it's low-impact so I'll add it to 0.17.

I only wonder if it is possible to define a deprecated alias for one bitcoinj version cycle. I know protobuf has `[deprecated = true]` field option which generates `@Deprecated` annotations. However is it possible to declare two enum names to the same code and declare one of them deprecated?

schildbach:
> However is it possible to declare two enum names to the same code and declare one of them deprecated?

Answering my own question: It's possible via `option allow_alias = true;` as an enum option. I just tested this protobuf and it passed tests:

```
  enum Type {
    option allow_alias = true;
    […]
    ENCRYPTED = 2;
    ENCRYPTED_SCRYPT_AES = 2 [deprecated = true];
    […]
  }
```

core/src/main/proto/wallet.proto
```diff
@@ -90,7 +90,7 @@ message Key {
     ORIGINAL = 1;
 
     /** Encrypted with Scrypt and AES - Original bitcoin secp256k1 curve */
```
schildbach, line 92: I think this comment should be changed too, like you did in your other PR. And maybe mention that the details of the encryption are declared via `Wallet.EncryptionType`.

schildbach:
Merged.

PR #3400, opened 2024-05-18
Simplified and improved kotlin forwarding service

ItoroD:
- One main difference to note in the kotlin version of forwarding service is that in kotlin its not the best best practice to use multiple constructors (secondary constructors). Hence default values have been used for some member variables of config class.
- With this we can initialize config as if it had multiple constructors

### DIFF
--- a/examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
+++ b/examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
@@ -91,6 +91,7 @@ class ForwardingService(val config: Config) : Closeable {
      * @param prevBalance wallet balance before this transaction (unused)
      * @param newBalance wallet balance after this transaction (unused)
      */
+    @Suppress("UNUSED_PARAMETER")
     private fun coinForwardingListener(wallet: Wallet, incomingTx: Transaction, prevBalance: Coin, newBalance: Coin) {
         // Incoming transaction received, now "compose" (i.e. chain) a call to wait for required confirmations
         // The transaction "incomingTx" can either be pending, or included into a block (we didn't see the broadcast).
@@ -130,7 +131,7 @@ class ForwardingService(val config: Config) : Closeable {
      * @param sendRequest requested transaction parameters
      * @return A future for a TransactionBroadcast object that completes when relay is acknowledged by peers
      */
-    open fun send(wallet: Wallet, sendRequest: SendRequest?): CompletableFuture<TransactionBroadcast?>? {
+    fun send(wallet: Wallet, sendRequest: SendRequest?): CompletableFuture<TransactionBroadcast?>? {
         return wallet.sendTransaction(sendRequest)
             .thenCompose { broadcast: TransactionBroadcast ->
                 println("Transaction ${broadcast.transaction().txId} is signed and is being delivered to ${wallet.network()}...\n")
--- a/examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
+++ b/examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
@@ -14,8 +14,14 @@
  * limitations under the License.
  */
 
-import org.bitcoinj.base.*
-import org.bitcoinj.core.*
+import org.bitcoinj.base.BitcoinNetwork
+import org.bitcoinj.base.Address
+import org.bitcoinj.base.Coin
+import org.bitcoinj.base.AddressParser
+import org.bitcoinj.core.Context
+import org.bitcoinj.core.Transaction
+import org.bitcoinj.core.TransactionBroadcast
+import org.bitcoinj.core.TransactionConfidence
 import org.bitcoinj.kits.WalletAppKit
 import org.bitcoinj.utils.BriefLogFormatter
 import org.bitcoinj.wallet.CoinSelector
@@ -24,49 +30,38 @@ import org.bitcoinj.wallet.Wallet
 import org.bitcoinj.wallet.listeners.WalletCoinsReceivedEventListener
 import java.io.Closeable
 import java.io.File
-import java.util.Objects.*
+import java.util.Objects
 import java.util.concurrent.CompletableFuture
 
-data class Config(
-    val network: BitcoinNetwork,              // Network to operate on
-    val forwardingAddress: Address,           // Address to forward to
-    val walletDirectory: File,                // Directory to create wallet files in
-    val walletPrefix: String,                 // Prefix for wallet file names
-    val requiredConfirmations: Int,           // Required number of tx confirmations before forwarding
-    val maxConnections: Int                   // Maximum number of connections
-)
 
+val NETS = BitcoinNetwork.strings().joinToString(separator = "|")
+val USAGE = "Usage: address-to-forward-to $NETS"
 fun main(args: Array<String>) { //pass the network and forwarding address in this order [address, network]
     // This line makes the log output more compact and easily read, especially when using the JDK log adapter.
     BriefLogFormatter.init()
     Context.propagate(Context())
 
-    val addressAndNetwork = parseArgs(args)
-    val network: BitcoinNetwork = addressAndNetwork[1] as BitcoinNetwork
-    val forwardingAddress: Address = addressAndNetwork[0] as Address
+    if (args.isEmpty() || args.size > 2) {
+        System.err.println(USAGE)
+        System.exit(1)
+    }
 
-    val config = Config(network, forwardingAddress, File("."), getPrefix(network), 1, 4)
+    // If only an address is provided, derive network from the address
+    // If address and network provided, use network and validate address against network
+    val forwardingAddress: Address = AddressParser.getDefault().parseAddress(args[0])
+    val config = if (args.size == 1) Config(forwardingAddress) else
+        Config(forwardingAddress, BitcoinNetwork.fromString(args[1]).orElseThrow())
+    println("Network: ${config.network}")
+    println("Forwarding address: ${config.forwardingAddress}")
 
-    println("Network: ${config.network}");
-    println("Forwarding address: ${config.forwardingAddress}");
+    // Create the service, which will listen for transactions and forward coins until closed
+    ForwardingService(config).use { forwardingService ->
+        // After we start listening, we can tell the user the receiving address
+        println(forwardingService.status())
+        println("Press Ctrl-C to quit.")
 
-    // Create and start the WalletKit
-    val walletAppKit: WalletAppKit =
-        WalletAppKit.launch(network, config.walletDirectory, getPrefix(network), config.maxConnections)
-    val forwardingService = ForwardingService(config, walletAppKit.wallet())
-
-    walletAppKit.use { walletAppKit ->
-        forwardingService.use { forwardingService ->
-            // After we start listening, we can tell the user the receiving address
-            println("Waiting to receive coins on: ${walletAppKit.wallet().currentReceiveAddress()}")
-            println("Press Ctrl-C to quit.")
-
-            // Wait for Control-C
-            try {
-                Thread.sleep(Long.MAX_VALUE)
-            } catch (ignored: InterruptedException) {
-            }
-        }
+        // Wait for Control-C
+        Thread.sleep(Long.MAX_VALUE)
     }
 }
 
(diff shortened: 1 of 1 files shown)


examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
```diff
@@ -114,72 +123,55 @@ class ForwardingService(val config: Config, val wallet: Wallet) : Closeable {
...
+data class Config(
+    val forwardingAddress: Address,                                              // Address to forward to
+    val network: BitcoinNetwork = forwardingAddress.network() as BitcoinNetwork, // Network to operate on
+    val walletDirectory: File = File("."),                              // Directory to create wallet files in
+    val walletPrefix: String = getPrefix(network),                               // Prefix for wallet file names
+    val requiredConfirmations: Int = 1,                                          // Required number of tx confirmations before forwarding
+    val maxConnections: Int = 4                                                  // Maximum number of connections
+) {
+    init {
+        network.checkAddress(forwardingAddress)
```
msgilligan, line 169: It's nice how the Kotlin `data class` can be built with default values and not require extra constructors, but doing this validation here seems a little weird, for two reasons:
1) It seems like the check is run after the member fields are assigned.
2) It appears to run `checkAddress` redundantly in the case where `network` is read from `forwardingAddress`

I'm not sure what the alternative is here, but maybe there is a way to address these two (relatively minor) issues?

msgilligan (changes requested):
Overall, this looks great!

I am being exceptionally picky about the `ForwardingService` because for many it will be their first impression of bitcoinj (and maybe Kotlin or modern Java) and also because if we try really hard and can't eliminate any unwieldy code then we need to look at improving the bitcoinj API itself to fix them.

I really appreciate your work and your willingness to keep fine-tuning this!

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
```diff
@@ -24,63 +24,59 @@ import org.bitcoinj.wallet.Wallet
...
-data class Config(
-    val network: BitcoinNetwork,              // Network to operate on
-    val forwardingAddress: Address,           // Address to forward to
-    val walletDirectory: File,                // Directory to create wallet files in
-    val walletPrefix: String,                 // Prefix for wallet file names
-    val requiredConfirmations: Int,           // Required number of tx confirmations before forwarding
-    val maxConnections: Int                   // Maximum number of connections
-)
 
+private val NETS = BitcoinNetwork.strings().joinToString(separator = "|")
```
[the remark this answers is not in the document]
ItoroD: I agree

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
```diff
@@ -102,7 +98,20 @@ class ForwardingService(val config: Config, val wallet: Wallet) : Closeable {
...
+                // Use a CoinSelector that only returns wallet UTXOs from the incoming transaction.
+                sendRequest.coinSelector = CoinSelector.fromPredicate { output: TransactionOutput ->
+                    Objects.equals(
+                        output.parentTransactionHash,
+                        incomingTx.txId
+                    )
+                }
+                send(wallet, sendRequest)
+                //forward(wallet, incomingTx, config.forwardingAddress)
+                //return send(wallet,sendRequest)
```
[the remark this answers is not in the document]
ItoroD: yes, will remove

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
```diff
@@ -24,63 +24,59 @@ import org.bitcoinj.wallet.Wallet
...
-                Thread.sleep(Long.MAX_VALUE)
-            } catch (ignored: InterruptedException) {
-            }
-        }
+    // If only an address is provided, derive network from the address
+    // If address and network provided, use network and validate address against network
+    val forwardingAddress: Address = AddressParser.getDefault().parseAddress(args[0])
+    val config = if (args.size == 1) Config(forwardingAddress) else Config(
+        forwardingAddress, BitcoinNetwork.fromString(args[1]).orElseThrow()
+    )
```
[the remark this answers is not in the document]
ItoroD: will do

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
```diff
@@ -102,7 +98,20 @@ class ForwardingService(val config: Config, val wallet: Wallet) : Closeable {
...
+                    Objects.equals(
+                        output.parentTransactionHash,
+                        incomingTx.txId
+                    )
+                }
+                send(wallet, sendRequest)
+                //forward(wallet, incomingTx, config.forwardingAddress)
+                //return send(wallet,sendRequest)
             }
             .whenComplete { broadcast: TransactionBroadcast?, throwable: Throwable ->
```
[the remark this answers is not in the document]
ItoroD: yes, will update

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
```diff
@@ -114,72 +123,55 @@ class ForwardingService(val config: Config, val wallet: Wallet) : Closeable {
...
         return wallet.sendTransaction(sendRequest)
             .thenCompose { broadcast: TransactionBroadcast ->
                 println("Transaction ${broadcast.transaction().txId} is signed and is being delivered to ${wallet.network()}...\n")
                 broadcast.awaitRelayed() // Wait until peers report they have seen the transaction
             }
+            .whenComplete { broadcast: TransactionBroadcast?, throwable: Throwable? ->
+                if (broadcast != null) {
+                    println("Sent ${broadcast.transaction().outputSum.toFriendlyString()} onwards and acknowledged by peers, via transaction ${broadcast.transaction().txId}\n")
+                }
+                if (throwable != null) {
```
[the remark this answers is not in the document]
ItoroD: Its fine to use else here. Should have switched that back to else after I made it nullable

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
ItoroD: I see your point. In this case secondary constructor route is the way to go

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
```diff
@@ -24,63 +24,58 @@ import org.bitcoinj.wallet.Wallet
 import org.bitcoinj.wallet.listeners.WalletCoinsReceivedEventListener
 import java.io.Closeable
 import java.io.File
-import java.util.Objects.*
+import java.util.*
```
msgilligan, line 27: We generally try to avoid using wildcard imports...

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
```diff
@@ -114,72 +120,61 @@ class ForwardingService(val config: Config, val wallet: Wallet) : Closeable {
...
+    val network: BitcoinNetwork = forwardingAddress.network() as BitcoinNetwork, // Network to operate on
+    val walletDirectory: File,                                                   // Directory to create wallet files in
+    val walletPrefix: String = getPrefix(network),                               // Prefix for wallet file names
+    val requiredConfirmations: Int = 1,                                          // Required number of tx confirmations before forwarding
+    val maxConnections: Int = 4                                                  // Maximum number of connections
+) {
+    constructor(forwardingAddress: Address) : this(
+        forwardingAddress,
+        forwardingAddress.network() as BitcoinNetwork
+    )
```
msgilligan, line 168: Is this constructor necessary?

msgilligan (changes requested):
Getting close! This is looking good!

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
ItoroD: Yes, I noticed  you do the address check in a second constructor that has (address, network) as parameters. So this constructor for when only address is passed, calls the other constructor that checks address before address is set on member variables. 

since forwardingAddress.network() as BitcoinNetwork is now being done in the constructor, i should also take it out of the default set value for network

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
ItoroD: will look out those

ItoroD:
> Overall, this looks great!
> 
> I am being exceptionally picky about the `ForwardingService` because for many it will be their first impression of bitcoinj (and maybe Kotlin or modern Java) and also because if we try really hard and can't eliminate any unwieldy code then we need to look at improving the bitcoinj API itself to fix them.
> 
> I really appreciate your work and your willingness to keep fine-tuning this!

Thanks! @msgilligan

msgilligan (approved):
Approved.

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
```diff
@@ -114,72 +125,61 @@ class ForwardingService(val config: Config, val wallet: Wallet) : Closeable {
...
-    fun forward(
-        wallet: Wallet,
-        incomingTx: Transaction,
-        forwardingAddress: Address?
-    ): CompletableFuture<TransactionBroadcast?> {
-        // Send coins received in incomingTx onwards by sending exactly the outputs that have been sent to us
-        val sendRequest = SendRequest.emptyWallet(forwardingAddress)
-        sendRequest.coinSelector = forwardingCoinSelector(incomingTx.txId)
-        println("Creating outgoing transaction for $forwardingAddress...\n")
+    open fun send(wallet: Wallet, sendRequest: SendRequest?): CompletableFuture<TransactionBroadcast?>? {
```
schildbach, line 133: I get this compiler warning:

> 'open' has no effect in a final class

Is it intended? Shall we remove the "open"?

schildbach (approved):
I wonder can we do anything about these warnings as well?

> w: file:///bitcoinj/examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt:94:81 Parameter 'prevBalance' is never used
> w: file:///bitcoinj/examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt:94:100 Parameter 'newBalance' is never used

Obviously we can't change the method signature, but is it possible to declare the parameters as "intentionally not being used"?

schildbach:
One nitpick about the commit message: we use present tense to describe the semantics of the commit, rather than past tense. So "simplify coin listener" rather than "simplified…".

examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt
ItoroD: yes I can take "open" out. open allows classes and functions to be inherited or overridden. This is not necessary here

ItoroD:
> One nitpick about the commit message: we use present tense to describe the semantics of the commit, rather than past tense. So "simplify coin listener" rather than "simplified…".

Noted. Thanks

ItoroD:
> I wonder can we do anything about these warnings as well?
> 
> > w: file:///bitcoinj/examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt:94:81 Parameter 'prevBalance' is never used
> > w: file:///bitcoinj/examples-kotlin/src/main/kotlin/org/bitcoinj/ForwardingService.kt:94:100 Parameter 'newBalance' is never used
> 
> Obviously we can't change the method signature, but is it possible to declare the parameters as "intentionally not being used"?

We can use the "Suppress" annotation like so

```
@Suppress("UNUSED_PARAMETER")
private fun coinForwardingListener(wallet: Wallet, incomingTx: Transaction, prevBalance: Coin, newBalance: Coin)
```

msgilligan:
> Obviously we can't change the method signature, but is it possible to declare the parameters as "intentionally not being used"?

Can't they be replaced with `_` or prefixed with `_`?

Note that this convention is now in Java 22 and later: https://openjdk.org/jeps/456

ItoroD:
> > Obviously we can't change the method signature, but is it possible to declare the parameters as "intentionally not being used"?
> 
> Can't they be replaced with `_` or prefixed with `_`?
> 
> Note that this convention is now in Java 22 and later: https://openjdk.org/jeps/456

Yes the underscore exists in kotlin but it cannot be used on parameter of a function. Its used in special cases like parameter of a lamda function. 

Also in java 22 I doubt you can use the underscore in a method param. Its used in other cases in java 22 as well. (not in method params) to the best of my knowledge

schildbach:
@msgilligan Are you fine with the added `@Suppress`? Or should we do without and live with the warnings?

msgilligan:
> @msgilligan Are you fine with the added `@Suppress`? Or should we do without and live with the warnings?

I'm fine with it for now. I spent some time looking at alternatives and while doing so discovered that the neither the Kotlin nor the Java example correctly removes the listener. The problem is that when creating the method references now objects are created. So the listener passed in to remove does not match the one stored in the list...

schildbach:
Merged.

PR #3401, opened 2024-05-26
Added code showing some features that the wallet kit has

ItoroD:
Not much changes to be made here. Almost identical to java kit sample. Small differences: 

- main is a top level method 
- trailing lambda function 

I wonder if we want to use Config class for some of the config variables like in forwarding service. (It might be overkill for this example though).

### DIFF
--- a/examples-kotlin/src/main/kotlin/org/bitcoinj/Kit.kt
+++ b/examples-kotlin/src/main/kotlin/org/bitcoinj/Kit.kt
@@ -1,0 +1,70 @@
+package org.bitcoinj
+
+import org.bitcoinj.base.BitcoinNetwork
+import org.bitcoinj.base.Coin
+import org.bitcoinj.core.Transaction
+import org.bitcoinj.crypto.ECKey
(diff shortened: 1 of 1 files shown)


msgilligan (commented):
This looks good. But looking at it and comparing it to the Java version makes me aware of the deficiencies of the Java version that are now replicated here.

I'm wondering what we can do to improve them both.

ItoroD:
> This looks good. But looking at it and comparing it to the Java version makes me aware of the deficiencies of the Java version that are now replicated here.
> 
> I'm wondering what we can do to improve them both.

Thanks.

I expected this : D . I am open to your improvement suggestions (java and kotlin). I have looked at the issue link you shared. I will sleep on this and give my suggestions tomorrow.

ItoroD:
@msgilligan  Since we plan on working on wallet interface and not urgently, what do you suggest the way forward for this PR and other potential Kotlin examples? Going head with examples will likely mean a change later when we finally get wallet interface out. Do we want to still add the samples the way they are for now? (I mean using the current walletAppKit)

PR #3407, opened 2024-06-01
ForwardingService.java: remove listener properly

msgilligan:
The listener wasn't being removed properly, because using a method reference in both add and remove generated two different objects and therefore the remove method wasn't finding and removing the listener that was previously added.

By saving the listener created by the method reference, we can properly remove it later.


schildbach:
Merged.

Issue #3410, opened 2024-06-23
java.lang.IllegalStateException: Ran out of space to store chain work!


snex:
This is happening when trying to add to SPV chain file (from https://github.com/bisq-network/bisq/)

Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #033[0;39m#033[31mJun-23 08:44:26.728 [NioClientManager] WARN  org.bitcoinj.core.PeerSocketHandler: [127.0.0.1]:8333 -  java.lang.IllegalStateException: Ran out of space to store chain work!
Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #011at com.google.common.base.Preconditions.checkState(Preconditions.java:510)
Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #011at org.bitcoinj.core.StoredBlock.serializeCompact(StoredBlock.java:121)
Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #011at org.bitcoinj.store.SPVBlockStore.put(SPVBlockStore.java:201)
Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #011at org.bitcoinj.core.BlockChain.addToBlockStore(BlockChain.java:93)
Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #011at org.bitcoinj.core.AbstractBlockChain.connectBlock(AbstractBlockChain.java:543)
Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #011at org.bitcoinj.core.AbstractBlockChain.add(AbstractBlockChain.java:469)
Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #011at org.bitcoinj.core.AbstractBlockChain.add(AbstractBlockChain.java:361)
Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #011at org.bitcoinj.core.BlockChain.add(BlockChain.java:170)
Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #011at org.bitcoinj.core.Peer.endFilteredBlock(Peer.java:1072)
Jun 23 08:44:26 bisq bisq-Bisq.desktop[263666]: #011at org.bitcoinj.core.Peer.processMessage(Peer.java:454)
[...]

PR #3411, opened 2024-06-23
StoredBlock: use ByteUtils.bigIntegerToBytes to convert chainWork

msgilligan:
This is a fix for Issue #3410 that will allow main net blocks above 849,137 to be processed. As of block 849,138 we can no longer fit total chainwork in a 12-byte **signed** field. This fix "kicks the can down the road" by making the field 12-bytes **unsigned**.

We should open a new issue to address the long term need for bigger values. Note that converting the field to 12-byte unsigned precludes us from using the most-significant bit as a flag for a new format, but we should be able to pick some arbitrary value, say 0xA0 as a version flag and declare that values less than 0xA0 are "unversioned".

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/StoredBlock.java
+++ b/core/src/main/java/org/bitcoinj/core/StoredBlock.java
@@ -126,9 +126,7 @@ public class StoredBlock {
 
     /** Serializes the stored block to a custom packed format. Used by {@link CheckpointManager}. */
     public void serializeCompact(ByteBuffer buffer) {
-        byte[] chainWorkBytes = getChainWork().toByteArray();
-        checkState(chainWorkBytes.length <= CHAIN_WORK_BYTES, () ->
-                "ran out of space to store chain work!");
+        byte[] chainWorkBytes = ByteUtils.bigIntegerToBytes(getChainWork(), CHAIN_WORK_BYTES);
         if (chainWorkBytes.length < CHAIN_WORK_BYTES) {
             // Pad to the right size.
             buffer.put(EMPTY_BYTES, 0, CHAIN_WORK_BYTES - chainWorkBytes.length);


schildbach (commented):
Fix looks good!

Probably doesn't matter, but this will change a thrown `IllegalStateException` to `IllegalArgumentException`.

Thanks for the quick reaction!

msgilligan:
I did some quick testing using `WalletTemplate`. Without the fix you can see the old exception in the debug log, the wallet keeps running, but never syncs. When I apply the fix and re-run `WalletTemplate` using the same chain state file it recovers and syncs properly.

Note also that I opened Issue #3415 because we'll need more than 12 bytes to represent chainWork at some point. (Probably months or maybe 1-2 years?)

schildbach:
I just verified (on master) that `org.bitcoinj.tools.BuildCheckpoints` still generates checkpoints after applying this patch. However the last checkpoint on mainnet is at height 846,720 – that's still a block with low enough chain work to not trigger the bug.

aeidelman:
Thanks, @msgilligan, for addressing this quickly. We are following this closely from Rootstock, as this has [impacted some core services in the network](https://blog.rootstock.io/noticia/incident-report-rootstock-peg-in-peg-out-service-outage-on-june-24th/). Happy to collaborate from our team in regards to a long-term solution if that's useful.

msgilligan:
> Thanks, @msgilligan, for addressing this quickly. We are following this closely from Rootstock, as this has [impacted some core services in the network](https://blog.rootstock.io/noticia/incident-report-rootstock-peg-in-peg-out-service-outage-on-june-24th/). Happy to collaborate from our team in regards to a long-term solution if that's useful.

You're welcome, thanks for the note! The short term fix (serializing to a 12-byte _unsigned_ value) doubles the amount of chainwork that can be represented so should last for (at least) 6 months.

We are using Issue #3415 to track a longer-term solution, which we will include in our 0.17 release.  @schildbach has started work on it and has begun submitting PRs. Your team can help by commenting on #3415 or by reviewing any of the PRs.

There is also discussion in our [Matrix space](https://matrix.to/#/#bitcoinj:matrix.org), so joining that would also be helpful.

The 0.17 and 0.18 releases will be a significant step forward in simplicity and modularity and if your team can contribute to those release that would be awesome!

schildbach:
@aeidelman Out of curiosity, did Rootstock update to a version with the long-term fix (so 0.16.5+ or 0.17+) ? I'm asking because the short-term fix (this PR) [expired about 24 hours ago](https://github.com/bitcoinj/bitcoinj/issues/3975).

PR #3412, opened 2024-06-23
StoredBlock: use Utils.bigIntegerToBytes to convert chainWork (Backport)

msgilligan:
This is a fix for Issue #3410 that will allow main net blocks above 849,137 to be processed. As of block 849,138 we can no longer fit total chainwork in a 12-byte *signed* field. This fix "kicks the can down the road" by making the field 12-bytes *unsigned*.

We should open a new issue to address the long term need for bigger values. Note that converting the field to 12-byte unsigned precludes us from using the most-significant bit as a flag for a new format, but we should be able to pick some arbitrary value, say 0xA0 as a version flag and declare that values less than 0xA0 are "unversioned".

This is a **BACKPORT** to the 0.16 branch. In the master/0.17 branch the bigIntegerToBytes method has been moved to base.internal.BytUtils, so this commit differs in that one respect.


schildbach (commented):
Fix looks good!

PR #3413, opened 2024-06-23
gradle.yml: explicitly use ubuntu-22.04, macos-12, windows-2019

msgilligan:
This is a backport of a similar change made on `master`.


msgilligan:
@schildbach : it looks like their is an issue with JNI and LevelDB on `windows-2022` and `windows-latest`.  We must have either removed those tests or updates something in the `master` branch.

It's also very frustrating running CI on the `release-0.16` branch because there are intermittent test failures that we have fixed on `master`.

Issue #3415, opened 2024-06-24
Upgrade serialization of chainWork field of `StoredBlock` to support more than 12 bytes (96 bits)


msgilligan:
**NOTE: The short-term, critical issue that occurred on June 24, 2024 has been fixed** in the [0.16.3](https://github.com/bitcoinj/bitcoinj/releases/tag/v0.16.3) release, see Issue #3410

This issue is about being prepared for when chainwork doubles again (in roughly 1 year)

As of block 849,138 chainWork has reached a value of `800028897aa1dce163ead2cf` which is too big to store in 12 **signed** bytes, though it will fit in 12 **unsigned** bytes. This was reported in Issue #3410 and fixed by PR #3411 (which converted the serialization to output an unsigned value.)

This fix buys us some time, but we will need a longer-term situation.

A proper solution might be to add a version field at the start of each `StoredBlock` record -- or perhaps the start of the file or in some metadata somewhere.  It could be as simple as `0xFF01` indicating a new version. But we will need to do something before the chain work value reaches roughly double what it is now.

For now, I'm going to mark this as targeted for the 0.17 release, but we may be able to postpone until 0.18.

ulserGit:
Good afternoon, my blockchain download process stopped at block 849137
I'm using BitcoinJ 16.2 from the Maven repository
How can I solve the problem?

msgilligan:
> my blockchain download process stopped at block 849137 I'm using BitcoinJ 16.2 from the Maven repository How can I solve the problem?

There will be a bitcoinj 0.16.3 release soon.  See Issue #3410 which has been fixed on both `master` and `release-0.16` branches.

ulserGit:
> > my blockchain download process stopped at block 849137 I'm using BitcoinJ 16.2 from the Maven repository How can I solve the problem?
> 
> There will be a bitcoinj 0.16.3 release soon. See Issue #3410 which has been fixed on both `master` and `release-0.16` branches.

Thank you) how long wait release in to maven repo?

msgilligan:
**Bitcoin Core** uses a 32-byte (256-bit) value using types uint256 and arith_uint256. So maybe we should just serialize the value as 32-bytes.

`SPVBlockStore` is formatted according to the following comment:

```
    // File format:
    //   4 header bytes = "SPVB"
    //   4 cursor bytes, which indicate the offset from the first kb where the next block header should be written.
    //   32 bytes for the hash of the chain head
    //
[...]

PR #3417, opened 2024-06-24
StoredBlock: Update comment about CHAIN_WORK_BYTES

msgilligan:
The original comment, written in 2011 said "12 bytes should be plenty for now". As of June 2024, we know that 12 bytes is not enough.

See Issue #3415.


schildbach:
Merged.

PR #3419, opened 2024-06-25
.gitlab-ci.yml: remove build with OpenJDK 21 under Debian Trixie

schildbach:
Trixie doesn't carry OpenJDK 21 any more, so our GitLab CI builds started failing. Because OpenJDK support under Debian is a moving target, it's better to remove it until Trixie is released. Otherwise, CI builds will start failing at any time.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -37,17 +37,6 @@ bookworm-jdk17:
     - gradle --version
     - sha256sum core/build/libs/*.jar wallettool/build/install/wallet-tool/bin/*  wallettool/build/install/wallet-tool/lib/*.jar
 
-trixie-jdk21:
-  image: debian:trixie-slim
-  before_script:
-    - apt-get update
-    - apt-get -y install openjdk-21-jdk-headless gradle
-  script:
-    - gradle build :bitcoinj-core:publishToMavenLocal :bitcoinj-wallettool:installDist --init-script build-scan-agree.gradle --scan --stacktrace
-  after_script:
-    - gradle --version
-    - sha256sum core/build/libs/*.jar wallettool/build/install/wallet-tool/bin/*  wallettool/build/install/wallet-tool/lib/*.jar
-
 sast:
   stage: test
 


schildbach:
Instead, we could perhaps add a build under `default-jdk-headless` on Debian testing, unstable or experimental. It would also fail from time to time, but at least it would follow changes and failures would potentially be repaired semi-automatically.

schildbach:
Successful build on GitLab: https://gitlab.com/bitcoinj/bitcoinj/-/commit/c12fa3bf74cd19135c874ec21714b7ab72043eba

msgilligan (approved):
Removing it seems like the best short-term option. Approved.

msgilligan:
> Instead, we could perhaps add a build under `default-jdk-headless` on Debian testing, unstable or experimental. It would also fail from time to time, but at least it would follow changes and failures would potentially be repaired semi-automatically.

That is one possibility.

There are several other solutions (and they are not mutually exclusive):

1. Add Trixie back and always bump to the latest Trixie JDK whenever it changes (in the worst case after a break)
2. Use NixOS (or Nix package manager under Debian) in the Docker image used for the GitLab CI builds.
3. Use GitHub Actions to test with newer JDKs (e.g. 21)

schildbach:
Merged.


> * Add Trixie back and always bump to the latest Trixie JDK whenever it changes (in the worst case after a break)

This is what I want to avoid, as it will constantly break CI builds. Also think of the release branches that will stay red for a long time.

> * Use NixOS (or Nix package manager under Debian) in the Docker image used for the GitLab CI builds.

Might be an option, although not for 0.17 and also I'm not confident with NixOS yet.

> * Use GitHub Actions to test with newer JDKs (e.g. 21)

Agreed, and are we not already doing that? Note this issue is only about GitLab.

msgilligan:
> > * Add Trixie back and always bump to the latest Trixie JDK whenever it changes (in the worst case after a break)
> 
> This is what I want to avoid, as it will constantly break CI builds. Also think of the release branches that will stay red for a long time.

I agree, but I don't think it would happen that often. And combined with using `default-jdk-headless` it might be acceptable.

> 
> > * Use NixOS (or Nix package manager under Debian) in the Docker image used for the GitLab CI builds.
> 
> Might be an option, although not for 0.17 and also I'm not confident with NixOS yet.

I agree. I'm going to start experimenting with using NixOS images in Docker/Podman. I'll let you know how that goes. And also we don't have to _replace_ the Debian images with NixOS, we can always have both. I think it's valuable to test on Debian.

> 
> > * Use GitHub Actions to test with newer JDKs (e.g. 21)
> 
> Agreed, and are we not already doing that? 

Yes. We currently have JDK 21 in our Matrix.

msgilligan:
Tangentially related: I'm wondering if Trixie will go final with JDK 25.  JDK 25 is scheduled for September 2025 and Trixie has been including EA JDKs. So, if Trixie ships after September 2025, JDK 25 will be the current LTS Java at the time Trixie goes final.

PR #3421, opened 2024-06-25
AddressParserProvider: deprecate fromNetworks()

msgilligan:
It was already package-private and was only used by deprecated methods in `AddressParser`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/base/DefaultAddressParserProvider.java
+++ b/core/src/main/java/org/bitcoinj/base/DefaultAddressParserProvider.java
@@ -76,10 +76,11 @@ class DefaultAddressParserProvider implements AddressParser.AddressParserProvide
     }
 
     /**
-     * Dynamically create a new AddressParser using a snapshot of currently configured networks
-     * from Networks.get().
+     * Dynamically create a new {@link AddressParser} using a snapshot of currently configured networks
+     * from {@code Networks.get()}. This is only used by deprecated methods in {@link AddressParser}.
      * @return A backward-compatible parser
      */
+    @Deprecated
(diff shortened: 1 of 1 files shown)


schildbach (approved):
Seems reasonable.

core/src/main/java/org/bitcoinj/base/DefaultAddressParserProvider.java
```diff
@@ -79,7 +79,9 @@ public AddressParser forNetwork(Network network) {
      * Dynamically create a new AddressParser using a snapshot of currently configured networks
      * from Networks.get().
      * @return A backward-compatible parser
+     * @deprecated This is only used by deprecated methods in {@link AddressParser}
```
schildbach, line 82: Do you mind if we move that comment to a real comment, rather than JavaDoc. I think it should not go into the `@deprecated` JavaDoc tag. I'd probably remove that tag (but keep the annotation).

msgilligan: I don't mind doing it. I'll go ahead and change it and force-push. But, why do you think so?

schildbach: Because the message in the JavaDoc tag usually tells consumers what to use instead. In this case, there is no alternative.

msgilligan: Fixed and force-pushed.

schildbach (approved):

schildbach:
Merged.

PR #3424, opened 2024-06-27
StoredBlock: rework comment about usage of serializeCompact(), deserializeCompact()

schildbach:
The explicit list of usages was already stale. Rather, just mention it's being used internally.


msgilligan (approved):
Approved.

schildbach:
Merged.

PR #3425, opened 2024-06-27
StoredBlock: add serializeCompactV2(), deserializeCompactV2()

schildbach:
The old format will soon run out of bytes for the chain work value.

Also deprecates the old methods and adds tests for the V2 format.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/StoredBlock.java
+++ b/core/src/main/java/org/bitcoinj/core/StoredBlock.java
@@ -40,10 +40,18 @@ import static org.bitcoinj.base.internal.Preconditions.checkState;
 public class StoredBlock {
 
     // A BigInteger representing the total amount of work done so far on this chain. As of June 22, 2024, it takes 12
-    // unsigned bytes to store this value, so we need to create an updated storage format soon.
-    private static final int CHAIN_WORK_BYTES = 12;
-    private static final byte[] EMPTY_BYTES = new byte[CHAIN_WORK_BYTES];
-    public static final int COMPACT_SERIALIZED_SIZE = Block.HEADER_SIZE + CHAIN_WORK_BYTES + 4;  // for height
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/core/StoredBlock.java
```diff
@@ -40,10 +40,16 @@
...
+    // unsigned bytes to store this value, so developers should use the V2 format.
+    private static final int CHAIN_WORK_BYTES_V1 = 12;
+    // A BigInteger representing the total amount of work done so far on this chain.
+    private static final int CHAIN_WORK_BYTES_V2 = 32;
+    // Used for padding.
+    private static final byte[] EMPTY_BYTES = new byte[CHAIN_WORK_BYTES_V2]; // fit larger format
+    /** Number of bytes serialized by {@link #serializeCompact(ByteBuffer)} */
+    public static final int COMPACT_SERIALIZED_SIZE = Block.HEADER_SIZE + CHAIN_WORK_BYTES_V1 + /*height*/ 4;
+    /** Number of bytes serialized by {@link #serializeCompactV2(ByteBuffer)} */
+    public static final int COMPACT_SERIALIZED_SIZE_V2 = Block.HEADER_SIZE + CHAIN_WORK_BYTES_V2 + /*height*/ 4;
```
msgilligan, line 52: Since we have the `/*height*/` comment in two places now, maybe we should use a constant named something like `HEIGHT_SIZE` instead?

msgilligan (changes requested):
Looks good. I have one minor suggestion.

core/src/main/java/org/bitcoinj/core/StoredBlock.java
schildbach: Added.

schildbach:
Rebased.

msgilligan (approved):
Looks good!

schildbach:
Merged.

PR #3428, opened 2024-06-28
SPVBlockStore: support 32 byte chain work

schildbach:
Existing V1 files are automatically converted to V2 format.

Followup of #3436.

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
@@ -60,8 +60,10 @@ public class SPVBlockStore implements BlockStore {
     public static final int DEFAULT_CAPACITY = 10000;
     @Deprecated
     public static final String HEADER_MAGIC = "SPVB";
-    // Magic header.
-    private static final byte[] MAGIC_HEADER = HEADER_MAGIC.getBytes(StandardCharsets.US_ASCII);
+    // Magic header for the V1 format.
+    static final byte[] HEADER_MAGIC_V1 = HEADER_MAGIC.getBytes(StandardCharsets.US_ASCII);
+    // Magic header for the V2 format.
+    static final byte[] HEADER_MAGIC_V2 = "SPV2".getBytes(StandardCharsets.US_ASCII);
 
     protected volatile MappedByteBuffer buffer;
     protected final NetworkParameters params;
@@ -95,6 +97,7 @@ public class SPVBlockStore implements BlockStore {
     // Used to stop other applications/processes from opening the store.
     protected FileLock fileLock = null;
     protected RandomAccessFile randomAccessFile = null;
+    private final FileChannel channel;
     private int fileLength;
 
     /**
@@ -119,11 +122,13 @@ public class SPVBlockStore implements BlockStore {
         Objects.requireNonNull(file);
         this.params = Objects.requireNonNull(params);
         checkArgument(capacity > 0);
+
         try {
             boolean exists = file.exists();
+
             // Set up the backing file, empty if it doesn't exist.
             randomAccessFile = new RandomAccessFile(file, "rw");
-            FileChannel channel = randomAccessFile.getChannel();
+            channel = randomAccessFile.getChannel();
 
             // Lock the file.
             fileLock = channel.tryLock();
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/bitcoinj/store/SPVBlockStoreTest.java
```diff
@@ -190,4 +194,34 @@ public void oneStoreDelete() throws Exception {
...
+        Block genesisBlock = TESTNET.getGenesisBlock();
+        StoredBlock storedGenesisBlock = new StoredBlock(genesisBlock.cloneAsHeader(), genesisBlock.getWork(), 0);
+        Sha256Hash genesisHash = storedGenesisBlock.getHeader().getHash();
+        ((Buffer) buffer).position(SPVBlockStore.FILE_PROLOGUE_BYTES);
+        buffer.put(genesisHash.getBytes());
+        storedGenesisBlock.serializeCompact(buffer);
+        buffer.putInt(4, buffer.position()); // ring cursor
+        ((Buffer) buffer).position(8);
+        buffer.put(genesisHash.getBytes()); // chain head
+        raf.close();
```
msgilligan, line 215: I would consider putting this into a helper method for readability.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -131,44 +135,91 @@ public SPVBlockStore(NetworkParameters params, File file, int capacity, boolean
...
+                    buffer.get(record);
+                    buffer.position(FILE_PROLOGUE_BYTES + i * RECORD_SIZE_V2);
+                    buffer.put(record, 0, 32); // hash
+                    buffer.put(zeroPadding);
+                    buffer.put(record, 32, RECORD_SIZE_V1 - 32); // work, height, block header
+                }
+
+                // migrate cursor
+                int cursorRecord = (getRingCursor() - FILE_PROLOGUE_BYTES) / RECORD_SIZE_V1;
+                setRingCursor(FILE_PROLOGUE_BYTES + cursorRecord * RECORD_SIZE_V2);
```
msgilligan, line 191: Could the migration be moved to a static helper method for readability? If this change is merged as is the constructor will be 110 lines of code.

I'm also wondering if the upgrade should be made (effectively) atomic. Obviously this would require more free storage to do, but it would protect for the case of power-failure in the middle of conversion, etc.

msgilligan (changes requested):
I reviewed this fairly thoroughly and I don't see any major issues (although I worry about a power-failure, etc. during the upgrade process) but have a few suggestions for readability.

It might also be a good idea to add some additional tests. The only migration test is for a block store with a single record.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -131,44 +135,91 @@ public SPVBlockStore(NetworkParameters params, File file, int capacity, boolean
                 throw new ChainFileLockedException("Store file is already locked by another process");
 
             // Ensure expected file size, grow if desired.
-            fileLength = getFileSize(capacity);
-            if (!exists) {
-                log.info("Creating new SPV block chain file " + file);
-                randomAccessFile.setLength(fileLength);
+            this.fileLength = getFileSize(capacity);
```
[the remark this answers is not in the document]
schildbach: Removed via 626e5fb638d9148bc680d567e46d7a8104c3b421.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -327,10 +378,24 @@ public void close() throws BlockStoreException {
...
+    //   4 magic header bytes = "SPV2"
+    //   4 cursor bytes, which indicate the offset from the first kb where the next block header should be written.
+    //   32 bytes for the hash of the chain head
+    //
+    // For each header (148 bytes)
+    //   32 bytes hash of the header
+    //   32 bytes of chain work
+    //    4 bytes of height
+    //   80 bytes of block header data
+
```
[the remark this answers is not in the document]
schildbach: I deliberately decided against this, because the v2 format is the relevant one in future and v1 documentation is only there for historic purposes. We could decide to remove the v1 doc though.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
schildbach: Extracted helper method via 7383f1d25526c1e96ba331bbb865413359d6a1c5.

Regarding the atomicity, effectively we'd need to create a second file, migrate into that without touching the old file and then rename the new file over the old. We do this with wallet saves. However since migration happens only once and presumibly at app start I would not expect crashing processes etc. Do you think this is a real issue?

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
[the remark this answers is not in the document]
schildbach: But then again, yeah maybe you're right.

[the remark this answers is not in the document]
schildbach: See 38d7d6c786832310d4f8ac405632256e1df0ac3e.

schildbach:
I added a couple of commits on top (which I will squash before merging).

schildbach:
> It might also be a good idea to add some additional tests. The only migration test is for a block store with a single record.

Do you have a specific test case in mind?

Note that there are 3 records even if two of them are empty. For example if the migration had a bug that would leak record 1 or 2 into 0 (e.g. by running the migration the other direction) it should show up in the assert because it would zero some bytes. It's quite an effort to create another test block deterministically and I couldn't see how it would increase test coverage.

That said, we could relatively easy use either two identical genesis blocks, or one each of different networks (ignoring that this would never happen in real use).

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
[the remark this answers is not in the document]
msgilligan: Look great! Resolved.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
msgilligan: > Extracted helper method via [7383f1d](https://github.com/bitcoinj/bitcoinj/commit/7383f1d25526c1e96ba331bbb865413359d6a1c5).

Looks good!

msgilligan: > However since migration happens only once and presumibly at app start I would not expect crashing processes etc. Do you think this is a real issue?

Well Bitcoin Wallet has a large installed base. If a phone runs out of battery or crashes for some reason during the upgrade, what will happen? How do you recover from a corrupted SPVStore file?

msgilligan:
> Do you have a specific test case in mind?

A few:

1. Trying to convert when the disk is full
2. Recovering from a failure during convert
3. Testing with live (or simulated) data from main net (this would/could have caught the total chainwork issue)

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
schildbach: > How do you recover from a corrupted SPVStore file?

The blockchain will be reset and the wallet replayed.

schildbach:
1 and 2 are somewhat the same, and we could add "out of RAM" to this list. I have no idea how to test for random failures properly, unless we clutter our production code with (conditional) throw statements. Is there at least a way to simulate "disk full", e.g. by writing to a directory with a quota? I'm not aware of such a possibility, at least not via the Java API and portable between various filesystems.

3 is in the realm of integration tests I'd say. Unit tests cannot depend on being fed with live data, and mining >12 bytes worth of chain work isn't feasible either.

msgilligan:
> 1 and 2 … I have no idea how to test for random failures properly,

> 3 is in the realm of integration tests I'd say. 

I think all these cases are out-of-scope for this PR. I just wanted to answer your question about additional test cases as best I could.

> Unit tests cannot depend on being fed with live data, and mining >12 bytes worth of chain work isn't feasible either.

We could (at some time in the future) create our own Test Runner node that maintains and has access to a fully-synced main net and runs some tests against it. But, again, this is out-of-scope for this PR.

msgilligan (approved):
Approved.

schildbach:
Merged.

PR #3430, opened 2024-07-01
BuildCheckpoints: create mixed-format checkpoints

schildbach:
Create checkpoints with 12 (or less) bytes of work in the old V1 format, otherwise V2. This will make checkpoint files easier to review.

This un-deprecates the V1 format in StoredBlock, since it's no longer planned to remove that in the forseeable future.

This also reverts recent changes to the bundled checkpoints (commit de8afc67ca60642e8d18ed183d069144cc48e254).


msgilligan (approved):
Looks good!

schildbach:
Merged.

PR #3431, opened 2024-07-12
SPVBlockStore: deprecate constant `HEADER_MAGIC`

schildbach:
It's an implementation detail. We should be switching this to private in future.


msgilligan (approved):
Approved.

schildbach:
Merged.

PR #3432, opened 2024-07-12
SPVBlockStore: compare magic header as bytes, not string

schildbach:
This gets rid of several bytes to/from string conversions.

Followup of #3431.

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
@@ -16,6 +16,7 @@
 
 package org.bitcoinj.store;
 
+import org.bitcoinj.base.internal.ByteUtils;
 import org.bitcoinj.core.Block;
 import org.bitcoinj.core.NetworkParameters;
 import org.bitcoinj.core.ProtocolException;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -175,9 +179,8 @@ else if (fileLength < randomAccessFile.length())
     }
 
     private void initNewStore(Block genesisBlock) throws Exception {
-        byte[] header;
-        header = HEADER_MAGIC.getBytes(StandardCharsets.US_ASCII);
-        buffer.put(header);
+        ((Buffer) buffer).rewind();
```
msgilligan, line 182: Isn't this an unrelated change that should be in a separate commit?

msgilligan (changes requested):
I think there is one unrelated change in this PR.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
schildbach: Indeed. Removed.

msgilligan (approved):
Looks good!

schildbach:
Merged.

PR #3433, opened 2024-07-14
SPVBlockStore: rewind buffer directly before accesses to magic header

schildbach:
This is meant to make sure the buffer position is always right.

Followup to #3432

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
@@ -162,6 +162,7 @@ public class SPVBlockStore implements BlockStore {
             // Check or initialize the header bytes to ensure we don't try to open some random file.
             if (exists) {
                 byte[] currentHeader = new byte[4];
+                ((Buffer) buffer).rewind();
                 buffer.get(currentHeader);
                 if (!Arrays.equals(currentHeader, MAGIC_HEADER))
                     throw new BlockStoreException("Magic header expected, got: " + ByteUtils.formatHex(currentHeader));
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -162,6 +162,7 @@ else if (fileLength < randomAccessFile.length())
             // Check or initialize the header bytes to ensure we don't try to open some random file.
             if (exists) {
                 byte[] currentHeader = new byte[4];
+                ((Buffer) buffer).rewind();
```
msgilligan, line 165: I don't think calling `rewind` is necessary, because `channel.map()` has just been called and `buffer` was just initialized.

(It also is not necessary in the not-exists case, but since `initNewStore` is called elsewhere the rewind is needed in `initNewStore`)

msgilligan (changes requested):
I think there is a redundant `rewind` that should be removed.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
schildbach: It's meant to be defensive, as in my other PR I'm moving stuff around and suddenly this rewind() became essential.

msgilligan: It's ok to leave it there, but `buffer` is initialized just 5 lines above.

msgilligan (approved):
Approved, as is. But if you want to remove what I consider an unnecessary `rewind` that is fine with me, too.

PR #3444, opened 2024-08-14
ECKey, BtcFormat: fix bad JavaDoc @link tags

msgilligan:
JDK 23 fixes a bug that allowed these invalid links: https://bugs.openjdk.org/browse/JDK-8164094

This PR fixes the invalid links and does not require JDK 23, but does fix our build so it will work with JDK 23.


schildbach:
Merged.

PR #3445, opened 2024-08-15
TorUtils: Move to o.b.core.internal

msgilligan:
It is only used by PeerAddress, so it makes sense for it to be in the same package as PeerAddress. This removes a Guava dependency from the o.b.crypto package/module.


schildbach:
Makes sense. Merged.

PR #3449, opened 2024-08-27
settings.gradle: require Gradle 7.3 or later

schildbach:
Two reasons for picking that particular version:
* it enables us to replace all previous Gradle-related conditionals in this file
* it's required for official support of Java 17

Gitlab pipeline is fine: https://gitlab.com/bitcoinj/bitcoinj/-/pipelines/1428473105

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,13 +1,12 @@
+/*
+ * Settings file for Gradle 7.3 or later.
+ * For building with Debian Gradle 4.4, use the settings-debian.gradle settings.
+ */
+
 import org.gradle.util.GradleVersion
 
 // Minimum Gradle version for build
-def minGradleVersion = GradleVersion.version("4.4")
-// Minimum Gradle version for JUnit5
-def minJunit5GradleVersion = GradleVersion.version("4.6")
-// Minimum Gradle version for builds of JavaFX 17 module using JDK 17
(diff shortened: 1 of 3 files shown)


schildbach:
Ah, and it should be noted that though this PR simplifies `settings.gradle` to a large degree by replacing some conditionals by just one at the beginning, there are still a couple of Gradle version conditionals in various `build.gradle` files.

schildbach:
We cannot support the special case that is meant to be close to Debian Gradle, so I removed it. Debian Gradle is tested via the GitLab Runners though.

schildbach:
I've added an experimental commit on top for testing with Gradle 7.3.3.

schildbach:
I've just reverted all the experiments with adding Gradle 7.3 to the build matrixes. The resulting CI problems need further examination, but I think they're out of scope for this PR. Sorry for the noise.

settings.gradle
```diff
@@ -1,13 +1,12 @@
+/*
+ * Setting file for Gradle 7.3 or later.
```
msgilligan, line 2: Setting**s** file

msgilligan (changes requested):
I found one trivial typo (setting vs settings)  Otherwise approved!

msgilligan (approved):

schildbach:
Merged.

PR #3452, opened 2024-08-28
build.gradle: update Gradle plugin for GraalVM to 0.10.2

schildbach:
I assume updating to most recent 0.10.2 would also bump our GraalVM requirement to 22 – seems excessive as our highest Java version is currently 21.

https://graalvm.github.io/native-build-tools/latest/index.html#changelog

### DIFF
--- a/wallettool/build.gradle
+++ b/wallettool/build.gradle
@@ -4,7 +4,7 @@ plugins {
     id 'java'
     id 'application'
     id 'org.asciidoctor.jvm.convert' version '3.3.2' apply false
-    id 'org.graalvm.buildtools.native' version '0.9.27' apply false
+    id 'org.graalvm.buildtools.native' version '0.10.2' apply false
 }
 
 def annotationProcessorMinVersion = GradleVersion.version("4.6")


schildbach:
I think this should be backported to 0.17 as well.

msgilligan:
> I assume updating to most recent 0.10.2 would also bump our GraalVM requirement to 22 … https://graalvm.github.io/native-build-tools/latest/index.html#changelog

They're actually talking about GraalVM 22.3, which is a version from before they switched to releasing GraalVM using the JDK version number. (They were using the release year and zero-based quarter, so 22.3 was the 4th quarterly release of 2022.) So there is actually a GraalVM 22.3 for JDK 17. 

So we should be able to use the latest version of the native image plugin.

wallettool/build.gradle
```diff
@@ -4,7 +4,7 @@ plugins {
     id 'java'
     id 'application'
     id 'org.asciidoctor.jvm.convert' version '3.3.2' apply false
-    id 'org.graalvm.buildtools.native' version '0.9.27' apply false
+    id 'org.graalvm.buildtools.native' version '0.9.28' apply false
```
msgilligan, line 7: I think we can (and should) use version 0.10.2 here. See my comment on the main thread.

msgilligan (changes requested):

wallettool/build.gradle
schildbach: Ok, I changed it to 0.10.2.

msgilligan (approved):
Looks like 0.10.2 works. Approved!

schildbach:
Merged.

PR #3454, opened 2024-08-28
core/build.gradle,gradle.yml remove test(On)Jdk8

msgilligan:
Since we are planning on dropping support for Jdk8 in the next release and the current implementation of this in GitHub Actions is blocking us from testing on macos-14 (ARM64), let's just remove this.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           gradle-version: ${{ matrix.gradle }}
       - name: Run Gradle
-        run: gradle -PtestJdk8=true build --init-script build-scan-agree.gradle --scan --info --stacktrace
+        run: gradle build --init-script build-scan-agree.gradle --scan --info --stacktrace
       - name: Upload Test Results and Reports
         uses: actions/upload-artifact@v4
         if: always()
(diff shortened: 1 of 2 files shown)


schildbach:
How do you mean "drop JDK8 support"? We still need Android support, and that currently means something close to Java 8 language and Java 8 API support. The `testJdk8` option was meant to test if we have any regressions on that front.

msgilligan:
> How do you mean "drop JDK8 support"? 

I thought we were going to use JDK 11 for `bitcoinj-core` in release 0.18.

> We still need Android support, and that currently means something close to Java 8 language and Java 8 API support. The `testJdk8` option was meant to test if we have any regressions on that front.

I created PR #3456 which runs tests on JDK 11 to look for regressions that could affect (a more recent version of) the Android API.

An additional option (which we should consider independently) is getting a test workflow that actually tests in an Android emulator.

schildbach:
I've read through #3345 but I wonder if we can simply disable `testJdk8` on the condition that we're running on MacOS? I'd be fine with restricting `testJdk8` to Linux as well.

schildbach (changes requested):
Sorry, I think we should generally not remove these tests. I'd be ok with removing them just for macos, or for any non-linux.

msgilligan:
Closing.

PR #3456, opened 2024-08-28
core/build.gradle,gradle.yml: test(On)Jdk11

msgilligan:
Continue using Java Toolchain support in Gradle to test on an earlier JDK, but switch from JDK 8 to
JDK 11.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           gradle-version: ${{ matrix.gradle }}
       - name: Run Gradle
-        run: gradle -PtestJdk8=true build --init-script build-scan-agree.gradle --scan --info --stacktrace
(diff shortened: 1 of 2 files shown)


schildbach (commented):
For Android, this change could be considered in 2 years at best.

schildbach:
I'd expect the macOS-13 image, which we now switched to via #3537, will stay around for at least a year. Hopefully by then we've started the 0.19 cycle. So I agree with targetting 0.19.

msgilligan:
Closing this PR given that we have PR #3960 which allows us to test on macOS-15 Intel architecture where JDK 8 is supported.

msgilligan:
Well maybe closing this was premature, PR #3960 is failing because Gradle Toolchains says it can't find a JDK.

msgilligan:
Closing in favor of PR #3962 and now that we use `testMinJDK` with a parameterized minimum JDK version.

PR #3455, opened 2024-08-28
graalvm.yml,gradle.yml: Upgrade to macOS-14

msgilligan:
This is a draft because it is a child of PR #3454

PR #3457, opened 2024-08-28
 graalvm.yml,gradle.yml: Upgrade to macOS-14

msgilligan:
This is the same as PR #3455 except it is a child of PR #3456


msgilligan:
Closing in favor of PR #3962 which adds arm64. (GitHub provides "community" runners for macOS-15 and both aarch64 and Intel.)

PR #3458, opened 2024-08-29
build.gradle: target Java 17 for wallettool, tools, examples and integration-test

schildbach:
Notably missing is `core`.

I admit I've forgot why we have `options.compilerArgs.addAll(['--release', '11'])`, and if they could perhaps be removed now. But I updated them anyway.

### DIFF
--- a/examples-kotlin/build.gradle
+++ b/examples-kotlin/build.gradle
@@ -7,16 +7,16 @@ dependencies {
 }
 compileKotlin {
     kotlinOptions {
-        jvmTarget = "11"
+        jvmTarget = "17"
     }
 }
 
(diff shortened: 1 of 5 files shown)


msgilligan:
> I've forgot why we have options.compilerArgs.addAll(['--release', '11']), and if they could perhaps be removed now.

This is the new way to specify the Java version when compiling, which unlike the old way tells the compiler to compile against the API of that version in addition to generating byte code for that version.

It's actually the "target version" that we should eventually remove.

msgilligan (approved):
Approved!

schildbach:
Merged.

PR #3460, opened 2024-08-29
build.gradle: migrate `--release` compiler option to `compileJava.options.release` for wallettemplate, examples-kotlin

schildbach:
For the remaining modules, we're still using Gradle 4.4 so we can't do the same.

Removing sourceCompatibility, targetCompatibility is for the next PR.

### DIFF
--- a/examples-kotlin/build.gradle
+++ b/examples-kotlin/build.gradle
@@ -12,11 +12,11 @@ compileKotlin {
 }
 
 sourceCompatibility = 17
+compileJava.options.release = 17
 compileJava.options.encoding = 'UTF-8'
 compileTestJava.options.encoding = 'UTF-8'
 javadoc.options.encoding = 'UTF-8'
 
 compileJava {
-    options.compilerArgs.addAll(['--release', '17'])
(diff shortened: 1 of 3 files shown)


msgilligan:
Why only for two modules?

msgilligan (changes requested):
Why only for two modules?

schildbach:
> Why only for two modules?

For the remaining modules, we're still using Gradle 4.4 so we can't do the same.

schildbach:
We could add it for integration-test as well I guess, but currently it has no `--release`.

msgilligan:
> > Why only for two modules?
> 
> For the remaining modules, we're still using Gradle 4.4 so we can't do the same.

Right. Sorry, my mistake.

msgilligan:
> We could add it for integration-test as well I guess, but currently it has no `--release`.

Yeah, I think we should add it there.

schildbach:
Ok, I snuck it in for `integration-test` though strictly it's no migration.

schildbach:
Merged.

PR #3463, opened 2024-08-29
build.gradle: remove `sourceCompatibility`, `targetCompatibility`

schildbach:
We have defined `release` one way or another, and on JDK 17 that should suffice.


msgilligan (approved):
Approved!

schildbach:
Merged.

PR #3464, opened 2024-08-29
base: remove all (deprecated) usage of Networks.get()

msgilligan:
Remove `DefaultAddressParserProvder.fromNetworks()` and everything that depends on it. All methods being removed were previously deprecated.

Removing `DefaultAddressParserProvder.fromNetworks()` propagates to the 3 AddressParser getLegacy() methods and several other deprecated methods in Address, LegacyAddress, and Segwit Address.

This change means that nothing in `o.b.base` will use the global list of supported networks stored in `Networks`. This is a mechanism we previously added to support other Bitcoin-compatible networks and their addressing schemes (e.g. LiteCoin.) "Alt networks" are still supported, but applications using them will need to create their own implementations of AddressParser to do it. We hope to eventually remove entirely the global state returned by Networks.get(), but this PR is just about removing its usage from `o.b.base`.

Note: this also removes all uses of javax.annotation.Nullable from o.b.base. There are still instances of @NonNull which we should probably leave in until we move to JSpecify.

### DIFF
--- a/core/src/test/java/org/bitcoinj/base/LegacyAddressTest.java
+++ b/core/src/test/java/org/bitcoinj/base/LegacyAddressTest.java
@@ -121,45 +121,12 @@ public class LegacyAddressTest {
     }
 
     @Test
-    @Deprecated
-    // Test a deprecated method just to make sure we didn't break it
-    public void getNetworkViaParameters() {
-        NetworkParameters params = LegacyAddress.getParametersFromAddress("17kzeh4N8g49GFvdDzSf8PjaPfyoD1MndL");
-        assertEquals(MAINNET.id(), params.getId());
-        params = LegacyAddress.getParametersFromAddress("n4eA2nbYqErp7H6jebchxAN59DmNpksexv");
-        assertEquals(TESTNET.id(), params.getId());
-    }
-
-    @Test
     public void getNetwork() {
         AddressParser parser = AddressParser.getDefault();
         Network mainNet = parser.parseAddress("17kzeh4N8g49GFvdDzSf8PjaPfyoD1MndL").network();
         assertEquals(MAINNET, mainNet);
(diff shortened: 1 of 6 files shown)


schildbach:
Hmm, I wonder why `BitcoinURITest.testConvertToBitcoinURI()` still works? It registers a mock network and then relies on it near the end of the method.

core/src/test/java/org/bitcoinj/base/LegacyAddressTest.java
```diff
@@ -139,29 +129,6 @@ public void getNetwork() {
         assertEquals(TESTNET, testNet);
     }
 
-    @Test
```
schildbach, line 142: If we remove this test, shouldn't we better remove the entire cap to register other networks? (We can still deprecate these methods in 0.17 last minute.)

msgilligan: I'm not sure where else in `core` relies on this list. This PR is just to remove use of registering Network from address parsing and related functionality in base.

msgilligan:
> Hmm, I wonder why `BitcoinURITest.testConvertToBitcoinURI()` still works? It registers a mock network and then relies on it near the end of the method.

Because it is not in `o.b.base` and this PR only removes use of `Networks.get()` from `base`.

core/src/test/java/org/bitcoinj/base/LegacyAddressTest.java
msgilligan: > I'm not sure where else in core relies on this list. 

For example `BitcoinURITest.testConvertToBitcoinURI()` as you point out. There may be other places as well. We should remove these separately and probably on a case-by-case basis.

msgilligan:
The most notable remaining use of `Networks.network` (private) is in `Networks.find()` which is used by `NetworkParameters.of()` which is used in about 40 places.

schildbach:
Merged.

PR #3470, opened 2024-08-30
LegacyAddressTest: Use MockAltNetwork directly 

msgilligan:
This simplifies the test, removes a dependency on core.NetworkParameters,
and also demonstrates using an Alt Network using only `base` classes.

This is child of PR #3469


schildbach:
Merged.

PR #3474, opened 2024-08-30
AddressData: extracted from SegwitAddressTest

msgilligan:
Put it in base.internal so it can also be used by tests in the core module.


schildbach (changes requested):
Moving test data from test to main seems not the right thing to do, honestly. And we should not misuse "internal" for placing thing we don't know where to put them in.

msgilligan:
See PR #3476

msgilligan:
Closing in favor of PR #3476

PR #3476, opened 2024-08-31
add test-support module for test data and utilities

msgilligan:
extract AddressData inner class from SegwitAddressTest and move to test-support.

There are a few other classes we can move here, but those will be in  separate PRs.

### DIFF
--- a/test-support/build.gradle
+++ b/test-support/build.gradle
@@ -1,0 +1,22 @@
+plugins {
+    id 'java-library'
+}
+
+dependencies {
+    api project(':bitcoinj-core')
+}
+
+tasks.withType(JavaCompile) {
+    options.compilerArgs.addAll(['--release', '8'])
+    options.compilerArgs << '-Xlint:deprecation'
+    options.encoding = 'UTF-8'
+}
+
+ext.moduleName = 'org.bitcoinj.test-support'
+
+jar {
+    inputs.property("moduleName", moduleName)
+    manifest {
+        attributes 'Automatic-Module-Name': moduleName
+    }
+}
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -14,6 +14,8 @@ dependencies {
     api 'com.google.protobuf:protobuf-javalite:4.27.3'
     implementation 'org.slf4j:slf4j-api:2.0.13'
     implementation 'net.jcip:jcip-annotations:1.0'
+
+    testImplementation project(':bitcoinj-test-support')
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.easymock:easymock:5.4.0'
     testImplementation 'com.fasterxml.jackson.core:jackson-databind:2.17.2'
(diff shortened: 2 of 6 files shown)


test-support/build.gradle
```diff
@@ -0,0 +1,67 @@
+import org.gradle.util.GradleVersion
+
+plugins {
+    id 'java-library'
+    id 'maven-publish'
```
schildbach, line 5: What do we need the maven publishing for?

test-support/build.gradle
```diff
@@ -0,0 +1,67 @@
...
+}
+
+version = '0.18-SNAPSHOT'
```
schildbach, line 8: We don't need to version the module I'd say.

test-support/build.gradle
```diff
@@ -0,0 +1,67 @@
...
+            from components.java
+            artifact sourcesJar
+            artifact javadocJar
+
+            pom {
+                description = 'A Java Bitcoin library'
+            }
+        }
+    }
+}
```
schildbach, line 67: I would think this is all not needed.

schildbach (changes requested):
I will probably use two commits:
* Add empty submodule
* Move first class into submodule

test-support/build.gradle
```diff
@@ -0,0 +1,67 @@
+import org.gradle.util.GradleVersion
+
+plugins {
+    id 'java-library'
```
schildbach, line 4: I'm not sure, but maybe we only need the `java` plugin here.

test-support/build.gradle
schildbach: Sorry, the `dependencies` section is indeed needed.

schildbach: I'm talking about lines 20-67 that I think are not needed.

test-support/build.gradle
msgilligan: I think it is fine to publish this as it might be useful to us or others. But I guess that also implies that it is "supported" and reduces our flexibility in making changes.

We can remove the publishing and add it later if we really need it.

test-support/build.gradle
msgilligan: We don't need the version if we don't publish. I'll remove it.

test-support/build.gradle
msgilligan: I would leave the automatic module name just to reserve the name. It may be useful of we start running tests in modular mode.

But the rest can be removed since we're not publishing.

test-support/build.gradle
msgilligan: Yeah, we can remove the `maven-publish` plugin, since we're not publishing.

test-support/build.gradle
msgilligan: Fixed and force-pushed.

test-support/build.gradle
msgilligan: Fixed and force-pushed.

test-support/build.gradle
msgilligan: Fixed and force-pushed.

schildbach:
Merged.

PR #3479, opened 2024-09-02
Remove ListenableCompletableFuture (Breaking, but previously announced change)

msgilligan:
Release 0.17 was the transition release to allow users to migrate from `ListenableFuture` to `CompletableFuture`. In Release 0.18 we are removing `ListenableCompletableFuture` and `ListenableCompletionStage` and changing method signatures to use `CompletableFuture`.

This is technically a breaking change and, of necessity, we weren't able to mark everything that is going away as "deprecated", but we did (ahem, _will_) put a special notice in the 0.17 release notes and the migration should not be too difficult.


msgilligan:
I opened Issue #3480 to remind us to warn more sternly about this in the 0.17 release notes.

schildbach:
I just quickly skimmed over this, and I give my "concept ack". Getting rid of our `ListenableCompletableFuture` was always the plan, and 0.18 seems like the right time to do it.

msgilligan:
I suppose it is also not too late to add @Deprecation tags to `ListenableCompletableFuture` and elsewhere. I believe we decided not to do that because deprecation warnings would get out of control, but we could reconsider that decision and add them now.

schildbach:
Since our active development on 0.17 is almost done we could indeed consider to add deprecatons. Would you like to prepare a PR against master to merge, and 0.17 to cherry-pick, and then rebase this on top? I assume this PR would then remove all the added deprecation warnings again.

schildbach:
With #3508 merged, would you like to rebase this?

msgilligan:
> With #3508 merged, would you like to rebase this?

@schildbach Done!

schildbach:
Merged to master.

PR #3502, opened 2024-09-02
Block: remove reference to deprecated getTimeSeconds()

schildbach:
If approved, I'll merge this to 0.17 as well.

PR #3506, opened 2024-09-02
AbstractBlockChain: migrate private getMedianTimestampOfRecentBlocks() to Instant

schildbach:
Child of #3502.

Will backport this to 0.17 if possible.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -600,7 +600,7 @@ public abstract class AbstractBlockChain {
                         block.getHashAsString(), filteredTxHashList.size(), filteredTxn.size());
                 for (Sha256Hash hash : filteredTxHashList) log.debug("  matched tx {}", hash);
             }
-            if (expensiveChecks && block.time().getEpochSecond() <= getMedianTimestampOfRecentBlocks(head, blockStore))
+            if (expensiveChecks && !block.time().isAfter(getMedianTimestampOfRecentBlocks(head, blockStore)))
                 throw new VerificationException("Block's timestamp is too early");
 
             // BIP 66 & 65: Enforce block version 3/4 once they are a supermajority of blocks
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
```diff
@@ -600,7 +600,7 @@ private void connectBlock(final Block block, StoredBlock storedPrev, boolean exp
                         block.getHashAsString(), filteredTxHashList.size(), filteredTxn.size());
                 for (Sha256Hash hash : filteredTxHashList) log.debug("  matched tx {}", hash);
             }
-            if (expensiveChecks && block.time().getEpochSecond() <= getMedianTimestampOfRecentBlocks(head, blockStore))
+            if (expensiveChecks && block.time().compareTo(getMedianTimestampOfRecentBlocks(head, blockStore)) <= 0)
```
msgilligan, line 603: We should be using `.isBefore()` and `.isAfter()` in code like this, rather than `.compareTo()`.  I probably should have noticed/commented earlier, but late is better than never.

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
```diff
@@ -832,7 +832,7 @@ private void handleNewBestChain(StoredBlock storedPrev, StoredBlock newChainHead
             for (Iterator<StoredBlock> it = newBlocks.descendingIterator(); it.hasNext();) {
                 cursor = it.next();
                 Block cursorBlock = cursor.getHeader();
-                if (expensiveChecks && cursorBlock.time().getEpochSecond() <= getMedianTimestampOfRecentBlocks(cursor.getPrev(blockStore), blockStore))
+                if (expensiveChecks && cursorBlock.time().compareTo(getMedianTimestampOfRecentBlocks(cursor.getPrev(blockStore), blockStore)) <= 0)
```
msgilligan, line 835: isBefore/isAfter. It's hard for me to even read this code to decide which one it should be 🤣

msgilligan (changes requested):
`isBefore`/`isAfter` rather than `compareTo`

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
schildbach: Are you ok with `!isAfter()` then? Or would you craft an `isBefore() || equals()`? That's why I went with `compareTo()`.

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
schildbach: See above.

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
schildbach: Pushed a commit on top that uses `!isAfter()`.

msgilligan (approved):

core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
msgilligan: I think `!isAfter()` is pretty readable. In my mind I read it as "is not after" and it makes sense.

schildbach:
Merged to master and 0.17.

PR #3507, opened 2024-09-03
NetworkParameters: migrate constant `BIP16_ENFORCE_TIME` to Instant

schildbach:
If possible, I'll backport this to 0.17 as well.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+++ b/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
@@ -35,6 +35,7 @@ import org.bitcoinj.utils.VersionTally;
 
 import javax.annotation.Nullable;
 import java.math.BigInteger;
+import java.time.Instant;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/NetworkParameters.java
```diff
@@ -117,7 +118,7 @@ protected NetworkParameters(Network network) {
      * network rules in a soft-forking manner, that is, blocks that don't follow the rules are accepted but not
      * mined upon and thus will be quickly re-orged out as long as the majority are enforcing the rule.
      */
-    public static final int BIP16_ENFORCE_TIME = 1333238400;
+    public static final Instant BIP16_ENFORCE_TIME = Instant.ofEpochSecond(1333238400);
```
msgilligan, line 121: Since this is `public`, maybe it should be deprecated and replaced with `BIP16_ENFORCE_INSTANT`?

msgilligan (changes requested):
We should probably do this migration in a non-breaking way.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
schildbach: I decided against it, as it's extremely unlikely someone is using that constant.

And migration is very straightforward.

msgilligan: Fair enough.

msgilligan (approved):

schildbach:
Merged to master and 0.17.

PR #3508, opened 2024-09-03
ListenableComplet{ableFuture,ionStage}: add deprecations/warnings

msgilligan:
If we deprecate these two classes, there will be no way for callers to get rid of the deprecation warnings.

Instead, I've added **bolded** warnings in the JavaDoc for each class **and** deprecated all their methods.

This is a candidate for backporting/cherry-picking to the 0.17 branch. Pending PR #3479 (on `master`) will remove both of these classes, but we should rebase PR #3479 after this is merged anyway.

### DIFF
--- a/core/src/main/java/org/bitcoinj/utils/ListenableCompletableFuture.java
+++ b/core/src/main/java/org/bitcoinj/utils/ListenableCompletableFuture.java
@@ -21,21 +21,25 @@ import java.util.concurrent.CompletableFuture;
 
 /**
  * A {@link CompletableFuture} that is also a {@link com.google.common.util.concurrent.ListenableFuture} for migration
- * from Guava {@code ListenableFuture} to {@link CompletableFuture}.
+ * from <b>Guava</b> {@code ListenableFuture} to {@link CompletableFuture}. This allows clients of <b>bitcoinj</b> to change the type
+ * of variables receiving {@code Future}s from bitcoinj methods. You must switch from Guava's
+ * {@link com.google.common.util.concurrent.ListenableFuture} (and related types) to Java 8's {@link CompletableFuture}.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/utils/ListenableCompletableFuture.java
```diff
@@ -52,7 +59,9 @@ public static <T> ListenableCompletableFuture<T> completedFuture(T value) {
      * @param throwable the exceptions
      * @param <T> the type of the expected value
      * @return the completed CompletableFuture
+     * @deprecated Use {@link FutureUtils#failedFuture(Throwable)} or if JDK 9+ use {@code CompletableFuture.failedFuture}
```
schildbach, line 62: FutureUtils is an internal class, so it's probably not ok to recommend to use it?

schildbach (changes requested):

core/src/main/java/org/bitcoinj/utils/ListenableCompletableFuture.java
msgilligan: Good catch. I was very much in an "internal state of mind" both when I wrote these JavaDoc originally and when editing them.

Fixed and force-pushed.

msgilligan:
> > noticed that some When the migration to {@link CompletableFuture} is finished paragraphs should be removed
> 
> Should they really be removed on this PR? This is what gets merged to 0.17 mainly, and on 0.17 the migration won't be finished – or did I get that part wrong?

I think they should be removed in this PR. Yes, technically the migration isn't finished, but those comments are entirely replaced by the added `@deprecation` tags.

schildbach:
Merged.

PR #3511, opened 2024-09-13
Simplify Gradle version comparison in settings.gradle and all build.gradle

panic08:
I think we should change the version comparison in Gradle to use the “<” operator, since using the comparison operator in GradleVersion under the hood would result in “compareTo” as it is. Also, the code would look more concise that way, I think.

### DIFF
--- a/base/build.gradle
+++ b/base/build.gradle
@@ -39,7 +39,7 @@ test {
 
 def gradleVersionToolchains = GradleVersion.version("6.7")
 
-if (GradleVersion.current().compareTo(gradleVersionToolchains) > 0) {
+if (GradleVersion.current() > gradleVersionToolchains) {
     // If the Gradle Java Toolchains feature is available, run tests on older JDKs
     System.err.println "Adding 'testOnJdk8' task, because ${GradleVersion.current()}"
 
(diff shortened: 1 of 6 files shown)


msgilligan (approved):
I think sometimes we forget that we are using Groovy. This is definitely more readable.

msgilligan (changes requested):
On second thought, for consistency, this PR should also update line 11 of `settings-debian.gradle` to use a Groovy `==` comparison.

Can you make that change and force-push?

msgilligan (approved):
Looks good! Thanks!

panic08:
**Made a force push as I accidentally committed a commit unrelated to this PR**

schildbach (commented):
I wonder if we should replace those `.compareTo()`s in the various `build.gradle` files as well, for consistency. It's the same Groovy afaik.

panic08:
Replacing `.compareTo()` in other `build.gradle` I suppose would be a good idea. Since we are changing .compareTo() for ease of readability, I suppose it would be just as good if we did it in other `build.gradle` as well

msgilligan (approved):
Even better!

msgilligan:
Thanks @panic08 !

(This is like when your high school teacher catches you eating M&M's and says "Why don't you share them with the whole class?" 🤣)

PR #3512, opened 2024-09-14
ScriptChunk: replace/deprecate write()

msgilligan:
Deprecate write(OutputStream) and replace with existing toByteArray() method. Internally implement a write(ByteBuffer) method.

Update Script to use chunk.toByteArray()

We should also consider back-porting this change to 0.17, so we could remove `write(OutputStream)` altogether in 0.18.


msgilligan:
> Concept looks good!
> 
> I think however we should try to keep `toByteArray()` free from calls to the deprecated method.

`toByteArray()` calls `write(ByteBuffer buf)` which is not deprecated -- `write(OutputStream stream)` is deprecated.

schildbach:
Merged to master and 0.17.

PR #3519, opened 2024-09-22
Create PeerMonitor class with Java FX

ItoroD:
Its been mentioned a while ago to create a java fx version of the peer monitor class. Here is a first draft of the peer montior class using JavaFX. I noticed a few things when I created this class with JavaFX.

- The `PeerMonitorFx` fails to start up and gives an error that says 

           > JavaFX runtime components are missing, and are required to run this application

> I found that, when we have a main inside the class where we extend the JavaFX `Application` class we get the error above because of how the `FXLauncher` works and the pre-checks for JavaFX components that happens when the main launcher sees that the application is a javafx application. See [here](https://edencoding.com/runtime-components-error/). Looks like this has been the case since JavaFx was seperated out of the Java runtime.

- I checked our `wallettemplate` module which has javafx as well. When I run the main in there, I get the same error
      

          > JavaFX runtime components are missing, and are required to run this application

- One work around I did for the `PeerMonitorFx` class in this PR was to create a second class with a main `PeerMonitorMain` which tricks the LauncherHelper into not realising the app is an instance of a JavaFX Application. (not an ideal solution)

- We might want to use modules i.e module-info.java to explicitly specify the JavaFx components required. ONly problem with modules is that when you have interdependent modules/packages like we have a bitcoinj, you might have to add a module-info file for every bitcoinj child module. Alternatively, a developer has to really know what he is doing to and always makes sure he has vm arguments like 

> _--module-path /path/to/javafx-sdk-14/lib --add-modules javafx.controls,javafx.fxml_

 in IDE or cmd before running bitcoinj FX applications.

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -1,5 +1,11 @@
 plugins {
     id 'java'
+    id 'org.openjfx.javafxplugin' version '0.0.14'
+}
+
+javafx {
+    version = '17'
+    modules = [ 'javafx.controls', 'javafx.fxml' ]
 }
 
 dependencies {
--- a/examples/src/main/java/org/bitcoinj/examples/PeerMonitorFx.java
+++ b/examples/src/main/java/org/bitcoinj/examples/PeerMonitorFx.java
@@ -1,0 +1,248 @@
+package org.bitcoinj.examples;
+
+import javafx.application.Application;
+import javafx.application.Platform;
+import javafx.beans.property.SimpleStringProperty;
+import javafx.concurrent.Task;
+import javafx.geometry.Insets;
+import javafx.scene.Scene;
+import javafx.scene.control.*;
+import javafx.scene.control.cell.PropertyValueFactory;
+import javafx.scene.layout.BorderPane;
+import javafx.scene.layout.HBox;
+import javafx.stage.Stage;
+import org.bitcoinj.base.BitcoinNetwork;
+import org.bitcoinj.base.Coin;
+import org.bitcoinj.base.Network;
+import org.bitcoinj.core.AddressMessage;
+import org.bitcoinj.core.Peer;
+import org.bitcoinj.core.PeerAddress;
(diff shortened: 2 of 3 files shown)


ItoroD (commented):
The sizing of the JavaFX UI window should be smaller. I will update this

ItoroD:
I have updated the size of the scene. 

@msgilligan any thoughts on the comments above

msgilligan:
> 
> * The `PeerMonitorFx` fails to start up and gives an error that says
>   ```
>        > JavaFX runtime components are missing, and are required to run this application
>   ```
> 
> * I checked our `wallettemplate` module which has javafx as well. When I run the main in there, I get the same error
>   ```
>       > JavaFX runtime components are missing, and are required to run this application
>   ```

Take a look at the latest version of the `wallettemplate` app. It is now a modular app and can be linked with `jlink` so does not need a separate "main" class.


> * We might want to use modules i.e org.beryx.jlink to explicitly specify the JavaFx components required. ONly problem with modules is...

I think before this PR is merged, we should move it to it's own bitcoinj submodule so we can add the `application` and `org.beryx.jlink` plugins, a `module-info.java`, etc.

Note that `PeerMonitorFX.java` can also be made into a JBang application and JBang will make it modular automatically. Adding the following lines to the top of the file makes it runnable as a JBang script:

```
//JAVA 17+
//DEPS org.openjfx:javafx-controls:21:${os.detected.jfxname}
//DEPS org.openjfx:javafx-graphics:21:${os.detected.jfxname}
//DEPS org.bitcoinj:bitcoinj-core:0.17
//DEPS org.slf4j:slf4j-jdk14:2.0.17
```

msgilligan:
@ItoroD See PR #3832 for changes I made to move this to its own module and some other minor improvements.

Let's see what @schildbach says about creating a new module for it and if he agrees, you can update this PR with some of the work from #3832.

PR #3525, opened 2024-09-27
graalvm.yml: Use setup-java GitHub Action

msgilligan:
graalvm.yml: Use setup_java GitHub Action

**UPDATED:**

We can now use the same setup_java GitHub Action
in gradle.yml and graalvm.yml. This commit switches
to the setup_java action from the setup-graalvm action
and also changes some variable names to more closely
match gradle.yml.

The setup_java Action at this time only supports _Oracle_ GraalVM which
is licensed under the GFTC license:
https://www.oracle.com/downloads/licenses/graal-free-license.html

See: https://github.com/actions/setup-java/issues/742

So the benefits of using the same action for JDK and GraalVM
have to be weighed against using only a non-FOSS version of GraalVM.

Oracle GraalVM has stopped providing updates under the free-beer license
for JDK 17, so I removed `17` from the matrix. It is also possible
to use the last version: `17.0.12`

The setup_graalvm action supports the Community Edition, Oracle GraalVM,
and some other options.

(It might even be possible/desirable to combine the
two workflows in the future, but let's keep this PR simple)

### DIFF
--- a/.github/workflows/graalvm.yml
+++ b/.github/workflows/graalvm.yml
@@ -9,20 +9,19 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-24.04, macOS-13]
-        java: [ '17', '21' ]
-        distribution: [ 'graalvm-community' ]
+        java: [ '21' ]
+        distribution: [ 'graalvm' ]
         gradle: ['8.12.1']
       fail-fast: false
     name: ${{ matrix.os }} JDK ${{ matrix.java }}.${{ matrix.distribution }}
     steps:
       - name: Git checkout
         uses: actions/checkout@v4
-      - name: Set up GraalVM
-        uses: graalvm/setup-graalvm@v1
+      - name: Set up JDK (GraalVM)
+        uses: actions/setup-java@v4
         with:
-          java-version: ${{ matrix.java }}
           distribution: ${{ matrix.distribution }}
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          java-version: ${{ matrix.java }}
       - name: Set up Gradle
         uses: gradle/actions/setup-gradle@v4
         with:


schildbach (approved):
Looks good. Should we backport this to 0.17?

msgilligan:
Although this ran successfully on GitHub actions when I first submitted it, as of today it is failing. Probably because of this issue: https://github.com/actions/setup-java/issues/692.  See the comment that says it is happening for `GraalVM`, too: https://github.com/actions/setup-java/issues/692#issuecomment-2417522553

schildbach:
Can you rebase this, mainly to trigger the builds?

msgilligan:
> Although this ran successfully on GitHub actions when I first submitted it, as of today it is failing. Probably because of this issue: [actions/setup-java#692](https://github.com/actions/setup-java/issues/692). See the comment that says it is happening for `GraalVM`, too: [actions/setup-java#692 (comment)](https://github.com/actions/setup-java/issues/692#issuecomment-2417522553)

This is because _Oracle_ GraalVM stopped making updates to their free-beer distribution of 17.0.x. I have updated the description (first comment) on this PR to discuss the issue.

@schildbach Perhaps we should just stick with the `setup_graalvm` action for now?

schildbach:
> Oracle GraalVM has stopped providing updates under the free-beer license for JDK 17, so I removed 17 from the matrix. It is also possible to use the last version: 17.0.12

Just for clarification: does this mean that instead of removing `17` we could change it to `17.0.12`? If so, why don't we prefer this?

msgilligan:
> Just for clarification: does this mean that instead of removing `17` we could change it to `17.0.12`?

Yes, we could. But then we'd be using something that from our perspective (i.e. as FOSS developers without a support contract) 17.x is unmaintained.

> If so, why don't we prefer this?

I would definitely prefer to use the `setup_java` action if we could use it why the Community Edition of GraalVM. I opened this issue: https://github.com/actions/setup-java/issues/742

I originally thought it would be OK to use Oracle GraalVM and switch to `setup_java` because I was thinking it was "open enough", but on more careful consideration I feel we'd be better off testing with Community Edition because it is truly FOSS and is usable in more places and by more projects.

As far as I know, nobody is using bitcoinj in production with GraalVM anywhere. if they were, it would be good to test with the version of GraalVM that they are using, and if that were the Oracle version, I would say we should test it.

But for now, until https://github.com/actions/setup-java/issues/742 is resolved, I'd say lets stick with Community Edition and using  different "setup actions" in our two workflows.

schildbach:
So we keep this open and unmerged for now?

schildbach:
We could do the `java-version`→`java` rename however.

msgilligan:
Yeah, I'll make a PR for the `java-version` → `java` rename. Then we can rebase this PR and leave it open.

If/when https://github.com/actions/setup-java/issues/742 is resolved, we can update this PR to use Community Edition.

msgilligan:
(`graalvm.yml` was merged into `build-gradle.yml` since this PR was made)

Closing in favor of:

*  PR #4228

PR #3534, opened 2024-10-04
Extract Conversion Logic and Overflow Check in ExchangeRate

panic08:
The logic of check overflow and conversion have been moved to separate methods. This will get rid of code duplication in `ExchangeRate` class

### DIFF
--- a/core/src/main/java/org/bitcoinj/utils/ExchangeRate.java
+++ b/core/src/main/java/org/bitcoinj/utils/ExchangeRate.java
@@ -53,11 +53,10 @@ public class ExchangeRate {
      */
     public Fiat coinToFiat(Coin convertCoin) {
         // Use BigInteger because it's much easier to maintain full precision without overflowing.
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/utils/ExchangeRate.java
```diff
@@ -68,19 +67,37 @@ public Fiat coinToFiat(Coin convertCoin) {
...
             return Coin.valueOf(converted.longValue());
         } catch (IllegalArgumentException x) {
             throw new ArithmeticException("Overflow: " + x.getMessage());
         }
     }
 
+    /**
+     * Checks for overflow in the BigInteger value.
+     */
+    private void checkOverflow(BigInteger value) {
```
schildbach, line 86: Perhaps add a throws clause here?

schildbach (changes requested):
Looks good. One comment inline.

panic08:
I also removed the `@throws` comment above the coinToFiat() method

Issue #3523, opened 2024-09-25
Script creationTime behavior may have changed. We need to verify behavior.


msgilligan:
In our refactoring, we may have changed the behavior of `Script` with respect to unknown creation times.

`@Nullable private final Instant creationTime` is never `null` despite the `@Nullable` annotation and the `public Optional<Instant> creationTime()` getter.

There are factory methods and deprecated constructors that set `creationTime` to `TimeUtils.currentTime()`. Is there a use-case where we need to represent the creation time as "unknown"?

In a private conversation, Andreas said:

> I'd assume that if a script doesn't associate a key then the creation time is unknown. Or if the creation time of an associated key is unknown itself. If I understand the field comment correctly, it's always about keys not the script itself

This recent PR is also relevant:

https://github.com/bitcoinj/bitcoinj/pull/3513

We should verify what the past behavior was and what the correct behavior should be and update the code and JavaDoc to match.

schildbach:
I looked at `release-0.16` and the behaviour is inconsistent:

- The constructor `public Script(byte[] programBytes, long creationTimeSeconds)` uses the supplied creation time.
- The constructor `public Script(byte[] programBytes)` initialized it to zero.
- The package-private constructor `Script(List<ScriptChunk> chunks)` is used only by `ScriptBuilder.build()` and inititalizes the creation time to the current time.

I'm not sure if we should try to preserve this behaviour.

schildbach:
`ScriptBuilder` has two related JavaDocs:

> If this is not set, the current time is used by the builder.

This was added in 838d14d027b63381ac0317fa51900a3f9035ad8d, after 0.16.

> The creation time will be the current time.

This was added in 15ee38aae08372d8524d2fd1b916ac3aca88d190, also after 0.16.

schildbach:
Investigated more into the usages of `Script.setCreationTimeSeconds()`, and it's only used for watching addresses like I suspected.

schildbach:
To sum it up, I think we should make clear that for generic case, scripts don't have a creation time and thus it's `null` not  current time. Only if a script is a scriptPubKey, it could be set to the creation date of the public/private keypair and thus the corresponding address.

schildbach:
Furthermore, we should perhaps think about removing `@Deprecated public Script(byte[] program)` again?

msgilligan:
> Investigated more into the usages of `Script.setCreationTimeSeconds()`, and it's only used for watching addresses like I suspected.

[...]

PR #3535, opened 2024-10-17
Script, ScriptBuilder: don't use current time as a default for creationTime

schildbach:
Don't use current time as a default for creationTime. Rather, leave it `null`.

In future, we might get rid of this field in the generic case, and use subclasses just for the watching wallets usecase.

See #3523 and #3541.

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/Script.java
+++ b/core/src/main/java/org/bitcoinj/script/Script.java
@@ -240,7 +240,7 @@ public class Script {
      * @return script that wraps the chunks
      */
     public static Script of(List<ScriptChunk> chunks) {
-        return of(chunks, TimeUtils.currentTime());
+        return of(chunks, null);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


schildbach:
I believe after this PR, the only behaviour change is when you call `ScriptBuilder.build()` without having called `.creationTime()`, the resulting script would now not have the current time any more. None of the unit tests seem concerned about this.

schildbach:
I just pushed a commit with JavaDoc changes to creation time for both the related methods and fields.

schildbach:
If this has been merged to 0.17, we should not forget to port this to master as well (probably minus the deprecated methods).

schildbach:
> I just pushed a commit with JavaDoc changes to creation time for both the related methods and fields.

Extracted this commit to its own PR: #3541

schildbach:
I rebased this PR, updated its description and switched it to "ready for review".

PR #3542, opened 2024-11-08
SegwitAddress: allow instantiation of P2A (pay-to-anchor) address

johnzweng:
Hi! 🙂 

## Overview

As you may remember, in pullrequest #2663, I introduced a length check for the witness programs of v1 addresses, allowing only 32-byte long programs. Reviewing our discussion from back then, we decided to merge this check even though, according to [BIP 341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#script-validation-rules), such v1 witness programs aren’t prohibited. Instead, they produce "anyone-can-spend" outputs (per consensus rules).

Additionally spending from these outputs is disallowed by policy (non-standard), so we came to the conclusion that it probably might not make sense to instantiate such `SegwitAddress` (especially when bitcoinJ is used in the context of a client).

## Change in this PR

In this PR, I’ve relaxed the restriction slightly to permit one specific v1 witness program shorter than 32 bytes: the P2A (pay-to-anchor) output script.

A P2A output script is a segwit output with a witness version of `1` and a hardcoded witness program of `[0x4e, 0x73]`.

You can find the definition of this output type [here in Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/455fca86cfada1823aa28615b5683f9dc73dbb9a/src/script/script.cpp#L216-L222). Since the merge of https://github.com/bitcoin/bitcoin/pull/30352 ([released in 28.0](https://bitcoincore.org/en/releases/28.0/)) spending from this specific output script is now allowed by policy.

## Rationale

Following our discussion when we introduced the taproot witness program length check in #2663, where we talked that bitcoinJ should follow policy, and the fact that spending from P2A is now allowed by policy, I allowed creation of this address. 

## References

- Discussion about ephemeral anchor outputs on delving: https://delvingbitcoin.org/t/segwit-ephemeral-anchors/160
- The P2A address on mempool.space: [bc1pfeessrawgf](https://mempool.space/address/bc1pfeessrawgf)

### DIFF
--- a/base/src/main/java/org/bitcoinj/base/SegwitAddress.java
+++ b/base/src/main/java/org/bitcoinj/base/SegwitAddress.java
@@ -50,6 +50,11 @@ public class SegwitAddress implements Address {
     public static final int WITNESS_PROGRAM_LENGTH_TR = 32;
     public static final int WITNESS_PROGRAM_MIN_LENGTH = 2;
     public static final int WITNESS_PROGRAM_MAX_LENGTH = 40;
+    /**
+     * <p>P2A (pay-to-anchor) Script</p>
+     * Defined <a href="https://github.com/bitcoin/bitcoin/blob/455fca86cfada1823aa28615b5683f9dc73dbb9a/src/script/script.cpp#L216-L222">here in Bitcoin Core</a>:
+     */
+    private static final byte[] P2A_SCRIPT = {0x4e, 0x73};
 
 
     /**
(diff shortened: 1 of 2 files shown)


schildbach (approved):
Looks good!

johnzweng:
Thank you for looking into it so quickly.

schildbach:
Merged to master.

schildbach:
This didn't make it into v0.17-rc1, but if I get another review I think we could add it to release (maybe after a short rc2 cycle) or to the inevitable 0.17.1.

base/src/main/java/org/bitcoinj/base/SegwitAddress.java
```diff
@@ -166,14 +171,26 @@ private SegwitAddress(Network network, int witnessVersion, byte[] witnessProgram
...
                     "Invalid length for address version 1: " + witnessProgram.length);
         this.network = normalizeNetwork(Objects.requireNonNull(network));
         this.witnessVersion = (short) witnessVersion;
         this.witnessProgram = Objects.requireNonNull(witnessProgram);
     }
 
+    /**
+     * Check if this script is a P2A (pay-to-anchor) output script.
+     *
+     * @return <code>true</code> if the given wittnessProgram and wittnessVersion represent a Pay-to-anchor output script, <code>false</code> otherwise
```
schildbach, line 188: Nitpick: it's "witness"

schildbach: Fixed on master via f0ddbcb91c2ee66595a19f74bffbe6a620d1e29d.

base/src/main/java/org/bitcoinj/base/SegwitAddress.java
```diff
@@ -50,6 +50,11 @@ public class SegwitAddress implements Address {
     public static final int WITNESS_PROGRAM_LENGTH_TR = 32;
     public static final int WITNESS_PROGRAM_MIN_LENGTH = 2;
     public static final int WITNESS_PROGRAM_MAX_LENGTH = 40;
+    /**
+     * <p>P2A (pay-to-anchor) Script</p>
```
msgilligan, line 54: We're moving away from XHTML-style Javadoc. There should just be a `<p>` tag here following the first line.

But since we're doing a back port, let's leave it as it is to keep the diffs simpler. We can fix it on `master` later.

msgilligan (commented):
Looks like I had a partial review here "in limbo"

msgilligan:
@schildbach did this make it into v0.17.1?

schildbach:
> @schildbach did this make it into v0.17.1?

I initially picked it, but then took it out when we were afraid some of the picked changes need more testing before merged to a stable release.

msgilligan:
Shall we try to add it to 0.17.2?

@schildbach I tagged it with milestone 0.17.2. We can change it to something else if necessary.

Issue #3388, opened 2024-04-28
Reduce (default) verbosity of PeerGroup logging.


msgilligan:
PeerGroup logging is perhaps too verbose by default. We should probably do one or more of the following:

1. Change the default log level for the PeerGroup messages
2. Review the current log message and change the log level of certain messages
3. Possibly consider a context-specific verbosity flag for PeerGroup
4. Document how to adjust PeerGroup logging.

msgilligan:
We've merged a handful of PRs that should help with this:

* PR #3590
* PR #3591
* PR #3592
* PR #3593
* PR #3594

msgilligan:
I'm running WalletTemplate on MAINNET after merging the above PRs and they are definitely an improvement.

However, I'm still seeing at least two messages that seem unnecessarily verbose:

```
22:21:03.624 42 PeerGroup$2.go: Waiting 9999 ms before next connect attempt to [178.131.96.87]:18333
22:21:03.624 42 PeerGroup$2.go: Waiting 9999 ms before next connect attempt to [178.131.96.87]:18333
22:21:03.624 42 PeerGroup$2.go: Waiting 9999 ms before next connect attempt to [178.131.96.87]:18333
22:21:03.624 42 PeerGroup$2.go: Waiting 9999 ms before next connect attempt to [178.131.96.87]:18333
22:21:03.624 42 PeerGroup$2.go: Waiting 9999 ms before next connect attempt to [178.131.96.87]:18333
22:21:03.625 42 PeerGroup$2.go: Waiting 9998 ms before next connect attempt to [178.131.96.87]:18333
```

This one should probably be changed from `log.info` to `log.debug`.

And:

```
22:25:04.972 44 Peer.setBloomFilter: Peer{[72.211.1.222]:18333, version=70016, subVer=/Satoshi:25.0.0/, services=NETWORK, WITNESS, NETWORK_LIMITED, time=2025-02-19T22:25:05Z, height=3803483}: Peer does not support bloom filtering.
22:25:04.972 44 PeerGroup.handlePeerDeath: [72.211.1.222]:18333: Peer died      (6 connected, 19 pending, 12 max)
22:25:04.972 44 ConnectionHandler.writeBytes: Error writing message to connection, closing connection
java.nio.channels.CancelledKeyException
        at java.base/sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:75)
[...]

PR #3551, opened 2024-12-16
PeerGroup: Reduce logging verbosity

ipolonskij:
This PR addresses #3388. Out of the suggested approached, option 2 was selected. Log messages have been reviewed and the log level of certain messages has been reduced.

Looking forward for your feedback and suggestions.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/PeerGroup.java
+++ b/core/src/main/java/org/bitcoinj/core/PeerGroup.java
@@ -329,7 +329,7 @@ public class PeerGroup implements TransactionBroadcaster {
                 if (numAdded >= MAX_ADDRESSES_PER_ADDR_MESSAGE)
                     break;
             }
-            log.info("{} gossiped {} addresses, added {} of them to the inactive pool", peer.getAddress(),
+            log.debug("{} gossiped {} addresses, added {} of them to the inactive pool", peer.getAddress(),
                     addresses.size(), numAdded);
         }
     }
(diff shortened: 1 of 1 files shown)


msgilligan:
@ipolonskij For some reason I'm just noticing this now. Thanks for the PR.

@schildbach What are your thoughts on this PR and the related issue (#3388)?

schildbach:
Are these messages really a problem? I'd assume (and experience in real) they get logged just once per peer, which I consider not excessive.

msgilligan:
> Are these messages really a problem? I'd assume (and experience in real) they get logged just once per peer, which I consider not excessive.

I haven't tested this PR, so I'm not sure how much this PR changes things. But I know that I personally find the default verbosity of the logs of-putting and I know of at least a handful of cases where new **bitcoinj** developers were confused or annoyed by them.

Issue #3388 takes more of a big picture or high-level view. I feel pretty strongly that we need to do _something_.

schildbach:
Usually it's the exceptions (with stack trace) that do off-put people, including me. So I just contributed two PRs for those: #3590 and #3591.

Regarding the logging in PeerGroup (this PR), I find it quite useful. The one message I'd like to see reduced is this:

```
PeerGroup: [PeerGroup Thread] Waiting 8517 ms before next connect attempt to [75.155.110.24]:18333
```

However I'm pretty sure the reason for this appearing so often is a bug, which we would simply hide by hiding this message. If it were to appear only once after a peer death, I'd assume it would be fine. I'd prefer fixing the bug rather than hiding, it just never seemed serious enough to me.

Other messages in this PR, e.g. about peers connecting and disconnecting and the download peer changing I consider important enough for log level info.

ipolonskij:
Sorry, I missed your comments.

To be very honest, the log level is inho a question of taste. it is very hard to define objective criterias for the log level. Therefore I do not claim to have chosen all the right log level in this PR, rather my preferences given the issue. If in doubt maybe it's better to be rather verbose than too concise?

msgilligan:
> Sorry, I missed your comments.
> 
> To be very honest, the log level is imho a question of taste. it is very hard to define objective criterias for the log level. Therefore I do not claim to have chosen all the right log level in this PR, rather my preferences given the issue. If in doubt maybe it's better to be rather verbose than too concise?

This is a tricky issue, that's (one reason) why we are moving slow on it.

See Issue #3388.

Issue #1820, opened 2019-04-24
WalletTemplate: Replace Font Awesome usage with 2 icons


msgilligan:
Currently the `wallettemplate` module depends upon FontAwesome (`'de.jensd:fontawesomefx:8.0.0'`) for two small icons: 

1. The 'copy' icon (for copying the displayed Bitcoin address to the clipboard)
2. The 'QR code' icon (for showing the displayed Bitcoin address as a QR Code)

FontAwesome allows `Label` components to be creating using these glyphs. I propose removing the FontAwesome dependency and replacing the glyphs with images and changing from `Label` to `ImageView` (or maybe `Button`)

I think the Font Awesome free license will allow us to use the same images, so that's probably the easiest way to go.

msgilligan:
If we do this `wallettemplate` will only depend on `com.google.zxing:core` and `slf4j`!

schildbach:
Sounds reasonable. We can also use material design icons if it fits the style.

msgilligan:
Thanks @xotonic!  They don't have to look exactly the same as before, but they should look nice. I'm also happy to approve (if @schildbach agrees) a PR with the code changes and temporary icons that can be updated later.

schildbach:
We could use https://pictogrammers.com/library/mdi/icon/content-copy/ and https://pictogrammers.com/library/mdi/icon/qrcode/

These material design icons are available for download in various vector and raster formats.

The `fontawesomefx` dependency is a problematic one, as the project is not maintained any more and their latest versions are only available via Bintray (which as been sunset a while ago). We should get rid of it.

msgilligan:
@ItoroD Do you think you might want to try fixing this?

ItoroD:
Yes, will be happy to take a look.

msgilligan:
Fixed by PR #3567

PR #3566, opened 2025-01-29
wallettemplate: Modular build using jlink plugin

msgilligan:
I don't know why fontawesomefx is breaking the build. It worked in PR #1779.

I guess the first/next step is to do the removal of fontawesomefx.

PR #3567, opened 2025-01-29
WalletTemplate: Replace-Font-Awesome

ItoroD:
This PR addresses issue #1820


schildbach:
Awesome! (-:

Can you say where you've got the icons from? (Because of licensing.)

ItoroD:
Thanks @schildbach . I got it from pictogrammers see the links below

https://pictogrammers.com/library/mdi/icon/content-copy/
https://pictogrammers.com/library/mdi/icon/qrcode/

based on your comment from [here](https://github.com/bitcoinj/bitcoinj/issues/1820#issuecomment-1637060238)

schildbach (changes requested):
I tested this PR and it generally works. A couple of remarks however:

- If your goal is to get rid of fontawesome, then you need to remove it from `wallettemplate/src/main/java/module-info.java` as well.
- We could however do two commits: one for replacing the icons, and one for removing fontawesome
- To the right of the icons there are strings like "<COP…" (see your screenshot). I didn't check, is this a regression to before? Then perhaps try to remove it?

ItoroD:
> If your goal is to get rid of fontawesome, then you need to remove it from wallettemplate/src/main/java/module-info.java as well.


@schildbach This PR was based off main which does not have the module-info.java in template yet. I see @msgilligan had already created a PR for that here #3566 . My thinking was that if this is merged, then #3566 can be updated to take out fontawesome.

> To the right of the icons there are strings like "<COP…" (see your screenshot). I didn't check, is this a regression to before? Then perhaps try to remove it?

Yes I can take out the weird text there. Although that is how its always been. I also wondered if it was intentional

schildbach:
> Yes I can take out the weird text there. Although that is how its always been. I also wondered if it was intentional

Ah no, in this case please leave it as is. Your PR is not about fixing this.

ItoroD:
> Ah no, in this case please leave it as is. Your PR is not about fixing this, I only want to avoid regressions.

All right. Any other thing I need to do here ?

msgilligan:
> Any other thing I need to do here ?

It looks good to me. But there are two things I think you should do:

1. Squash the PR down to a single commit
2. Add some documentation about where the .png images came from. I think the best place would as an XML comment inside the FXML, just above where each one is used.  @schildbach What do you think?

schildbach:
I'm on this already.

schildbach:
Merged!

msgilligan:
> > To the right of the icons there are strings like "<COP…" (see your screenshot). I didn't check, is this a regression to before? Then perhaps try to remove it?
> 
> Yes I can take out the weird text there. Although that is how its always been. I also wondered if it was intentional

I don't remember the `<COPY>` and `<QR>` labels being displayed before. Perhaps something changed in JavaFX that is now causing them to be shown. But we should fix that in another PR.

ItoroD:
So in the fxml file there is a text property that causes the `<COPY> and <QR>` to show 

` text="&lt;COPY&gt;" >`

```
<Label fx:id="copyWidget" onMouseClicked="#copyWidgetClicked" text="&lt;COPY&gt;" />
<Label fx:id="qrCode" onMouseClicked="#showQRCode" text="&lt;QR&gt;" />
```

This existed in code before any change was made. Taking the text property out stops that from displaying in the UI

schildbach:
> I want to clarify if the `<COPY> and <QR>` are supposed to be there or not?

I'd say not. You want to submit PR?

ItoroD:
Yes. will create one

ItoroD:
See pr [here](https://github.com/bitcoinj/bitcoinj/pull/3568)

PR #3569, opened 2025-01-29
wallettemplate/build.gradle: better slf4j dependency declaration

msgilligan:
There is a compile-time dependency on slf4j-api and a runtimeOnly dependency on slf4j-jdk14. This commit updates Gradle to declare this explicitly.


schildbach:
Merged.

PR #3576, opened 2025-02-05
Transaction: when serializing a transaction re-use buffer as much as possible

schildbach:
This introduces a ByteBuffer-based `write()`, which calls `write()` on TransactionInput, TransactionOutput, etc. Allocation of intermediate buffers is avoided.

This is used for getTxId(), getWTxId() for faster id generation.

Subclass `TransactionTest.HugeDeclaredSizeTransaction` is also changed to `write()`.

~If this is merged, I'll backport it to 0.17.~

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -53,6 +53,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.math.BigInteger;
 import java.math.RoundingMode;
+import java.nio.BufferOverflowException;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.time.Instant;
@@ -337,13 +338,9 @@ public class Transaction extends BaseMessage {
             if (!hasWitnesses() && cachedWTxId != null) {
                 cachedTxId = cachedWTxId;
             } else {
-                ByteArrayOutputStream baos = new ByteArrayOutputStream();
-                try {
-                    bitcoinSerializeToStream(baos, false);
-                } catch (IOException e) {
-                    throw new RuntimeException(e); // cannot happen
-                }
-                cachedTxId = Sha256Hash.wrapReversed(Sha256Hash.hashTwice(baos.toByteArray()));
+                ByteBuffer buf = ByteBuffer.allocate(messageSize(false));
+                write(buf, false);
+                cachedTxId = Sha256Hash.wrapReversed(Sha256Hash.hashTwice(buf.array()));
             }
         }
         return cachedTxId;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1424,33 +1423,47 @@ protected void bitcoinSerializeToStream(OutputStream stream) throws IOException
...
-     * <a href="https://en.bitcoin.it/wiki/Protocol_documentation#tx">classic format</a>, depending on if segwit is
-     * desired.
+     * Write this transaction into the given buffer.
+     *
+     * @param buf buffer to write into
+     * @return the buffer
+     * @throws BufferOverflowException if the transaction doesn't fit the remaining buffer
      */
-    protected void bitcoinSerializeToStream(OutputStream stream, boolean useSegwit) throws IOException {
+    protected ByteBuffer write(ByteBuffer buf, boolean useSegwit) throws BufferOverflowException {
```
msgilligan, line 1432: I think this can (and should) be made `private`

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1424,33 +1423,47 @@ protected void bitcoinSerializeToStream(OutputStream stream) throws IOException
...
+                in.getWitness().write(buf);
         }
         // lock_time
-        writeInt32LE(vLockTime.rawValue(), stream);
+        ByteUtils.writeInt32LE(vLockTime.rawValue(), buf);
+        return buf;
+    }
+
+    /**
+     * Serialize according to <a href="https://github.com/bitcoin/bips/blob/master/bip-0144.mediawiki">BIP144</a> or the
```
msgilligan, line 1459: This comment should probably be moved up to the ByteBuffer-based `write` which actually implements the BIP.

msgilligan (changes requested):
Minor changes requested.

schildbach:
First of all: rebased.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: It's protected, because it's overridden in `HugeDeclaredSizeTransaction` in a similar fashion that the previous `.bitcoinSerializeToStream`.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: Changed: https://github.com/bitcoinj/bitcoinj/compare/79b8257961f637276473a2b82d6738ff8baffdc5..e20e5948dd9714e5160815b2d36f2a789ca7ea1f

schildbach:
I guess I changed my mind about this PR and *not* backport it to 0.17. Since we cache txids now I think there is not much performance benefit to speeding up serialization now.

And for 0.18, I think we'd want to open up the `.write()` method to public like on our other protocol data classes.

schildbach:
Closing in favor of #3678, #3679 and the various PRs we already merged in the past few days.

PR #3577, opened 2025-02-05
TransactionOutput: make field `scriptBytes` immutable

schildbach:

PR #3578, opened 2025-02-05
TransactionOutput: make field `value` immutable

schildbach:
Because tweaking is necessary for fee calculation logic, these usages
have been changed to produce new TransactionOutputs instead and
replace them in transactions as needed.

This PR is a child of #3577, but could be changed to merge individually.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -1365,8 +1365,9 @@ public class Transaction extends BaseMessage {
                             BigInteger.valueOf(output.getValue().getValue()),
                             bosHashOutputs
                     );
-                    bosHashOutputs.write(VarInt.of(output.getScriptBytes().length).serialize());
-                    bosHashOutputs.write(output.getScriptBytes());
+                    byte[] scriptBytes = output.getScriptBytes();
+                    bosHashOutputs.write(VarInt.of(scriptBytes.length).serialize());
+                    bosHashOutputs.write(scriptBytes);
                 }
                 hashOutputs = Sha256Hash.hashTwice(bosHashOutputs.toByteArray());
             } else if (basicSigHashType == SigHash.SINGLE.value && inputIndex < outputs.size()) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1011,6 +1011,20 @@ public TransactionOutput addOutput(TransactionOutput to) {
...
 
+    /**
+     * Replaces an already added output.
+     *
+     * @param index  index of output to replace
+     * @param output output to replace with
+     */
+    public void replaceOutput(int index, TransactionOutput output) {
+        TransactionOutput oldOutput = outputs.remove(index);
+        checkState(oldOutput.isAvailableForSpending());
```
msgilligan, line 1022: Why are you checking this? Maybe we should add a comment explaining why?

core/src/main/java/org/bitcoinj/core/TransactionOutput.java
```diff
@@ -169,14 +169,14 @@ public Coin getValue() {
...
     /**
-     * Sets the value of this output.
+     * Returns a clone of this output, with given value. The typical use case is fee calculation.
+     *
+     * @param value value for the clone
+     * @return clone of output, with given value
      */
-    public void setValue(Coin value) {
+    public TransactionOutput withValue(Coin value) {
         Objects.requireNonNull(value);
```
msgilligan, line 178: non-null `value` is checked (implicitly) by the called constructor, so this could be removed.

msgilligan (approved):
This is nicely done! Approve!

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: I thought that if people start replacing outputs in buried in their wallet things will go wrong. Those txns should really be considered immutable.

msgilligan: Makes sense. I think you should say something like that in a comment.

core/src/main/java/org/bitcoinj/core/TransactionOutput.java
schildbach: True, but isn't it convenient if the called method is on top of the stack trace?

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: Will do.

core/src/main/java/org/bitcoinj/core/TransactionOutput.java
msgilligan: I wonder if the JIT is smart enough to figure out that this only needs to be done once in that use case?

schildbach: I'd assume no, but is that really relevant for performance? It's just a null check.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: I added this: https://github.com/bitcoinj/bitcoinj/compare/2d465f91971cb35818b217647229db29ce22ce7d..9daf18eb0fd5d2b3bf2463f9c24522d11d8c88af

schildbach:
Merged.

PR #3579, opened 2025-02-06
TransactionInput: make field `sequence` immutable

schildbach:
Because tweaking is necessary in cases like unit tests, these usages have been changed to produce new inputs instead and replace them in transactions as needed.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionInput.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionInput.java
@@ -73,7 +73,7 @@ public class TransactionInput {
     @Nullable private Transaction parent;
 
     // Allows for altering transactions after they were broadcast. Values below NO_SEQUENCE-1 mean it can be altered.
-    private long sequence;
+    private final long sequence;
     // Data needed to connect to the output of the transaction we're gathering coins from.
     private TransactionOutPoint outpoint;
     // The "script bytes" might not actually be a script. In coinbase transactions where new coins are minted there
@@ -124,18 +124,32 @@ public class TransactionInput {
     }
 
     public TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBytes,
+                            TransactionOutPoint outpoint, long sequence) {
+        this(parentTransaction, scriptBytes, outpoint, sequence, null);
+    }
+
+    public TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBytes,
                             TransactionOutPoint outpoint, @Nullable Coin value) {
         this(parentTransaction, scriptBytes, outpoint, NO_SEQUENCE, value);
     }
 
-    private TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBytes,
-                            TransactionOutPoint outpoint, long sequence, @Nullable Coin value) {
+    /** internal use only */
+    public TransactionInput(Transaction parentTransaction, byte[] scriptBytes, TransactionOutPoint outpoint,
+                            long sequence, @Nullable Coin value) {
+        this(parentTransaction, null, scriptBytes, outpoint, sequence, value, null);
+    }
+
+    private TransactionInput(@Nullable Transaction parentTransaction, @Nullable Script scriptSig, byte[] scriptBytes,
+                            TransactionOutPoint outpoint, long sequence, @Nullable Coin value,
+                            @Nullable TransactionWitness witness) {
         checkArgument(value == null || value.signum() >= 0, () -> "value out of range: " + value);
         parent = parentTransaction;
-        this.scriptBytes = scriptBytes;
-        this.outpoint = outpoint;
+        this.scriptSig = scriptSig != null ? new WeakReference<>(scriptSig) : null;
+        this.scriptBytes = Objects.requireNonNull(scriptBytes);
+        this.outpoint = Objects.requireNonNull(outpoint);
         this.sequence = sequence;
         this.value = value;
+        this.witness = witness;
     }
 
     /**
@@ -143,12 +157,13 @@ public class TransactionInput {
      */
     TransactionInput(Transaction parentTransaction, TransactionOutput output) {
         this(parentTransaction,
-                EMPTY_ARRAY,
+                null, EMPTY_ARRAY,
                 output.getParentTransaction() != null ?
                         new TransactionOutPoint(output.getIndex(), output.getParentTransaction()) :
                         new TransactionOutPoint(output),
                 NO_SEQUENCE,
-                output.getValue());
+                output.getValue(),
+                null);
     }
 
     /**
@@ -239,15 +254,22 @@ public class TransactionInput {
     }
 
     /**
+     * Returns a clone of this input, with a given sequence number.
+     * <p>
      * Sequence numbers allow participants in a multi-party transaction signing protocol to create new versions of the
      * transaction independently of each other. Newer versions of a transaction can replace an existing version that's
      * in nodes memory pools if the existing version is time locked. See the Contracts page on the Bitcoin wiki for
      * examples of how you can use this feature to build contract protocols.
+     *
+     * @param sequence sequence number for the clone
+     * @return clone of input, with given sequence number
      */
-    public void setSequenceNumber(long sequence) {
+    public TransactionInput withSequence(long sequence) {
         checkArgument(sequence >= 0 && sequence <= ByteUtils.MAX_UNSIGNED_INTEGER, () ->
                 "sequence out of range: " + sequence);
-        this.sequence = sequence;
+        Script scriptSig = this.scriptSig != null ? this.scriptSig.get() : null;
+        return new TransactionInput(this.parent, scriptSig, this.scriptBytes, this.outpoint, sequence, this.value,
+                this.witness);
     }
 
     /**
--- a/core/src/test/java/org/bitcoinj/wallet/DefaultRiskAnalysisTest.java
+++ b/core/src/test/java/org/bitcoinj/wallet/DefaultRiskAnalysisTest.java
@@ -90,7 +90,7 @@ public class DefaultRiskAnalysisTest {
         assertNull(analysis.getNonFinal());
 
         // Set a sequence number on the input to make it genuinely non-final. Verify it's risky.
-        input.setSequenceNumber(TransactionInput.NO_SEQUENCE - 1);
+        tx.replaceInput(0, input.withSequence(TransactionInput.NO_SEQUENCE - 1));
         analysis = DefaultRiskAnalysis.FACTORY.create(wallet, tx, NO_DEPS);
         assertEquals(RiskAnalysis.Result.NON_FINAL, analysis.analyze());
         assertEquals(tx, analysis.getNonFinal());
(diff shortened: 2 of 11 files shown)


core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -115,40 +115,45 @@ public static TransactionInput read(ByteBuffer payload, Transaction parentTransa
...
-        this(parentTransaction, scriptBytes, outpoint, NO_SEQUENCE, value);
+        this(parentTransaction, null, scriptBytes, outpoint, NO_SEQUENCE, value, null);
     }
 
-    private TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBytes,
-                            TransactionOutPoint outpoint, long sequence, @Nullable Coin value) {
+    /** used internally */
+    public TransactionInput(@Nullable Transaction parentTransaction, @Nullable Script scriptSig, byte[] scriptBytes,
+                            TransactionOutPoint outpoint, long sequence, @Nullable Coin value,
+                            @Nullable TransactionWitness witness) {
```
msgilligan, line 134: I suggest making this 7-arg constructor private and add a 5-arg constructor specifically for WalletProtobufSerializer and mark it as `/** internal use only **/`.

(We could even make the 5-arg constructor private and then put a static method in an internal-only util class if we really wanted to lock it down.)

msgilligan (changes requested):
This looks good! But, I think we should make the constructor changes (private 7-arg constructor, etc) before merging.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -115,40 +115,45 @@ public static TransactionInput read(ByteBuffer payload, Transaction parentTransa
...
         checkArgument(value == null || value.signum() >= 0, () -> "value out of range: " + value);
         parent = parentTransaction;
-        this.scriptBytes = scriptBytes;
+        this.scriptSig = scriptSig != null ? new WeakReference<>(scriptSig) : null;
+        this.scriptBytes = Objects.requireNonNull(scriptBytes);
         this.outpoint = outpoint;
```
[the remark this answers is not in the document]
schildbach: Changed: https://github.com/bitcoinj/bitcoinj/compare/6ee35cf0b92f935d89ec8ea6a6945085d157b398..091a578e3ae22f7aafd331505d8b1d840eba18ca

core/src/main/java/org/bitcoinj/core/TransactionInput.java
schildbach: Changed: https://github.com/bitcoinj/bitcoinj/compare/6ee35cf0b92f935d89ec8ea6a6945085d157b398..091a578e3ae22f7aafd331505d8b1d840eba18ca

core/src/test/java/org/bitcoinj/wallet/DefaultRiskAnalysisTest.java
```diff
@@ -104,7 +104,8 @@ public void nonFinal() {
     @Test
     public void selfCreatedAreNotRisky() {
         Transaction tx = new Transaction();
-        tx.addInput(MAINNET.getGenesisBlock().getTransactions().get(0).getOutput(0)).setSequenceNumber(1);
+        TransactionInput input = tx.addInput(MAINNET.getGenesisBlock().getTransactions().get(0).getOutput(0));
+        tx.replaceInput(0, input.withSequence(1));
```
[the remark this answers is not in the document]
schildbach: Changed: https://github.com/bitcoinj/bitcoinj/compare/091a578e3ae22f7aafd331505d8b1d840eba18ca..15e38c5334843f47fcdad7233860e0a4117a9f4c

integration-test/src/test/java/org/bitcoinj/core/PeerTest.java
```diff
@@ -732,8 +732,7 @@ private void checkTimeLockedDependency(boolean shouldAccept) throws Exception {
         t2.setLockTime(999999);
         // Add a fake input to t3 that goes nowhere.
         Sha256Hash t3 = Sha256Hash.of("abc".getBytes(StandardCharsets.UTF_8));
-        t2.addInput(new TransactionInput(t2, new byte[]{}, new TransactionOutPoint(0, t3)));
-        t2.getInput(0).setSequenceNumber(0xDEADBEEF);
+        t2.addInput(new TransactionInput(t2, new byte[]{}, new TransactionOutPoint(0, t3)).withSequence(0xDEADBEEF));
```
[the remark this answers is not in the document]
schildbach: I'm not sure about this. Added every permutation of arguments as a constructor isn't sustainable. I had hoped that the `with*` methods are something that someday could lead to a builder for TransactionInput (though I expect this to change).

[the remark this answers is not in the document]
msgilligan: Yeah, Java lacks a good mechanism (e.g. named arguments) to control constructor explosion. I'm also not a fan of the builder pattern (though I agree it is sometimes necessary.) Generally, I prefer having a few too many constructors than most of the common alternatives.

In this case, I'll leave it to your judgement what you want to do.

core/src/test/java/org/bitcoinj/wallet/DefaultRiskAnalysisTest.java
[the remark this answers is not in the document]
msgilligan: The change is a slight improvement, but I think the code would be much better with a constructor that takes a sequence. I guess longer-term some form of builder could solve the problem. But 1-3 extra constructors is IMO better/simpler than moving to a builder.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -115,40 +115,50 @@ public static TransactionInput read(ByteBuffer payload, Transaction parentTransa
         TransactionOutPoint outpoint = TransactionOutPoint.read(payload);
         byte[] scriptBytes = Buffers.readLengthPrefixedBytes(payload);
         long sequence = ByteUtils.readUint32(payload);
-        return new TransactionInput(parentTransaction, scriptBytes, outpoint, sequence, null);
+        return new TransactionInput(parentTransaction, null, scriptBytes, outpoint, sequence, null, null);
```
msgilligan, line 118: I would consider using the 5-argument "internal use only" constructor in these 3 places.

msgilligan (changes requested):
One minor comment (use the "internal" constructor in 3 places), and I' hoping you'll consider one more time adding a constructor that takes `sequence`

integration-test/src/test/java/org/bitcoinj/core/PeerTest.java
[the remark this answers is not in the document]
schildbach: Changed in https://github.com/bitcoinj/bitcoinj/compare/15e38c5334843f47fcdad7233860e0a4117a9f4c..542f34cf7fa52ba065aa0e4650c3bddc6aba0e12.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
schildbach: Changed in https://github.com/bitcoinj/bitcoinj/compare/542f34cf7fa52ba065aa0e4650c3bddc6aba0e12..0b597887e953c7445e2ea0ab6c750143f30d882d, but what are the other two places?

core/src/test/java/org/bitcoinj/wallet/DefaultRiskAnalysisTest.java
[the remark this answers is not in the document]
schildbach: I don't understand your comment in this particular context. Are you talking about another variant of `tx.addInput()`?

core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -115,40 +115,55 @@ public static TransactionInput read(ByteBuffer payload, Transaction parentTransa
...
         byte[] scriptBytes = Buffers.readLengthPrefixedBytes(payload);
         long sequence = ByteUtils.readUint32(payload);
-        return new TransactionInput(parentTransaction, scriptBytes, outpoint, sequence, null);
+        return new TransactionInput(parentTransaction, scriptBytes, outpoint, sequence);
     }
 
     public TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBytes,
                             TransactionOutPoint outpoint) {
-        this(parentTransaction, scriptBytes, outpoint, NO_SEQUENCE, null);
+        this(parentTransaction, null, scriptBytes, outpoint, NO_SEQUENCE, null, null);
```
msgilligan, line 123: Can call the 5-arg constructor.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -115,40 +115,55 @@ public static TransactionInput read(ByteBuffer payload, Transaction parentTransa
...
+    }
+
+    public TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBytes,
+                            TransactionOutPoint outpoint, long sequence) {
+        this(parentTransaction, null, scriptBytes, outpoint, sequence, null, null);
```
msgilligan, line 128: Can call the 5-arg constructor.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
```diff
@@ -115,40 +115,55 @@ public static TransactionInput read(ByteBuffer payload, Transaction parentTransa
...
     }
 
     public TransactionInput(@Nullable Transaction parentTransaction, byte[] scriptBytes,
                             TransactionOutPoint outpoint, @Nullable Coin value) {
-        this(parentTransaction, scriptBytes, outpoint, NO_SEQUENCE, value);
+        this(parentTransaction, null, scriptBytes, outpoint, NO_SEQUENCE, value, null);
```
msgilligan, line 133: Can call the 5-arg constructor.

core/src/test/java/org/bitcoinj/wallet/DefaultRiskAnalysisTest.java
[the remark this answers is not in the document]
msgilligan: I think so. I don't like how we're using `add` followed by `replace`.  We' re mutating both the input and the transaction as part of tx/txInput creation. It seems there should be a way to do this without the 2 mutations.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
schildbach: I think the reasoning was we should migrate these constructors to usecase-specific static constructors anyway and the only one that remains is the private one.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
schildbach: Changed in https://github.com/bitcoinj/bitcoinj/compare/0b597887e953c7445e2ea0ab6c750143f30d882d..249b2a98453de7dfb90a36f10fc6b7dbf72af04d.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
schildbach: Changed in https://github.com/bitcoinj/bitcoinj/compare/0b597887e953c7445e2ea0ab6c750143f30d882d..249b2a98453de7dfb90a36f10fc6b7dbf72af04d.

core/src/main/java/org/bitcoinj/core/TransactionInput.java
schildbach: Changed in https://github.com/bitcoinj/bitcoinj/compare/0b597887e953c7445e2ea0ab6c750143f30d882d..249b2a98453de7dfb90a36f10fc6b7dbf72af04d.

core/src/test/java/org/bitcoinj/wallet/DefaultRiskAnalysisTest.java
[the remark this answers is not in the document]
schildbach: I think I'll prefer to defer this to our TransactionBuilder refactoring. I'm pretty sure at least for transactions we'll have one. There is no way we can do with mutations of constructors only.

schildbach:
I decided to merge this.

PR #3580, opened 2025-02-06
TransactionInput: make field `witness` immutable

schildbach:
Because tweaking is necessary for transaction signing, these usages
have been changed to produce new inputs instead and replace them in
transactions as needed.

This is a child of #3579, but could be made to merge independently.


msgilligan (approved):
LGTM

schildbach:
Rebased.

schildbach:
Merged.

PR #3584, opened 2025-02-08
TransactionInput: make field `scriptBytes` immutable

schildbach:
Because tweaking is necessary for transaction signing, these usages
have been changed to produce new inputs instead and replace them in
transactions as needed.

This is a child of #3580, but could be made to merge individually.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -902,20 +902,20 @@ public class Transaction extends BaseMessage {
         if (ScriptPattern.isP2PK(scriptPubKey)) {
             TransactionSignature signature = calculateSignature(inputIndex, sigKey, scriptPubKey, sigHash,
                     anyoneCanPay);
-            input.setScriptSig(ScriptBuilder.createInputScript(signature));
+            input = input.withScriptSig(ScriptBuilder.createInputScript(signature));
             input = input.withoutWitness();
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1195,7 +1195,7 @@ public Sha256Hash hashForSignature(int inputIndex, byte[] connectedScript, byte
             // EC math so we'll do it anyway.
             for (int i = 0; i < tx.inputs.size(); i++) {
                 TransactionInput input = tx.getInput(i);
-                input.clearScriptBytes();
+                input = input.withoutScriptBytes();
```
msgilligan, line 1198: These two "without" calls could be chained. I would even consider chaining the interior of the `for` to one statement.

msgilligan (approved):
LGTM. But at least consider using chaining.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: Very valid comment. I deliberately choose against chaining this (and others in this PR) because of the simplicity (reviewability) of the diff. I thought the chaining can be done later, maybe on a case-by-case basis. Or maybe we rewrite the entire Transaction(Builder) class anyway…

schildbach:
Merged.

PR #3585, opened 2025-02-10
Transaction: cache transaction IDs

schildbach:
Without caching, transactions are serialized way too often.

This effectively reverts b8d3c4a641f1d6d7410f7152e5bd383b1f00c9ce, but invalidation is slightly different.

This PR has multiple commits, but **only the last commit is relevant for review**. All previous commits except the last are just backports of recent immutability work from master. We need that because it would be hard to detect mutations of dependant objects in order to properly invalidate caches.

If this is merged, I'll forward-port the relevant commit to master as well.


msgilligan (approved):
LGTM

PR #3599, opened 2025-02-19
GHA graalvm.yml: rename matrix variable java-version to java

msgilligan:
This is a component of PR #3525 that helps make our `graalvm.yaml` and `gradle.yaml`workflow files more similar.


schildbach:
Merged to maser and 0.17.

PR #3602, opened 2025-02-24
GHA graalvm.yml: move java-version line

msgilligan:
This is cosmetic only, but matches the order in `gradle.yml`, to make diffs easier.


schildbach:
Merged.

PR #3603, opened 2025-02-24
GHA gradle.yml: job name conform with graalvm.yml

msgilligan:
(The graalvm name is a little more compact and we should harmonize the two to make it easier to compare these files)


schildbach:
Merged.

PR #3604, opened 2025-02-25
Bech32: Use Optional in verifyChecksum (private)

msgilligan:
This is our last annotated @Nullable param or return value in  the base module!


schildbach:
#3606 was merged, so I guess this one is obsolete.

PR #3605, opened 2025-02-25
base: remove last @nonnull annotations & findbugs dependency

msgilligan:
A previous PR removed the last @nullable annotation.

This PR removes the @nonnull annotations as we can now assume that
everything in this module is non-nullable. If we discover otherwise
we can add JSpecify and mark them with that. We may want to use
JSpecify anyway, but I'll leave that for a separate PR.

This is a child of PR #3604


msgilligan:
I'll re-create this one based on PR #3606

msgilligan:
This is replaced by PR #3608

PR #3606, opened 2025-02-25
Bech32: private verifyChecksum throws on invalid

msgilligan:
This eliminates the last @Nullable annotation in the `base` module.

Since this is a private method and is called for every address that is parsed, let's not use Optional<>, but move the throw from decode() to verifyChecksum().

This is an alternative to PR #3604.


schildbach:
Merged.

PR #3607, opened 2025-02-25
Bech32: use ByteUtils.concat in verifyChecksum()

msgilligan:
This is a child of PR #3606, though could be made independent.


schildbach:
Merged.

PR #3608, opened 2025-02-25
base: remove last @Nonnull annotations & findbugs dependency

msgilligan:
A previous PR removed the last @Nullable annotation.

This PR removes the @Nonnull annotations as we can now assume that everything in this module is non-nullable. If we discover otherwise we can add JSpecify and mark them with that. We may want to use JSpecify anyway, but I'll leave that for a separate PR.

### DIFF
--- a/base/build.gradle
+++ b/base/build.gradle
@@ -9,7 +9,6 @@ version = '0.18-SNAPSHOT'
 
 dependencies {
     implementation 'org.slf4j:slf4j-api:2.0.16'
-    implementation 'com.google.code.findbugs:jsr305:3.0.2'
 
     testImplementation project(':bitcoinj-test-support')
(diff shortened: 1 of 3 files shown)


msgilligan:
I think the reason these methods got the `@Nonnul`l annotation is because they were replacements for deprecated methods that took `NetworkParameters` and the deprecated methods _did allow_ `null` parameters. The `@Nonnull` was added to make it clear that the new methods behaved differently.

msgilligan:
See these follow-on PRs for the use of JSpecify:

* PR #3835
* PR #3836

msgilligan:
I now think this should be rebased and merged _after_ PR #3834, so I have marked it as _DRAFT_.

schildbach (approved):

PR #3609, opened 2025-02-26
ByteUtils: rename maxBytes parameter of bigIntegerToBytes()

schildbach:
It's a maximum allowed size rather than a constant size. The `maxBytes` name makes this clearer.


msgilligan (approved):
Approved.

schildbach:
Merged.

PR #3610, opened 2025-02-27
modular-architecture.md: Update for planned 0.18 release

msgilligan:
The main change here is that we are replacing the plan of migrating the o.b.crypto module to it's own JAR and instead focusing on migrating to the secp256k1-jdk API.


schildbach:
Merged, and I have a followup PR #3617.

PR #3616, opened 2025-02-28
wallettemplate: build for JDK 21+

msgilligan:


schildbach:
I agree that at *some* time we can/should advance our JDK requirements, and I agree auxillary subprojects (wallettemplate, examples, etc.) would be a good start.

However, I believe first some CI builds should run under JDK 21+, so that the subprojects stay covered by the CI. This likely means Debian Trixie as underlying OS (or using Nix?), but Trixie is not even in freeze.

Also, don't forget to update the README alongside in the same PR.

msgilligan:
Closing in favor of PR #3976

PR #3622, opened 2025-03-09
Transaction: do a better job at showing invalid scriptSigs in toString()

schildbach:
This is a child of #3621, but could be merged individually if desired.

Previously, it was just showing an exception message. This is improved as follows:
    
- if an input scriptSig cannot be parsed, the raw bytes are shown instead
- also, the input value is shown if available
- for the coinbase input, which may be arbitrary bytes, the exception message is suppressed

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -774,7 +774,7 @@ public class Transaction extends BaseMessage {
                 s.append("in   ");
 
                 try {
-                    s.append(in.getScriptSig());
+                    s.append(in.getScriptSig()); // throws ScriptException
                     final Coin value = in.getValue();
                     if (value != null)
                         s.append("  ").append(value.toFriendlyString());
@@ -809,8 +809,16 @@ public class Transaction extends BaseMessage {
                             s.append(", has RLT");
                         s.append('\n');
                     }
-                } catch (Exception e) {
-                    s.append("[exception: ").append(e.getMessage()).append("]\n");
+                } catch (ScriptException e) {
+                    s.append(ByteUtils.formatHex(in.getScriptBytes()));
+                    if (!isCoinBase()) {
+                        final Coin value = in.getValue();
+                        if (value != null)
+                            s.append("  ").append(value.toFriendlyString());
+                        s.append('\n');
+                        s.append(indent).append("        exception: ").append(e.getMessage());
+                    }
+                    s.append('\n');
                 }
             }
         } else {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -811,9 +810,16 @@ public String toString(@Nullable AbstractBlockChain chain, @Nullable Network net
                         s.append('\n');
                     }
                 } catch (Exception e) {
```
msgilligan, line 812: I might be helpful to know what type of exceptions are being caught here.  It's a large `try`-block and it is hard to know which operations might be causing exceptions and what the exceptions are.

I was able to change the type from `Exception` to `RuntimeException` and the code still compiles and passes tests and I think that change would be slightly helpful for people reading  the code. It might also be helpful to catch more specific exceptions and/or use multiple catch blocks to break down where in the code the exception might be occurring. As the code now stands there could be fatal exceptions that are ignored.

I realize this might be out-of-scope for this PR, but these issue make the PR harder to review.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -811,9 +810,16 @@ public String toString(@Nullable AbstractBlockChain chain, @Nullable Network net
...
-                    s.append("[exception: ").append(e.getMessage()).append("]\n");
+                    s.append(ByteUtils.formatHex(getInput(0).getScriptBytes()));
+                    if (!isCoinBase()) {
```
msgilligan, line 814: Shouldn't we be using `in` instead of `getInput(0)` here?

msgilligan (changes requested):
Maybe I'm just confused but it seems like we should be hex-dumping the current input from the for loop. I also think the exception-handling code be made a little more clear, at least by changing the caught type to `RuntimeException`.

More tests would also be nice.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: Absolutely! Thanks for spotting this.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: You're right, I hate these kind of big catch blocks too. But I don't know how to do it better in Java, as there is no "otherwise { }" (or similar). I thought the only line method call that can throw is the getScriptSig() call, so I now changed the catch to ScriptException. However, that uncovered an IndexOutOfBounds in a test…

…leading to this bugfix, which we'll need to merge first: #3635

schildbach: An alternative to the catch block would be a "isScriptSigValid()" that the getScriptSig() could be guarded with, but that would lead to everything parsed twice.

schildbach:
I rebased this, which fixes the IndexOutOfBounds exception. The other review comment has also been taken care of. Does this qualify for approval? (-:

msgilligan (approved):
Approved.

schildbach:
Merged.

PR #3623, opened 2025-03-09
Transaction: include coinbase input in toString()

schildbach:
This is a child of #3621 and #3622, but could be merged individually.

It used to be suppressed, presumably because coinbase scriptSigs may contain arbitrary bytes which could confuse our script parser.


msgilligan (approved):
Approved. (Assuming review on #3622 is resolved)

schildbach:
Merged.

PR #3625, opened 2025-03-09
ScriptBuilder: remove methods that add chunks at a specified index

schildbach:
Currently, there is no usecase for that.

PR #3626, opened 2025-03-09
ScriptBuilder: make bigNum() public

schildbach:
This is a child of #3625, but it could be made to be merged independently if so desired.

We will need this to construct genesis block scriptSigs, which use a pushdata of a small value that would, optimally encoded, be pushed via OP_n.


msgilligan (approved):
Approved.

schildbach:
Merged.

PR #3627, opened 2025-03-09
Block: construct genesis block coinbase from individual values

schildbach:
This is a child of #3626 and #3625.

In particular,
    
- the input script is constructed from the difficulty target and the genesis message
- the output script is constructed from the pubkey via ScriptBuilder.createP2PKOutputScript()

The added constants will be futher moved. This PR is preparing the stage for future refactorings.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -32,7 +32,6 @@ import org.bitcoinj.crypto.ECKey;
 import org.bitcoinj.params.BitcoinNetworkParams;
 import org.bitcoinj.script.Script;
 import org.bitcoinj.script.ScriptBuilder;
-import org.bitcoinj.script.ScriptOpCodes;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -43,6 +42,7 @@ import java.io.OutputStream;
 import java.math.BigInteger;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -239,22 +239,19 @@ public static Block createGenesis(Instant time, long difficultyTarget, long nonc
...
 
     private static List<Transaction> genesisTransactions() {
-        Transaction tx = Transaction.coinbase(genesisTxInputScriptBytes);
-        tx.addOutput(new TransactionOutput(tx, FIFTY_COINS, genesisTxScriptPubKeyBytes));
+        byte[] messageBytes = GENESIS_MESSAGE.getBytes(StandardCharsets.US_ASCII);
+        Script scriptSig = // TODO find out what the pushdata(4) is supposed to mean
+                new ScriptBuilder().bigNum(STANDARD_MAX_DIFFICULTY_TARGET).bigNum(4).data(messageBytes).build();
+        Transaction tx = Transaction.coinbase(scriptSig.program());
+        tx.addOutput(new TransactionOutput(
+                tx, FIFTY_COINS, ScriptBuilder.createP2PKOutputScript(GENESIS_OUTPUT_PUBKEY).program()));
```
msgilligan, line 247: I think we should keep the `private static final` constants `genesisTxInputScriptBytes` and `genesisTxScriptPubKeyBytes`, but calculate them as above.

I think this will make the code easier to read and also allow the compiler to better optimize.

msgilligan (changes requested):
Looks good, but I think we can keep more values as static constants.

schildbach:
Rebased.

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: This is how the method signature will look like a near-future commit:

`private static List<Transaction> genesisTransactions(String message, byte[] outputPubKey)`

msgilligan: I see. Since this method is `static` then it should be possible to even create the `List` of transactions at compile/init time.

Is there ever a case where there is more than one transaction? I wonder if we should change the method to return a single transaction (and then convert to a singleton list before calling new Block() or whatever uses it.)

schildbach: > Is there ever a case where there is more than one transaction? I wonder if we should change the method to return a single transaction (and then convert to a singleton list before calling new Block() or whatever uses it.)

Currently not, but there could. Moving the list to the user would duplicate code. But I'm not sure myself, I thought this through a couple of days ago and decided to not change the return type for now because I'm missing the clear vision of that part of the API.

schildbach: > I see. Since this method is `static` then it should be possible to even create the `List` of transactions at compile/init time.

Currently the genesis blocks are lazily generated but then re-used. We'll get to that in upcoming PRs.

schildbach:
Merged.

PR #3630, opened 2025-03-10
DefaultCoinSelector: in compareByDepth() compare transaction hashes directly

schildbach:
Before this change, it was converting them to BigInteger.


msgilligan (approved):
Approved.

PR #3632, opened 2025-03-10
SPVBlockStore: fix get() locates an invalid block if called with zero hash

schildbach:
This happens if the store is relatively fresh and has not yet fully wrapped around at least once. Thus, entire entries of the ring buffer are still zeroed, and in particular the hash field of store entries will be zero. In consequence, get() is locating and returning an invalid block when asked to look for the zero hash.

Valid blocks that hash to zero will require an astronomical amount of mining. So we fix this bug by hardcoding the zero hash to never be found in our store.

Includes a test for this edge case.

This PR should probably be backported to 0.17.


msgilligan (approved):
Approved.

schildbach:
Merged to master and 0.17.

PR #3633, opened 2025-03-10
StoredBlock: if getPrev() called on a genesis block, cut short

schildbach:
Don't try to locate a previous block of the genesis block, using an arbitrary "previous block hash".

This could be backported to 0.17.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/StoredBlock.java
+++ b/core/src/main/java/org/bitcoinj/core/StoredBlock.java
@@ -129,7 +129,9 @@ public class StoredBlock {
      * @return the previous block in the chain or null if it was not found in the store.
      */
     public StoredBlock getPrev(BlockStore store) throws BlockStoreException {
-        return store.get(getHeader().getPrevBlockHash());
+        return height > 0 ?
+                store.get(header.getPrevBlockHash()) :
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/StoredBlock.java
```diff
@@ -129,7 +129,10 @@ public StoredBlock build(Block block) throws VerificationException {
      * @return the previous block in the chain or null if it was not found in the store.
      */
     public StoredBlock getPrev(BlockStore store) throws BlockStoreException {
-        return store.get(getHeader().getPrevBlockHash());
+        if (height == 0)
+            // Genesis blocks have no previous block.
+            return null;
+        return store.get(header.getPrevBlockHash());
```
msgilligan, line 135: I would suggest using a single `return` statement and a ternary operator here.

msgilligan (changes requested):
Minor change requested.

core/src/main/java/org/bitcoinj/core/StoredBlock.java
schildbach: Changed.

schildbach:
Merged to master and 0.17.

PR #3635, opened 2025-03-13
TransactionOutPoint: clarify JavaDoc of getConnectedOutput()

schildbach:


schildbach:
Hmm, thinking of it again, maybe it's not the right fix. Because according to the JavaDoc, null means "we don't have the link". It doesn't mean "we have the link, but it's invalid".

msgilligan (approved):
~~Approved.~~

schildbach:
I just checked, and none of our own usages of the method seem to expect or handle IndexOutOfBoundsException in some way – except the Transaction.toString() which was conveniently swallowing it.

schildbach:
I just rebased this and stripped it down to just the JavaDoc clarification.

msgilligan (approved):
Approve.

schildbach:
Merged.

PR #3636, opened 2025-03-13
TransactionTest: fix testWitnessSignatureP2WPKH() connected output

schildbach:
txIn1 is referencing an output at index 1, but when we're creating a mock for the connected transaction the output was inadvertently at index 0. This is fixed by inserting a dummy output before the referenced one.


msgilligan (approved):
Approved.

schildbach:
Merged to master and 0.17.

PR #3637, opened 2025-03-14
TransactionOutPoint: when constructing with known fromTx, do more integrity checks

schildbach:
- the output with given index has to exist in fromTx, and
- that output needs to match connectedOutput if given
    
This restriction causes a test to be unconstructable. It is replaced by the next best thing, but it's unclear if the test remains meaningful at all.

Child of #3636.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
@@ -91,6 +91,13 @@ public class TransactionOutPoint {
         checkArgument(index >= 0 && index <= ByteUtils.MAX_UNSIGNED_INTEGER, () ->
                 "index out of range: " + index);
         this.index = index;
+        if (fromTx != null) {
+            TransactionOutput outputFromTx = fromTx.getOutput(index);
+            Objects.requireNonNull(outputFromTx);
+            if (connectedOutput != null) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -93,6 +93,13 @@ private TransactionOutPoint(Sha256Hash hash, long index, @Nullable Transaction f
         this.index = index;
         this.fromTx = fromTx;
         this.connectedOutput = connectedOutput;
+        if (fromTx != null) {
+            TransactionOutput outputFromTx = fromTx.getOutput(index);
+            Objects.requireNonNull(outputFromTx);
+            if (connectedOutput != null) {
+                checkArgument(connectedOutput.equals(outputFromTx), () -> "mismatched connected output");
+            }
+        }
```
msgilligan, line 102: This looks good, but I would put the validation before the assignment to object members, as we typically do.

msgilligan (approved):
Approved, with one minor suggestion.

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
schildbach: Moved up.

schildbach:
Merged.

PR #3639, opened 2025-03-15
TestNet3Params: simplify calculation of timeDelta

msgilligan:
Using Duration.between() is slightly simpler, it also allows easier access to the result as a Duration if we move to stronger types here in the future.

### DIFF
--- a/core/src/main/java/org/bitcoinj/params/TestNet3Params.java
+++ b/core/src/main/java/org/bitcoinj/params/TestNet3Params.java
@@ -28,6 +28,7 @@ import org.bitcoinj.store.BlockStore;
 import org.bitcoinj.store.BlockStoreException;
 
 import java.math.BigInteger;
+import java.time.Duration;
 import java.time.Instant;
 
 import static org.bitcoinj.base.internal.Preconditions.checkState;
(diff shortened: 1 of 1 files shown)


schildbach:
Yes, I saw that, and the following value clamps too, but I thought better keep it aligned to the code in Bitcoin Core.

schildbach:
One issue with the java.time API is you need to be very careful to use only seconds not milliseconds. Otherwise, the calculations might be off by a bit, leading to an unexpected difficulty change by a bit. Which – even for an SPV client – means stalling the chain.

schildbach:
Have you run `BuildCheckpoints` successfully on Testnet3 with this change? Ideally, comparing to a previous `org.bitcoin.test.checkpoints.txt` (there is one in core).

msgilligan:
> Have you run BuildCheckpoints successfully on Testnet3 with this change?

No. And you make several good points. I'll downgrade this to _DRAFT_ and we can focus on the other stuff.

msgilligan:
> Have you run `BuildCheckpoints` successfully on Testnet3 with this change? Ideally, comparing to a previous `org.bitcoin.test.checkpoints.txt` (there is one in core).

Could we make this an integration test? Or create something similar?

schildbach:
> > Have you run `BuildCheckpoints` successfully on Testnet3 with this change? Ideally, comparing to a previous `org.bitcoin.test.checkpoints.txt` (there is one in core).
> 
> Could we make this an integration test? Or create something similar?

If we were willing to include headers of an entire blockchain in our repo, then yes I think that's possible.

schildbach:
Having looked at the difficulty transition code again yesterday, I now think this is doable. We just need to be careful.

And maybe we should convert the SPACING constants to Duration at the same time?

schildbach:
This change is contained in https://github.com/bitcoinj/bitcoinj/pull/3646.

PR #3631, opened 2025-03-10
WIP: Difficulty: new class representing the difficulty target of a block

schildbach:
TODO add back deprecated methods
TODO test more edge cases
TODO check bounds of `of*()` static constructors, but what are they?

PR #3640, opened 2025-03-15
Block: remove exception-safe usage mode from `checkProofOfWork()`

schildbach:
This removes both the boolean argument and the boolean return type. It always throws if difficulty target is not met. If it returns normally, the target is met.
    
The only consumer of the exception-safe usage mode is switched to using `Difficulty.isMetBy()` directly.
    
Also, this method is switched to private visibility.

This PR is a child of #3631.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -434,7 +434,7 @@ public class Block implements Message {
         while (true) {
             try {
                 // Is our proof of work valid yet?
-                if (checkProofOfWork(false))
+                if (difficultyTarget.isMetByWork(getHash()))
                     return;
                 // No, so increment the nonce and try again.
                 setNonce(getNonce() + 1);
@@ -455,11 +455,16 @@ public class Block implements Message {
         return difficultyTarget.asInteger();
     }
 
-    /** Returns true if the hash of the block is OK (lower than difficulty target). */
-    protected boolean checkProofOfWork(boolean throwException) throws VerificationException {
+    /**
+     * Checks the proof of work of this block. If this method returns without throwing, the hash of this block is OK
+     * (lower than difficulty target).
+     *
+     * @throws VerificationException if there is not enough proof of work
+     */
+    private void checkProofOfWork() throws VerificationException {
         // shortcut for unit-testing
         if (Context.get().isRelaxProofOfWork())
-            return true;
+            return;
 
         // This part is key - it is what proves the block was as difficult to make as it claims
         // to be. Note however that in the context of this function, the block can claim to be
@@ -469,15 +474,10 @@ public class Block implements Message {
         //
         // To prevent this attack from being possible, elsewhere we check that the difficultyTarget
         // field is of the right value. This requires us to have the preceding blocks.
-        if (!difficultyTarget.isMetByWork(getHash())) {
+        if (!difficultyTarget.isMetByWork(getHash()))
             // Proof of work check failed!
-            if (throwException)
-                throw new VerificationException("Hash is higher than target: " + getHashAsString() + " vs "
-                        + difficultyTarget.toIntegerString());
-            else
-                return false;
-        }
-        return true;
+            throw new VerificationException("Hash is higher than target: " + getHashAsString() + " vs "
+                    + difficultyTarget.toIntegerString());
     }
 
     private void checkTimestamp() throws VerificationException {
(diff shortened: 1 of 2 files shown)


schildbach:
I might look into why the unit-tests are calling `solve()` in the first place. If they wouldn't, then the relaxProofOfWork check in that method would not be necessary.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -434,12 +434,14 @@ public String toString() {
      */
     @VisibleForTesting
     public void solve() {
+        if (Context.get().isRelaxProofOfWork())
```
msgilligan, line 437: It sure would be nice to get rid of `Context.get()` here, if at all possible.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -434,12 +434,14 @@ public String toString() {
...
+            return;
```
msgilligan, line 438: It's also nice to avoid short-circuiting returns.

core/src/main/java/org/bitcoinj/core/Block.java
msgilligan: Could we encode the fact that we are unit-testing and want relaxed PoW in the `Block` or `Difficulty` classes?

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -460,11 +462,16 @@ public BigInteger getDifficultyTargetAsInteger() {
...
 
-    /** Returns true if the hash of the block is OK (lower than difficulty target). */
-    protected boolean checkProofOfWork(boolean throwException) throws VerificationException {
+    /**
+     * Checks the proof of work of this block. If this method returns without throwing, the hash of this block is OK
+     * (lower than difficulty target).
+     *
+     * @throws VerificationException if there is not enough proof of work
+     */
+    private void checkProofOfWork() throws VerificationException {
```
msgilligan, line 471: I would consider changing this to always return a `boolean` and have higher-level methods decide to throw an exception.

msgilligan (approved):
I'm uneasy with some of the coding patterns we are using (and are not new to this PR) and am wondering if/when we can move away from them, but that is probably out-of-scope for this PR. So I'm reluctantly approving.

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: Actually, this *is* the higher level method that decides to throw. The method with a boolean is now in Difficulty.

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: I love them (-:

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: I'd really like to get rid of it here too. I want to invest some time to see if we can make the affected unit tests "easier", meaning they use an easier difficulty.

I would rather not make "unit test mode" a property of Difficulty or Block.

schildbach:
Rebased.

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: The main problem I have with this short-circuit is that we (implicitly) call solve() but at the same time we say "don't solve".

msgilligan: Anyway, no reason to hold up the merge.

schildbach:
Even though you approved this PR already, I've reworked it so that the Context-check for relaxProofOfWork could be removed from solve(). That method is now only called if the test requires it, and if UnitTestParams (easiest difficulty) are used.

This is now a child of #3653 and #3654.

schildbach:
> I'm uneasy with some of the coding patterns we are using (and are not new to this PR) and am wondering if/when we can move away from them, but that is probably out-of-scope for this PR. So I'm reluctantly approving.

I assume you refer to the `check*()` methods that either return (void) or throw. We have a couple of them, and I think they're fine if at least they follow the same pattern. What I really hated was this dual-use `checkProofOfWork()` that could be configured via boolean argument wether it throws or returns a boolean. This is eliminated with this PR, and I think this is a win – even if the result is a plain `check*()` method.

schildbach:
I decided to merge this. We can add further improvements on how the block checking works, but it will likely affect a broader scope that just this method.

PR #3646, opened 2025-03-16
NetworkParameters: convert spacing and timespans to `java.time.Duration`

schildbach:
This is two commits:

* BitcoinNetworkParams: move spacing, timespan, interval constants from NetworkParameters
  Also rename `INTERVAL` to `INTERVAL_BLOCKS`.

* NetworkParameters: convert spacing and timespans to `java.time.Duration`

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+++ b/core/src/main/java/org/bitcoinj/core/NetworkParameters.java
@@ -30,6 +30,7 @@ import org.bitcoinj.base.utils.MonetaryFormat;
 import org.bitcoinj.utils.VersionTally;
 
 import java.math.BigInteger;
+import java.time.Duration;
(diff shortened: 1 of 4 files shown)


schildbach:
Damn, the referencing of superclass to subclass at init time is clearly not liked by the VM at runtime.

Thinking about the moving of the constants again, maybe we want to move them (and their variable friends and accessors) to `BitcoinNetwork`? Or, if we don't want to grow that class too much, create a new hierarchy just for "blockchain parameters"?

schildbach:
I converted this to draft because currently it doesn't work as planned. Maybe it's better to merge the Difficulty refactor first, then decide on where the constants should _really_ go.

But I certainly want to revisit especially the "move to java.time.Duration" commit again.

PR #3644, opened 2025-03-15
TestNet3Params: extract method backwardsSkipMindiffBlocks()

schildbach:

PR #3647, opened 2025-03-17
DifficultyTransitions: interface and implementations for difficulty transitions

schildbach:
Extract logic from BitcoinNetworkParameters and TestNet3Params.

~~This is a child of #3631.~~

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/AbstractBlockChain.java
@@ -18,6 +18,7 @@
 package org.bitcoinj.core;
 
 import com.google.common.annotations.VisibleForTesting;
+import org.bitcoinj.base.BitcoinNetwork;
 import org.bitcoinj.base.Difficulty;
 import org.bitcoinj.base.Network;
 import org.bitcoinj.base.Sha256Hash;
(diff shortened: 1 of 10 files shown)


msgilligan (changes requested):
This looks good.

One minor request: At least a dozen imports in `BitcoinNetworkParams` are no longer needed because the method was removed. I think those imports should be removed in this PR.

The rest is just a comment, not a request for changes:

I also am wondering if using inheritance is the best approach long-term for `DifficultyTransitions`, but I agree that it is the best approach for this PR and for migration away from `BitcoinNetworkParams`.

With the differences between "standard" and testnet3 in separate files it is harder to compare them, I think. (And to review this PR I had to copy the moved methods into separate editor buffer to diff them.)

Perhaps if we factor out logic from standard `checkDifficultyTransitions` into (private) methods (similar to what we did with PR #3644, etc) we can simplify the logic enough to find some common patterns with the testnet3 case and combine to a single file and use conditionals for the differences. (I am not certain this will work, just thinking it might be worth a try.)

Simplifying the `checkDifficultyTransitions` by extracting methods would be a useful improvement even if we stick with using inheritence.

schildbach:
Rebased.

schildbach:
> I also am wondering if using inheritance is the best approach long-term for DifficultyTransitions, but I agree that it is the best approach for this PR and for migration away from BitcoinNetworkParams.

Me too, but the alternative is an "if ladder" in (or close to) Block. The specs kind of inherit from each other: testnet3 has the same rules as mainnet, except something. Testnet4 has same rules as testnet3, except something.

schildbach:
> Simplifying the checkDifficultyTransitions by extracting methods would be a useful improvement even if we stick with using inheritence.

Generally agree, though we already did some of that and I hoped it's "enough".

schildbach:
I have addressed the imports issue, and rebased. Time for another review and maybe merge.

msgilligan (approved):
Approved. I agree that the OOP approach is the best way to go for now.

We can look at extracting methods within the OOP implementation and/or an alternate approach later. Maybe _after_ TESTNET4 is added.

coinfetti:
> Approved. I agree that the OOP approach is the best way to go for now.
> 
> We can look at extracting methods within the OOP implementation and/or an alternate approach later. Maybe _after_ TESTNET4 is added.

I was looking for TestNet4 implementation, and only found this post from you. Do you know what the progress is on that? And will it be propably included in 0.17.1?

PR #3650, opened 2025-03-24
Block: add precondition to test helper `addCoinbaseTransaction()`

schildbach:
It requires the block to contain no transactions, so that a coinbase can be added. This is already true for all current usages.

This change allows us to get rid of another `unCacheTransactions()`.


msgilligan (approved):
Approved.

schildbach:
Merged.

PR #3652, opened 2025-03-24
FakeTxBuilder: move to `test-support` module

schildbach:
It's needed in both in the `bitcoinj-core` unit tests and in the `integration-test` subproject, but it can't stay in the `bitcoinj-core` main class path.


msgilligan (approved):
Approved.

msgilligan:
PR #3850 was merged to eliminate circular dependencies (for Maven) so we need to rethink this PR.

PR #3653, opened 2025-03-24
FakeTxBuilder: do not solve fake blocks

schildbach:
Because `solve()` in almost all cases requires too much work, in future it should be called explicitly by the tests that depend on proper proof of work.

Most tests already use `Context.relaxProofOfWork` to skip the check. Only PeerTest depends on it, so it is changed to solve one block manually (on easiest difficulty).

Some method names were stripped of the adjective "solved" and the JavaDocs amended appropriately.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -820,7 +820,7 @@ public class Block implements Message {
     private static final byte[] pubkeyForTesting = new ECKey().getPubKey();
 
     /**
-     * Returns a solved block that builds on top of this one. This exists for unit tests.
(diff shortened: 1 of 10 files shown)


schildbach:
I've added a commit on top that removes implicit block solving from `Block.createNextBlock()` as well:

- efb1588bdcb450fb77625648bc8c28c333ad01a8

msgilligan (approved):
Approved. This is a very nice improvement and should help us get to unmodifiable/immutable blocks!

schildbach:
Merged.

schildbach:
> should help us get to unmodifiable/immutable blocks

Note this is only getting rid of (most, not all) Block mutations in the context of tests. Not sure if we ever can get rid of `solve()` entirely, but at least it can be part of a block builder I guess.

PR #3657, opened 2025-03-24
FilteredBlock: make fields immutable

msgilligan:
All the serialized fields are now immutable. (There is still a cachedTransactionsHashes that is nullable and mutable)


schildbach:
Merged.

PR #3658, opened 2025-03-24
BaseMessage: remove unused logger

msgilligan:


schildbach (changes requested):
Shouldn't the imports be removed in the same commits?

msgilligan:
> Shouldn't the imports be removed in the same commits?

Fixed and force-pushed.

schildbach:
Merged.

PR #3659, opened 2025-03-24
BaseMessage: convert to interface

msgilligan:
This is done in two commits:

1. Remove `final` modifier form `serialize()` (see the  commit description 30c553535351303f66def120393379f3acb3b8d2) on that commit)
2. Convert `BaseMessage` to `interface` and have all subclasses `implement` rather than `extends` (82e03fb991eb48cf08229f27214dfdc38282e9ca)

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AddressMessage.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressMessage.java
@@ -31,7 +31,7 @@ import static org.bitcoinj.base.internal.Preconditions.check;
  * Abstract superclass for address messages on the P2P network, which contain network addresses of other peers. This is
  * one of the ways peers can find each other without using the {@link PeerDiscovery} mechanism.
  */
-public abstract class AddressMessage extends BaseMessage {
+public abstract class AddressMessage implements BaseMessage {
 
     protected static final long MAX_ADDRESSES = 1000;
     protected List<PeerAddress> addresses;
--- a/core/src/main/java/org/bitcoinj/core/AddressV1Message.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressV1Message.java
@@ -54,7 +54,7 @@ public class AddressV1Message extends AddressMessage {
     }
 
     @Override
-    protected void bitcoinSerializeToStream(OutputStream stream) throws IOException {
+    public void bitcoinSerializeToStream(OutputStream stream) throws IOException {
         if (addresses == null)
             return;
         stream.write(VarInt.of(addresses.size()).serialize());
--- a/core/src/main/java/org/bitcoinj/core/AddressV2Message.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressV2Message.java
@@ -55,7 +55,7 @@ public class AddressV2Message extends AddressMessage {
     }
 
     @Override
-    protected void bitcoinSerializeToStream(OutputStream stream) throws IOException {
+    public void bitcoinSerializeToStream(OutputStream stream) throws IOException {
         if (addresses == null)
             return;
         stream.write(VarInt.of(addresses.size()).serialize());
--- a/core/src/main/java/org/bitcoinj/core/BaseMessage.java
+++ b/core/src/main/java/org/bitcoinj/core/BaseMessage.java
@@ -27,7 +27,7 @@ import java.io.OutputStream;
  * <p>
  * Instances of this class are not safe for use by multiple threads.
  */
-public abstract class BaseMessage implements Message {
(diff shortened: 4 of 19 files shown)


schildbach:
What do we gain from this change?

msgilligan:
> What do we gain from this change?

In general: Less inheritance, more flexibility. I think this will really help with future refactoring.

For example, the current implementation (pretty much) forces you to serialize with an `OutputStream` because of inheritance. (although the first commit, not the second is the one that really fixes that.)

After this PR it's possible for a `Message` to be subclass of some other class _and_ implement `BaseMessage`.

After this PR, It might also make sense to rename `BaseMessage` to `StreamSerializedMessage` or something similar.

Note also that `default int messageSize()` could be moved to `Message`.

schildbach:
BaseMessage is currently not an abstraction we need anywhere in our code, so I see no need in having both Message _and_ BaseMessage. It's only the typical `Abstract*` class that helps the various implementations a bit.

If we _can_ fold BaseMessage into Message without duplicating code into the implementations, imho we should do that.

We could also simply revert 1fef4e2cdbff30560b1fc2ac7442cc81935c7e9e to make it a one (abstract) class again.

Regarding the duality of serialize() and bitcoinSerializeToStream(), the final modifier is meant to prevent implementations from becoming a mishmash of implementing both (or sometimes one, sometimes the other) again. We came from that mess, and I wanted to prevent it happening again. I'm not saying OutputStreams are great, but at the moment that's preferred way to implement serialization. We can take another look at ByteBuffer, or maybe some form of socket, but I suspect that's another discussion.

msgilligan:
> BaseMessage is currently not an abstraction we need anywhere in our code (in terms of using it as a type), so I see no need in having both Message _and_ BaseMessage interfaces.

Agreed. Moving to an interface is step towards either eliminating `BaseMessage` or perhaps using it as a Mix-in for OutputStream-based serialization.

> BaseMessage is only the typical `Abstract*` class that helps the various implementations a bit.

I generally see this as an anti-pattern, something to generally be avoided. At least when it only helps "a bit". In general, I would rather see a small amount of duplicate code than a layer of inheritance. There are some cases where using an abstract base class makes sense, but these days I would lean towards either using composition or a mix-in interface in most cases.

> If we _can_ fold BaseMessage into the rest of the Message hierarchy without duplicating code into the implementations, I'd not be against that. However it would spoil the essence of interfaces a bit, because it would then contain a bit of implementation (default methods).

I think it's OK to use default methods and the mix-in pattern. I'm also not opposed to some code duplication to avoid inheritance.

> Regarding the duality of serialize() and bitcoinSerializeToStream(), the final modifier is meant to prevent  ...  that mess

I think that was a good decision and was helpful during our refactoring.

> OutputStreams ... at the moment ... preferred way to implement serialization. We can take another look at ByteBuffer ...

If we make our current `OutputStream`-based serialization optional (perhaps by putting it in a mix-in class) each `Message` implementation can have its own method of serialization. So we could move some to `ByteBuffer` where that works best. Some like `Ping` or `Pong` can be almost trivially simple.

schildbach:
If you ask me, `extends EmptyMessage` seems hard to beat when it comes to simpleness. But let's stay with the `BaseMessage` discussion in this PR, ok?

schildbach:
I could imagine doing the following to get rid of `BaseMessage`:

1. Remove the default implementation of `BaseMessage.messageSize()`. Implement it more efficiently in the concrete message classes. This on its own could be useful, as it's more efficient – but of course more code.
2. Optional, to prevent consumers to have to use classic Java I/O in future: Refactor `BaseMessage.bitcoinSerializeToStream(OutputStream stream)` and use `ByteBuffer` instead. Yes, this would affact all subclasses. We use `ByteBuffer` for reading from bytes already, so this is something I always thought of. 
3. Maybe pick a nicer name for the method (from 2) and make it available in the `Message` interface.
4. Keep the old `byte[] serialize()` around as a `default` implementation.

I think then we could simply delete `BaseMessage` and **not** make it an interface. All direct subclasses would then be free to extend something else.

schildbach:
When deciding step 2, I think we should avoid excessive instantiation. For example, a Block with 100 Transaction should *not* create 101 ByteBuffers. But I think it's easy to do if consumers of the API either continue using `byte[] serialize()` or construct their own ByteBuffer by asking for the correct `messageSize()` first.

schildbach:
Steps 1-4 from above are implemented. I think this can be closed now.

msgilligan:
`BaseMessage` has been removed! Closing.

PR #3660, opened 2025-03-25
Experiment: Create SignedTransactionSubclass

msgilligan:
This experiment shows that we can mostly assume that any transaction that is deserialized via `read()` can be assumed signed and mostly immutable.

The "WalletTransaction" methods (to set purpose, exchange rate, and memo) were left mutable in this experiment. A separate effort to create a wallet transaction that stores mutable wallet annotated info can address those fields.

Besides the changes to Transaction and BitcoinSerializer, the only other changes needed were to test code (which mutates deserialized and roundtrip transactions, so the `.asUnsigned()` conversion method needed to be called.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
+++ b/core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
@@ -61,6 +61,7 @@ public class BitcoinSerializer extends MessageSerializer {
         names.put(Block.class, "block");
         names.put(GetDataMessage.class, "getdata");
         names.put(Transaction.class, "tx");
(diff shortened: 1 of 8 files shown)


schildbach:
> This experiment shows that we can mostly assume that any transaction that is deserialized via read() can be assumed signed and mostly immutable.

Ignoring tests and (unwanted) malleability for a second, read() is used for deserializing Bitcoin messages from the network. And afaik txns from the network are always signed and immutable.

schildbach:
> SignedTransaction

When I thought about the first class to split out, I thought maybe CommittedTransaction would be a good pick. But "signed" seems reasonable too, maybe better.

PR #3662, opened 2025-03-26
BaseMessage: remove default implementation of `messageSize()`

schildbach:
Implement it in the concrete implementations. This avoids needless serialization and instantiation of buffers.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/AddressV1Message.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressV1Message.java
@@ -54,6 +54,16 @@ public class AddressV1Message extends AddressMessage {
     }
 
     @Override
+    public int messageSize() {
+        if (addresses == null)
+            return 0;
+        return VarInt.sizeOf(addresses.size()) +
+                addresses.stream()
+                        .mapToInt(addr -> addr.getMessageSize(1))
+                        .sum();
+    }
+
+    @Override
     protected void bitcoinSerializeToStream(OutputStream stream) throws IOException {
         if (addresses == null)
             return;
--- a/core/src/main/java/org/bitcoinj/core/AddressV2Message.java
+++ b/core/src/main/java/org/bitcoinj/core/AddressV2Message.java
@@ -55,6 +55,16 @@ public class AddressV2Message extends AddressMessage {
     }
 
     @Override
+    public int messageSize() {
+        if (addresses == null)
+            return 0;
+        return VarInt.sizeOf(addresses.size()) +
+                addresses.stream()
+                        .mapToInt(addr -> addr.getMessageSize(2))
+                        .sum();
+    }
+
+    @Override
     protected void bitcoinSerializeToStream(OutputStream stream) throws IOException {
         if (addresses == null)
             return;
--- a/core/src/main/java/org/bitcoinj/core/BloomFilter.java
+++ b/core/src/main/java/org/bitcoinj/core/BloomFilter.java
@@ -167,6 +167,14 @@ public class BloomFilter extends BaseMessage {
         return helper.toString();
     }
 
+    @Override
+    public int messageSize() {
+        return Buffers.lengthPrefixedBytesSize(data) +
+                4 + // hashFuncs
+                4 + // nTweak
+                1; // nFlags
+    }
+
     /**
      * Serializes this message to the provided stream. If you just want the raw bytes use {@link #serialize()}.
      */
--- a/core/src/main/java/org/bitcoinj/core/FilteredBlock.java
+++ b/core/src/main/java/org/bitcoinj/core/FilteredBlock.java
@@ -68,6 +68,12 @@ public class FilteredBlock extends BaseMessage {
     }
 
     @Override
+    public int messageSize() {
+        return Block.HEADER_SIZE +
+                merkleTree.messageSize();
+    }
+
+    @Override
     public void bitcoinSerializeToStream(OutputStream stream) throws IOException {
         if (header.getTransactions() == null)
             header.bitcoinSerializeToStream(stream);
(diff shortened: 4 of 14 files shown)


core/src/main/java/org/bitcoinj/core/AddressV1Message.java
```diff
@@ -53,6 +53,17 @@ public void addAddress(PeerAddress address) {
         addresses.add(address);
     }
 
+    @Override
+    public int messageSize() {
+        if (addresses == null)
+            return 0;
```
msgilligan, line 59: Can we guarantee that `addresses` is never null, but is an empty list when there are zero addresses? This could be a separate PR (either before or after this one)

core/src/main/java/org/bitcoinj/core/AddressV1Message.java
```diff
@@ -53,6 +53,17 @@ public void addAddress(PeerAddress address) {
...
+        int size = VarInt.sizeOf(addresses.size());
+        for (PeerAddress addr : addresses) {
+            size += addr.getMessageSize(1);
+        }
+        return size;
+    }
+
```
msgilligan, line 66:
```
         return (addresses != null)
                ?  VarInt.sizeOf(addresses.size()) +
                    addresses.stream()
                        .mapToInt(addr -> addr.getMessageSize(1))
                        .sum()
                : 0;
```

or simpler if we can get rid of the null check.

core/src/main/java/org/bitcoinj/core/AddressV2Message.java
```diff
@@ -54,6 +54,17 @@ public void addAddress(PeerAddress address) {
...
+    public int messageSize() {
+        if (addresses == null)
+            return 0;
+        int size = VarInt.sizeOf(addresses.size());
+        for (PeerAddress addr : addresses) {
+            size += addr.getMessageSize(2);
+        }
+        return size;
+    }
+
```
msgilligan, line 67: Same as above.

core/src/main/java/org/bitcoinj/core/BloomFilter.java
```diff
@@ -167,6 +167,15 @@ public String toString() {
...
     }
 
+    @Override
+    public int messageSize() {
+        int size = VarInt.sizeOf(data.length) + data.length;
+        size += 4; // hashFuncs
+        size += 4; // nTweak
+        size += 1; // nFlags
+        return size;
+    }
```
msgilligan, line 177: I would put the `return` on the first line of the function, but still keep the commented lines, but just as:

```
  + 4   // hashFuncs
```

core/src/main/java/org/bitcoinj/core/FilteredBlock.java
```diff
@@ -67,6 +67,13 @@ public FilteredBlock(Block header, PartialMerkleTree pmt) {
         this.merkleTree = pmt;
     }
 
+    @Override
+    public int messageSize() {
+        int size = Block.HEADER_SIZE;
+        size += merkleTree.messageSize();
+        return size;
+    }
```
msgilligan, line 75: Same here, `return` on the first line.

core/src/main/java/org/bitcoinj/core/RejectMessage.java
```diff
@@ -112,6 +112,18 @@ public RejectMessage(RejectCode code, @Nullable Sha256Hash rejectedMessageHash,
...
+    public int messageSize() {
+        int rejectedMessageLength = rejectedMessage.getBytes(StandardCharsets.UTF_8).length;
+        int size = VarInt.sizeOf(rejectedMessageLength) + rejectedMessageLength;
+        size++; // code
+        int reasonLength = reason.getBytes(StandardCharsets.UTF_8).length;
+        size += VarInt.sizeOf(reasonLength) + reasonLength;
+        if ("block".equals(rejectedMessage) || "tx".equals(rejectedMessage))
+            size += Sha256Hash.LENGTH;
+        return size;
+    }
```
msgilligan, line 125: This one is a little trickier, but still could be written in a more functional style.

msgilligan (approved):
Looks good, but I'd prefer if we wrote these in a functional style. I think all (or all but 1) of them could be written with `return` on the first line.

schildbach:
To me, it's important that you can easily compare the size calculation with the actual serialization. This is why it uses a similar format. Is it possible to keep that similarity in a functional style?

Also, all of the pre-existing messageSize() were written that way.

core/src/main/java/org/bitcoinj/core/AddressV1Message.java
schildbach: I stumbled across this as well. It may be a (pre-existing) bug.

schildbach:
I think this PR paves the way to implementing your suggestions: #3665

schildbach:
I rebased this on master and rewrite all methods to at least use functional style (not 100% though).

I'm requesting a review again because you said you approved inadvertently.

core/src/main/java/org/bitcoinj/core/BloomFilter.java
schildbach: Done.

core/src/main/java/org/bitcoinj/core/FilteredBlock.java
schildbach: Done.

core/src/main/java/org/bitcoinj/core/AddressV1Message.java
schildbach: Except for the terniary this is done.

core/src/main/java/org/bitcoinj/core/RejectMessage.java
schildbach: Except for the terniary this is done.

msgilligan (approved):
Approved.

schildbach:
Merged. I'll work on step 2 tomorrow.

PR #3663, opened 2025-03-27
ECKey: use ByteBuffer in `formatMessageForSigning()`

schildbach:


msgilligan (approved):
Approved.  (Buffers.*lengthPrefixedBytes* is a little long for something used repeatedly in the same method, maybe in the future we can figure out a reasonable way to shorten that.)

PR #3664, opened 2025-03-27
Buffers: add helpers `lengthPrefixedBytesSize()` and `lengthPrefixedStringSize()`

schildbach:
Those determine the number of bytes that are written by `writeLengthPrefixedBytes()` and `writeLengthPrefixedString()`, respectively.

Also use the new helpers where appropriate.

### DIFF
--- a/base/src/main/java/org/bitcoinj/base/internal/Buffers.java
+++ b/base/src/main/java/org/bitcoinj/base/internal/Buffers.java
@@ -74,6 +74,17 @@ public class Buffers {
     }
 
     /**
+     * Determines the number of bytes written by {@link #writeLengthPrefixedBytes(ByteBuffer, byte[])}. A typical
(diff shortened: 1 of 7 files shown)


schildbach:
In case you're wondering why the String-based methods are not used yet, that's because the serialization code that they're meant for is still OutputStream-based – but once we've migrated that to ByteBuffer they will be used as well.

schildbach:
Thinking about method names again. Maybe:

- `prefixedBytesLength()` → `lengthPrefixedBytesSize()`
- `prefixedStringLength()` → `lengthPrefixedStringSize()`

msgilligan (approved):
Approved!  (At some point in the future, it might be worth finding a short name than `lengthPrefixedBytes` -- I'd even be ok with `lbp` for this)

schildbach:
Merged.

PR #3665, opened 2025-03-27
Block, Transaction, TransactionInput, TransactionOutput, TransactionWitness: make `messageSize()` more functional

schildbach:
* replace variable re-assignments by concatenating additions
* use streams to walk and sum the various lists
    
Note: conditionals could be replaced with terniary expressions, but the resulting expression seems too difficult to understand as a whole.

This is another child of #3664.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -278,10 +278,10 @@ public class Block extends BaseMessage {
         int size = HEADER_SIZE;
         List<Transaction> transactions = getTransactions();
         if (transactions != null) {
-            size += VarInt.sizeOf(transactions.size());
-            for (Transaction tx : transactions) {
-                size += tx.messageSize();
-            }
+            size += VarInt.sizeOf(transactions.size()) +
+                    transactions.stream()
+                            .mapToInt(Transaction::messageSize)
+                            .sum();
         }
         return size;
     }
--- a/core/src/main/java/org/bitcoinj/core/PartialMerkleTree.java
+++ b/core/src/main/java/org/bitcoinj/core/PartialMerkleTree.java
@@ -154,11 +154,10 @@ public class PartialMerkleTree {
      * @return size of the serialized message in bytes
      */
     public int messageSize() {
-        int size = Integer.BYTES; // transactionCount
-        size += VarInt.sizeOf(hashes.size());
-        size += hashes.size() * Sha256Hash.LENGTH;
-        size += Buffers.lengthPrefixedBytesSize(matchedChildBits);
-        return size;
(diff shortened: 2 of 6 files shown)


schildbach:
It's probably not much of an issue, but I wonder how instance allocation differs between a foreach and a stream with a couple of operators. It's probably just one iterator for a foreach, and multiple instances for our "streamed sums"?

core/src/main/java/org/bitcoinj/core/PartialMerkleTree.java
```diff
@@ -154,11 +154,10 @@ public byte[] serialize() {
...
     public int messageSize() {
-        int size = Integer.BYTES; // transactionCount
-        size += VarInt.sizeOf(hashes.size());
-        size += hashes.size() * Sha256Hash.LENGTH;
-        size += Buffers.lengthPrefixedBytesSize(matchedChildBits);
-        return size;
+        return Integer.BYTES + // transactionCount
+                VarInt.sizeOf(hashes.size()) +
+                hashes.size() * Sha256Hash.LENGTH +
+                Buffers.lengthPrefixedBytesSize(matchedChildBits);
```
msgilligan, line 160: I prefer the `+` at the beginning of the line on these multiline sums. At some point we should probably standardize this and/or look at what other projects are suggesting.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -278,10 +278,10 @@ public int messageSize() {
...
         List<Transaction> transactions = getTransactions();
         if (transactions != null) {
-            size += VarInt.sizeOf(transactions.size());
-            for (Transaction tx : transactions) {
-                size += tx.messageSize();
-            }
+            size += VarInt.sizeOf(transactions.size()) +
+                    transactions.stream()
+                            .mapToInt(Transaction::messageSize)
+                            .sum();
```
msgilligan, line 284: I'll make a follow-on to this with the ternary (if you haven't already) for us to look at.

core/src/main/java/org/bitcoinj/core/PartialMerkleTree.java
msgilligan: I do love methods that are a single `return` expression!

(In a more modern functional language, the `return wouldn't even be necessary, of course)

msgilligan (approved):
Approved!

core/src/main/java/org/bitcoinj/core/Block.java
msgilligan: https://github.com/bitcoinj/bitcoinj/pull/3666/commits/30101a7b0858e326263cf0e4f16758574cbb5720

schildbach:
Merged.

PR #3666, opened 2025-03-27
Block: Simplify messageSize() with ternary

msgilligan:
~~Child of PR #3665~~

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -280,14 +280,13 @@ public class Block implements Message {
 
     @Override
     public int messageSize() {
-        int size = HEADER_SIZE;
-        if (!isHeaderOnly()) {
-            size += VarInt.sizeOf(transactions.size()) +
-                    transactions.stream()
-                            .mapToInt(Transaction::messageSize)
-                            .sum();
-        }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -275,15 +275,14 @@ private static List<Transaction> genesisTransactions() {
...
         List<Transaction> transactions = getTransactions();
-        if (transactions != null) {
-            size += VarInt.sizeOf(transactions.size()) +
-                    transactions.stream()
-                            .mapToInt(Transaction::messageSize)
-                            .sum();
-        }
-        return size;
+        return (transactions == null)
+                ? HEADER_SIZE
```
schildbach, line 280: Add HEADER_SIZE first, and not inside the terniary? That would be more to the point.

schildbach (commented):
Generally terniaries are syntactially so different to the `if` statements in the serialization that it's difficult to compare. It's already borderline for foreach vs stream, but if vs terniary is too much imho.

So I intentionally excluded them from my conversions for now and would try again once if or switch expressions become a reality.

schildbach:
I think this should be rebased, and then I'll take another look at how `messageSize()` can be compared to `write()` in a code review.

PR #3667, opened 2025-03-28
Wallettool: Modularize `call` method

ItoroD:
This PR is based off @msgilligan issue #2509. 

The PR focuses on making the Call method more modular so its easier to maintain. This is a part of the general Idea of improving wallettools as a whole. The plan is to make improvements to the module where necessary, and I will be doing this in parts. 

I will be making this into a draft for now as I can see a few places that still need improvement in the Call method.


wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -83,7 +80,6 @@
 import java.time.LocalDate;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
-import java.util.Collections;
 import java.util.List;
```
msgilligan, line 83: The removal of unused imports is unrelated to these changes so should be in a separate commit. Good catch though.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -186,7 +182,7 @@ public class WalletTool implements Callable<Integer> {
     private String pubKeyStr;
     @CommandLine.Option(names = "--privkey", description = "Specifies a WIF-, hex- or base58-encoded private key.")
     private String privKeyStr;
-    @CommandLine.Option(names = "--addr", description ="Specifies a Bitcoin address, either segwit or legacy.")
+    @CommandLine.Option(names = "--addr", description = "Specifies a Bitcoin address, either segwit or legacy.")
```
msgilligan, line 185: Can you please make a separate commit with these whitespace error corrections (these two and any others you can find)?

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -325,53 +321,78 @@ public static void main(String[] args) {
...
 
-        if (debugLog) {
-            BriefLogFormatter.init();
-            log.info("Starting up ...");
-        } else {
-            // Disable logspam unless there is a flag.
-            java.util.logging.Logger logger = LogManager.getLogManager().getLogger("");
-            logger.setLevel(Level.SEVERE);
-        }
+        setupLogging();
```
msgilligan, line 327: You should pass `debugLog` as a parameter to his method, rather than have it reference a member. This will begins the separation of pico-cli UX from the actual operation of the wallet implementation.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -325,53 +321,78 @@ public static void main(String[] args) {
...
 
-        if (action == ActionEnum.RAW_DUMP) {
-            // Just parse the protobuf and print, then bail out. Don't try and do a real deserialization. This is
-            // useful mostly for investigating corrupted wallets.
-            try (FileInputStream stream = new FileInputStream(walletFile)) {
-                Protos.Wallet proto = WalletProtobufSerializer.parseToProto(stream);
-                proto = attemptHexConversion(proto);
-                System.out.println(proto.toString());
-                return 0;
-            }
```
msgilligan, line 372: I think the RAW_DUMP is handled early so that it can be printed even if some of the consistency checks fail.  Moving RAW_DUMP to a method is a good idea, but moving it into `processAction` will prevent people from using this RAW_DUMP to dump/debug a corrupted wallet. If you could make just the move to a function a separate commit that would be great.

msgilligan (changes requested):
I was only able to do a partial review before the diff got out of sync and I found a major logic change (not doing RAW_DUMP before everything else) so please send me a series of smaller PRs. Start with the really easy ones like the whitespace changes. Thanks!

PR #3668, opened 2025-03-28
Transaction: don't wrap Coin values in BigInteger

schildbach:
This is likely a remnant from the past when we expressed Coin values as BigInteger.

As a side effect, we can get rid of a ByteUtils helper.


msgilligan (approved):
Approved.

schildbach:
Merged.

PR #3669, opened 2025-03-28
Transaction: optimize `getWeight()` to skip serialization

schildbach:
Now that we have accurate messageSize(), why not use it? We need to distinguish between segwit and non-segwit message sizes though.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -357,6 +357,16 @@ public class Transaction extends BaseMessage {
     }
 
     /**
+     * Determines if segwit serialization should be used for this transaction, based on whether it has any witnesses
+     * and on protocol version.
+     *
+     * @return {@code true} if segwit serialization should be used, {@code false} otherwise
+     */
+    private boolean useSegwitSerialization() {
+        return hasWitnesses() && allowWitness(protocolVersion);
+    }
+
+    /**
      * Returns the witness transaction ID (aka witness ID) as per BIP144. For transactions without witness, this is the
      * same as {@link #getTxId()}.
      *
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -390,16 +390,9 @@ private void invalidateCachedTxIds() {
...
-            stream.reset();
-            bitcoinSerializeToStream(stream, true);
-            final int totalSize = stream.size();
-            return baseSize * 3 + totalSize;
-        } catch (IOException e) {
-            throw new RuntimeException(e); // cannot happen
-        }
+        int baseSize = messageSize(false);
+        int totalSize = messageSize(true);
+        return baseSize * 3 + totalSize;
```
msgilligan, line 395: With a ternary this could be functional-style with a single `return` statement.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1485,7 +1478,10 @@ public synchronized Sha256Hash hashForWitnessSignature(
 
     @Override
     public int messageSize() {
-        boolean useSegwit = hasWitnesses() && allowWitness(protocolVersion);
+        return messageSize(hasWitnesses() && allowWitness(protocolVersion));
```
msgilligan, line 1481: I kinda miss the `useSegwit` name here. Maybe we could add a comment to help out the reader?

msgilligan (approved):
Approved. One ternary suggestion and one comment request.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: Yes, we can do that. I decided to do the simple refactor first, for easy reviewability.

schildbach:
I rebased this on #3672.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: The line now looks like this:

`return messageSize(useSegwitSerialization());`

Ok?

msgilligan (approved):
Approved! Very nice improvement!

schildbach:
Merged.

PR #3670, opened 2025-03-28
Convert BaseMessage to StreamSerializedMessage mix-in

msgilligan:
This is just a demonstration of the mix-in approach. As previously discussed, exposing `void bitcoinSerializeToStream(OutputStream stream)` is an unfortunate side-effect. And because we're moving to `ByteBuffer`-based serialization this is moot. So after @schildbach looks at this, we'll just close this issue.


schildbach:
If so, I'd perhaps pick a non-substantive name, like `BitcoinSerializable`.

msgilligan:
Yeah, this was meant as a throw-away. Just to demonstrate what the mix-in would look like. Especially since you have one PRs for `ByteBuffer`-based serialization. We could consider doing this for `ByteBuffer` and also using it for our data/sub-message classes, too.

msgilligan:
Closing as obsolete.

PR #3675, opened 2025-03-29
Wallettool refactor: Correct whitespaces and remove unused import

ItoroD:
This pr is based off feedback from #3667. Feedback [here](url)


msgilligan (approved):
Approved!

schildbach:
I merged both commits.

The single whitespace alone I probably would have rejected, it's such an extremely small change. What I usually do in such "cleanup" cases is that I try to clean up an entire file, or even entire package. Which I just did for `WalletTool` and I found many more whitespace anomalies and fixed them in the same commit.

ItoroD:
@schildbach thanks for the feedback. Will keep that in mind for next time.

PR #3677, opened 2025-03-29
Wallettool modularize `call()` method

ItoroD:
This pr is based on feedback [here ](https://github.com/bitcoinj/bitcoinj/pull/3667#discussion_r2019570790)in PR #3667

This contains 2 commits that
- Moves log setup to its own method
- Moves default chain setup to its own method

### DIFF
--- a/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
+++ b/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
@@ -332,10 +332,7 @@ public class WalletTool implements Callable<Integer> {
 
         setupLogging(debugLog);
         params = NetworkParameters.of(net);
-        String fileName = String.format("%s.chain", net);
-        if (chainFile == null) {
-            chainFile = new File(fileName);
-        }
+        setDefaultChainFile(net);
         Context.propagate(new Context());
 
         if (conditionStr != null) {
@@ -465,6 +462,12 @@ public class WalletTool implements Callable<Integer> {
         shutdown();
 
         return 0;
+    }
+
+    private void setDefaultChainFile(BitcoinNetwork net) {
+        if (chainFile == null) {
+            chainFile = new File(String.format("%s.chain", net));
+        }
     }
 
     private void setupLogging(boolean debugLog){
--- a/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
+++ b/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
@@ -332,7 +332,11 @@ public class WalletTool implements Callable<Integer> {
 
         setupLogging(debugLog);
         params = NetworkParameters.of(net);
-        setDefaultChainFile(net);
+
+        if(chainFile == null){
+            chainFile = getDefaultChainFile(net);
+        }
+        getDefaultChainFile(net);
         Context.propagate(new Context());
 
         if (conditionStr != null) {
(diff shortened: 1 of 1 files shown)


schildbach:
Now that #3675 is merged to master, can you rebase this PR on master?

ItoroD:
Yes I will do so and force push

schildbach:
Just in case you are not yet aware, we *do* have tests for WalletTool (in the usual place), so theoretically we *could* add a test that RAW_DUMP still works on unconsistent wallets.

ItoroD:
> Just in case you are not yet aware, we _do_ have tests for WalletTool (in the usual place), so theoretically we _could_ add a test that RAW_DUMP still works on unconsistent wallets. Or, even better, verify that it had worked like that _before_ your change.

Good idea. I believe this should be for a seperate PR right? As I have not touched RAW_DUMP here

schildbach:
> Good idea. I believe this should be for a seperate PR right? As I have not touched RAW_DUMP here

Yes, most likely. But I don't want to derail you from what you were initially targetting at.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -474,6 +464,23 @@ else if (feeSatPerVbyteStr != null)
         return 0;
     }
 
+    private void setDefaultChainFile(BitcoinNetwork net) {
+        if (chainFile == null) {
+            chainFile = new File(String.format("%s.chain", net));
+        }
+    }
+
+    private void setupLogging(boolean debugLog){
```
msgilligan, line 473: This should be made a `static` method since it doesn't rely on instance state. The `static` modifier makes it more obvious no state is being used and may make future refactoring easier.

msgilligan:
More testing is definitely needed. By factoring most of the logic into a separate class (independent of picocli) writing tests should become easier.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -474,6 +464,23 @@ else if (feeSatPerVbyteStr != null)
         return 0;
     }
 
+    private void setDefaultChainFile(BitcoinNetwork net) {
+        if (chainFile == null) {
+            chainFile = new File(String.format("%s.chain", net));
+        }
+    }
```
msgilligan, line 471: Moving this code to a method isn't much of an improvement, since it is only 2 or 3 lines of code _and_ it mutates state. My suggestion is to add a `static` method `getDefaultChainFile(BitcoinNetwork net)` that returns `File`. That will fully encapsulate the naming (and eventually file path calculation) of the chain file. We'll still have the 3-line `if` but it will be more clear to readers what is being mutated.

msgilligan (changes requested):
Almost there. I have two requests.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
ItoroD: The idea for moving the code to a method is not limited to reducing the number of lines, reusability is something I was going for as well. for example `if (!walletFile.exists())` and it body was called 2 or 3 times in the call method, I intend to move that out too. 

I do agree making this block of code a static method is way better. I will update this.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
ItoroD: Noted. Will update

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
ItoroD: @msgilligan Also, given that the `chainFile` is a non-static field we might want to pass the file in the static method. Like this `getDefaultChainFile(BitcoinNetwork net, File chainFile)` (updated your method suggesting with additional param)

ItoroD: Alright. Understood

ItoroD: Should I close this PR? @msgilligan

msgilligan: I'm just noticing this. Yes, if there's nothing left here that we need you should close it. Otherwise rebase it.

PR #3678, opened 2025-03-29
TransactionTest: rewrite tests to do without `HugeDeclaredSizeTransaction`

schildbach:
Three tests are changed to manually construct the out-of-spec serialized states designed to overwhelm `Transaction.read()`.

As a consequence, `HugeDeclaredSizeTransaction` is removed.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/BlockTest.java
+++ b/core/src/test/java/org/bitcoinj/core/BlockTest.java
@@ -38,17 +38,16 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.OutputStream;
 import java.math.BigInteger;
+import java.nio.Buffer;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.time.Instant;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.List;
 
+import static org.bitcoinj.base.internal.Preconditions.checkState;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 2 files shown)


schildbach:
I pushed a commit on top which does a very similar thing for a Block test, removing an anonymous subclass of Block.

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -669,100 +669,62 @@ static <T, R extends T> IntFunction<R[]> genericArray(IntFunction<T[]> arrayCrea
         return size -> (R[]) arrayCreator.apply(size);
     }
 
-    @Test
+    @Test(expected = BufferUnderflowException.class)
```
msgilligan, line 672: I suppose using the `expected` feature to catch and validate the exception could be a separate commit.

core/src/test/java/org/bitcoinj/core/TransactionTest.java
```diff
@@ -669,100 +669,62 @@ static <T, R extends T> IntFunction<R[]> genericArray(IntFunction<T[]> arrayCrea
...
+        ByteBuffer buf = ByteBuffer.allocate(4 + huge.getSizeInBytes() + zero.getSizeInBytes() + 4);
+        writeInt32LE(1, buf); // version
+        huge.write(buf); // txin count, txins intentionally missing
+        zero.write(buf); // txout count, no txouts
+        writeInt32LE(0, buf); // lock_time
+        checkState(!buf.hasRemaining());
+        ((Buffer) buf).rewind();
+
+        // make sure this safely throws BufferUnderflowException rather than running out of memory
+        Transaction.read(buf);
```
msgilligan, line 686: This is definitely the right approach, but it is hard for me to quickly review this at the moment. It might be helpful if you created a method (or maybe even a class) for creating invalid `ByteBuffer`s for testing. It might even have the same parameters as before (e.g. `hackInputsSize`)

msgilligan (changes requested):
The overall approach is great (eliminating the subclass and eliminating the need for  `Transaction` and `Block` objects that are invalid.

I would be willing to approve as-is, but got interrupted in the middle of reviewing. If you agree with my request and feel like making those changes, I'll review again. If you don't want to make those changes, I can do a more thorough review on the code as is and if I don't find any further issues, I will approve.

schildbach:
I amended both commits. I've uppercase the word "HUGE" in comments to emphasize the actual hack vs regular data.

In the second commit, I'm now re-using existing header serialization so it's really just "header + hack".

core/src/test/java/org/bitcoinj/core/TransactionTest.java
schildbach: I've not yet done this second commit (actually 2 seconds commits). I hope it's understandable as is now. But if you think it's necessary, I can do it.

core/src/test/java/org/bitcoinj/core/BlockTest.java
```diff
@@ -38,17 +38,16 @@
 import org.junit.Test;
 
 import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.OutputStream;
 import java.math.BigInteger;
+import java.nio.Buffer;
```
msgilligan, line 42: It's so nice to see `IOException` disappearing along with the change to `Buffer`.

core/src/test/java/org/bitcoinj/core/TransactionTest.java
msgilligan: I don't think it's necessary.

msgilligan (approved):
Approved! Nice work!

schildbach:
Merged.

PR #3681, opened 2025-03-30
HDPath code cleanup

msgilligan:
Improvements to HDPath as a series of 8 commits.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/ChildNumber.java
+++ b/core/src/main/java/org/bitcoinj/crypto/ChildNumber.java
@@ -73,9 +73,10 @@ public class ChildNumber implements Comparable<ChildNumber> {
      */
     public static ChildNumber parse(String str) {
         boolean isHard = str.endsWith("H");
-        if (isHard) str = str.substring(0, str.length() - 1).trim();
-        int nodeNumber = Integer.parseInt(str);
-        return new ChildNumber(nodeNumber, isHard);
+        String nodeNumber = (isHard)
+                ? str.substring(0, str.length() - 1)
+                : str;
+        return new ChildNumber(Integer.parseInt(nodeNumber.trim()), isHard);
     }
 
     /** Returns the uint32 encoded form of the path element, including the most significant bit. */
--- a/core/src/main/java/org/bitcoinj/crypto/HDPath.java
+++ b/core/src/main/java/org/bitcoinj/crypto/HDPath.java
@@ -50,6 +50,8 @@ public class HDPath extends AbstractList<ChildNumber> {
     private static final char PREFIX_PRIVATE = 'm';
     private static final char PREFIX_PUBLIC = 'M';
     private static final char SEPARATOR = '/';
+    private static final String PREFIX_PRIVATE_STRING = Character.toString(PREFIX_PRIVATE);
+    private static final String PREFIX_PUBLIC_STRING = Character.toString(PREFIX_PUBLIC);
     private static final InternalUtils.Splitter SEPARATOR_SPLITTER = s -> Stream.of(s.split("/"))
             .map(String::trim)
             .collect(Collectors.toList());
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -172,13 +174,13 @@ public static HDPath m(ChildNumber... children) {
...
         List<String> parsedNodes = SEPARATOR_SPLITTER.splitToList(path);
-        boolean hasPrivateKey = false;
-        if (!parsedNodes.isEmpty()) {
-            final String firstNode = parsedNodes.get(0);
-            if (firstNode.equals(Character.toString(PREFIX_PRIVATE)))
-                hasPrivateKey = true;
-            if (hasPrivateKey || firstNode.equals(Character.toString(PREFIX_PUBLIC)))
-                parsedNodes.remove(0);
+
+        String firstNode = parsedNodes.get(0);
```
schildbach, line 178: How do you know the list isn't empty at this point?

schildbach (commented):
I merged all except 2 commits and added one review comment.

Can you rebase this on master?

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: We don't. But the prefix comparisons that follow work correctly if `firstNode` is `null`.

schildbach: If parserNodes is an empty list, it will throw at .get(0)?

core/src/main/java/org/bitcoinj/crypto/ChildNumber.java
```diff
@@ -73,9 +73,10 @@ public ChildNumber(int i) {
      */
     public static ChildNumber parse(String str) {
         boolean isHard = str.endsWith("H");
-        if (isHard) str = str.substring(0, str.length() - 1).trim();
-        int nodeNumber = Integer.parseInt(str);
-        return new ChildNumber(nodeNumber, isHard);
+        String nodeNumber = (isHard)
+                ? str.substring(0, str.length() - 1)
+                : str;
+        return new ChildNumber(Integer.parseInt(nodeNumber.trim()), isHard);
```
schildbach, line 79: I just analyzed for a while why you changed the logic around `trim()` and then noticed the somewhat hidden comment about the bugfix.

I think if we encounter such bugs, we should fix them on a separate commit/PR, and preferable directly based on master so that the chance of a clean backport is as high as possible.

Anyway, I split this commit into two commits and merged.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: Oops. I wasn't thinking correctly, I was thinking it would return `null` for an empty `List`.

One could argue that an empty string (and thus an empty `parsedNode` list) is invalid and _should_ throw an exception, but that would be a breaking change and would break current unit tests.

Let me rethink this.

msgilligan: Update: it didn't require much rethinking, just a test for empty. Fixed and force-pushed.

core/src/main/java/org/bitcoinj/crypto/ChildNumber.java
msgilligan: Yeah, I should have used a separate commit or at least mentioned the fix in the 1-line summary.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -172,13 +174,13 @@ public static HDPath m(ChildNumber... children) {
...
     public static HDPath parsePath(@Nonnull String path) {
         List<String> parsedNodes = SEPARATOR_SPLITTER.splitToList(path);
-        boolean hasPrivateKey = false;
-        if (!parsedNodes.isEmpty()) {
-            final String firstNode = parsedNodes.get(0);
-            if (firstNode.equals(Character.toString(PREFIX_PRIVATE)))
-                hasPrivateKey = true;
-            if (hasPrivateKey || firstNode.equals(Character.toString(PREFIX_PUBLIC)))
-                parsedNodes.remove(0);
+
```
schildbach, line 177: Honestly, I think this is a perfect example where an early exit would be much more readable:

`if (parsedNodes.isEmpty()) return new HDPath(false, Collections.emptyList());`

…and the entire rest of the method wouldn't need to deal with this case.

schildbach: By the way, maybe our `splitToList()` guarantees a non-empty list? I vaguely remember that sometimes `"".split()` returns `[ "" ]`. But this would need to be documented first, and maybe test cases added if not already present.

msgilligan: I'm rethinking the approach further and going in a slightly different direction. Starting with: https://github.com/bitcoinj/bitcoinj/pull/3685

msgilligan: And we should definitely add more test cases. Note also, that 1b3863c8376e0ada3ffd87cae2e7736fbfeebe80 allow us to add tests at a lower level.

msgilligan:
If PR #3686 (or similar) is merged, we can close this one.

msgilligan:
PR #3686 was merged. Closing.

PR #3682, opened 2025-03-30
StoredUndoableBlock: make immutable

schildbach:
Also add `@Nullable` annotations, require non-null arguments to constructors, and make a field private.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/StoredUndoableBlock.java
+++ b/core/src/main/java/org/bitcoinj/core/StoredUndoableBlock.java
@@ -19,7 +19,9 @@ package org.bitcoinj.core;
 
 import org.bitcoinj.base.Sha256Hash;
 
+import javax.annotation.Nullable;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Contains minimal data necessary to disconnect/connect the transactions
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/StoredUndoableBlock.java
```diff
@@ -27,14 +28,18 @@
...
     
     // Only one of either txOutChanges or transactions will be set
-    private TransactionOutputChanges txOutChanges;
-    private List<Transaction> transactions;
+    @Nullable
+    private final TransactionOutputChanges txOutChanges;
+    @Nullable
+    private final List<Transaction> transactions;
     
     public StoredUndoableBlock(Sha256Hash hash, TransactionOutputChanges txOutChanges) {
```
msgilligan, line 45: `txOutChanges` should be annotated as nullable or non-nullable as well as should `transactions` in the other constructor. Or, if we are trying to consistently have unannotated parameters be considered non-nullable, we should at least add some JavaDoc to the two constructors explaining their use-cases.

msgilligan: We could also consider a private canonical constructor with a pre-condition that enforces that both can't be null or both can't be set.

msgilligan (changes requested):
This looks good, but I have a few minor request. We also might want to update the commit description to say "make shallowly immutable".

msgilligan:
As I was writing my review you force-pushed checks with `Objects.requireNonNull()` that address the main issue I was raising. I still think JavaDoc on the contructors might be good, and a private canonical constructor is also worth considering.

schildbach:
> We also might want to update the commit description to say "make shallowly immutable".

Good idea. I added that wording in the commit description and the class-level JavaDoc.

I assume we use "shallowly immutable" if there are contained objects that are *not* immutable (and Java collections won't be for quite a while…)?

core/src/main/java/org/bitcoinj/core/StoredUndoableBlock.java
schildbach: I would add JavaDocs to the constructors if I would understand the usecases. I'm currently happy we're able to make this class shallowly immutable without any tests failing.

schildbach: The private canonical constructor I think is a candidate for when we migrate the existing constructors to static constructors. But that's out of scope I'd say.

msgilligan:
> I assume we use "shallowly immutable" if there are contained objects that are _not_ immutable (and Java collections won't be for quite a while…)?

Yes. But I would say an object is "effectively deeply immutable" if it makes defensive copies of mutable contained objects and never allows the outside world to access them.

msgilligan (approved):
Approved!

schildbach:
Merged.

PR #3684, opened 2025-03-30
Block: make field `prevBlockHash` immutable

schildbach:
Also rename accessor to `prevBlockHash()`. Keep the old as deprecated.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -697,9 +697,19 @@ public class Block implements Message {
         hash = null;
     }
 
-    /** Returns the version of the block data structure as defined by the Bitcoin protocol. */
-    public long getVersion() {
+    /**
+     * Returns the version of the block data structure.
+     *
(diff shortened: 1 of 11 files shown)


schildbach:
I pushed a commit on top that renames setVersion() → version(), because the field is already immutable.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -696,16 +704,17 @@ public long getVersion() {
 
     /**
      * Returns the hash of the previous block in the chain, as defined by the block header.
+     *
+     * @return hash of the previous block
      */
-    public Sha256Hash getPrevBlockHash() {
+    public Sha256Hash prevBlockHash() {
```
msgilligan, line 720: We know we're operating on `Block`, can we shorten this to `block.prevHash()`?

msgilligan (approved):
Approved, with a suggestion to shorten `prevBlockHash()` to `prevHash()`.

schildbach:
Merged.

PR #3687, opened 2025-03-31
Block: add method `sigOpCount()` that sums up SigOps of transactions

schildbach:
Use it where appropriate.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -497,14 +497,20 @@ public class Block implements Message {
                     TimeUtils.dateTimeFormat(allowedTime), allowedTime.toEpochMilli()));
     }
 
+    /**
+     * Sums up all SigOps in all transactions of this block.
+     *
+     * @return sum of SigOps
+     */
+    public int sigOpCount() {
+        Objects.requireNonNull(transactions);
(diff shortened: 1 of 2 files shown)


schildbach:
Rebased.

msgilligan (approved):
Approved. Nice!

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -497,14 +497,19 @@ private void checkTimestamp() throws VerificationException {
                     TimeUtils.dateTimeFormat(allowedTime), allowedTime.toEpochMilli()));
     }
 
+    /**
+     * Sums up all SigOps in all transactions of this block.
+     *
+     * @return sum of SigOps
+     */
+    public int sigOpCount() {
+        return transactions.stream().mapToInt(Transaction::getSigOpCount).sum();
```
msgilligan, line 506: Needs null check!

msgilligan (changes requested):
Nul-check!

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: Added, although it would throw NPE immediately at the beginning of the existing statement.

msgilligan (approved):
Approved!

PR #3689, opened 2025-03-31
Block: add a test helper `replaceTransactions()`

schildbach:
This is a more powerful replacement to the former helper `addTransaction(tx, runSanityChecks)`. Use it to get rid of direct mutations of the `transactions` field.


schildbach:
Merged.

PR #3692, opened 2025-03-31
HDPath: add Prefix.unspecified()

msgilligan:
Add a factory method to return the default prefix, to be used when a prefix is unspecified. (I would have named the method `default`, but that is a reserved word.)


schildbach (commented):
I now suspect that our behaviour to default on public (or private, for that matter) prefix might be wrong. Don't absolute paths always have a prefix, either `M` or `m`? I'd say no prefix means relative, and all other prefixes would need to throw.

However, I'm no expert on these paths so I may misunderstand something.

msgilligan:
I'll mark this as DRAFT until we resolve other issues with HDPath.

PR #3694, opened 2025-04-01
DeterministicKey: Deprecate List<ChildNumber> constructors

msgilligan:


msgilligan:
I believe this has been made obsolete by other work we have done, and _more importantly_ I think we should **not use deprecation** for all the constructors and methods that were using `List<ChildNumber>`.

msgilligan:
Closing as obsolete.

PR #3695, opened 2025-04-01
HDPath: Make `Prefix` public, `prefix()` accessor, new `of` variant

msgilligan:
* Make the nested `Prefix` type public
* Add `.prefix()` accessor
* Add a public `of` method that takes `Prefix`
* Use the new methods internally where appropriate


schildbach:
Merged.

PR #3696, opened 2025-04-01
HDPath: Enforce maximum path length in constructor

msgilligan:


msgilligan (commented):
There may be an off-by-one error in the way the limit is enforced. If there is more than one possible limit, we should probably enforce the most-restrictive limit in the interest of safety.

schildbach:
Note that this issue affects only the serialization of deterministic keys into base58-encoded strings.

Inside a wallet for example, I'd say the path has no length limit (or it's much higher):

```
message DeterministicKey {
  // Random data that allows us to extend a key. Without this, we can't figure out the next key in the chain and
  // should just treat it as a regular ORIGINAL type key.
  required bytes chain_code = 1;

  // The path through the key tree. Each number is encoded in the standard form: high bit set for private derivation
  // and high bit unset for public derivation.
  repeated uint32 path = 2;

  // How many children of this key have been issued, that is, given to the user when they requested a fresh key?
  // For the parents of keys being handed out, this is always less than the true number of children: the difference is
  // called the lookahead zone. These keys are put into Bloom filters so we can spot transactions made by clones of
  // this wallet - for instance when restoring from backup or if the seed was shared between devices.
  //
  // If this field is missing it means we're not issuing subkeys of this key to users.
  optional uint32 issued_subkeys = 3;
  optional uint32 lookahead_size = 4;

  /**
   * Flag indicating that this key is a root of a following chain. This chain is following the next non-following chain.
   * Following/followed chains concept was used for married keychains, where the set of keys combined together to produce
   * a single P2SH multisignature address. It is currently unused, but this flag is preserved.
   */
  optional bool isFollowing = 5;

  // Number of signatures required to spend. This field was needed only for married keychains to reconstruct KeyChain
  // and represents the N value from N-of-M CHECKMULTISIG script. It is currently unused, but this number is preserved.
  // For regular single keychains it will always be 1.
  optional uint32 sigsRequiredToSpend = 6 [default = 1];
}
```

Also, the BIP32 spec seems to impose no limit. So we have the problem that long paths could already be used and working fine, and with this change we'd make them unvalid. I can't think of a usecase for such long paths, but at least we should consider the possibility. Maybe make it a warning?

schildbach:
I wonder if wallet descriptors can replace base58-encoded deterministic keys? If so, we could consider to deprecate the latter.

schildbach (changes requested):
This is missing a test or two.

Also, I wonder if this is the right place to impose a limit. If we want to prevent the derivation of deterministic keys that cannot be serialized, we should place the limit at key derivation imho.

msgilligan:
> This is missing a test or two.
> 
> Also, I wonder if this is the right place to impose a limit. If we want to prevent the derivation of deterministic keys that cannot be serialized, we should place the limit at key derivation imho.

I think you are correct. This is the wrong place to impose a limit. Closing...

PR #3698, opened 2025-04-02
LazyECPoint: make immutable

schildbach:
~~Child of #3697.~~

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
+++ b/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
@@ -26,10 +26,13 @@ import java.util.Arrays;
 import java.util.Objects;
 
 /**
- * A wrapper around a SECP256K1 ECPoint that delays decoding of the point for as long as possible. This is useful because point
- * encode/decode in Bouncy Castle is quite slow especially on Dalvik, as it often involves decompression/recompression.
+ * A wrapper around a SECP256K1 ECPoint that delays decoding of the point for as long as possible. This is useful
+ * because point encode/decode in Bouncy Castle is quite slow especially on Dalvik, as it often involves
+ * decompression/recompression.
+ * <p>
+ * Apart from the lazy field {@link #point}, instances of this class are immutable.
  */
-public class LazyECPoint {
+public final class LazyECPoint {
     private static final ECCurve curve = ECKey.CURVE.getCurve();
 
     // bits will be null if LazyECPoint is constructed from an (already decoded) point


schildbach:
I pushed a commit on top that re-uses a LazyECPoint, rather than cloning it.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -172,7 +172,7 @@ public DeterministicKey(List<ChildNumber> childNumberPath,
     
     /** Clones the key */
     public DeterministicKey(DeterministicKey keyToClone, DeterministicKey newParent) {
-        super(keyToClone.priv, keyToClone.pub.get(), keyToClone.pub.isCompressed());
+        super(keyToClone.priv, keyToClone.pub);
```
msgilligan, line 175: I'm not sure what is going on here. Partly, I'm confused by the `/** Clones the key */` comment which is not clear to me which key is being cloned. And with the change there is no cloning going on.

msgilligan (changes requested):
I think the comment on the "clone constructor" should be changed and/or clarified.

I also object to the "make immutable" title of the first commit when it is really just making it "final".

schildbach:
> I also object to the "make immutable" title of the first commit when it is really just making it "final".

Sealing the class is one of the preconditions of making immutable, right? The other fields were already final, so the change is minimal.

For me, the only question was if cache fields (like `point`) are allowed in immutable classes. I think for performance reasons we can allow them, as long as the code guarantees that they get a value derived only from the other final fields, and the value never changes once assigned.

schildbach:
> I think the comment on the "clone constructor" should be changed and/or clarified.

I have commits about the clone constructor coming up. It should be made a static `withSomething()`, and the "something" is also not entirely clear to me but we'll figure this out.

However, this PR is not about the clone constructor. It's about re-using of LazyECPoint, rather than disassembling it into components and then reassembling them in the called constructor. I think because LazyECPoint is now immutable, we can safely re-use it and it will make our code cleaner.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
schildbach: > I think the comment on the "clone constructor" should be changed and/or clarified.

I have commits about the clone constructor coming up. It should be made a static `withSomething()`, and the "something" is also not entirely clear to me but we'll figure this out.

However, this commit is not about the clone constructor. It's about re-using of LazyECPoint, rather than disassembling it into components and then reassembling them in the called constructor. I think because LazyECPoint is now immutable, we can safely re-use it and it will make our code cleaner.

schildbach:
Rebased.

schildbach:
I decided to merge the second commit so we can carry on with adding a canonical constructor to DeterministicKey and discuss about the definition of immutability later.

schildbach:
Rebased on master. This is now just the sealing of the class and the usual JavaDoc comment to declare immutability.

I think we should merge this, or otherwise find a more fitting wording for immutable classes that maintain some form of internal cache.

msgilligan (approved):
Approved.

msgilligan:
> JavaDoc comment to declare immutability.
> 
> or otherwise find a more fitting wording for immutable classes that maintain some form of internal cache.

I think the wording is close enough.

[JEP 502](https://openjdk.org/jeps/502) adds a type that could be used to declare:

```
 private final StableValue<ECPoint> point = StableValue.of();
 ```

The JEP refers to this as _deferred immutability_.

schildbach:
Merged.

PR #3699, opened 2025-04-02
ECKey: introduce two static constructors `random()`

schildbach:
This is composed of 2 commits:

- rename of a constant
- introduce the two static constructors

~~This is a child of #3698.~~


msgilligan (approved):
Approved

PR #3701, opened 2025-04-02
ECKey: make constructor based on `ECPoint` private

schildbach:
Change two usages to construct the `LazyECPoint` themselves.

This is a child of #3700.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -94,7 +94,7 @@ public class DeterministicKey extends ECKey {
                             byte[] chainCode,
                             BigInteger priv,
                             @Nullable DeterministicKey parent) {
-        super(priv, ECKey.publicPointFromPrivate(priv), true);
+        super(priv, new LazyECPoint(ECKey.publicPointFromPrivate(priv), true));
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -93,7 +93,7 @@ public DeterministicKey(HDPath hdPath,
                             byte[] chainCode,
                             BigInteger priv,
                             @Nullable DeterministicKey parent) {
-        super(priv, ECKey.publicPointFromPrivate(priv), true);
+        super(priv, new LazyECPoint(ECKey.publicPointFromPrivate(priv), true));
```
msgilligan, line 96: Why not create a 1-parameter constructor that takes a private key? It could then do the `new LazyECPoint(ECKey.publicPointFromPrivate(priv), true)` internally and simplify the callers.

msgilligan (approved):
Approved.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
schildbach: Sorry, I missed that comment earlier. I plan to modernize the DeterrministicKey constructors once we've got a canonical constructor. And this (simple) PR was a precondition for having one.

PR #3702, opened 2025-04-02
DeterministicKey: make field `parentFingerprint` immutable

schildbach:
Also use one single constructor that is able to set an arbitrary parent fingerprint.

Child of #3701.


schildbach:
This is most likely bugged, because it uses `M` (capital) prefix for private deterministic keys. Need to debug that part first.

schildbach:
Closing, in favor to #3707 and various preceding work.

PR #3704, opened 2025-04-02
HDPath: implement `equals()` and `hashCode()`

schildbach:
Take the `hasPrivateKey` (prefix) into account.

Also add a test, and change some that don't work any more.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/HDPath.java
+++ b/core/src/main/java/org/bitcoinj/crypto/HDPath.java
@@ -169,6 +169,20 @@ public abstract class HDPath extends AbstractList<ChildNumber> {
         public HDPartialPath asPartial() {
             return new HDPartialPath(this.childNumbers);
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if ((o == null) || getClass() != o.getClass()) return false;
+            HDFullPath other = (HDFullPath) o;
+            return Objects.equals(this.hasPrivateKey, other.hasPrivateKey) &&
+                    super.equals(other);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(this.hasPrivateKey, super.hashCode());
+        }
     }
 
     public static class HDPartialPath extends HDPath {
@@ -540,7 +554,7 @@ public abstract class HDPath extends AbstractList<ChildNumber> {
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-        if ((o == null) || !(o instanceof HDPath)) return false;
+        if ((o == null) || getClass() != o.getClass()) return false;
         HDPath other = (HDPath) o;
         return Objects.equals(this.childNumbers, other.childNumbers);
     }
--- a/core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
+++ b/core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
@@ -278,7 +278,6 @@ public class HDPathTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


schildbach:
This currently fails tons of tests, but I believe we need to fix this. It hides actual bugs, e.g. this one:

https://github.com/bitcoinj/bitcoinj/issues/3705

msgilligan:
PR #3721 has a YOLO `equals()` and `hasCode()`. It also sorts out the various use cases so that full/partial and public/private are resolved, so that tests that compare `HDPath` instances with `equals()` are passing.

msgilligan (approved):
Approved. But I'm not sure we want to merge this independently of PR #3723 which fixes the tests.

msgilligan:
@schildbach I created PR #3729 to partially implement this change (but continue ignoring `hasPrivateKey`). Once that is merged we can rebase this PR upon that one.

msgilligan:
@schildbach This can be rebased now

schildbach:
Rebased.

msgilligan:
> Rebased.

According to the timeline it looks like you rebased this _after_  PR #3729 was merged, but when I look at the diff or the warning about merge conflicts it look like it was rebased before 3729 was merged.

Do you mind rebasing again so we can get a clean diff of the changes to `equals` and `hashCode`?

schildbach:
Rebased on master.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -162,6 +162,20 @@ public HDFullPath parent() {
...
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if ((o == null) || getClass() != o.getClass()) return false;
+            HDFullPath other = (HDFullPath) o;
+            return Objects.equals(this.hasPrivateKey, other.hasPrivateKey) &&
+                    super.equals(other);
+        }
```
msgilligan, line 173: Can't this be shortened to:
```
            if (!super.equals(o)) return false;
            HDFullPath other = (HDFullPath) o;
            return Objects.equals(this.hasPrivateKey, other.hasPrivateKey);
```

msgilligan (changes requested):
One minor suggestion/request.

msgilligan:
@schildbach See PR #3750 this enables the full implementation of `equals()` and `hashCode()` and passes all tests.

schildbach:
Rebased.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
schildbach: Not sure, I'm following a pattern here that we established years ago – I think based on "Java best practices".

I would suggest that if we'd like to change this pattern, we should do it for the entire codebase and hence it would be out of scope for this PR.

msgilligan: The thing I like about the shortened version is hit makes it more clear that it's the same as `super` but adds the comparison of `hasPrivateKey`. Not a big deal, though.

msgilligan (approved):
Approved!

core/src/main/java/org/bitcoinj/crypto/HDPath.java
schildbach: > I think based on "Java best practices"

For reference: https://github.com/bitcoinj/bitcoinj/pull/65

schildbach:
Yipiieeh!!! Thanks for taking this effort, it's a huge improvement to test coverage I'd say.

PR #3706, opened 2025-04-03
DeterministicKey: mark parent as @nullable, make ascertainParentFingerprint() static

msgilligan:
These commits can be merged separately.

Marking parent as @Nullable is simple and standalone.

In `ascertainParentFingerprint()`, there is no advantage to accessing `parent` as a member, whereas a static method with `parent` as a parameter enables other uses. Specifically, when `ascertainParentFingerprint()` is non-static its use is limited to after any calls to other constructors (via  `this()),  which limits where it can be called.


schildbach:
Merged!

I had those two on my YOLO branch about a DeterministicKey canonical constructor as well (-:

PR #3708, opened 2025-04-03
DeterministicKey: add makeHDPath() static helper 

msgilligan:
Add a static helper method for use in constructors (and elsewhere)
that takes a List<ChildNumber> (which could already be an HDPath) and
returns it unmodified if it's an instance of HDPath. Otherwise it
creates one of the correct prefix based upon whether `priv != null`.

Child of PR #3707


msgilligan:
Closing as this is no longer needed since we are not using full paths as constructor parameters for `DeterministicKey`.

PR #3710, opened 2025-04-03
Block: add `transaction(index)` accessor

schildbach:
Use it to prevent accessing `getTransactions()` whereever possible.

Child of #3687.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -497,14 +497,19 @@ public class Block implements Message {
                     TimeUtils.dateTimeFormat(allowedTime), allowedTime.toEpochMilli()));
     }
 
+    /**
+     * Sums up all SigOps in all transactions of this block.
+     *
+     * @return sum of SigOps
+     */
+    public int sigOpCount() {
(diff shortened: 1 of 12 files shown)


msgilligan (approved):
Approved!

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -802,6 +807,17 @@ public List<Transaction> getTransactions() {
...
 
+    /**
+     * Gets the transaction at the given index.
+     *
+     * @param index index of the transaction to get
+     * @return transaction
+     * @throws IndexOutOfBoundsException if the given index is out of bounds
+     */
+    public Transaction transaction(int index) {
+        return transactions.get(index);
```
msgilligan, line 818: Needs null-check

msgilligan (changes requested):
Null-check here, too.

core/src/main/java/org/bitcoinj/core/Block.java
schildbach: Added, though it would throw NPE anyway.

msgilligan (approved):
Approved.

PR #3709, opened 2025-04-03
DeterministicKey: add `depth` to `equals()` and `hashCode()`

schildbach:
This breaks a test, so for now marked as draft.

org.bitcoinj.wallet.WalletTest#encryptDecryptWalletWithArbitraryPathAndScriptType

PR #3711, opened 2025-04-03
DeterministicKeyChain: Fix depth issue in `toEncrypted()`

msgilligan:
In the `DeterministicKeyChain` constructor that uses a `KeyCrypter` to encrypt an existing keychain, there was a bug where encrypted non-leaf keys that are grandchildren or lower of the root were being constructed with their parent set to root rather than their direct encrypted parent.

This commit fixes the issue.

The problem was undetected until PR #3709 added `depth` to `DeterministicKey`'s `equals` and `hashCode`, which was than detecting the incorrect depth after a roundtrip.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
@@ -431,9 +431,11 @@ public class DeterministicKeyChain implements EncryptableKeyChain {
         basicKeyChain.importKey(rootKey);
 
         for (HDPath path : getAccountPath().ancestors()) {
-            encryptNonLeaf(aesKey, chain, rootKey, path);
+            DeterministicKey parent = this.getKeyByPath(path.parent());
+            encryptNonLeaf(aesKey, chain, parent, path);
         }
-        DeterministicKey account = encryptNonLeaf(aesKey, chain, rootKey, getAccountPath());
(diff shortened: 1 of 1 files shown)


schildbach:
Thanks a lot for delving into this, and finding the cause of the issue!

Do you think there could be protobuf-serialized wallets out there with "corrupted" depth?

schildbach:
I just dived into the call hierarchy of encryptNonLeaf() and found parent is only used for these things:

- checking that the parent is always encrypted
- copying time from the parent
- deriving the new depth from the parent (+1)

I think those are all quite low risk of regression, except perhaps the depth change itself. But I think this cannot void any keys.

msgilligan:
> I think those are all quite low risk of regression, except perhaps the depth change itself. 

If you look at the failing unit/integration test, the incorrect depth shows up after `encryptedWallet.encrypt(PASSWORD1)`, but when the wallet is serialized and then deserialized (`encryptedWallet = roundTrip(encryptedWallet)`)  the correct depth is present after deserialization and a comparison of before and after serialization/deserialization fails (`assertEquals(watchingKey, roundTrippedActiveKeyChain.getWatchingKey());`) with the deserialized value being correct.

schildbach:
I can assert that with this fix, and #3709 cherry-picked on top, all tests pass.

schildbach (approved):
I think we should merge this, to master and 0.17.

PR #3712, opened 2025-04-03
Block: add iteration helper `forEachTransaction()`

schildbach:
Use it to prevent accessing `getTransactions()`

Child of #3687 and #3710.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -50,6 +50,7 @@ import java.util.EnumSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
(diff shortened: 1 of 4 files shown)


msgilligan (approved):
Approved!

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -818,6 +819,15 @@ public Transaction transaction(int index) {
         return transactions.get(index);
     }
 
+    /**
+     * Performs the given action for each transaction in this block.
+     *
+     * @param action action to be performed for each transaction
+     */
+    public void forEachTransaction(Consumer<Transaction> action) {
+        transactions.forEach(action);
```
msgilligan, line 828: Needs null check!

msgilligan (changes requested):
Oops, just found an issue.

schildbach:
At the moment, I'm not so sure if I'd like to merge this. It uses a lamda for the action, and in some cases this complicates usages to the point where I didn't replace `getTransactions()` with usage of this method. This is due to the `final` requirement when accessing variables outside of the lamda. So this method is a bit less universal than I hoped.

Also, I wonder if maybe exposing a stream (of transaction) would be more useful? What do you think?

msgilligan (approved):
Approved.

schildbach:
Rebased.

PR #3713, opened 2025-04-03
ChildKeyDerivationTest: add a few asserts on depth to `pubOnlyDerivation()`

schildbach:
This is another test that fails, but shouldn't afaict.


schildbach:
I cherry-pick #3707 on top, and as expected the test is fixed.

schildbach:
Closing in favor of #3725.

PR #3714, opened 2025-04-03
HDPath: add a precondition to the canonical constructor

schildbach:
Make sure `list` can't be null.

PR #3715, opened 2025-04-03
DeterministicKey: add a canonical constructor and use it

schildbach:
Child of #3714.

On top of the canonical constructor commit, there are two fixes and then the corresponding tests that uncovered the bugs.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -71,13 +71,8 @@ public class DeterministicKey extends ECKey {
                             LazyECPoint publicAsPoint,
                             @Nullable BigInteger priv,
                             @Nullable DeterministicKey parent) {
-        super(priv, publicAsPoint.compress());
-        checkArgument(chainCode.length == 32);
-        this.parent = parent;
-        this.childNumberPath = HDPath.M(Objects.requireNonNull(childNumberPath));
-        this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
-        this.depth = parent == null ? 0 : parent.depth + 1;
-        this.parentFingerprint = (parent != null) ? parent.getFingerprint() : 0;
+        this(priv, publicAsPoint.compress(), parent == null ? 0 : parent.depth + 1, parent,
+                parent != null ? parent.getFingerprint() : 0, chainCode, HDPath.M(childNumberPath), null, null);
     }
 
     public DeterministicKey(List<ChildNumber> childNumberPath,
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -160,25 +145,37 @@ public DeterministicKey(List<ChildNumber> childNumberPath,
...
-        this.parentFingerprint = this.parent.getFingerprint();
+    }
+
+    /**
+     * Canonical constructor.
+     *
+     * @param priv            private key, or {@code null} if public key only
+     * @param pub             public key, corresponding to private key (if present)
+     * @param depth           depth of this key in the path, {@code 0} means master key
+     * @param parent          parent deterministic key, or {@code null} if unknown or master key
```
msgilligan, line 166: @param parentFingerPrint is missing

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -160,25 +145,37 @@ public DeterministicKey(List<ChildNumber> childNumberPath,
...
+     * @param chainCode       32 bytes of chain code
+     * @param childNumberPath path leading up to this key
+     */
+    private DeterministicKey(@Nullable BigInteger priv, LazyECPoint pub, int depth, @Nullable DeterministicKey parent,
+                             int parentFingerprint, byte[] chainCode, HDPath childNumberPath) {
```
msgilligan, line 171: I would actually name the `HDPath` variable `hdPath` or something similar.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -160,25 +145,37 @@ public DeterministicKey(List<ChildNumber> childNumberPath,
...
+        super(priv, pub);
+        checkArgument(chainCode.length == 32);
+        this.depth = depth;
+        this.parent = parent;
+        this.parentFingerprint = parentFingerprint;
```
msgilligan, line 176: Shouldn't you be calling ` ascertainParentFingerprint`?

msgilligan (commented):
This is only a partial review.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
schildbach: Addressed.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
schildbach: Addressed, but for now only in the new constructor.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
schildbach: Addressed.

schildbach:
I also added `@Nullable EncryptedData encryptedPrivateKey, @Nullable KeyCrypter keyCrypter` to the canonical constructor, in case you've reviewed the old version.

schildbach:
I just added – as [second commit](https://github.com/bitcoinj/bitcoinj/pull/3715/commits/8493a6d4432b01bb034ee0a25c24fda0fa1596c0) – a warning about large depth.

schildbach:
I decided to un-YOLO this PR by moving all the bugfixes to their own PR: #3718 

So this PR is now a single commit and should be easier to review.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -71,13 +71,8 @@ public DeterministicKey(List<ChildNumber> childNumberPath,
...
                             @Nullable DeterministicKey parent) {
-        super(priv, publicAsPoint.compress());
-        checkArgument(chainCode.length == 32);
-        this.parent = parent;
-        this.childNumberPath = HDPath.M(Objects.requireNonNull(childNumberPath));
-        this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
-        this.depth = parent == null ? 0 : parent.depth + 1;
-        this.parentFingerprint = (parent != null) ? parent.getFingerprint() : 0;
+        this(priv, publicAsPoint.compress(), parent == null ? 0 : parent.depth + 1, parent,
+                parent != null ? parent.getFingerprint() : 0, chainCode, HDPath.M(childNumberPath), null, null);
```
msgilligan, line 75: Since `priv` is nullable, `makeHDPath` 7bc87972523443ab910146b6622eb6acce767ae7 could be used here. Probably in a follow-on commit.

msgilligan (approved):
Looks good! Approved!

PR #3716, opened 2025-04-03
DeterministicKey: `withParent()`, `withoutParent()`

schildbach:
Those methods are in two separate commits, could be merged individually.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -149,7 +149,8 @@ public class DeterministicKey extends ECKey {
                 chainCode, HDPath.M(childNumberPath), null, null);
     }
 
-    /** Clones the key */
+    /** @deprecated use {@link #withParent(DeterministicKey)} */
+    @Deprecated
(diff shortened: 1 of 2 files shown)


schildbach:
TODO

- In withParent(), I think the depth need to be re-calculated.
- withoutParent() needs some JavaDoc work

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -149,12 +149,6 @@ public DeterministicKey(List<ChildNumber> childNumberPath,
                 chainCode, HDPath.M(childNumberPath), null, null);
     }
 
-    /** Clones the key */
-    public DeterministicKey(DeterministicKey keyToClone, DeterministicKey newParent) {
```
msgilligan, line 153: You don't want to `@Deprecate` this constructor?

msgilligan (approved):
Approved. (with or without adding deprecation to the removed constructor)

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
schildbach: Yes, thanks. It doesn't hurt.

schildbach:
Merged.

PR #3717, opened 2025-04-04
DeterministicKey: rewrite `withoutPrivateKey()`

schildbach:
* use canonical constructor to clear both cleartext and encrypted private keys and the keyCrypter
* be more strict about returning same instance
* amend the JavaDoc
    
Also rename the method from `dropPrivateBytes()` and keep the old method around as deprecated.

~~This a child of #3715.~~


msgilligan (approved):
Approved.

schildbach:
Merged.

PR #3718, opened 2025-04-04
ChildKeyDerivationTest: amend testSerializationMainAndTestNetworks() to round-trip serialization

schildbach:
~~This is a child of #3715.~~

### DIFF
--- a/core/src/test/java/org/bitcoinj/crypto/ChildKeyDerivationTest.java
+++ b/core/src/test/java/org/bitcoinj/crypto/ChildKeyDerivationTest.java
@@ -224,16 +224,25 @@ public class ChildKeyDerivationTest {
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


msgilligan (commented):
This looks good, but I think it should be rebased because some of these change are duplicated from already approved PRs.

schildbach:
Rebased. The thing is the test will only surface the bug (fixed in the second commit) once we've fully fixed equals() / hashCode(). We should continue to work towards that.

msgilligan (approved):
Approved.

msgilligan:
The bug fixed in the second commit is also fixed in PR #3764 in this commit e3a7a26d2a23e80e42319aa6082949ebbede7af3 along with several similar bugs.

msgilligan (changes requested):
This should wait until after PR #3764 is merged and this is rebased.

PR #3720, opened 2025-04-04
HDPath: Deprecated public constructor, add `Prefix.M` and `Prefix.m` shortcuts

msgilligan:
2 commits

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/HDPath.java
+++ b/core/src/main/java/org/bitcoinj/crypto/HDPath.java
@@ -88,7 +88,7 @@ public class HDPath extends AbstractList<ChildNumber> {
             .map(String::trim)
             .collect(Collectors.toList());
     private final boolean hasPrivateKey;
-    private final List<ChildNumber> unmodifiableList;
+    private final List<ChildNumber> childNumbers;
 
     /** Partial path with BIP44 purpose */
     public static final HDPath BIP44_PARENT = m(ChildNumber.PURPOSE_BIP44);
@@ -105,7 +105,7 @@ public class HDPath extends AbstractList<ChildNumber> {
      */
     public HDPath(boolean hasPrivateKey, List<ChildNumber> list) {
         this.hasPrivateKey = hasPrivateKey;
-        this.unmodifiableList = Collections.unmodifiableList(new ArrayList<>(Objects.requireNonNull(list)));
+        this.childNumbers = Collections.unmodifiableList(new ArrayList<>(Objects.requireNonNull(list)));
     }
 
     /**
@@ -263,7 +263,7 @@ public class HDPath extends AbstractList<ChildNumber> {
      * @return A new immutable path
      */
     public HDPath extend(ChildNumber child1, ChildNumber... children) {
-        List<ChildNumber> mutable = new ArrayList<>(this.unmodifiableList); // Mutable copy
+        List<ChildNumber> mutable = new ArrayList<>(this.childNumbers); // Mutable copy
         mutable.add(child1);
         mutable.addAll(Arrays.asList(children));
         return new HDPath(this.hasPrivateKey, mutable);
@@ -276,7 +276,7 @@ public class HDPath extends AbstractList<ChildNumber> {
      * @return A new immutable path
      */
     public HDPath extend(HDPath path2) {
-        List<ChildNumber> mutable = new ArrayList<>(this.unmodifiableList); // Mutable copy
+        List<ChildNumber> mutable = new ArrayList<>(this.childNumbers); // Mutable copy
         mutable.addAll(path2);
         return new HDPath(this.hasPrivateKey, mutable);
     }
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -107,7 +107,9 @@ public String toString() {
      *
      * @param hasPrivateKey Whether it is a path to a private key or not
      * @param list List of children in the path
+     * @deprecated Use {@link HDPath#of(Prefix, List)} or another static constructor
      */
+    @Deprecated
```
schildbach, line 112: Shouldn't we make the other constructor the canonical one, before deprecating this one? Otherwise, this constructor will still be used by us – despite the deprecation.

schildbach: There are even more usages, imho we should all migrate them in the same commit.

msgilligan: It's deprecated for being made private, not for removal.

(However, we may have to remove it without deprecation if we implement HDFullPath and HDPartialPath subclasses)

schildbach (changes requested):
I have added some comments.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
schildbach: So this will stay the canonical constructor?

msgilligan: Probably, because internally we are using a primitive `boolean` for memory reasons.

schildbach: Ok. Still I think the references from tests should be migrated to some public non-deprecated constructor, if that's possible. Otherweise we can't make it private anyway, would need to change it to package-private.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -55,6 +55,11 @@ public enum Prefix {
 
         private final char symbol;
 
+        /** Convenience shortcut for {@link Prefix#PUBLIC} */
+        public static Prefix M = Prefix.PUBLIC;
+        /** Convenience shortcut for {@link Prefix#PRIVATE} */
+        public static Prefix m = Prefix.PRIVATE;
```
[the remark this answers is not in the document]
msgilligan: > Unless we have a very compelling reason, like a fluent, almost DSL-like notation within our tests (for example)

That's what I'm going for. I think all Bitcoin developers working with paths will recognize the difference between 'M' and 'm' in this context.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: I thought I had done that, will double check.

msgilligan:
Making _DRAFT_ for now.

The `Prefix.M` / `Prefix.m` addition is controversial and not needed for short-term, so we can shelve that for now.

The deprecated constructor may have to be eliminated entirely do to the addition of full-path and partial-path subclasses.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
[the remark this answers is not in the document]
schildbach: Could you give an example how the end result would look like? I was thinking about something like:

```
HDPath path = m / 0 / 1' / 0;
```

But compiler-safe, so no usage of `parse("m/…")`.

In this example, the M vs m would be clearly visible. But as soon as it's inbetween Java boilerplate, imho the difference is easily overlooked.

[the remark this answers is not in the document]
msgilligan: How about this:
```
return path.prefix() == M ?
                  ? "public"
                  : "private"
```    
or:

```
var prefix = M;
var path = HDPath.of(prefix);
```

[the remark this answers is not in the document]
msgilligan: But I do think we should defer this discussion until later, when we have finished some other refactoring. I don't think this change is that important at the moment. Because we could add `String`-parameter version of the `M()` and `m()` static constructors and have:

```
var path = M("/0H/0H);
```

msgilligan:
@schildbach I accidentally push a bunch more commits on this branch/PR. I'll sort it out tomorrow.

core/src/main/java/org/bitcoinj/crypto/HDPath.java
msgilligan: > So this will stay the canonical constructor?

Yes, it stays the canonical constructor in PR #3721, but it is made private. (which means unless we change something there will be no deprecation)

msgilligan: >> I think the references from tests should be migrated to some public non-deprecated constructor

> I thought I had done that, will double check.

I double-checked and I don't see any.

I'm going to "Resolve this conversation", but we can continue it on PR #3722 if you like.

msgilligan:
Ok, this is rebased and only has a single commit now. I changed the title as well.

The proposed addition is not very important and is also controversial, so I'll leave this as draft.

PR #3721, opened 2025-04-05
WIP/YOLO: Use HDPath and subclasses everywhere

msgilligan:

PR #3722, opened 2025-04-05
HDPath: Deprecate (the only) public constructor

msgilligan:
Replace with m() and M() factory methods in some tests.

This is cherry-picked from PR #3720 as a standalone. I want to cherry-pick some other commits (that were accidentally added to it) and then rebase PR #3721.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/HDPath.java
+++ b/core/src/main/java/org/bitcoinj/crypto/HDPath.java
@@ -102,7 +102,9 @@ public class HDPath extends AbstractList<ChildNumber> {
      *
      * @param hasPrivateKey Whether it is a path to a private key or not
      * @param list List of children in the path
+     * @deprecated Use {@link HDPath#of(Prefix, List)} or another static constructor
      */
+    @Deprecated
     public HDPath(boolean hasPrivateKey, List<ChildNumber> list) {
         this.hasPrivateKey = hasPrivateKey;
         this.childNumbers = Collections.unmodifiableList(new ArrayList<>(Objects.requireNonNull(list)));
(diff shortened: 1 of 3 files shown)


msgilligan:
Note the following _resolved_ conversation from the previous PR: https://github.com/bitcoinj/bitcoinj/pull/3720#discussion_r2029042716

schildbach:
> Copilot reviewed 3 out of 3 changed files in this pull request and generated no comments.
> Comments suppressed due to low confidence (1)
> 
> **core/src/main/java/org/bitcoinj/crypto/HDPath.java:105**
> 
>     * [nitpick] The deprecation javadoc suggests using HDPath.of(Prefix, List) while tests use the m() and M() factory methods. Consider updating the documentation to reference the actual factory methods in use for consistency.
> 
> 
> ```
> * @deprecated Use {@link HDPath#of(Prefix, List)} or another static constructor
> ```

Not sure which tests you are talking about. There are some references from static constructors, but tests have been migrated to use non-deprecated constructors.

schildbach (approved):
Looks good!

The remaining references to the deprecated constructor cannot be avoided I guess, so we have to live with a few additional deprecation warnings until the next cycle.

msgilligan:
> The remaining references to the deprecated constructor cannot be avoided I guess, so we have to live with a few additional deprecation warnings until the next cycle.

Unfortunately, I think we're going to have to remove this constructor without deprecation in this release. But I doubt many **bitcoinj** applications are constructing HDPaths from (`boolean`, `List<ChildNumber>`)

msgilligan:
> > Comments suppressed due to low confidence (1)
> >  tests use the m() and M() factory methods. 
> 
> Not sure which tests [Copilot is] talking about. There are some references from static constructors, but tests have been migrated to use non-deprecated constructors.

Note that @Copilot said "low confidence". I think what Copilot did was notice that to replace the deprecated constructor, this PR migrated to the `.M()` and `.m()` constructors but that the JavaDoc didn't explicitly list those constructors. I think the JavaDoc is fine as-is because there are a large number of static constructors and the only one worth specifically mentioning is the one that is a direct replacement for the deprecated constructor.

schildbach:
Merged.

PR #3723, opened 2025-04-05
HDPath: make abstract, add HDFullPath and HDPartialPath subclasses

msgilligan:
This is a child of PR #3722 and adds two new commits:

* 0566b9c3280cb10ab285ac266c02d36b4008ee8b - make abstract and add `HDFullPath` subclass
* 12dba13dbc6cdd1aae42a1281eedced19ca52e93 - add `HDPartialPath` subclass

The first of the two is not much use without the second, but it's useful to break the refactoring into two steps.

Before we merge this we should have a discussion about how we want to name the subclasses:

`HDPath.HDFullPath` and `HDPath.HDPartialPath` is pretty verbose. We could shorten it to `HDPath.Full` and `HDPath.Partial` or something similar. Or we could move `HDFullPath` and HDPartialPath` into standalone files so they are no longer nested classes with nested names.

Note that the idea of having `HDFullPathPublic` and `HDFullPathPrivate` subclasses of an `HDFullPath` (or similar) was considered and rejected in favor of continuing to use an internal boolean to distinguish.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/HDPath.java
+++ b/core/src/main/java/org/bitcoinj/crypto/HDPath.java
@@ -102,7 +102,9 @@ public class HDPath extends AbstractList<ChildNumber> {
      *
      * @param hasPrivateKey Whether it is a path to a private key or not
      * @param list List of children in the path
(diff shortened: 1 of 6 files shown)


schildbach:
Merged.

PR #3725, opened 2025-04-05
DeterministicKey: fix depth issue with `dropParent()`

schildbach:
This method should not change anything but the parent. The canonical constructor is used to assure that.

Also adds asserts to `ChildKeyDerivationTest.pubOnlyDerivation()` to surface the issue.


schildbach:
Merged.

PR #3726, opened 2025-04-05
HDPath: add isEmpty()

msgilligan:
This is a step towards not being a `List` implementation.


schildbach:
Why is isEmpty() a step towards not being a `List` implementation? I mean, List has `isEmpty()`, so this seems like a step towards, not away from.

schildbach:
("away from" would be a better wording than "step towards not", I'd say.)

msgilligan:
> Why is isEmpty() a step towards not being a `List` implementation? I mean, List has `isEmpty()`, so this seems like a step towards, not away from.

Because it's a useful method to have. When we no longer extend `AbstractList` we will keep this method but remove the `@Override` annotation.

schildbach:
Merged.

PR #3727, opened 2025-04-05
HDPath: fix toString for HDPartialPath, refactor and improve tests

msgilligan:
4 commits to cleanup tests, fix `HDPath.toString()` for `HDPartialPath`, and test the fix.

... and a 5th commit to test `parsePath` with the `toStringTestVectors`, too.

PR #3728, opened 2025-04-05
HDPath: parsePath fix parse of partial paths

msgilligan:
~~Child of PR #3727~~

### DIFF
--- a/core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
+++ b/core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
@@ -205,8 +205,13 @@ public class HDPathTest {
     @Test
     @Parameters(method = "toStringTestVectors, parseTestVectors")
     public void testParsePath(PathVector tv) {
-        HDPath.HDFullPath path = HDPath.parsePath(tv.pathString);
+        HDPath path = HDPath.parsePath(tv.pathString);
         assertEquals(tv.path, path);
+        // TODO: This is a reminder to FULLY implement (including hasPrivateKey) equals and hashCode on HDPath and subclasses
+        // Until equals and hashCode are FULLY implemented let's test `hasPrivateKey()` manually
+        if (path instanceof HDPath.HDFullPath) {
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
```diff
@@ -203,8 +203,13 @@ private PathVector[] parseTestVectors() {
     @Test
     @Parameters(method = "toStringTestVectors, parseTestVectors")
     public void testParsePath(PathVector tv) {
-        HDPath.HDFullPath path = HDPath.parsePath(tv.pathString);
+        HDPath path = HDPath.parsePath(tv.pathString);
         assertEquals(tv.path, path);
+        // TODO: This is a reminder to implement equals and hashCode on HDPath and subclasses
```
schildbach, line 208: Can we merge equals/hashCode first, then?

msgilligan: Yes, we did.

schildbach: Not the full version.

msgilligan: I think we have to implement (almost) all of the change in PR #3721 before we can start comparing `hasPrivateKey` in `equals`/`hashCode` -- the code and tests are full of assumptions left over from when we were using `List<ChildNumber>` everywhere. We have to declare stricter types and make sure we convert between them before doing a full comparison.

msgilligan: I'm also happy to remove the comment and "kludge" if you will merge this PR without them. We don't really need the reminder we have pending PRs with the fix.

schildbach:
I merged, with the kludge and a slightly shorter TODO comment.

Sorry for the delay. I was trying to understand if we can fix whatever is needed for proper equals()/hashCode() first, then add equals()/hashCode(), then merge this PR without kludge/comment.

PR #3730, opened 2025-04-06
HDPath: add asPartial()

msgilligan:
Child of PR #3728


schildbach:
Merged, out of sequence.

PR #3736, opened 2025-04-06
AesKey: remove deprecations and obsolete JavaDoc

msgilligan:
2 commits: 1 for the removed methods, 1 for the JavaDoc changes.

PR #3737, opened 2025-04-07
ASEKey: implement Java SecretKey interface

msgilligan:
This is a child of PR #3736


schildbach (commented):
Interesting. What ability do we gain with this?

msgilligan:
> Interesting. What ability do we gain with this?

I'm not exactly sure. Possible interoperability with Java Cryptography APIs that use AES? There may also be some Bouncy Castle methods that interoperate with the built-in Java Cryptography interface.

I certainly don't feel this is urgent. It's just applying some of the things I've learned about Java Cryptography while working on the secp256k1 API.

PR #3735, opened 2025-04-06
WIP: ECKey, LazyECPoint implement ECPublicKey

msgilligan:
This needs review and tests to verify the points and encoding are correct for Java Crypto.

PR #3738, opened 2025-04-07
ECKey: add getter for Secp256k1PrivKey interface 

msgilligan:
This is a child of PR #3735


msgilligan:
This was an exploration that went in a bad direction. The current plan is to entirely replace `LazyECPoint` when we migrate to `secp256k1-jdk`.

PR #3739, opened 2025-04-07
LazyECPoint: Deprecate all unused delegated methods

msgilligan:
We are migrating away from using (even internally) unwrapped Bouncy Castle APIs and towards an abstraction that will allow either Bouncy Castle or secp256k1 (C library) and possibly others in the future.


msgilligan:
See related PR #3754

schildbach (approved):
Looks good.

Did you also consider to immediately remove the methods? It's that sort of API that probably should never have been public, and I suspect/hope that nobody used it anyway. But I may be wrong on that.

msgilligan:
> Did you also consider to immediately remove the methods? It's that sort of API that probably should never have been public, and I suspect/hope that nobody used it anyway. But I may be wrong on that.

I was hoping you were gonna say that! I have a follow-on PR that does exactly that: https://github.com/bitcoinj/bitcoinj/pull/3740

PR #3740, opened 2025-04-07
LazyECPoint: remove deprecated (unused delegated) methods

msgilligan:
This is a child of PR #3739.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
+++ b/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
@@ -18,11 +18,9 @@ package org.bitcoinj.crypto;
 
 import org.bitcoinj.crypto.secp.Secp256k1Constants;
 import org.bouncycastle.math.ec.ECCurve;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
```diff
@@ -96,119 +94,19 @@ public byte[] getEncoded() {
...
-        return get().twice();
-    }
-
-    public ECFieldElement getAffineYCoord() {
-        return get().getAffineYCoord();
-    }
-
-    public ECFieldElement getAffineXCoord() {
-        return get().getAffineXCoord();
-    }
```
schildbach, line 206: These two methods are used in the new test (#3754).

schildbach (changes requested):
Can you rebase this, then fix the test (see comment).

Other than that, pre-approved.

msgilligan:
> Can you rebase this, then fix the test (see comment).

Rebased (and still broken.) I'll make a separate PR to change the test so it does not require those two methods.

schildbach:
Merged.

PR #3741, opened 2025-04-07
Experiment: make LazyECPoint EAGER

msgilligan:
This is just an experiment. It doesn't seem to affect the duration of unit tests, but we might want to do some more focused testing to see what the performance impact might be on lower-powered Android devices.


msgilligan:
The first two commits of this PR were merged long ago. The third commit (which actually makes `LazyECPoint` eager) is now provided by PR #4278 which supersedes this PR.

PR #3742, opened 2025-04-07
HDPath.HDPartialPath: add asFull()/asPublic()/asPrivate()

msgilligan:
To encourage use of stronger typing, these methods are only being added to HDPartialPath.  (This could change if we really need it.)


schildbach:
Merged.

PR #3743, opened 2025-04-07
DeterministicHierarchy: Use HDPartialPath as type of keys map

msgilligan:
Child of PR #3742 (but doesn't strictly need to be)


schildbach:
Merged.

PR #3745, opened 2025-04-07
DeterministicKey: Use HDPartialPath for internal childNumberPath

msgilligan:
This is a child of PR #3743 (though it need not be)

The idea here is somewhat inspired by the concept of _database normalization_. If we know from other state variables of the `DeterministicKey` whether it has a public or private path, we don't need to replicate that state in an internal member field.

I think this is a clear win, _unless_ there is a corner case I'm not thinking of (e.g. the case where a key does not know the full path from the root, or where it doesn't know whether it's parents have the private key available, etc.)

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -58,7 +58,7 @@ public class DeterministicKey extends ECKey {
 
     @Nullable
     private final DeterministicKey parent;
-    private final HDPath childNumberPath;
+    private final HDPath.HDPartialPath childNumberPath;
     private final int depth;
     private final int parentFingerprint; // 0 if this key is root node of key hierarchy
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -186,18 +186,26 @@ private DeterministicKey(@Nullable BigInteger priv, LazyECPoint pub, int depth,
         this.parent = parent;
         this.parentFingerprint = ascertainParentFingerprint(parent, parentFingerprint);
         this.chainCode = Arrays.copyOf(chainCode, chainCode.length);
-        this.childNumberPath = Objects.requireNonNull(hdPath);
+        this.childNumberPath = Objects.requireNonNull(hdPath).asPartial();
```
schildbach, line 189: I'm not sure if in these cases we should keep the `requireNonNull()` – a NPE will be thrown anyway. But it certainly doesn't hurt either.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -186,18 +186,26 @@ private DeterministicKey(@Nullable BigInteger priv, LazyECPoint pub, int depth,
...
-     * Returns the path through some {@link DeterministicHierarchy} which reaches this keys position in the tree.
-     * A path can be written as 0/1/0 which means the first child of the root, the second child of that node, then
+     * Returns the {@link HDPath.HDFullPath} through the {@link DeterministicHierarchy} to this key's position in the tree.
+     * A path can be written as {@code M/0/1/0} which means the first child of the root, the second child of that node, then
      * the first child of that node.
+     * @return A full path starting with {@code 'm'} or {@code 'M'} depending upon whether ths private key is available.
      */
-    public HDPath getPath() {
-        return childNumberPath;
+    public HDPath.HDFullPath getPath() {
```
schildbach, line 200: We might want to rename this getter to `path()` in a future PR.

schildbach:
All merged.

PR #3750, opened 2025-04-09
HDPath: implement equals() and hashCode() after fixing tests

msgilligan:
This branch begins with cherry-picks of:

1. PR #3748 
2. PR #3745
3. PR #3747
4. PR #3749
5. PR #3751

**Update:** The above PRs have all been merged. There are 4 commits remaining:

3 commits to fix tests by being more precise about HDPath types.

The final commit adds the full `equals()` / `hasCode()` and passes all tests!

PR #3752, opened 2025-04-09
HDPath: Stop implementing List

msgilligan:
This is a child of PR #3750


msgilligan:
This PR is superseded by PR #3764. Once that one is merged we should rebase and see if there is anything left here that we still want.

PR #3753, opened 2025-04-09
HDPath: use int[] internally

msgilligan:
~~This is a child of PR #3752~~ 

~~This should be rebased on PR #3764 once it is (hopefully) merged.~~

This consists of two commits: the first converts to `int[]` internally but with some intermediate usage of `List`. The second removes those intermediate `List` conversions to improve performance.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/HDPath.java
+++ b/core/src/main/java/org/bitcoinj/crypto/HDPath.java
@@ -20,8 +20,6 @@ import org.bitcoinj.base.internal.StreamUtils;
 import org.bitcoinj.base.internal.InternalUtils;
 
 import javax.annotation.Nonnull;
-import java.util.AbstractList;
-import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/crypto/HDPath.java
```diff
@@ -87,7 +85,9 @@ public String toString() {
     private static final InternalUtils.Splitter SEPARATOR_SPLITTER = s -> Stream.of(s.split(SEPARATOR))
             .map(String::trim)
             .collect(Collectors.toList());
-    protected final List<ChildNumber> childNumbers;
+
+    //private final List<ChildNumber> childNumbers;
```
msgilligan, line 89: This commented out line should be removed.

schildbach: I agree.

PR #3754, opened 2025-04-10
LazyECPoint: implement java.s.i.ECPublicKey

msgilligan:
Implement `java.security.interfaces.ECPublicKey` to help us migrate away from leaking Bouncy Castle APIs into our API.

It needs some constant definitions for the SECP256K1 curve. These are provided by a new, temporary sub-package `secp` that should eventually be replaced by definitions in the secp256k1-jdk API.

This PR/commit is IMO the minimal, first step to begin migration to Java Cryptography types and secp256k1-jdk.

Note that eventually `LazyECPoint` (or its possible renamed version or successor) will want to implement a Secp256K1-specific `ECPublicKey` subtype. But for now, I think it is simpler to just use the built-in Java type. We can narrow it later.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
+++ b/core/src/main/java/org/bitcoinj/crypto/LazyECPoint.java
@@ -16,12 +16,14 @@
 
 package org.bitcoinj.crypto;
 
+import org.bitcoinj.crypto.secp.Secp256k1Constants;
 import org.bouncycastle.math.ec.ECCurve;
 import org.bouncycastle.math.ec.ECFieldElement;
 import org.bouncycastle.math.ec.ECPoint;
 
 import javax.annotation.Nullable;
 import java.math.BigInteger;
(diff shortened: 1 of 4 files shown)


msgilligan:
@schildbach I think this is both important and related to the other refactoring we are doing. The `ECKey`/`DeterministicKey` constructor hierarchy is currently using `LazyECPoint` in many places. So we need to at least start thinking about other types we will use for representing keys.

This is related to PR #3739 that deprecates (unused) Bouncy Castle methods in `LazyECPoint`.

msgilligan:
I pushed a simple test on top.

msgilligan:
> why is infinity broken, and what's needed to fix it?

BC and JCA use slightly different formats for storing infinity and since infinity is a special case, it requires a special case in the conversion. I added the special case, reactivated the test, and force-pushed.

schildbach:
Merged.

PR #3756, opened 2025-04-10
Experiment: ECKey store `compressed`

msgilligan:
This experimental branch/PR moves the `compressed` `boolean` variable from `LazyECPoint` to `ECKey`.

`compressed` is serialized with/from `ECKey` and is used in the `ECKey` constructors. It will simplify the code (including the constructors) if `compressed` is made a direct property of `ECKey` and `LazyECPoint` does not keep this state, but has the ability to serialize in either compressed or uncompressed format.

This PR is a (currently) child of the following (in order):

* https://github.com/bitcoinj/bitcoinj/pull/3739
* https://github.com/bitcoinj/bitcoinj/pull/3740
* https://github.com/bitcoinj/bitcoinj/pull/3741

Not all of these changes are required to get to the result of `ECKey` storing `compressed` directly, but most of them are. I mention this because we have not yet measured the performance impact of removing the laziness of `LazyECPoint`, so we probably will want to skip that commit as we work to apply these changes one-by-one.


msgilligan:
This should be rebased once PR #4278 is merged.

PR #3758, opened 2025-04-12
Wallettool: subcommand methods

ItoroD:
This PR is an attempt to show the usage of subcommands in wallettool. I have been able to migrate `dump, create, and current-receive-addr` to subcommands. Currently we can run the commands like this:

`wallettool --wallet=<myWalletPath> --net=<myNetwork> --dump-privkeys dump`

My next step is to be able to pass specific options to the subcommand that need those options. Those options should no longer be global but specific to the particular subcommand. So that we can have something like. 

`wallettool --wallet=<myWalletPath> --net=<myNetwork> dump  --dump-privkeys`

This can easily be done. But before I do I want to take my time to note what options should be global and what options should be local. (by global I mean after main command `wallettool` and local after subcommand `dump, create...`

### DIFF
--- a/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
+++ b/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
@@ -66,6 +66,9 @@ import org.bitcoinj.wallet.WalletProtobufSerializer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import picocli.CommandLine;
+import picocli.CommandLine.Spec;
+import picocli.CommandLine.Model.CommandSpec;
+import picocli.CommandLine.ParseResult;
 
 import javax.annotation.Nullable;
 import java.io.File;
@@ -99,6 +102,10 @@ import static org.bitcoinj.base.Coin.parseCoin;
  */
 @CommandLine.Command(name = "wallet-tool", usageHelpAutoWidth = true, sortOptions = false, description = "Print and manipulate wallets.")
 public class WalletTool implements Callable<Integer> {
+
+    @Spec
+    CommandSpec spec;
+
     @CommandLine.Parameters(index = "0", description = "Action to perform. Valid values:%n" +
             "  dump                 Loads and prints the given wallet in textual form to stdout. Private keys and seed are only printed if --dump-privkeys is specified. If the wallet is encrypted, also specify the --password option to dump the private keys and seed.%n" +
             "                       If --dump-lookahead is present, also show pregenerated but not yet issued keys.%n" +
@@ -316,7 +323,10 @@ public class WalletTool implements Callable<Integer> {
     }
 
     public static void main(String[] args) {
-        int exitCode = new CommandLine(new WalletTool()).execute(args);
+        WalletTool walletTool = new WalletTool();
+        int exitCode = new CommandLine(walletTool)
+                .setExecutionStrategy(walletTool::executionStrategy)
+                .execute(args);
         System.exit(exitCode);
     }
 
@@ -329,6 +339,17 @@ public class WalletTool implements Callable<Integer> {
             System.err.println("Could not understand action name " + actionStr);
             return 1;
         }
+        return 0;
+    }
+
+    private int executionStrategy(ParseResult parseResult) {
+        init(); // Custom initialization logic before subcommand execution
+        int commandResult = new CommandLine.RunLast().execute(parseResult);
+        return handleCleanup(commandResult);
+    }
+
+    private void init() {
+        CommandSpec currentSpec = spec;
 
         if (debugLog) {
             BriefLogFormatter.init();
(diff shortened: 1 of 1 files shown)


schildbach:
I looked at the first two commits in this series. I think we can merge the first 776a498389139a30cfb4c45f56944cdd42903bf3 right away, if so desired.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -453,27 +453,28 @@ private static ByteString bytesToHex(ByteString bytes) {
         return ByteString.copyFrom(ByteUtils.formatHex(bytes.toByteArray()).getBytes());
     }
 
-    private void upgrade() {
+    private Integer upgrade() {
```
schildbach, line 409: Could this also be an `int`?

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -348,77 +369,65 @@ public Integer call() throws IOException, BlockStoreException {
...
-
-        if (action == ActionEnum.RAW_DUMP) {
-            // Just parse the protobuf and print, then bail out. Don't try and do a real deserialization. This is
-            // useful mostly for investigating corrupted wallets.
-            return rawDumpWallet();
+        if(!"create".equals(currentSpec.name())){
+            if (!walletFile.exists()) {
+                System.err.println("Specified wallet file " + walletFile + " does not exist. Try wallet-tool --wallet=" + walletFile + " create");
+                System.exit(1);
+            }
```
schildbach, line 320: Hmm, I wonder if we could keep this init code closer to the actual sub-action(s)? Especially if those at some point would live in their own classes each, it would be nice to have the init (and cleanup) code there as well.

schildbach: For init code, that is shared between several actions, we could use shared init methods for just a single pupose. For example, the block I've marked above we could put into a dedicated `initWallet()` and explicitly call it from each action? In the end, I think we would not need neither if-ladder nor an execution strategy. What do you think?

ItoroD: > Hmm, I wonder if we could keep this init code closer to the actual sub-action(s)? Especially if those at some point would live in their own classes each, it would be nice to have the init (and cleanup) code there as well.

yes I can move it closer. About moving sub-actions to their own classes, I will PR a design doc soon for subcommand transition as @msgilligan encouraged. There are 2 main ways for sub commands. first is as methods and second is as individual classes. This PR shows as methods. Might be worth showing in another PR how it will look as classes. What do you think ?

ItoroD:
> I looked at the first two commits in this series. I think we can merge the first [776a498](https://github.com/bitcoinj/bitcoinj/commit/776a498389139a30cfb4c45f56944cdd42903bf3) right away, if so desired.

yes please

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
ItoroD: Yes. Why only this one ?

msgilligan: I'm sure @schildbach means all of them. He usually does!

msgilligan: Screenshot 2025-04-14 at 11 45 41 AM

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -1098,6 +1045,66 @@ private void dumpWallet() throws BlockStoreException {
...
+                coinSelector = null;
+            }
+            send(coinSelector, outputsStr, feePerVkb, lockTimeStr, allowUnconfirmed);
+        } else {
+            System.err.println("You must specify at least one --output=addr:value.");
+            return 1;
+        }
+        return 0;
+    }
+
```
msgilligan, line 1077: Is it possible to make this method `static`? Or are there just too many "globals"?

Eventually we want to split this into two methods, one that parses the command options and generates any error messages (and maybe pico-cli can do all that work for us, maybe not) and another that actually executes the command.

Just moving this code to a method makes things a little bit easier to read, but you still have to pay attention to where all the variables are coming from. But, if there's no easy way to make this static and/or pass variables as parameters, I'm ok with this as a first step.

msgilligan (changes requested):

msgilligan (changes requested):
Definitely make `rawDumpWallet()` `static`. And see if you can make it work with `send`...

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
msgilligan: > > Hmm, I wonder if we could keep this init code closer to the actual sub-action(s)? Especially if those at some point would live in their own classes each, it would be nice to have the init (and cleanup) code there as well.
> 
>I will PR a design doc soon for subcommand transition as @msgilligan encouraged. 

The document could be as simple as a list of all the commands with their global and local options and maybe a note or paragraph on each one.

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -1098,6 +1045,66 @@ private void dumpWallet() throws BlockStoreException {
...
     }
 
+    private Integer rawDumpWallet() throws IOException {
+        try (FileInputStream stream = new FileInputStream(walletFile)) {
+            Protos.Wallet proto = WalletProtobufSerializer.parseToProto(stream);
+            proto = attemptHexConversion(proto);
+            System.out.println(proto.toString());
+            return 0;
+        }
+    }
```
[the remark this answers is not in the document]
ItoroD: The problem with doing this is that whatever you pass as param to a subcommand method like `rawDumpWallet()` it will be required as either command-line option or param that you might include in your command call in cli (if you tag args as @option its not required of course). Passing wallet file makes sense but just know we have now made it a local option for raw-dump command by doing this. I suggest for now we leave it as is until I share subcommand docs for each command

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
ItoroD: pico-cli can handle the parsing errors. 

About passing variables I think [this ](https://github.com/bitcoinj/bitcoinj/pull/3758/files/776a498389139a30cfb4c45f56944cdd42903bf3#r2044288884)applies here too

ItoroD:
> I looked at the first two commits in this series. I think we can merge the first [776a498](https://github.com/bitcoinj/bitcoinj/commit/776a498389139a30cfb4c45f56944cdd42903bf3) right away, if so desired.

I have done seperate PR [here](https://github.com/bitcoinj/bitcoinj/pull/3774) for the initial comment

PR #3762, opened 2025-04-13
DeterministicKeyChain: store accountPath as HDPartialPath, add accountFullPath(), fix and improve tests

msgilligan:
This is a series of 5 related commits that: fix several _incorrect_ tests, fix a few bugs, improve tests to use full path where they were only using partial paths, and prepare for deprecation of `getAccountPath()` for replacement by `accountFullPath()`

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
@@ -128,7 +128,7 @@ public class DeterministicKeyChain implements EncryptableKeyChain {
     @Nullable private DeterministicKey rootKey;
     @Nullable private final DeterministicSeed seed;
     private final ScriptType outputScriptType;
-    private final HDPath accountPath;
+    private final HDPath.HDPartialPath accountPath;
 
     // Paths through the key tree. External keys are ones that are communicated to other parties. Internal keys are
     // keys created for change addresses, coinbases, mixing, etc - anything that isn't communicated. The distinction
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -450,10 +450,16 @@ protected DeterministicKeyChain(KeyCrypter crypter, AesKey aesKey, Deterministic
...
+    public HDPath.HDPartialPath getAccountPath() {
         return accountPath;
     }
 
+    public HDPath.HDFullPath accountFullPath() {
+        boolean hasPrivateKey = !getWatchingKey().isWatching();
+        return accountPath.asFull(hasPrivateKey ? HDPath.Prefix.PRIVATE : HDPath.Prefix.PUBLIC);
+    }
+
+
```
schildbach, line 462: I think this could be extracted from the first commit and merged with the usages in tests (basically the other 4 commits).

There is a surplus newline here.

schildbach (changes requested):
The first commit seems to be trying to do too much at the same time. See my comment for a suggestion on what to extract and merge ahead, and then we'll see what's left.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
msgilligan: Resolved in new PR #3765

msgilligan:
This PR is replaced by:

* PR #3765 
* PR #3766 

and should probably be closed.

msgilligan:
Closing in favor of PRs #3765 and #3566

PR #3764, opened 2025-04-13
HDPath: Do not inherit from List

msgilligan:
~~This is a child of PR #3763~~

~~The first commit is a duplicate of PR #3770, so this should be rebased after that is merged.~~

There are two preparatory commits to remove the last usages of `List<ChildNumber>` and dependencies on `HDPath` implementing `List`. The third commit simply removes the `extend AbstractList<ChildNumber>` and removes `@Override` annotations on methods that exist in `AbstractList `.


schildbach (changes requested):
The remaining commit is a really nice one. If I understand right, it's the climax of many PRs of the last weeks.

However, I think the commit message (mainly first line) could be made to not simply repeat the removal of the `extends` (which you'd see in a diff anyway). I often used the verb "to divorce" in cases like this, but perhaps there are better wordings?

PR #3765, opened 2025-04-14
DeterministicKeyChain: add accountFullPath(), fix and improve tests

msgilligan:
This is a series of 4 related commits that: fix several incorrect tests, fix a few bugs, improve tests to use full path where they were only using partial paths.


schildbach:
Merged.

PR #3775, opened 2025-04-18
UTXO: Remove address property (BREAKING)

msgilligan:
UTXOProvider is only used by the FullPrunedBlockStore tests, does not support segwit, and there has been a TODO in this class since 2015 to remove the address property which is implemented as a String which can also be set to "".

It is also not easy to provide a deprecation because of circular dependencies on the current behavior.


msgilligan:
@schildbach rebased.

schildbach (approved):

PR #3776, opened 2025-04-18
Wallet: deprecate UTXOProvider support

msgilligan:
Deprecate getUTXOProvider/setUTXOProvider and protected methods that use the UTXOProvider methods.

We have plans to refactor and simplify the wallet interface and removing this mode of operation (which is rarely used outside of tests) will help us move in that direction.

Longer term, we hope to provide a more flexible wallet interface that will make it even easier to use other sources of UTXO information.


schildbach (approved):
In general, I'm fine with the removal.

However, I've got an use-case: emptying a paper wallet. In this case, UTXOs are requested (via Electrum API) and passed into this API. If all goes well, the resulting wallet should be usable, e.g. for emptying. In theory.

In practise, I currently do not use this API and instead fake an incoming transaction with the UTXOs. But I always wanted to migrate my code to this API now marked for deprecation. We should keep this in mind in future.

schildbach:
Rebased.

schildbach (approved):

PR #3777, opened 2025-04-18
TransactionOutPoint: static from() and of() constructors

msgilligan:
`of()` constructor for a standalone outpoint, `from()` constructor that is connected to what it was constructed from.

Deprecate the two constructors that take a connected "fromTx" or a connected output and replace them with factory methods.

Deprecate (index, hash) constructor. Create private (hash, index) constructor. Replace with .of(hash, index) factory method.

This will allow us to factor out an interface or abstract base class in the future. It also completes our migration to hash parameter before index parameter.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
@@ -75,6 +75,8 @@ public class TransactionOutPoint {
         return new TransactionOutPoint(index, hash);
     }
 
+    /** @deprecated use {@link TransactionOutPoint#from(Transaction, long)} */
+    @Deprecated
     public TransactionOutPoint(long index, Transaction fromTx) {
         this(fromTx.getTxId(), index, fromTx, null);
     }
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -70,21 +70,46 @@ public class TransactionOutPoint {
...
+
+    /**
+     * @deprecated Use {@link TransactionOutPoint#from(TransactionOutput)}
+     */
+    @Deprecated
     public TransactionOutPoint(TransactionOutput connectedOutput) {
         this(connectedOutput.getParentTransactionHash(), connectedOutput.getIndex(), null, connectedOutput);
     }
 
+    public static TransactionOutPoint of(Sha256Hash hash, long index) {
```
schildbach, line 101: This needs a JavaDoc. If this constructor is for unconnected outpoints (as you say in the commit description), I think we could pick a less generic method name, e.g. `unconnected()`?

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -70,21 +70,46 @@ public class TransactionOutPoint {
...
+        return new TransactionOutPoint(hash, index);
+    }
+
+    public static TransactionOutPoint from(Transaction fromTx, long index) {
```
schildbach, line 105: Needs JavaDoc.

schildbach (changes requested):

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
msgilligan: JavaDoc added and force-pushed.

Since a `TransactionOutPoint` is defined as (TxID, index) I think the generic method name `of()` is the best choice. I'd be more open to renaming the `from()` static constructors to `connected()`, but I think `from()` is a better choice as it is simple and IMO more clear about what is happening. (see the added JavaDoc.)

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -70,21 +70,46 @@ public class TransactionOutPoint {
...
+        return new TransactionOutPoint (fromTx.getTxId(), index, fromTx, null);
+    }
+
+    public static TransactionOutPoint from(TransactionOutput connectedOutput) {
```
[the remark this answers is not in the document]
msgilligan: Added and force-pushed.

msgilligan:
Note: The branch name I created for this PR has a confusing typo: I typed `TransactionOutput-` instead of `TransactionOutPoint-`.

schildbach:
Rebased.

schildbach (approved):

PR #3778, opened 2025-04-18
TransactionOutPoint: make fromTx field private

msgilligan:
Provide a @Nullable getter.


core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -142,6 +142,11 @@ public TransactionOutput getConnectedOutput() {
         return null;
     }
 
+    @Nullable
+    public Transaction getFromTx() {
```
schildbach, line 146: Should we add a JavaDoc here?

msgilligan: Added and force-pushed.

schildbach:
Rebased.

schildbach (approved):

PR #3779, opened 2025-04-18
TransactionOutPoint: Use getConnectedOutput() where possible 

msgilligan:
2 commits:

* The first one makes a minor simplification in `getConnectedOutput()`
* The second replaces direct access to the `.connectedOutput` field with the accessor, except in one test where the behavior change causes a test to report failure.


msgilligan:
@schildbach After rebasing this PR, I reviewed the changes and decided to be a little more thorough in updating `TransactionInputTest` for the change from `getConnectedOutput()` from `. connectedOutput`. See PR #3815.

msgilligan:
This PR is almost completely replaced by the following PRs:

* PR #3815
* PR #3816
* PR #3817
* PR #3818

and those PRs contain improvements over this one.

The only thing this PR contains that those don't are changes internal to `TransactionOutPoint` to use `getConnectedOutput()` in `connectTransaction() and  `disconnectTransaction()`.  I'm not sure that we want those changes exactly as they are coded in this PR.

After the recommended PRs (3815-3818) are merged,  we should review the behavior of `connectTransaction()` and  `disconnectTransaction()` (the latter is unused) and decide what to do.

msgilligan:
@schildbach This is rebased and only the changes to `connectTransaction()` and `disconnectTransaction()` remain.

I made a separate PR to deprecate `disconnectTransaction()`: PR #3826 
(We may also want to rename `disconnectOutput()` to simply `disconnect()`)

For `connectTransaction()` we should review the behavior to make sure it is correct and then probably update the JavaDoc. It currently works as follows:

1. Creates and returns a new `TransactionOutPoint` with the specified transaction in the internal `fromTx` member.
2. The `TransactionOutPoint` constructor will throw `IllegalArgumentException` if the existing `connectedOutput` is not `null` and does not match the output at the same index in the newly provided transaction.

schildbach:
In C and D you mean "at the same time", I guess?

msgilligan:
> In C and D you mean "at the same time", I guess?

Yeah, C & D should be in the same commit.

schildbach:
> > In C and D you mean "at the same time", I guess?
> 
> Yeah, C & D should be in the same commit.

That's a misunderstanding. What I mean is the following. Do you mean this:

C. Document with comments that the fromTx and connectedOutput members can't/won't both be set **at the same time**.
D. Add a check to the (private) canonical constructor that disallows both the fromTx and connectedOutput being set **at the same time**.

schildbach:
> But I prefer (D)

From my gut feeling, I prefer D as well. But I'm open to alternatives as well.

PR #3780, opened 2025-04-18
TransactionOutPoint: have disconnectOutput() clear `fromTx`

msgilligan:
To be disconnected the returned/modified TransactionOutPoint should have neither a `fromTx` or a `connectedOutput`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
@@ -218,11 +218,11 @@ public class TransactionOutPoint {
     }
 
     /**
-     * Returns a copy of this outpoint, but with the connectedOutput removed.
-     * @return outpoint with removed connectedOutput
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
```diff
@@ -213,7 +213,7 @@ public RedeemData getConnectedRedeemData(KeyBag keyBag) throws ScriptException {
      * @return outpoint with removed connectedOutput
```
schildbach, line 213: I think the JavaDoc needs a change in the description of this semantics, too.

schildbach (changes requested):

core/src/main/java/org/bitcoinj/core/TransactionOutPoint.java
msgilligan: Fixed and force-pushed.

schildbach:
Rebased.

schildbach (approved):

PR #3781, opened 2025-04-22
ScriptTests: Use Jackson type mapping to make scripts more readable

msgilligan:


core/src/test/java/org/bitcoinj/script/ScriptTest.java
```diff
@@ -494,4 +487,56 @@ public void getToAddress() {
...
+        public final String hash;
+        public final long index;
+        public final String script;
+
+        public ScriptPubKeyEntry(String hash, long index, String script) {
+            this.hash = hash;
+            this.index = index;
+            this.script = script;
+        }
+    }
```
schildbach, line 537: I think these should implement toString() so that the original JSON line is output when an error occurs (as before this change).

msgilligan: This would already be implemented if we had `record` 🌈🦄

I will add it.

msgilligan:
@schildbach The reason I made this change was I was looking at PR #1930 and its use of GSON (Google JSON Library) and thinking that we should continue to use Jackson for reading JSON in tests, which lead me to this test which uses Jackson.

core/src/test/java/org/bitcoinj/script/ScriptTest.java
msgilligan: Fixed and force-pushed!

msgilligan: > This would already be implemented if we had `record` 🌈🦄

Actually, that's not quite true because the built-in `record` `toString()` would not use JSON format.

schildbach:
Rebased.

schildbach (approved):

PR #3785, opened 2025-04-24
Block: rename to `asHeader()` from `cloneAsHeader()`

schildbach:
Keep the old method around as deprecated.

Also make `writeHeader()` private. Any non-private usages are converted to `asHeader().write()`.

### DIFF
--- a/core/src/test/java/org/bitcoinj/core/FullBlockTestGenerator.java
+++ b/core/src/test/java/org/bitcoinj/core/FullBlockTestGenerator.java
@@ -1183,7 +1183,7 @@ public class FullBlockTestGenerator {
             checkState(b64Original.block.messageSize() == Block.MAX_BLOCK_SIZE);
 
             ByteBuffer buf = ByteBuffer.allocate(b64Original.block.messageSize() + 8);
-            b64Original.block.writeHeader(buf);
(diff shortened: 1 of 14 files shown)


core/src/test/java/org/bitcoinj/core/FullBlockTestGenerator.java
```diff
@@ -1183,7 +1183,7 @@ public boolean add(Rule element) {
             checkState(b64Original.block.messageSize() == Block.MAX_BLOCK_SIZE);
 
             ByteBuffer buf = ByteBuffer.allocate(b64Original.block.messageSize() + 8);
-            b64Original.block.writeHeader(buf);
+            b64Original.block.asHeader().write(buf);
```
msgilligan, line 1186: Doesn't this change behavior because the copy written to is unreferenced?

msgilligan (commented):
I have one question. See comment.

core/src/test/java/org/bitcoinj/core/FullBlockTestGenerator.java
schildbach: The target of the write in both cases is `buf`.

PR #3783, opened 2025-04-24
More HDPath simplification

msgilligan:
12 commits of HDPath cleanup, simplification, optimization, and JavaDoc improvements.

PR #3786, opened 2025-04-24
Remove incorrect usages of HDFullPath in deterministic key implementations

msgilligan:
This is child of #3783 (but doesn't strictly need to be)


schildbach:
Rebased.

schildbach (approved):

PR #3787, opened 2025-04-24
Block: rename to `transactions()` from `getTransactions()`

schildbach:
The new method doesn't return null any more and throws instead.

The old method is kept around as deprecated.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -805,10 +805,25 @@ public class Block implements Message {
         this.hash = null;
     }
 
-    /** Returns an unmodifiable list of transactions held in this block, or null if this object represents just a header. */
+    /**
+     * Returns an unmodifiable view of the transactions held in this block. The transactions themselves can change.
+     * This can only be called on regular (non-header-only) blocks.
+     * <p>
+     * Before using this method, consider using one of: {@link #transactionCount()}, {@link #transaction(int)},
+     * {@link #forEachTransaction(Consumer)}, {@link #findTransactions(Predicate)}, {@link #addTransaction(Transaction)}
+     *
+     * @return transactions in this block, can be empty
+     */
+    public List<Transaction> transactions() {
+        checkState(!isHeaderOnly(), () -> "block is header-only");
+        return Collections.unmodifiableList(transactions);
+    }
+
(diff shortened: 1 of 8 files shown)


msgilligan:
We're getting close to the point where we can create a header-only-block subclass that won't even have a `.transactions()` method.

schildbach:
This is now ready for review.

I tried to replace as many calls to `transactions()` as possible with more specific accessors prior to this PR. If you see more potential for the remaining usages, feel free to say so and we'll sneak in a PR or two.

schildbach:
> We're getting close to the point where we can create a header-only-block subclass that won't even have a `.transactions()` method. We need to try to eliminate the remaining mutable fields if possible (and it not create a mutable-test-block class or something.)

Yes, I see my work as a preparation for such a change. But I also think it's maybe not strictly necessary to use inheritance for this, or at least there is no pressure to do so?

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -805,10 +805,22 @@ void setNonce(long nonce) {
...
-    /** Returns an unmodifiable list of transactions held in this block, or null if this object represents just a header. */
+    /**
+     * Returns an unmodifiable list of transactions held in this block. This can only be called on regular
+     * (non-header-only) blocks.
+     *
+     * @return transactions in this block, can be empty
+     */
+    public List<Transaction> transactions() {
+        checkState(!isHeaderOnly(), () -> "block is header-only");
+        return Collections.unmodifiableList(transactions);
```
msgilligan, line 816: Just a reminder that the underlying list is technically still modifiable as are the transaction contained within.

core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -805,10 +805,22 @@ void setNonce(long nonce) {
         this.hash = null;
     }
 
-    /** Returns an unmodifiable list of transactions held in this block, or null if this object represents just a header. */
+    /**
+     * Returns an unmodifiable list of transactions held in this block. This can only be called on regular
```
msgilligan, line 809: unmodifiable view of this block list of...

msgilligan (approved):
Approved. With a reminder about underlying mutable that maybe should be mentioned in the JavaDoc.

PR #3788, opened 2025-04-24
ParseByteCacheTest, FullBlockTestGenerator: make use of `Block.hasTransactions()`

schildbach:
This prevents accessing all transactions in these cases.


msgilligan (approved):
Approved.

PR #3789, opened 2025-04-24
Block: add method `transactionCount()` to determine number of transactions

schildbach:
Use it to prevent accessing `getTransactions()` when possible.


msgilligan (approved):
Approved.

PR #3791, opened 2025-04-25
BIP85 example code and example test class

jeffmcclure:
Sample code for BIP85 key generation with tests that validate the original BIP85 test vector and also match https://iancoleman.io/bip39/


examples/build.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
     implementation project(':bitcoinj-core')
     implementation 'info.picocli:picocli:4.7.6'
     implementation 'org.slf4j:slf4j-jdk14:2.0.16'
+    testImplementation 'junit:junit:4.13.2'
```
msgilligan, line 9: It's great that you wrote tests!

I think examples can/should use JUnit 5, as we have been preferring JUnit 5 for tests outside of base/core (which must build with Debian "FrankenGradle" 4.4.x)

examples/src/main/java/org/bitcoinj/examples/BIP85.java
```diff
@@ -0,0 +1,111 @@
...
+    private static byte[] generateBIP85Entropy(byte[] privateKey) throws Exception {
+        // BIP-85: HMAC-SHA512 with key "bip-entropy-from-k"
+        String hmacKey = "bip-entropy-from-k";
+        Mac mac = Mac.getInstance("HmacSHA512");
+        SecretKeySpec secretKeySpec = new SecretKeySpec(hmacKey.getBytes(StandardCharsets.UTF_8), "HmacSHA512");
+        mac.init(secretKeySpec);
+        return mac.doFinal(privateKey);
+    }
+
+}
```
msgilligan, line 111: Missing LF at EOF.

examples/src/test/java/org/bitcoinj/examples/BIP85Test.java
```diff
@@ -0,0 +1,84 @@
...
+        String xprv = "xprv9s21ZrQH143K2LBWUUQRFXhucrQqBpKdRRxNVq2zBqsx8HVqFk2uYo8kmbaLLHRdqtQpUm98uKfu3vca1LqdGhUtyoFnCNkfmXRyPXLjbKb";
+        Network network = MainNetParams.get().network();
+        DeterministicKey masterKey = DeterministicKey.deserializeB58(xprv, network);
+        assertEquals("627ef3a6", BIP85.fingerprint(masterKey));
+
+        String childMnemonic = BIP85.deriveBIP85Mnemonic(masterKey, 24, 0);
+        assertEquals("puppy ocean match cereal symbol another shed magic wrap hammer bulb intact gadget divorce twin tonight reason outdoor destroy simple truth cigar social volcano",
+                childMnemonic);
+    }
+
```
msgilligan, line 83: Extra blank line.

examples/build.gradle
msgilligan: Actually, after looking at the rest of this PR I see that this is not an example app with a `main` method, but more of a library class that provides BIP85 functionality. So now, I'm thinking we should consider putting it in the `bitcoinj-core` module.

msgilligan (changes requested):
This is great to see! I'm not that familiar with BIP-85 (so will do some more research as homework) but I think this should probably go in `bitcoinj-core` (probably in the `org.bitcoinj.crypto` package) and not in examples. I've requested some trivial changes, but (assuming @schildbach agrees with me) I think we should move the BIP85.java class and its tests into a package inside the `core` module.

I'm also thinking a descriptive name like `DeterministicEntropy` might be better than naming the class `BIP85`. Do you have any suggestions?

jeffmcclure:
> This is great to see! I'm not that familiar with BIP-85 (so will do some more research as homework) but I think this should probably go in `bitcoinj-core` (probably in the `org.bitcoinj.crypto` package) and not in examples. I've requested some trivial changes, but (assuming @schildbach agrees with me) I think we should move the BIP85.java class and its tests into a package inside the `core` module.
> 
> I'm also thinking a descriptive name like `DeterministicEntropy` might be better than naming the class `BIP85`. Do you have any suggestions?

Sounds good on the package move.  I am new to the project, and not super familiar with all the source, so I figured submitting a sample code would have the highest probability of being accepted.   I agree that a utility function in the core packages should be useful to others.   I'll incorporate all your feedback and submit an updated PR.

jeffmcclure:
@msgilligan Updated javadoc.

jeffmcclure:
@msgilligan anything I can do to help get his pr merged?

msgilligan:
I just tagged this as a candidate for the `0.18` release. (I thought I had already done that.)

Andreas has been away and has a large backlog of PRs (mostly from me) to review. I feel those should be prioritized because  they contain extensive changes to the core, some will require rebasing, and additional changes to the core are going to be hard to make without rebasing.

This PR is two new files, so is not going to need rebasing. I'm pretty much OK with it as it is and have been waiting for Andreas to review it before commenting further.

The only thing I would change is that I would name the class `DeterministicEntropy` and leave the BIP-85 reference for the comments.

So, I think there is a good chance it will be merged and included in 0.18, but I can't speak for Andreas.

It might also be nice to see an example or an external use case, but that isn't required and probably shouldn't be part of this PR.

jeffmcclure:
@msgilligan Class renamed to DeterministicEntropy.   Regarding external use case, the test class provides good examples of using the class.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,202 @@
...
+
+import static org.bitcoinj.crypto.internal.CryptoUtils.sha256hash160;
+
+/**
+ * Utility class for working with BIP-85, generating wallets, etc. (<a href="https://github.com/bitcoin/bips/blob/master/bip-0085.mediawiki">https://github.com/bitcoin/bips/blob/master/bip-0085.mediawiki</a>).
+ * <p>
+ * See test class org.bitcoinj.crypto.DeterministicEntropyTest for examples on how to use this class.
+ */
+public class DeterministicEntropy {
+    private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DeterministicEntropy.class);
```
msgilligan: Minor request: We generally import this classes rather than use fully-qualified package names.

jeffmcclure:
@msgilligan the PR has the "changes requested" status. Is there something I need to do? I can't seem to clear the status, perhaps only you can.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,202 @@
+/*
+ * Copyright 2025 Jeff McClure.
```
msgilligan: For the last 5 years or so we are using "Copyright by the original author or authors" in all new files.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Parse a BIP-39 mnemonic seed phrase into a list of words.
+     * @param seedPhrase Space-separated BIP-39 mnemonic seed phrase. Words must be valid BIP-39 English words.
+     * @return List of words parsed from {@code seedPhrase}.  Leading, trailing and extra spaces are removed.
+     */
+    public static List<String> parseSeedPhrase(String seedPhrase) {
+        return Arrays.asList(seedPhrase.trim().split("\\s+"));
+    }
```
msgilligan, line 47: Does this duplicate functionality we already have? It is not specific to DeterministicEntropy.  If not, maybe it should be added elsewhere?

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * Derive a DeterministicKey from a BIP-39 mnemonic seed phrase and passphrase.
+     * @param seedPhrase Space-separated BIP-39 mnemonic seed phrase. Words must be valid BIP-39 English words.
+     * @param passphrase BIP-39 passphrase for the BIP-39 mnemonic seed phrase.  Use empty string if there is no passphrase.
+     * @return DeterministicKey for the BIP-39 mnemonic seed phrase and passphrase combination.
+     * @throws MnemonicException if the mnemonic words are invalid
+     */
+    static public DeterministicKey getPrivateKey(String seedPhrase, String passphrase) throws MnemonicException {
+        return getPrivateKey(parseSeedPhrase(seedPhrase), passphrase);
+    }
```
msgilligan, line 51: Does this duplicate functionality we already have? It is not specific to DeterministicEntropy. If not, maybe it should be added elsewhere?

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,202 @@
...
+    static public DeterministicKey getPrivateKey(List<String> mnemonicWords, String passphrase) throws MnemonicException {
+        MnemonicCode.INSTANCE.check(mnemonicWords); // Validate the mnemonic words
+
+        // Create a DeterministicSeed from the mnemonic words with passphrase
+        DeterministicSeed seed = DeterministicSeed.ofMnemonic(mnemonicWords, passphrase);
+
+        // Derive the master private key from the seed
+        assert seed.getSeedBytes() != null;
+        return HDKeyDerivation.createMasterPrivateKey(seed.getSeedBytes());
+    }
```
msgilligan, line 114: Does this duplicate functionality we already have? It is not specific to DeterministicEntropy. If not, maybe it should be added elsewhere?

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,202 @@
...
+     * @param deterministicKey DeterministicKey to calculate the fingerprint for.
+     * @return Fingerprint as a 4-byte hex string.  Example: "73c5da0a" or "00000000"
+     */
+    public static String fingerprint(DeterministicKey deterministicKey) {
+        // Calculate the fingerprint (first 4 bytes of HASH160 of the master public key)
+        byte[] pubKeyBytes = deterministicKey.getPubKey();
+        byte[] hash160 = sha256hash160(pubKeyBytes);
+        byte[] fingerprintBytes = Arrays.copyOfRange(hash160, 0, 4);
+        return new HexFormat().formatHex(fingerprintBytes);
+    }
```
msgilligan, line 126: Does this duplicate functionality we already have? It is not specific to DeterministicEntropy. If not, maybe it should be added elsewhere?

msgilligan:
> @msgilligan the PR has the "changes requested" status. Is there something I need to do? I can't seem to clear the status, perhaps only you can.

Looks like I had a pending review that was not completed, I added a few more comments and completed the review. I'm sorry for the confusion.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Change made

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Function removed.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Function removed

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Code removed

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Code removed

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Change made

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,154 @@
...
+     * Derive a BIP-85 key and return the corresponding BIP-39 mnemonic seed phrase.
+     *
+     * @param masterPrivateKey DeterministicKey to derive from.
+     * @param language         0 for English, see BIP-39 for other languages <a href="https://bips.xyz/85#bip39">https://bips.xyz/85#bip39</a>.
+     * @param wordCount        12, 18, or 24 for BIP-39 word counts.
+     * @param index            0 to 9999 for the index of the child key.
+     * @return Seed Phrase for the derived BIP-85 deterministic key.
+     */
+    public static String deriveBIP85Mnemonic(DeterministicKey masterPrivateKey, int language, int wordCount, int index) {
+        int[] derivationPath = {83696968, 39, language, wordCount, index};
```
msgilligan: We should probably define constants for these two numeric  values.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Derive a BIP-85 key and return the corresponding BIP-39 mnemonic seed phrase.
+     *
+     * @param masterPrivateKey DeterministicKey to derive from.
+     * @param language         0 for English, see BIP-39 for other languages <a href="https://bips.xyz/85#bip39">https://bips.xyz/85#bip39</a>.
+     * @param wordCount        12, 18, or 24 for BIP-39 word counts.
+     * @param index            0 to 9999 for the index of the child key.
+     * @return Seed Phrase for the derived BIP-85 deterministic key.
+     */
+    public static String deriveBIP85Mnemonic(DeterministicKey masterPrivateKey, int language, int wordCount, int index) {
```
msgilligan: This should probably return `List<String>`.

msgilligan: I'm also wondering if we should use stronger types (e.g. `enum`)  for `language` and for `wordCount`. bitcoinj seems to only support English for the word-list, btw.

(This is not a request for a change, just a topic for discussion.)

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * Derive the entropy for a BIP-85 derived key.
+     *
+     * @param masterPrivateKey DeterministicKey to derive from.
+     * @param derivationPath   path of derivation.
+     * @return Entropy for the derived BIP-85 deterministic key.
+     */
+    public static byte[] deriveBIP85Entropy(DeterministicKey masterPrivateKey, int[] derivationPath) {
```
msgilligan: This should take an `HDPath` (which in 0.18 will be `int[]` internally) rather than an `int[]`.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * Derive a BIP-39 mnemonic seed phrase for a BIP-85 derived key.
+     *
+     * @param masterPrivateKey DeterministicKey to derive from.
+     * @param derivationPath   path of derivation.  Example: {83696968, 39, 0, 12, 0} for BIP-39 English 12-word mnemonic, index 0.
+     * @return Seed Phrase for the derived BIP-85 deterministic key.
+     */
+    public static String deriveBIP85Mnemonic(DeterministicKey masterPrivateKey, int[] derivationPath) {
```
msgilligan: This should take an `HDPath`, not `int[]`. If adding a helper method to `HDPath` can make this easier, we should consider it. And also return a `List<String>`.

msgilligan: Since this is named `deriveBIP85Mnemonic` it should probably be next to the other methods with the same name in the source file.

jeffmcclure:
I'm new to the project, so I'm good with any needed changes.   BTW, please
let me know if you receive this... I am replying to the email from github
and I don't know if that gets the message to you.

Thank you!

- Jeff

msgilligan:
> I'm new to the project, so I'm good with any needed changes.

Great, thanks for bearing with me.

> BTW, please let me know if you receive this... I am replying to the email from github and I don't know if that gets the message to you. Thank you! - Jeff

Not only did I receive it, but it publicly posts to the issue.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Constants defined.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Returning DeterministicSeed now instead of mnemonic.
Added enum for language.
I don't think an enum makes sense for replacing the wordCount.  Added tests to ensure invalid wordCount throws an IllegalArgumentException.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Done

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Done.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -36,21 +38,55 @@
...
     private static final Logger logger = LoggerFactory.getLogger(DeterministicEntropy.class);
 
+    private static final int BIP85_PATH_ROOT = 83696968;
+    private static final int BIP39_APPLICATION_NUMBER = 39;
+
+    public enum Language {
+        English(0),
+        Japanese	(1),
+        Korean	(2),
+        Spanish	(3),
```
msgilligan: Something is weird with the whitespace here.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -36,21 +38,55 @@
...
-     * @param language         0 for English, see BIP-39 for other languages <a href="https://bips.xyz/85#bip39">https://bips.xyz/85#bip39</a>.
+     * @param language         Language <a href="https://bips.xyz/85#bip39">https://bips.xyz/85#bip39</a>.
      * @param wordCount        12, 18, or 24 for BIP-39 word counts.
      * @param index            0 to 9999 for the index of the child key.
      * @return Seed Phrase for the derived BIP-85 deterministic key.
      */
-    public static String deriveBIP85Mnemonic(DeterministicKey masterPrivateKey, int language, int wordCount, int index) {
-        int[] derivationPath = {83696968, 39, language, wordCount, index};
-        return deriveBIP85Mnemonic(masterPrivateKey, derivationPath);
+    public static List<String> deriveBIP85Mnemonic(DeterministicKey masterPrivateKey, Language language, int wordCount, int index) {
```
msgilligan: Wordcount should probably be an enum, too. That would prevent use of invalid values  at compile time. It also will prevent coding errors where `wordCount` and index are unintentionally swapped.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -36,21 +38,55 @@
...
+        List<ChildNumber> children = Stream.of(BIP85_PATH_ROOT, BIP39_APPLICATION_NUMBER, language.intValue, wordCount, index).map(ChildNumber::new).collect(Collectors.toList());
```
msgilligan: Shouldn't language and word count be "hardened" values, e.g. `0'`?  With the enum you could actually define them as hardened `ChildNumber`s internally.

Oh, I see the `.derive()` method you are calling below always uses the hardened value so that's why the tests work.

But I still think we should use hardened values in the `HDPath` for consistency.  (and the `.derive()` method as used below will throw if the hardened bit is already set)

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -61,24 +97,37 @@ public static String deriveBIP85Mnemonic(DeterministicKey masterPrivateKey, int
      * @param index            0 to 9999 for the index of the child key.
      * @return Seed Phrase for the derived BIP-85 deterministic key.
      */
-    public static String deriveBIP85Mnemonic(DeterministicKey masterPrivateKey, int wordCount, int index) {
-        int[] derivationPath = {83696968, 39, 0, wordCount, index};
-        return deriveBIP85Mnemonic(masterPrivateKey, derivationPath);
+    public static List<String> deriveBIP85Mnemonic(DeterministicKey masterPrivateKey, int wordCount, int index) {
```
msgilligan: Should we range-check the index?

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -61,24 +97,37 @@ public static String deriveBIP85Mnemonic(DeterministicKey masterPrivateKey, int
...
+            logger.debug("deriveBIP85Mnemonic() BIP-85 Child Private Key (hex): {}", hex.formatHex(fullEntropy));
+            logger.debug("deriveBIP85Mnemonic() BIP-85 Entropy (hex)a: {}", hex.formatHex(entropy));
+            logger.debug("deriveBIP85Mnemonic() BIP-85 Entropy 512bits (hex): {}", hex.formatHex(entropy512Bits));
+            logger.debug("deriveBIP85Mnemonic() BIP-85 Entropy full (hex): {}", hex.formatHex(fullEntropy));
+        }
+
+        return entropy;
+    }
+
+    private static byte[] getBytes(HDPath hdPath, byte[] entropy512Bits) {
```
msgilligan: Maybe pass in the length here so that this method doesn't need to know about HDPaths? I might also suggest calling it "truncate" or something similar, `getBytes()` is very generic.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
msgilligan: Also you might find the `.m(ChildNumber... children)` static constructor convenient here, especially if the enums and constants are `ChildNumber` (or can return `ChildNumber`) types.

msgilligan (changes requested):
Making progress!

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,182 @@
...
+        return DeterministicSeed.ofEntropy(data, "");
+    }
+
+    /**
+     * Perform a BIP-85 derivation and return the DeterministicSeed.
+     *
+     * @param masterPrivateKey DeterministicKey to derive from.
+     * @param language         Language <a href="https://bips.xyz/85#bip39">https://bips.xyz/85#bip39</a>.
+     * @param wordCount        12, 18, or 24 for BIP-39 word counts.
+     * @param index            0 to 9999 for the index of the child key.
```
msgilligan: I don't see the limit of 9999 as an upper-bound in the BIP-85. Where is this defined?

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Fixed

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: I just checked the bip and grok ai.   There is no real upper bound.   according to grok the upper bound is 2^31 - 1 (2,147,483,647).  Coldcard limits the index to 9999, but since that is not part of the standard I don't think we should enforce that limit.   The lower limit is zero and I'm adding that check.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: I just checked the bip and grok ai. There is no real upper bound. according to grok the upper bound is 2^31 - 1 (2,147,483,647). Coldcard limits the index to 9999, but since that is not part of the standard I don't think we should enforce that limit. The lower limit is zero and I'm adding that check.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: good ideas... done.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: enum added.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: I see the .m() method on HDPath but I don't see how that is better than HDPath.of()... let me know.

I tried creating hardened children, but it results in the following exception:

Most significant bit is reserved and shouldn't be set: -2063786680
java.lang.IllegalArgumentException: Most significant bit is reserved and shouldn't be set: -2063786680
	at org.bitcoinj.crypto.ChildNumber.<init>(ChildNumber.java:59)
	at org.bitcoinj.crypto.DeterministicKey.derive(DeterministicKey.java:527)
	at org.bitcoinj.crypto.DeterministicEntropy.deriveKey(DeterministicEntropy.java:168)
	at org.bitcoinj.crypto.DeterministicEntropy.deriveBIP85Entropy(DeterministicEntropy.java:127)

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -61,6 +61,19 @@ public enum Language {
         }
     }
 
+    public enum WordCount {
+        Twelve(12),
+        Eighteen(18),
+        TwentyFour(24);
+
+        private final int intValue;
```
msgilligan: You could either replace this with a `ChildNumber` value ~~or add an accessor/converter method~~ (I would keep it simple and just make it `private final ChildNumber value`).

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -81,24 +94,27 @@ public static DeterministicSeed deriveBIP85Seed(DeterministicKey masterPrivateKe
...
      * @param language         Language <a href="https://bips.xyz/85#bip39">https://bips.xyz/85#bip39</a>.
-     * @param wordCount        12, 18, or 24 for BIP-39 word counts.
+     * @param wordCount        Number of words in generated seed.
      * @param index            0 to 9999 for the index of the child key.
      * @return Seed for the BIP-85 derivation.
      */
-    public static DeterministicSeed deriveBIP85Seed(DeterministicKey masterPrivateKey, Language language, int wordCount, int index) {
-        List<ChildNumber> children = Stream.of(BIP85_PATH_ROOT, BIP39_APPLICATION_NUMBER, language.intValue, wordCount, index).map(ChildNumber::new).collect(Collectors.toList());
+    public static DeterministicSeed deriveBIP85Seed(DeterministicKey masterPrivateKey, Language language, WordCount wordCount, int index) {
+        // BIP-85 does not specify an upper limit on the index.  The index must be >= 0.
```
msgilligan, line 138: This code will work, but technically the index is 31-bits, with the upper limit being 2^31 - 1. The high-order bit indicates hardening. Since a Java `int` is unsigned 32-bits this code will work, but I think it will be more clear ~~if you test for `(a & HARDENED_BIT) != 0` -- see `ChildNumber.hasHardenedBit(int a)` (which is `private`)~~

msgilligan: or you could just use the `ChildNumber(int childNumber, boolean isHardened)` constructor with `isHardened = true`. See below.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -81,24 +94,27 @@ public static DeterministicSeed deriveBIP85Seed(DeterministicKey masterPrivateKe
...
+        if (index < 0) throw new IllegalArgumentException("index must be >= 0");
+        List<ChildNumber> children = Stream.of(BIP85_PATH_ROOT, BIP39_APPLICATION_NUMBER, language.intValue, wordCount.intValue, index)
+                .map(ChildNumber::new).collect(Collectors.toList());
```
msgilligan: ~~Use `.map(cn -> new ChildNumber(cn, true)` to set the hardened bit.~~

msgilligan: You could also use something like:

`HDFullPath path = HDPath.m(BIP85_PATH_ROOT, BIP85_PATH_ROOT, language.childNumber(), wordCount.childNumber(), new ChildNumber(index, true))`

_if_ the type of all those values is `ChildNumber`.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -125,10 +141,10 @@ private static byte[] deriveBIP85Entropy(DeterministicKey masterPrivateKey, HDPa
         return entropy;
     }
 
-    private static byte[] getBytes(HDPath hdPath, byte[] entropy512Bits) {
+    private static byte[] truncate(int wordCount, byte[] entropy512Bits) {
```
msgilligan: Actually if you pass the wordCount in as an enum, you can eliminate the switch statement.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: Done

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: done

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: done

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: done

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: done

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,221 @@
...
+        Korean(2),
+        Spanish(3),
+        ChineseSimplified(4),
+        ChineseTraditional(5),
+        French(6),
+        Italian(7),
+        Czech(8),
+        Portuguese(9);
+
+        private final int intValue;
```
msgilligan: I think the value should be stored as a `ChildNumber`

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,221 @@
...
+
+        Language(int intValue) {
+            this.intValue = intValue;
```
msgilligan: You can use new `ChildNumber(intValue, true)` directly here.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
msgilligan: As with `Language`, I think it will be best to store the value as a `ChildNumber` type.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,221 @@
...
+        }
+    }
+
+    public enum WordCount {
+        Twelve(12, 16),
+        Eighteen(18, 24),
+        TwentyFour(24, 32);
+
+        private final int intValue;
+        private final int bitCount;
```
msgilligan: Isn't this actually a count of bytes? It could be `byteCount` or `byteLength` might be even more descriptive.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,221 @@
...
+     * Derive the entropy for a BIP-85 derived key.
+     *
+     * @param masterPrivateKey DeterministicKey to derive from.
+     * @param hdPath           path of derivation.
+     * @return Entropy for the derived BIP-85 deterministic key.
+     */
+    private static byte[] deriveBIP85Entropy(DeterministicKey masterPrivateKey, HDPath hdPath) {
+        byte[] fullEntropy = deriveKey(masterPrivateKey, hdPath);
+        byte[] entropy512Bits = generateBIP85Entropy(fullEntropy);
+        byte[] entropy = Arrays.copyOf(entropy512Bits, WordCount.forWordCount(hdPath.children[3]).bitCount);
```
msgilligan: Instead of `.children[3]` (`.children` is meant for internal use only, but is technically available to other classes in the same package) you should use `.get(3)` which returns a `ChildNumber`.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,221 @@
...
+
+        private final int intValue;
+        private final int bitCount;
+
+        WordCount(int intValue, int bits) {
+            this.intValue = intValue;
+            this.bitCount = bits;
+        }
+
+        public static WordCount forWordCount(int wordCount) {
```
msgilligan, line 92: This isn't needed if you use the `.get(3)` method instead of `.children[3]` in the only caller.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,221 @@
...
+            logger.debug("deriveBIP85Entropy() BIP-85 Entropy (hex)a: {}", hex.formatHex(entropy));
+            logger.debug("deriveBIP85Entropy() BIP-85 Entropy 512bits (hex): {}", hex.formatHex(entropy512Bits));
+            logger.debug("deriveBIP85Entropy() BIP-85 Entropy full (hex): {}", hex.formatHex(fullEntropy));
+        }
+
+        return entropy;
+    }
+
+    private static byte[] deriveKey(DeterministicKey childKey, HDPath hdPath) {
+        for (int val : hdPath.children) {
```
msgilligan: Instead of .children[3] you should use `.list()` which returns a `List<ChildNumber>`.

msgilligan (changes requested):
I'm still hoping we can get to more consistent use of `ChildNumber`. I think it will make the code simpler and more robust.

My apologies for `.children` being accessible, `DeterministicEntropy` has access to it because it is in the same package, but it's really considered an internal detail of `HDPath`.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: done

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: done

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: renamed to byteLength

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: done

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: done

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: I don't understand how to do this without a switch statement and without an enum lookup.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
jeffmcclure: done

msgilligan:
I made two related, minor PRs:

* PR #3851 
* PR #3852 

Both of these changes were the result of reviewing this PR and looking at the interface and documentation of `HDPath` and `ChildNumber`.

jeffmcclure:
> I made two related, minor PRs:
> 
> * PR [HDPath: Reduce visibility of `children` #3851](https://github.com/bitcoinj/bitcoinj/pull/3851)
> * PR [ChildNumber: mark as value-based in JavaDoc #3852](https://github.com/bitcoinj/bitcoinj/pull/3852)
> 
> Both of these changes were the result of reviewing this PR and looking at the interface and documentation of `HDPath` and `ChildNumber`.

Looks good.  Looks like I don't need to make any changes in PR 3791 since I already moved off the children member.

jeffmcclure:
@msgilligan I see the PR if flagged with "changes requested" but I don't see anything.

msgilligan (commented):
To get this PR merged, I think two remaining things need to happen:

1. I want to make one more review pass actually working with the code in my IDE (I have a few ideas for further simplification that I would like to try)
2. We need to get a review from @schildbach, but I think he should wait until after I finished (1).

msgilligan:
> @msgilligan I see the PR if flagged with "changes requested" but I don't see anything.

I just now found the UI option to re-request my own review (I tried once before to make the "changes requested" go away and couldn't figure out how to do it.) See my recent "comment-only" review that explains what needs to happen.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
```diff
@@ -0,0 +1,217 @@
...
+
+        Language(int intValue) {
+            this.childNumber = createHardenedChildNumber(intValue);
+        }
+
+        public ChildNumber childNumber() {
+            return childNumber;
+        }
+    }
+
```
msgilligan, line 78:
```suggestion

    /**
     * Representation of standard word counts.
     */
```

msgilligan: This is just a test of the GitHub suggested change feature (which I've never used until now.)

msgilligan (commented):
Just a test. But we probably should add a comment here.

msgilligan:
Hi Jeff, thanks for your patience!

Here are my suggested changes: https://github.com/bitcoinj/bitcoinj/pull/3909

We also need to rebase onto `master`.

core/src/main/java/org/bitcoinj/crypto/DeterministicEntropy.java
msgilligan: This was just a test. Ignore for now.

jeffmcclure:
I'll get to them tonight.

Thank you!

- Jeff

msgilligan:
> I'll get to them tonight. Thank you! - Jeff

Awesome!

I thought that there was a way to associate the new PR with this one (like the suggest changes via web UI feature) and was going to add some comments about why I made the various changes, though in most cases you should be able to guess. If you have any questions let me know.

Also if PR #3910 is merged, you should be able use `CryptoUtils.hmacSha512` to simplify `generateBIP85Entropy`.

msgilligan:
Happy New Year @jeffmcclure ! Are you still interested and available to work on this?

We have finally merged several PRs to improve our `CryptoUtils` class to provide what you need for `HmacSHA512`. See:

* PR #3910
* PR #3911

jeffmcclure:
Happy New Year!  Yes, I'm still interested.  I'll get it done.

Thank you!

- Jeff

PR #3793, opened 2025-04-26
Move Wallet and BlockFileLoader integration tests to o.b.test package

msgilligan:
Our _integration_ tests should not live in `core` sub-packages and should be in their own package hierarchy.

(We should also require tests in this package use JUnit 5.)


schildbach:
Rebased.

schildbach (approved):

PR #3795, opened 2025-04-26
Wallettool: Make `--net` optional

ItoroD:


wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
```diff
@@ -379,12 +379,12 @@ public Integer call() throws IOException, BlockStoreException {
...
             return 1;
         }
-        if (wallet.network() != net) {
-            System.err.println("Wallet does not match requested network: " +
-                    wallet.network() + " vs " + net);
-            return 1;
+        Network detectedWalletNetwork = wallet.network();
+        if (detectedWalletNetwork != net) {
+            System.out.println("Detected Wallet network is " + detectedWalletNetwork);
+            net = (BitcoinNetwork) detectedWalletNetwork;
```
msgilligan, line 385: I think a non-matching network should remain an error. But of course, that only works if `net` has no default (or rather _defaults to the network of the existing wallet file)

msgilligan (changes requested):
I think this change should probably wait until after we refactor to subcommands. The most predictable behavior would be for `--net` to be removed from all commands except `create` which would either _require_ `--net` or default it to MAINNET (or maybe TESTNET.  We should probably subtly encourage people to use wallet-tool with TESTNET)

wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
ItoroD: Yes. That will work if `net` has no default. As right now its hard to tell if the value of `net` we are seeing was passed by the user of the default value. Maybe this should wait till we have completed subcommand design doc review.

PR #3798, opened 2025-04-27
PeerSocketHandler: minor cleanup

msgilligan:
Minor cleanup in `exceptionCaught()`


core/src/main/java/org/bitcoinj/core/PeerSocketHandler.java
```diff
@@ -244,9 +244,9 @@ private void exceptionCaught(Exception e) {
         String s = addr == null ? "?" : addr.toString();
         if (e instanceof IOException) {
             // Short message for network errors
-            log.info(s + " - " + e.getMessage());
+            log.info("{} - {}", s, e.getMessage());
         } else {
-            log.warn(s + " - ", e);
```
schildbach, line 249: This won't work. It should actually generate another warning in your IDE because the number of arguments doesn't match the number of template variables.

Maybe we should change the line to use `e.getMessage()` just like the above?

schildbach (changes requested):

core/src/main/java/org/bitcoinj/core/PeerSocketHandler.java
msgilligan: Ooops, typo. For some reason my IDE is not complaining. I'll fix it.

It seems Matt wanted the shorter message for network errors and the longer message for (less expected) errors. I think we should leave it that way (at least in this commit)

msgilligan: The IDE was right, I think: https://slf4j.org/faq.html#paramException

msgilligan: I changed the log message and force-pushed.

core/src/main/java/org/bitcoinj/core/PeerSocketHandler.java
```diff
@@ -244,9 +244,9 @@ private void exceptionCaught(Exception e) {
...
+            log.warn("Exception communicating with Peer at {}", s, e);
```
ronodhirSoumik, line 249: `Exception communicating with Peer at {}` ->
`Exception communicating with Peer at address {}`

core/src/main/java/org/bitcoinj/core/PeerSocketHandler.java
```diff
@@ -244,9 +244,9 @@ private void exceptionCaught(Exception e) {
         String s = addr == null ? "?" : addr.toString();
         if (e instanceof IOException) {
             // Short message for network errors
-            log.info(s + " - " + e.getMessage());
+            log.info("{} - {}", s, e.getMessage());
```
ronodhirSoumik, line 247: though it is not related to this change, but just to add a concern here, we are substituting the exact invalid input with ''?" which may be confusing. Can the log message be more elaborated in this case? Like-
`log.info("Invalid address[{}] - {}", s, e.getMessage());`

schildbach: Sure, though I'd investigate first when and why `peerAddress` can be `null` at all. Maybe we can eliminate that case? However, that's for a different PR (feel free to beat us at it).

schildbach (approved):

PR #3799, opened 2025-04-27
InboundMessageQueuer: use ConcurrentHashMap for mapPingFutures

msgilligan:


msgilligan:
> Can you add to the commit description about why we need this change to concurrent? Does this fix an issue?

I don't think it is strictly necessary to fix a bug. I am hoping it will magically fix our random tree-related test failures, but it probably won't. But using a concurrent map doesn't hurt.

(**Update: I should have been more clear here -- we are using non-thread-safe `HashMap` on multiple threads and we shouldn't be.**  But I have no evidence it is causing any specific failures we have seen.)

Should I just put that in the commit description?

msgilligan:
@schildbach I just discovered this old PR. There are potential concurrency issues with access to the `mapPingFutures` -- it is a `public` field that is being used in multiple threads.

I have updated the commit description (and also added a comment to describe the purpose of `mapPingFutures`.

msgilligan:
@schildbach @hasskell this _does not_ fix the "Address In Use" issue, because it happened on this PR and I re-ran the one failing test.

schildbach (approved):

PR #3802, opened 2025-04-28
FutureUtil: Improve JavaDoc and make more readable

msgilligan:
FutureUtils remains useful even after we've removed all the deprecated Guava Future adapters. But the JavaDoc needed an update and the code needed to be more readable in order to update the JavaDoc.

6 commits.


schildbach:
Rebased.

schildbach (approved):

Issue #3806, opened 2025-04-28
Warning from JDK about `sun.misc.Unsafe` being called from `protobuf-javalite`


msgilligan:
**bitcoinj** can build and run tests with Java 24 according to testing I did today with `Temurin-24.0.1+9` and Gradle 8.14.

However there is a warning from the build:

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::arrayBaseOffset has been called by com.google.protobuf.UnsafeUtil$MemoryAccessor (file:/Users/sean/.gradle/caches/modules-2/files-2.1/com.google.protobuf/protobuf-javalite/4.30.2/d11a96deaca12843aa9bf2b9f265c7bead7662cb/protobuf-javalite-4.30.2.jar)
WARNING: Please consider reporting this to the maintainers of class com.google.protobuf.UnsafeUtil$MemoryAccessor
```

This issue has been reported upstream to protobuf here: https://github.com/protocolbuffers/protobuf/issues/20760

msgilligan:
I suggest we upgrade to Gradle 8.14 and update our CI to use JDK 24 where it was previously using JDK 23 despite this warning. It's just a warning about future deprecation and protobuf should fix it before it becomes a problem. It certainly isn't a reason for us to not upgrade.  And JDK 23 is officially unsupported, so we should move to JDK 24.

We could add a JDK runtime flag to disable the warning, but I think we should leave it in as a reminder to update `protobuf-javalite`.

PR #3807, opened 2025-04-28
Update protobuf-javalite to 4.30.2, Gradle Protobuf Plugin to 0.9.5

msgilligan:
This is 2 commits: 1 for `protobuf-javalite` and 1 for the Gradle Protobuf Plugin. They are independent, but we might was well update both in the same PR.

Note the this update does not resolve Issue #3806 which is an upstream deprecation warning issue for Protobuf with Java 24.


schildbach (approved):

schildbach:
Rebased.

schildbach (approved):

PR #3815, opened 2025-05-18
TransactionInputTest: use getConnectedOutput() in disconnect tests

msgilligan:
This consists of two commits:

1. Improvements to tests and code comments to better test and document _current_ test behavior
2. The change to use `getConnectedOutput()` rather than `.connectedOutput` in `testStandardWalletDisconnect()`

Note that this is an improvement over changes in PR #3779 which simply removed the line:

```
    assertNull(txInToDisconnect.getOutpoint().connectedOutput);
```
from `testStandardWalletDisconnect()`

Also for reference, see related PR #3816 which uses `getConnectedOutput()`  to simplify `TransactionInput.disconnect()`.


schildbach:
Rebased.

PR #3816, opened 2025-05-18
TransactionInput: simplify disconnect() with outpoint.getConnectedOutput()

msgilligan:
`disconnect()` was duplicating the logic of `getConnectedOutput()` and can be simplified by using `getConnectedOutput()` directly.


schildbach:
Rebased.

schildbach (approved):

PR #3828, opened 2025-05-24
WIP: TransactionOutPointReference and subtypes

msgilligan:
Rename TransactionOutPoint to TransactionOutPointReference and 
make abstract.  Three subclasses:

* TransactionOutPoint: Contains only hash and index
* TransactionConnectedOutPoint: Has a connected fromTx
* OutputConnectedOutPoint: Has a connectedOutput

Some of the TransactionInput constructors now take the TransactionOutPoint
subtype and could be change to return a "disconnected" TransactionInput.

This approach has the following advantages:

1. Stronger typing helps make invalid states impossible to represent
2. TransactionOutPoint subtype uses less memory and is returned
   when calling read(ByteBuffer payload)
3. Eliminates external use of `null` values
4. Allows using types and pattern matching for conditional logic. 
   (This will be bigger advantage when using newer JDKs, but also
   moves the code towards a more modern "data-oriented programming"
   style, even where the JDK 8 syntax is slightly more verbose)
5. Will help us created simplified "disconnected" TransactionInput
   type and will be useful for creating immutable blocks and 
   transactions.

Disadvantages:

1. In some cases the "instanceOf" code seems more verbose than
   using null-checking.
2. The current names for the subclasses are little bit long/unwieldy,
   but this can be changed.

TODO:

1. Consider different names for the types
2. Consider not nesting one or more of the subtypes
3. Consider making the base type an interface not abstract class
4. Use the stronger types to implement disconnect TransactionInput
   and our other "immutable" block and transactions types
5. Replace the nullable getConnectedKey() and getConnectedRedeemData() methods

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/TransactionOutPointReference.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionOutPointReference.java
@@ -42,27 +42,59 @@ import static org.bitcoinj.base.internal.Preconditions.checkState;
  * 
  * <p>Instances of this class are not safe for use by multiple threads.</p>
  */
-public class TransactionOutPoint {
+public abstract class TransactionOutPointReference {
     public static final int BYTES = 36;
 
     /** Special outpoint that normally marks a coinbase input. It's also used as a test dummy. */
     public static final TransactionOutPoint UNCONNECTED =
-            TransactionOutPoint.of(Sha256Hash.ZERO_HASH, ByteUtils.MAX_UNSIGNED_INTEGER);
+            TransactionOutPointReference.of(Sha256Hash.ZERO_HASH, ByteUtils.MAX_UNSIGNED_INTEGER);
 
     /** Hash of the transaction to which we refer. */
-    private final Sha256Hash hash;
+    protected final Sha256Hash hash;
     /** Which output of that transaction we are talking about. */
-    private final long index;
+    protected final long index;
 
-    // `fromTx` and `connectedOutput` can both be `null`, but they both can't be non-`null`.
+    public static class TransactionOutPoint extends TransactionOutPointReference {
+        public TransactionOutPoint(Sha256Hash hash, long index) {
+            super(hash, index);
+        }
+    }
 
-    // This is not part of bitcoin serialization. It points to the connected transaction.
-    @Nullable
-    private final Transaction fromTx;
+    public interface HasConnectedOutput {
+        TransactionOutput connectedOutput();
+    }
 
-    // The connected output.
-    @Nullable
-    private final TransactionOutput connectedOutput;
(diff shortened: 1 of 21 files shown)


core/src/main/java/org/bitcoinj/core/TransactionOutPointReference.java
```diff
@@ -72,28 +106,14 @@ public class TransactionOutPoint {
     public static TransactionOutPoint read(ByteBuffer payload) throws BufferUnderflowException, ProtocolException {
         Sha256Hash hash = Sha256Hash.read(payload);
         long index = ByteUtils.readUint32(payload);
-        return TransactionOutPoint.of(hash, index);
+        return TransactionOutPointReference.of(hash, index);
```
msgilligan, line 109: Note that read is now returning the more compact `TransactionOutPoint` type (containing only hash and index)

core/src/main/java/org/bitcoinj/core/TransactionOutPointReference.java
```diff
@@ -207,8 +224,10 @@ public byte[] getConnectedPubKeyScript() {
      */
     @Nullable
     public ECKey getConnectedKey(KeyBag keyBag) throws ScriptException {
-        TransactionOutput connectedOutput = getConnectedOutput();
-        Objects.requireNonNull(connectedOutput, "Input is not connected so cannot retrieve key");
+        if (!(this instanceof HasConnectedOutput)) {
+            throw new NullPointerException("Input is not connected so cannot retrieve key");
+        }
+        TransactionOutput connectedOutput = ((HasConnectedOutput) this).connectedOutput();
```
msgilligan, line 230: This  method should be moved to one of the subtypes.

core/src/main/java/org/bitcoinj/core/TransactionOutPointReference.java
```diff
@@ -233,8 +252,10 @@ public ECKey getConnectedKey(KeyBag keyBag) throws ScriptException {
      */
     @Nullable
     public RedeemData getConnectedRedeemData(KeyBag keyBag) throws ScriptException {
-        TransactionOutput connectedOutput = getConnectedOutput();
-        Objects.requireNonNull(connectedOutput, "Input is not connected so cannot retrieve key");
+        if (!(this instanceof HasConnectedOutput)) {
+            throw new NullPointerException("Input is not connected so cannot retrieve key");
+        }
+        TransactionOutput connectedOutput = ((HasConnectedOutput) this).connectedOutput();
```
msgilligan, line 258: This  method should be moved to one of the subtypes.

msgilligan (commented):
A made some notes on a few lines.

PR #3830, opened 2025-05-27
WIP: Testnet4

schildbach:
This is early work on Testnet4.

When I last tested it several rebases ago, it successfully synced the Testnet4 chain. However, the specific rules for difficulty transitions are still missing so it could easily follow an invalid chain.

Also, a couple of refactorings / side-work need to be extracted from this PR and merged separately.


schildbach:
Rebased.

schildbach:
Rebased.

Issue #3106, opened 2023-05-29
Consider migrating to JSpecify nullability annotations


msgilligan:
It looks like the major players in the JDK ecosystem are finally converging on a standard for nullability annotations.

See:
* https://jspecify.dev/docs/start-here
* https://jspecify.dev/docs/api/org/jspecify/annotations/package-summary.html
* https://github.com/jspecify/jspecify/wiki/adoption

We need to keep an eye on this project and on the various tools that are migrating to it and then decide on which release (if any) we should migrate.

I'm going to mark this as "for consideration" in the 0.18 release, so we don't forget, but we can chose to postpone it if the tools aren't ready at that time.

msgilligan:
We have merged **JSpecify** support into secp256k1-jdk: https://github.com/bitcoinj/secp256k1-jdk/pull/64

msgilligan:
I changed the title from "Consider migrating" to "Migrate".

Guava 34.4.4 has been upgraded to use JSpecify (and no longer provides the transitive dependency for the nullablilty annotations we were using) -- See PR #3833 (the first commit of which upgrades to Guava 34.4.4)

PR #3608 removes all nullability annotations from `bitcoinj-base` although we could add (1?) JSpecify annotation(s) to declare the module as never returning or accepting null.

msgilligan:
There are essentially three steps to fully migrating to **JSpecify**:

1. Replace existing `com.google.code.findbugs:jsr305` (`base`, `core`, `integration-test`, and `wallettool`) and `jakarta.annotation:jakarta.annotation-api` (`wallettemplate`) annotations with the ones in JSpecify.
2. Add `@NullMarked` annotations to all packages and/or modules to indicate everything that is not marked is non-nullable.
3. Use a static analysis tool (e.g. Nullaway) to check for and eliminate nullability errors.

At this point the pathway to (1) is clear and we currently have PRs that will get us there.  (We also have some PRs that partially implement (2) and (3), but let's try to get (1) finished first and then we can rebase and discuss those PRs.)

To complete (1) we need to approve and merge the following PRs:

- [x] PR #3834 (migrates `base`, `core`, `integration-test`, and `wallettool`)
- [x] PR #3844 (migrates `wallettool` from Jakarta annotations)

Once (1) is completed we can pursue `@NullMarked` on a package-by-package and module-by-module basis and at some point we want to start enabling **NullAway** (or another tool to begin checking/enforcement.)

[...]

PR #3833, opened 2025-06-04
core: upgrade to Guava 33.4.8

msgilligan:
The first commit upgrades to 33.4.4:

Guava 33.4.4 uses JSpecify and removes a transitive dependency on com.google.code.findbugs:jsr305:3.0.2 which bitcoinj currently uses for null annotations.

We plan to switch bitcoinj to JSpecify, see Issue #3106, but this commit simply adds `com.google.code.findbugs:jsr305:3.0.2` as a direct dependency, so that bitcoinj can upgrade to JSpecify in a separate PR.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -11,7 +11,8 @@ version = '0.18-SNAPSHOT'
 dependencies {
     api project(':bitcoinj-base')
     api 'org.bouncycastle:bcprov-jdk15to18:1.80'
-    api 'com.google.guava:guava:33.4.0-android'
+    api 'com.google.guava:guava:33.4.4-android'
+    api 'com.google.code.findbugs:jsr305:3.0.2'
     api 'com.google.protobuf:protobuf-javalite:4.30.2'
     implementation 'org.slf4j:slf4j-api:2.0.16'
 


schildbach:
Actually shouldn't we, before doing anything else, add the missing `com.google.code.findbugs:jsr305` dependencies? We should not rely on transitive dependencies if we're using something directly, imho.

msgilligan:
> Actually shouldn't we, before doing anything else, add the missing `com.google.code.findbugs:jsr305` dependencies?

Done. There are now four commits in this PR and I updated the PR description accordingly. The first two commits should be able to be cherry-picked immediately.

schildbach:
> > Actually shouldn't we, before doing anything else, add the missing `com.google.code.findbugs:jsr305` dependencies?
> 
> Done. There are now four commits in this PR and I updated the PR description accordingly. The first two commits should be able to be cherry-picked immediately.

785fe99645da187ed8d7e23a6e6ed6c80ae5cc03 doesn't cover all subprojects that miss the dependency. I see JSR305 usage in integration-tests and wallet-tool too. (base is already covered)

msgilligan:
> [785fe99](https://github.com/bitcoinj/bitcoinj/commit/785fe99645da187ed8d7e23a6e6ed6c80ae5cc03) doesn't cover all subprojects that miss the dependency. See see JSR305 usage in integration-tests and wallet-tool too. (base is already covered)

Well I made it an API dependency to cover those projects, but I'll switch it to an `implementation` dependency and add it everywhere it is used.

schildbach:
Rebased.

schildbach:
I backported the "missing JSR305 dep" bugfix to 0.17 as well.

schildbach:
Rebased.

schildbach:
> The ~fourth~ **remaining** commit upgrades to Guava 33.4.8 and to get it working I had to disable a dependency constraint in `core/build.gradle` that probably _should not_ be disabled.

Yes, I would also like to keep the constraint – because this kind of "compatibility knowledge" could be valuable once we can upgrade the reference build to later Gradle versions (if we don't switch away to Maven that is).

msgilligan:
@schildbach See:

* PR #3877 
* PR #3878 
* PR #3879 

Those 3 should supersede this one, but I'll leave this one open until those are merged.

schildbach:
Rebased.

msgilligan:
Closing in favor of:

* PR #3877
* PR #3878
* PR #3879

PR #3834, opened 2025-06-05
Convert to JSpecify

msgilligan:
The first commit migrates the dependency and includes. The second fixes the JSpecify array annotations to indicate they mean it is the array reference that is nullable not the individual array items.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/ECKey.java
@@ -66,7 +66,7 @@ import org.bouncycastle.util.encoders.Base64;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.annotation.Nullable;
+import org.jspecify.annotations.Nullable;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.math.BigInteger;
@@ -1247,9 +1247,8 @@ public class ECKey implements EncryptableItem {
         return keyCrypter != null && encryptedPrivateKey != null && encryptedPrivateKey.encryptedBytes.length > 0;
     }
 
-    @Nullable
     @Override
-    public Protos.Wallet.EncryptionType getEncryptionType() {
+    public Protos.Wallet.@Nullable EncryptionType getEncryptionType() {
         return keyCrypter != null ? keyCrypter.getUnderstoodEncryptionType() : Protos.Wallet.EncryptionType.UNENCRYPTED;
     }
 
@@ -1258,8 +1257,7 @@ public class ECKey implements EncryptableItem {
      * to be derived (for the HD key case).
      */
     @Override
-    @Nullable
-    public byte[] getSecretBytes() {
+    public byte @Nullable [] getSecretBytes() {
         if (hasPrivKey())
             return getPrivKeyBytes();
         else
--- a/core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
+++ b/core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
@@ -44,8 +44,8 @@ import org.bitcoinj.wallet.WalletProtobufSerializer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
@@ -109,7 +109,7 @@ public class WalletAppKit extends AbstractIdleService implements Closeable {
     protected InputStream checkpoints;
     protected boolean blockingStartup = true;
     protected String userAgent, version;
-    @Nonnull protected WalletProtobufSerializer.WalletFactory walletFactory = WalletProtobufSerializer.WalletFactory.DEFAULT;
+    protected WalletProtobufSerializer.@NonNull WalletFactory walletFactory = WalletProtobufSerializer.WalletFactory.DEFAULT;
     @Nullable protected DeterministicSeed restoreFromSeed;
     @Nullable protected DeterministicKey restoreFromKey;
     @Nullable protected PeerDiscovery discovery;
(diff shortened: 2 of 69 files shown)


msgilligan:
@schildbach rebased!

schildbach:
Rebased.

core/src/main/java/org/bitcoinj/crypto/ECKey.java
```diff
@@ -1247,9 +1247,8 @@ public boolean isEncrypted() {
         return keyCrypter != null && encryptedPrivateKey != null && encryptedPrivateKey.encryptedBytes.length > 0;
     }
 
-    @Nullable
     @Override
-    public Protos.Wallet.EncryptionType getEncryptionType() {
+    public Protos.Wallet.@Nullable EncryptionType getEncryptionType() {
```
schildbach, line 1251: This seems weird. How does this even compile? I'd expect

`@Nullable public Protos.Wallet.EncryptionType…`

core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
```diff
@@ -109,7 +109,7 @@ public class WalletAppKit extends AbstractIdleService implements Closeable {
     protected InputStream checkpoints;
     protected boolean blockingStartup = true;
     protected String userAgent, version;
-    @Nonnull protected WalletProtobufSerializer.WalletFactory walletFactory = WalletProtobufSerializer.WalletFactory.DEFAULT;
+    protected WalletProtobufSerializer.@NonNull WalletFactory walletFactory = WalletProtobufSerializer.WalletFactory.DEFAULT;
```
schildbach, line 112: Similar here.

core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
```diff
@@ -281,7 +281,7 @@ public WalletAppKit setUserAgent(String userAgent, String version) {
      * @param walletFactory Factory for making new wallets (Use {@link WalletProtobufSerializer.WalletFactory#DEFAULT} for default behavior)
      * @return WalletAppKit for method chaining purposes
      */
-    public WalletAppKit setWalletFactory(@Nonnull WalletProtobufSerializer.WalletFactory walletFactory) {
+    public WalletAppKit setWalletFactory(WalletProtobufSerializer.@NonNull WalletFactory walletFactory) {
```
schildbach, line 284: Similar here.

core/src/main/java/org/bitcoinj/utils/Threading.java
```diff
@@ -80,8 +80,7 @@ public static void waitForUserCode() {
      * bitcoinj library code is run, setting it after you started network traffic and other forms of processing
      * may result in the change not taking effect.
      */
-    @Nullable
-    public static volatile Thread.UncaughtExceptionHandler uncaughtExceptionHandler;
+    public static volatile Thread.@Nullable UncaughtExceptionHandler uncaughtExceptionHandler;
```
schildbach, line 83: Similar here.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -1680,7 +1680,7 @@ public RiskAnalysis.Analyzer getRiskAnalyzer() {
      * @param delay How much time to wait until saving the wallet on a background thread.
      * @param eventListener callback to be informed when the auto-save thread does things, or null
      */
-    public WalletFiles autosaveToFile(File f, Duration delay, @Nullable WalletFiles.Listener eventListener) {
+    public WalletFiles autosaveToFile(File f, Duration delay, WalletFiles.@Nullable Listener eventListener) {
```
schildbach, line 1683: Similar here.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -5409,7 +5409,7 @@ private int estimateVirtualBytesForSigning(Script script) {
      * re-organisation of the wallet contents on the block chain. For instance, in future the wallet may choose to
      * optimise itself to reduce fees or improve privacy.</p>
      */
-    public void setTransactionBroadcaster(@Nullable org.bitcoinj.core.TransactionBroadcaster broadcaster) {
+    public void setTransactionBroadcaster(org.bitcoinj.core.@Nullable TransactionBroadcaster broadcaster) {
```
schildbach, line 5412: Similar here.

schildbach:
Ah seems you fix this with your second commit. Are these two meant to be squashed?

core/src/main/java/org/bitcoinj/crypto/ECKey.java
msgilligan: The new annotations are very specific at where in the "hierarchy" the nullability exists. This type of annotation is supported by the Java compiler.

msgilligan:
> Ah seems you fix this with your second commit. Are these two meant to be squashed?

I would probably squash them. But they could be left separate as nothing in the (current) build is broken when only applying the first. There may be warnings or errors when applying static analysis that we are not currently doing, though.

schildbach:
I still see

```
integration-test/build.gradle:    implementation 'com.google.code.findbugs:jsr305:3.0.2'
```

Is `integration-test` intentionally *not* migrated?

schildbach:
> > Ah seems you fix this with your second commit. Are these two meant to be squashed?
> 
> I would probably squash them. But they could be left separate as nothing in the (current) build is broken when only applying the first. There may be warnings or errors when applying static analysis that we are not currently doing, though.

I'm trying to understand this. Maybe it still builds, but the semantics between the current (jsr305) annotation and just the first commit applied seems to have changed. If I understand right, the second commit *has* to be applied in order to restore the current semantics.

If that's right, the second commit fixes the first and consequentially I'd squash them together, hiding the fact that a mistake had been made (fixing it before merge). Do you agree?

msgilligan:
> Is `integration-test` intentionally _not_ migrated?

That must be an oversight (or maybe there was a separate PR)

Update: I looked and don't see a separate PR, must just be an oversight on my part.

msgilligan:
> If that's right, the second commit fixes the first and consequentially I'd squash them together, hiding the fact that a mistake had been made (fixing it before merge). Do you agree?

I agree.  (I think the main reason I kept them separate was so you could see what was going on.)

msgilligan:
For reference the changes to the syntax of some of the annotation is a switch to "Type-use annotation syntax".

See:

* https://jspecify.dev/docs/user-guide/#type-use-annotation-syntax
* https://jcp.org/en/jsr/detail?id=308
* https://docs.oracle.com/javase/tutorial/java/annotations/type_annotations.html

schildbach:
I just ran the reference build successfully.

schildbach (approved):

PR #3843, opened 2025-06-18
wallettool-subcommands.adoc: fix Picocli capitalization

msgilligan:
In the picocli documentation the capitalization is "picocli", but when used at the start of a sentence it is "Picocli".

So in the title of our document we should use "Picocli".

cc: @ItoroD


ItoroD (approved):
LGTM !

thanks for this @msgilligan

schildbach (approved):

PR #3848, opened 2025-06-20
WIP Threading: Attempt to hide/replace UserThread nested class

msgilligan:
This is partly an attempt to remove use of Guava's `Uninterruptibles.putUninterruptibly` and `Uninterruptibles.takeUninterruptibly`, but also to see if simplification is possibly by using a built-in JDK `Executor` rather than using a custom one.

The first and second commit could possibly be cherry-picked and merged.


schildbach:
Cherry-picked the first commit.

Can you rebase? I think we removed a good chunk of the changed methods anyway.

Issue #3840, opened 2025-06-07
Migrate build (back) to Maven


msgilligan:
It seems like Gradle is really holding us back. Given that Debian 13 ("Trixie") was unable to include a newer Gradle than 4.4.x (the Debian port is known as "FrankenGradle") despite the valiant efforts of the Debian team.

I propose we migrate to Maven 3.9.9 (or 3.8 if we still want to build with Debian 12/Bookworm)

* It is available almost everywhere and already supports JDK 25.
* We will no longer have to straddle Gradle 4.4 to Gradle 8.14 (soon to be Gradle 9)
* We can upgrade all our tests to JUnit 5 (JUnit 4 just went into maintenance mode and JUnit 6 is nearing release)

There may be other advantages to upgrading, we can add them to the list of anyone can think of some.

I don't think there are any Gradle Plugins that we are using that don't have Maven equivalents, but I think we should do a Maven Proof-of-Concept port to see if there are any issues.

msgilligan:
Gradle Plugins Used:

* Application Plugin
* Maven Publish Plugin
* Gradle ProtoBuf Plugin (`core`)
* Kotlin Plugin (Kotlin examples)
* JavaFX Plugin (`wallettemplate`)
* B.A. JLink Plugin (`wallettemplate`)
* AsciiDoctor Plugin (for rendering pico-cli man-page documentation for `wallet-tool`)
* Enterprise Gradle Plugin (activated for Build Scans™ on CI builds)

Proposed Maven Plugins (Incomplete, TBD):

* [Maven plugin to publish Gradle Module Metadata](https://github.com/gradlex-org/gradle-module-metadata-maven-plugin)
* [Asciidoctor Maven Plugin](https://docs.asciidoctor.org/maven-tools/latest/plugin/introduction/)
* [Develocity Maven Extension](https://docs.gradle.com/develocity/maven-extension/current/) (for Build Scans) -- unlike the Gradle plugin this is not Open Source!

msgilligan:
For a Gradle "composite" build of a Maven child project there seem to be a a couple options:

1. (Worst case) Publish the Maven project to the local Maven repo and include from Gradle
2. Have Gradle "exec" the Maven build
3. Maintain a simple Gradle build just for base + core that can be included from Gradle
4. Maintain a simple ANT-based build that Gradle can import
5. Use IntelliJ to do it somehow, see: https://stackoverflow.com/a/45931222
[...]

PR #3850, opened 2025-07-02
test-support/build.gradle: depend on bitcoinj-base (was bitcoinj-core)

msgilligan:
This removes a circular dependency (in Maven, not Gradle):

core (tests) -> test-support -> core (main)

Gradle is able to understand that test-support is only used by tests in core, but Maven isn't.

This is a reasonable simplification, but may also be helpful with Issue #3840

### DIFF
--- a/test-support/build.gradle
+++ b/test-support/build.gradle
@@ -3,7 +3,7 @@ plugins {
 }
 
 dependencies {
-    api project(':bitcoinj-core')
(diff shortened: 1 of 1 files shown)


schildbach (approved):

msgilligan:
Note that this did not completely solve the problem for Maven. This follow-up PR does: PR #4089

PR #3855, opened 2025-07-18
Clean up, renaming findKeyFromPubHash to findKeyFromPubKeyHash

dukeh3:
Minor clean up, but I think its better that the interfaces are consistent across the different components.

So the hash for that pubKey should be named pubKeyHash not pubHash.
And the pubKey should be named pubKey, not pubkey.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -11,7 +11,7 @@ version = '0.18-SNAPSHOT'
 dependencies {
     api project(':bitcoinj-base')
     api 'org.jspecify:jspecify:1.0.0'
-    api 'org.bouncycastle:bcprov-jdk15to18:1.84'
+    api 'org.bouncycastle:bcprov-jdk18on:1.84'
     api 'com.google.guava:guava:33.5.0-android'
     api 'com.google.protobuf:protobuf-javalite:4.33.4'
 
--- a/core/src/main/java/org/bitcoinj/crypto/ChildNumber.java
+++ b/core/src/main/java/org/bitcoinj/crypto/ChildNumber.java
@@ -74,7 +74,7 @@ public class ChildNumber implements Comparable<ChildNumber> {
      * @throws NumberFormatException when parsing fails
      */
     public static ChildNumber parse(String str) {
-        boolean isHard = str.endsWith("H");
+        boolean isHard = str.endsWith("H") || str.endsWith("'");
         String nodeNumber = isHard ?
                 str.substring(0, str.length() - 1) :
                 str;
--- a/core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
+++ b/core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
@@ -235,6 +235,18 @@ public class HDPathTest {
             new PathVector (
                 "1 / 2 / 3 /",
                 HDPath.partial(new ChildNumber(1, false), new ChildNumber(2, false), new ChildNumber(3, false))
+            ),
+
+            // BIP-32 standard apostrophe notation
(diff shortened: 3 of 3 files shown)


core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
```diff
@@ -19,14 +19,8 @@
 
 import com.google.protobuf.ByteString;
 import org.bitcoinj.base.Network;
-import org.bitcoinj.crypto.AesKey;
+import org.bitcoinj.crypto.*;
```
msgilligan: Please don't change imports except as a direct result of the actual change in the PR. If imports are going to be changed for another reason, please put that in separate PR. Note: we periodically do import cleanup PRs.

core/src/main/java/org/bitcoinj/wallet/BasicKeyChain.java
```diff
@@ -218,7 +212,16 @@ public void importKey(ECKey key) {
         }
     }
 
+    /**
+     * Use {@link BasicKeyChain#findKeyFromPubKeyHash(byte[])}
```
msgilligan: The Use line should be preceded by the `@deprecated` JavaDoc tag.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Copyright by the original author or authors.
- * 
+ *
```
msgilligan: This and all other changes unrelated to the method rename should be removed form this PR.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -601,33 +615,50 @@ public DeterministicKey markKeyAsUsed(DeterministicKey k) {
         return k;
     }
 
-    public DeterministicKey findKeyFromPubHash(byte[] pubkeyHash) {
+    /**
+     * Use {@link DeterministicKeyChain#findKeyFromPubKeyHash(byte[])}
```
msgilligan: Add `@deprecated`

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -601,33 +615,50 @@ public DeterministicKey markKeyAsUsed(DeterministicKey k) {
...
         try {
-            return (DeterministicKey) basicKeyChain.findKeyFromPubKey(pubkey);
+            return (DeterministicKey) basicKeyChain.findKeyFromPubKey(pubKey);
         } finally {
             lock.unlock();
         }
     }
 
+    /**
+     * Ues {@link DeterministicKeyChain#markPubKeyHashAsUsed(byte[])}.
```
msgilligan: Add @deprecated, also fix the "Ues" typo.

msgilligan (changes requested):
Thanks for this. We absolutely should rename this for consistency. Thanks for keeping the old methods with the `@Deprecated` annotation. You also should add the `@deprecated` JavaDoc tag before the "Use xxx" comments.

For bitcoinj, we are very insistent upon an individual, incremental change (by PR or by commit) being limited to a single change. There should be no unrelated whitespace or formatting changes.  (Maybe a fix of a typo in a JavaDoc comment you are changing anyway, but nothing beyond that.) Not only does this make PRs easier to review, it results in a very clean Git history.

Thanks!

PR #3862, opened 2025-08-04
build.gradle: add hasConsistentArchives conditional for Gradle 9+

msgilligan:
Gradle 9 removes the deprecated `dirMode` and `fileMode` settings, but dirPermissions/unix() was not added until Gradle 8.3.

This is an experiment and there might be another mechanism and/or something that was added earlier than 8.3.

We could also consider bumping the minimum supported (non-reference-build) Gradle to 8.3 or even Gradle 9.0+ (for Gradle 9+ we'd change the conditional to do nothing for the non Debian-4.4 build)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,13 +25,21 @@ allprojects {
     // Ensure standard artifacts in all projects are built reproducibly
 
     tasks.withType(AbstractArchiveTask) {
-        preserveFileTimestamps = false
-        reproducibleFileOrder = true
+        // Gradle 9+ has consistent timestamps and file order in archives by default
+        boolean hasConsistentArchives = GradleVersion.current() >= GradleVersion.version("9.0")
+        if (!hasConsistentArchives) {
+            preserveFileTimestamps = false
+            reproducibleFileOrder = true
+        }
     }
 
     tasks.withType(Jar) {
-        dirMode = 0755
-        fileMode = 0644
+        // Gradle 9+ has consistent permissions in jars by default
+        boolean hasConsistentArchives = GradleVersion.current() >= GradleVersion.version("9.0")
+        if (!hasConsistentArchives) {
+            dirMode = 0755
+            fileMode = 0644
+        }
     }
 
     tasks.withType(Javadoc) {


msgilligan:
GitLab is passing: https://gitlab.com/bitcoinj/bitcoinj/-/pipelines/1965558926

schildbach:
I wonder what exactly fails the reproducibility if these lines are removed?

msgilligan:
> I wonder what exactly fails the reproducibility if these lines are removed?

Are there permission bits in the JARs?  (I forget)  I'm pretty sure there are and setting them explicitly might just be a safety mechanism to make sure they don't change accidentally.

msgilligan:
Note that for Gradle 9+ setting these properties can be removed as the defaults are now correct: https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#reproducible_archives_by_default

msgilligan:
We could support reproducible builds _only_ for reference builds and Gradle 9+ and that might simplify this PR a little.

schildbach:
~~I propose we make this three cases (instead of the two in this PR):~~

~~1. Gradle < 8.3~~
~~2. Gradle < 9~~
~~3. Gradle 9+ (new case: just use defaults)~~

~~Yes, this will add even more code. But it would be self-documenting, and at some point we can drop case 2.~~

schildbach:
Hmm wait.

If we can use dirMode/fileMode up to (excluding) Gradle 9, and Gradle 9 does the right thing anyway, then we can simplify this to one check – right?

schildbach:
Basically, change `hasDirPermission` to `hasDirMode` (Gradle <9) and guard the current (as of master) workaround with that condition. All other cases would not need to be considered, right?

msgilligan:
~~So you're saying we only want reproducible builds on Debian-Gradle 4.4 and Gradle 9+?~~

Nevermind. I understand what you mean. Let me fix it.

schildbach:
> So you're saying we only want reproducible builds on Debian-Gradle 4.4 and Gradle 9+?

No, I wasn't saying that (though I'm not ruling it off either).

I was saying that below Gradle 9, our current workaround does its job of working towards that aspect of reproducibility.

Gradle 9 and above does the right thing already (as per your comment, didn't verify this), so we don't need the workaround any more.

Am I missing something?

msgilligan:
I understand now. Good plan. I will fix it.

schildbach:
We can readjust later if reproducibility problems surface. (As it can always happen.)

msgilligan:
I just force-pushed a simple rebase.  (This will be followed shortly by an implementation of your suggestion.)

schildbach (approved):

PR #3863, opened 2025-08-06
wallettemplate/build.gradle: update jlink plugin to 3.1.2

msgilligan:
3.1.2 supports Gradle 9


msgilligan:
It looks like upgrading to BA JLink Plugin 3.1.2 breaks our build if Gradle is less than 8.11. (I ran tests locally and pushed extra commits testing various minimum Gradle versions in our matrix, to determine that 8.11.1 works, but 8.9 and earlier fail.)

Update: There is now a BA JLink Plugin 3.1.3 that should resolve this issue, see: https://github.com/beryx/badass-jlink-plugin/issues/306#issuecomment-3166494621

PR #3877, opened 2025-09-14
core, integration-test: improve guava android constraint

msgilligan:
Instead of forcing usage of the Android version of Guava on the core `implementation` classpath, do it on the `testImplementation` classpath of both `core` and `integration-test`.

This makes sure our tests run against the Android version, which is what we really want. But with this change we can use the JRE version in other modules, such as the wallettemplate module.

### DIFF
--- a/integration-test/build.gradle
+++ b/integration-test/build.gradle
@@ -1,3 +1,5 @@
+import org.gradle.util.GradleVersion
+
 plugins {
     id 'java'
     id 'java-library'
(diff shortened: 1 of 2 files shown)


integration-test/build.gradle
```diff
@@ -19,6 +21,21 @@ dependencies {
     testRuntimeOnly "org.junit.vintage:junit-vintage-engine:5.11.4"
 }
 
+// Prevent newer Gradle from switching to JRE version of Guava for building/running integration tests
+// We want our tests to run on the lowest-common denominator
+def gradleVersionTargetJVM = GradleVersion.version("7.0")
+if (GradleVersion.current() > gradleVersionTargetJVM) {
+    dependencies.constraints {
+        implementation("com.google.guava:guava") {
```
schildbach, line 30: According to the commit comment, this should be `testImplementation()` too?

msgilligan: Since this module is a test module and nothing will depend on it, I deliberately chose `implementation` so the Android version would be used even in the regular compile phase of this module. ~~But the comment is still accurate because it is a test module~~ [when I wrote this I thought I had updated the commit comment along with the code comments and I was mistaken.]

I can see how this might look like a mistake though. It wouldn't hurt to change it to `testImplementation` as I don't think there is any non-test code to compile. Alternatively I could update the comment to clarify this.

msgilligan: You are correct, the _commit comment_ was wrong. I have improved the code comments and fixed the commit comment.

schildbach:
Rebased.

PR #3882, opened 2025-09-20
wallettemplate: upgrade to JavaFX 25 (and JDK 23)

msgilligan:
JavaFX 25 requires JDK 23+, so required Java version is bumped to 23.


schildbach:
CI seems unhappy.

msgilligan:
> CI seems unhappy.

That's because I made the wallettemplate module build conditional on JDK 23, but the GHA `build.gradle` file unconditionally calls `jlink` and `installDist` in that module. 

I just force-pushed an attempt to resolve that issue.

msgilligan:
Splitting the `jlink` and `installDist` Gradle tasks into a separate, conditional GHA step fixed the issue.

schildbach:
What do we gain from JavaFX 25, and is it worth it trying to get it into 0.18 by introducing another build-time conditional?

Can it wait till 0.19, which will (maybe, no promises) require JDK 25?

msgilligan:
In my experience (and Sparrow somewhat confirms this) JavaFX apps need to be shipped with a recent (normally the latest LTS release) of JavaFX for best results, so we should try to update to the latest JavaFX with every release of bitcoinj.

JavaFX is also changing its supported JDK policy so that for JavaFX version N, they will only support JDK N, N-1, and N-2. In particular, I think they are starting to use FFM for _new_ or rewritten calls to the native GUI APIs.

I think build-time conditionals that include/exclude an entire module are much easier to manage and reason about than the various Gradle feature tests we have, so I am fine with requiring Java 25 for `wallettemplate` in 0.18.

schildbach:
17→25 is a huge jump, skipping 21 (also an LTS). I might be wrong, but 0.17 wasn't even able to build with 25 and now we're requiring it already.

We probably never talked about this, but I prefer to avoid situations where users are forced to update their software at the very same time when updating to the next version of out software. If possible, there should be a path where users can update their software first, still using our old version, and then – potentially much later – update to our new version without changing anything else.

On top of that, I like the fact we require 17 uniformly on all modules atm. But this is kind of personal taste…

So I propose moving this to 0.19.

msgilligan:
> 17→25 is a huge jump, skipping 21 (also an LTS).

JavaFX is (effectively) making the same jump, see the [Roadmap](https://gluonhq.com/products/javafx/). JavaFX 21 (LTS) requires JDK 17 and JavaFX 25 (LTS) requires JDK 23 (which is effectively JDK 25 if you want to use an LTS version.)

They have changed their policy and from now on, JavaFX `n` will require JDK `n - 2` or later. So this means we will have to make a big jump just like they did if we want to keep up. I'd say it is better to make the big jump now, rather than later. Their new policy (if you are targeting LTS releases) means that Java FX `n` LTS, requires JDK `n` LTS.

I think part of their reason for doing this is that FFM will really help them with making native UI library calls and the other reason is a reflection of two realities:

1. Most production-quality JavaFX applications these days **bundle a JRE** (e.g. Sparrow)
2. Most production-quality JavaFX applications **regularly update** to the latest versions (e.g. Sparrow)


> I prefer to avoid situations where users are forced to update their [JDK] software at the very same time when updating to the next version of our software.

I prefer to do it this way, too. However, for `wallettemplate` _only_ there are compelling reasons to make an exception:

1. JavaFX itself has skipped JDK 21 (see above) and we need to keep up (or catch up.)
2. JavaFX 21 is effectively unmaintained for `aarch64-linux` (I hope this won't happen with JavaFX 25...)
3. (I'm pretty sure that) Nobody is using `wallettemplate` for anything other than experimentation.
4. The wallettemplate app is _not_ (yet) structured as _a library_ that someone will have a Maven dependency on, so if they are doing anything with it they will have copy/pasted it into a new project. So an update to `wallettemplate` in our repository will not affect them.
5. The `walletemplate` is _not_ a real-world _application_ with uses other than testing bitcoinj. 
6. I would like `walletemplate` to become more useful and I think being on the latest FX/JDK will help us get there quicker. (By having a better feature-set and saving us from spending time testing and supporting earlier versions/JDKs.)

To the best of knowledge, We have never heard from anyone that has based an actual application on `wallettemplate` and would even be considering updating to a newer version of our source (I tried back in 2019, but gave up.) I want to get there, but we are not there yet.

> I like the fact we require 17 uniformly [for compiling] on all modules atm

I agree that is preferable in most cases.

> So I propose moving this to 0.19.

If you don't find my reasons above compelling enough, I would be OK with that. We also don't have to decide right now.

In the meantime I have two new PRs:

* PR #3913 
* PR #3914 

3913 permanently resolves issues we've been having with keeping explicit tasks (i.e. `module:tasks` and in this particular case the `installDist` task)  in CI YAML files in sync with modules that require specific JDK versions. It also make sure we call `installDist` for every module that supports it (we were missing `wallettool` in some cases.

3914 bumps `wallettemplate` to JDK 21 while leaving JavaFX version unchanged, so that would allow us to meet your criteria and upgrade to JavaFX 25 and JDK 25 in 0.19.

msgilligan:
Gluon says they will be releasing `aarch64-linux` binaries for all upcoming releases:

> Yes, we plan to build [aarch64-linux binaries] for all upcoming releases, but not supported by our commercial LTS.

https://foojay.social/@gluonhq@techhub.social/115458123913123657

msgilligan:
Rebased and force-pushed.

msgilligan:
Updated to JavaFX 25.0.1 and force-pushed.

msgilligan:
@schildbach If we want to support aarch64-linux, we pretty much need to bump to JavaFX 25. Did you get a chance to read these two comments:

* https://github.com/bitcoinj/bitcoinj/pull/3882#issuecomment-3336772108
* https://github.com/bitcoinj/bitcoinj/pull/3882#issuecomment-3462295716

?

schildbach:
Will merge once the tests are all green.

msgilligan:
> Will merge once the tests are all green.

"Merge queue" would do this automatically for you...

PR #3891, opened 2025-09-21
WIP: build with zulu and JDK 25 in the Matrix

msgilligan:
Unit test are failing on JDK 25 because EasyMock needs an upgrade, see: PR #3892

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -9,13 +9,18 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-24.04, macOS-13, windows-2022]
-        java: ['17', '21', '23']
+        java: ['17', '21', '23', '25']
         distribution: ['temurin']
         gradle: ['8.5', '8.14.3', '9.1.0']
         exclude:
           # Java 23+ requires Gradle 8.10+
           - java: '23'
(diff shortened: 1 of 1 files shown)


schildbach:
Rebased.

schildbach:
Why are we keeping JDK 23? It's not LTS, right?

schildbach:
Will temurin be added shortly? Can we wait for it, rather than switching to zulu?

msgilligan:
> Why are we keeping JDK 23? It's not LTS, right?

I want to keep 23 until we resolve the Temurin 24+/JLink issue. But, yes, the goal is to eventually remove 23 from the test matrix.

> Will temurin be added shortly? Can we wait for it, rather than switching to zulu?

I think Temurin 25 will be available within a week or so. And I think it is fine to wait for it.  (And use the time to try to solve the Temurin/JLink issue.)

Marking this PR as _DRAFT_.

msgilligan:
@schildbach See PR #3894 we can replace JDK 23 with 25 in our `graal.yml` builds.

msgilligan:
Rebased now that PR #3894 is merged.

msgilligan:
Closing in favor of PR #3903

PR #3896, opened 2025-09-22
build.gradle: Use -PtestMinJdk=8 instead of testJdk8

msgilligan:
Rather than use a boolean that runs hardcoded JDK8 tests, use a numeric parameter to set the minimum JDK to test on.

This makes it easier to change our minimum version and it also makes it possible to run the tests on the minimum available JDK on a particular platform or really any JDK you want.

### DIFF
--- a/base/build.gradle
+++ b/base/build.gradle
@@ -38,23 +38,22 @@ test {
 }
 
 def gradleVersionToolchains = GradleVersion.version("6.7")
+// Activate if `testMinJdk` is a integer > 0 in `gradle.properties` or set with `-P` on the command-line
+def minJdkVersion = Integer.valueOf((String) findProperty('testMinJdk') ?: "0")
 
-if (GradleVersion.current() > gradleVersionToolchains) {
+if (minJdkVersion > 0 && GradleVersion.current() > gradleVersionToolchains) {
     // If the Gradle Java Toolchains feature is available, run tests on older JDKs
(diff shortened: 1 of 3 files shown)


base/build.gradle
```diff
@@ -38,23 +38,22 @@ test {
 }
 
 def gradleVersionToolchains = GradleVersion.version("6.7")
+// Activate if `testMinJdk` is a integer > 0 in `gradle.properties` or `-PtestMinJdk=true` is on command-line
```
schildbach: > `-PtestMinJdk=true`

Really? I thought that argument is now an integer?

core/build.gradle
```diff
@@ -85,23 +85,22 @@ test {
 }
 
 def gradleVersionToolchains = GradleVersion.version("6.7")
+// Activate if `testMinJdk` is a integer > 0 in `gradle.properties` or `-PtestMinJdk=true` is on command-line
```
schildbach: Same here.

base/build.gradle
msgilligan: Good catch. I'll fix the comment.

msgilligan: Fixed and force-pushed.

core/build.gradle
msgilligan: Fixed and force-pushed.

PR #3898, opened 2025-09-22
gradle.yml: Set MIN_JDK: '9'

msgilligan:


msgilligan:
This is obsolete now with PR #3962. If we decide to use a minimum of JDK 9 we will do it differently than in this PR.

PR #3900, opened 2025-09-22
GHA gradle.yml: Temurin 23->24, remove jlink target

msgilligan:
Since the jlink plugin is broken with Temurin 24+ and we aren't  really using the jlink build for anything, let's turn jlink off so we can move off of EOL JDKs towards LTS JDK 25.

As a first step, let's move to Temurin 24 until Temurin 25 arrives.


msgilligan:
Closing in favor of PR ##3903 -- Temurin 25 is now available on GHA

PR #3901, opened 2025-09-22
gha gradle.yml: add separate JLink step for JDK 21, Gradle 9.1 only

msgilligan:


msgilligan:
~~I'm making this _DRAFT_~~: We ~~should probably~~ can eventually use the approach of PR #3913 for `jlink`, too. (i.e. always run the `jlink` task on the main `gradle build` step)

msgilligan:
Closing in favor of PR #3918

PR #3906, opened 2025-09-22
add flake.nix to setup a simple nix devshell

msgilligan:
The devshell using graalvm-ce to provide both a standard JDK and native image support.

Because Nixified paths confuse native-image's CheckToolchain, we will disabled it.

Among other advantages, this will provide a mechanism to get Gradle 9.1+ on Debian.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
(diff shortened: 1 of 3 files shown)


msgilligan:
I removed the "priority high" label from this one and am instead suggesting the slightly simpler, non-graalvm-supporting PR #3927 that provides a Zulu 25 JDK.

PR #3556, opened 2025-01-16
gradle.yml, graalvm.yml: add ubuntu-24.04-arm to matrix

schildbach:

PR #3908, opened 2025-09-25
gradle.yml, graalvm.yml: add ubuntu-24.04-arm to the build matrix

msgilligan:


schildbach:
Are these two commits related? I'd consider merging the second ahead if/when it builds.

msgilligan:
> Are these two commits related?

The first is a prerequisite for the second. We were unable to merge PR #3556 because JavaFX is not getting updates on aarch64-linux. Alternatively, we could downgrade JavaFX to 21.0.1 before adding aarch64-linux runners, but as you know I prefer to move to the latest JavaFX.

schildbach:
See discussion in https://github.com/bitcoinj/bitcoinj/pull/3882 because this seems to be the first commit, and more controversal than the second.

schildbach:
Rebased.

msgilligan:
Hmm. It's not working despite the aarch64 linux components being available for download here:
https://gluonhq.com/products/javafx/
(The aarch64 versions are listed as "unsupported", so maybe that means they're not on Maven Central?)

I also recall getting a response from Gluon on Mastodon saying they would be providing aarch64 builds of 25.0.x going forward.

So, I marked this as "WIP" until we find out what is going on.

msgilligan:
OK, JavaFX 25.0.2 has been available on Maven Central for some time and we have already upgraded to it.

But we have also combined our two GitHub Actions workflows/files to single workflow/file. So I'm closing this PR in favor of PR #4169 (which is a child of PR #4168)

PR #3910, opened 2025-09-25
CryptoUtils: Add Charset param to hmacSha512 method that takes a string

msgilligan:
~~Make the usage of Bouncy Castle HMacSHA512 a private implementation detail.~~ This PR was originally two commits, the first one making the core implementation private, the second doing some simplification. This PR now contains the second commit plus some additional improvements.

[The already merged commit] should allow PR #3791 to use `CryptoUtils` for hmacSha512, though this PR still contains a change to explicitly specify the CharSet (typically `US_ASCII`) which BIP85 requires.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/HDKeyDerivation.java
+++ b/core/src/main/java/org/bitcoinj/crypto/HDKeyDerivation.java
@@ -23,6 +23,7 @@ import org.bouncycastle.math.ec.ECPoint;
 
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.security.SecureRandom;
 import java.util.Arrays;
 import java.util.function.Supplier;
@@ -64,7 +65,7 @@ public final class HDKeyDerivation {
         checkArgument(seed.length > 8, () ->
                 "seed is too short and could be brute forced");
         // Calculate I = HMAC-SHA512(key="Bitcoin seed", msg=S)
-        byte[] i = CryptoUtils.hmacSha512("Bitcoin seed", seed);
+        byte[] i = CryptoUtils.hmacSha512("Bitcoin seed", StandardCharsets.US_ASCII, seed);
         // Split I into two 32-byte sequences, Il and Ir.
         // Use Il as master secret key, and Ir as master chain code.
         checkState(i.length == 64, () ->
--- a/core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
+++ b/core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
@@ -22,6 +22,7 @@ import org.bouncycastle.crypto.macs.HMac;
 import org.bouncycastle.crypto.params.KeyParameter;
 import org.bouncycastle.jcajce.provider.digest.SHA3;
 
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 
@@ -69,12 +70,15 @@ public class CryptoUtils {
     }
 
     /**
-     * Generate a MAC using a {@link String} key and data
-     * @param key The key
+     * Generate a MAC using a {@link String} tag and data. For example, this is used in our BIP-32
+     * implementation with a tag {@code String} of {@code "Bitcoin seed"}. The {@link Charset} parameter
+     * is typically {@link StandardCharsets#US_ASCII}.
+     * @param tag The tag
+     * @param charset The {@link Charset} to be used to encode the tag {@code String}
      * @param data The message data to process
      * @return The final result of the MAC operation
      */
-    public static byte[] hmacSha512(String key, byte[] data) {
-        return hmacSha512(key.getBytes(), data);
+    public static byte[] hmacSha512(String tag, Charset charset, byte[] data) {
+        return hmacSha512(tag.getBytes(charset), data);
     }
 }


schildbach:
Rebased.

core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
```diff
@@ -70,4 +70,8 @@ public static byte[] hmacSha512(HMac hmacSha512, byte[] input) {
     public static byte[] hmacSha512(byte[] key, byte[] data) {
         return hmacSha512(createHmacSha512Digest(key), data);
     }
+
```
schildbach: This new method could use a JavaDoc.

core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
```diff
@@ -70,4 +70,8 @@ public static byte[] hmacSha512(HMac hmacSha512, byte[] input) {
...
+    public static byte[] hmacSha512(String key, byte[] data) {
+        return hmacSha512(createHmacSha512Digest(key.getBytes()), data);
```
schildbach: Should we perhaps specify a Charset?

schildbach:
But I'll merge anyway, since this is just a refactoring.

schildbach:
I merged the first commit. For the second I'm a bit too tired atm. Maybe an inbetween JavaDoc commit would help (-:

core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
msgilligan: Agreed. Will add.

core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
msgilligan: We should. I will add that in a separate commit.

msgilligan:
This PR needs to be merged for PR #3791 to be rebased.

schildbach:
I merged 77dfe6f29f0e10305381df6a77e285225b0958b8, hope this will allow other PRs depending on this to be rebased.

Regarding the `byte[] hmacSha512(String key, byte[] data)` variant: I wonder if we should perhaps remove that method. We ourselves only have one usage, and US_ASCII would be fine (and more efficient) for that usage. However, other users who might be carelessly switching between default locales (Java versions, platform switches) could run into issues, with or without the change to consistently use UTF-8.

That, and I dislike the concept of storing keys in strings.

schildbach:
I quickly scanned BIP-85 and it seems they have a very similar usage of "String keys":

```
HMAC-SHA512(key="bip-entropy-from-k", msg=k)
```

But they fail to specify the encoding. Again, due to the fixed String constant it would not matter if US_ASCII or UTF-8 was used, but other encodings could yield incompatible results – UTF-16 for example? As indicated in my above comment, I tend to leave the String→byte[] conversion to the caller. What do you think?

schildbach:
(BIP32 also doesn't specify the encoding for the "Bitcoin seed" constant.) :cry:

schildbach:
An alternative to removing the method might be to keep it, but only allow Strings that yield the exact same bytes for various default platform encodings now and in past (tbd, but probably Android, Windows, Linux).

Or, yet another alternative, only allow US ASCII characters because that seems to mirror the intention in BIP32 and BIP85.

schildbach:
In any case, I'd perhaps clarify in the JavaDoc that the method is not meant to be used with arbitrary keys (e.g. user passwords). Rather, I think it's about constants specified in specs.

msgilligan:
I rebased and force-pushed.

We have (at least) four options:

1. Leave the code as is (with default conversion) but improve the Javadoc.
2. Merge this PR with the code as is (with hardcoded UTF8 conversion) but improve the Javadoc first.
3. Use something like `s.chars().allMatch(c -> c <= 127)` to enforce only US-ASCII strings
4. Change the `String` version of `hmacSha512()` to require a `CharSet` parameter

We should probably rename the variable `key` to `string` or `tag` or something like that to reflect the BIP-85 and BIP-32 use cases.

I think I prefer (4) -- force the caller to specify. It's an internal call and will be rarely used, so having the extra parameter is ok, I think.

schildbach:
I merged the JavaDoc commit as it describes the status quo.

I'm leaning towards discouraging the `public static byte[] hmacSha512(String key, byte[] data)` method to be used as a generic usecase. That implies a slight preference of the above option 3. And yes, renaming the `key` to something that reflects the use usecase, and maybe even mentioning the two existing usecases in the JavaDoc could help.

If we want to keep the method generic, then I'd pick 4 from the above options.

msgilligan:
I also renamed the PR now that the implementation details (well, mostly) have been hidden via commits cherry-picked from this PR.

msgilligan:
@schildbach This is rebased and updated to use option 4 (above) -- requiring a `Charset` parameter. This keeps the method generic (although it is considered internal and should not be widely used) and I prefer it to option 3 which would require a runtime check that could potentially throw an (unchecked) exception.

core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
```diff
@@ -69,12 +70,15 @@ public static byte[] hmacSha512(byte[] key, byte[] data) {
     }
 
     /**
-     * Generate a MAC using a {@link String} key and data
-     * @param key The key
+     * Generate a MAC using a {@link String} tag and data. For example, this is used in our BIP-32
+     * implementation with a tag {@code String} of {@code "Bitcoin seed"}. The {@link Charset} parameter
+     * is typically {@link StandardCharsets#UTF_8}.
```
schildbach: I would argue that this is usually US_ASCII, as all the usages we identified are restricted to that range. I would really like to discourage usage outside of the US_ASCII range.

core/src/main/java/org/bitcoinj/crypto/HDKeyDerivation.java
```diff
@@ -64,7 +65,7 @@ public static DeterministicKey createMasterPrivateKey(byte[] seed) throws HDDeri
         checkArgument(seed.length > 8, () ->
                 "seed is too short and could be brute forced");
         // Calculate I = HMAC-SHA512(key="Bitcoin seed", msg=S)
-        byte[] i = CryptoUtils.hmacSha512("Bitcoin seed", seed);
+        byte[] i = CryptoUtils.hmacSha512("Bitcoin seed", StandardCharsets.UTF_8, seed);
```
schildbach: Likewise, I'd change this to US_ASCII, even if in this case it would make no functional difference.

schildbach (commented):
See my comments. This is almost ready to merge. I'm ok with not forcing the US_ASCII-range, but at least we should discourage users from going outside of it via documentation.

core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
msgilligan: Fixed and force-pushed.

core/src/main/java/org/bitcoinj/crypto/HDKeyDerivation.java
msgilligan: Fixed and force-pushed.

PR #3911, opened 2025-09-25
CryptoUtils: Use built-in Java HmacSHA512

msgilligan:
~~This is a child of PR #3910 and will need a rebase after that one is merged.~~ Rebased.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
+++ b/core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
@@ -17,17 +17,16 @@ package org.bitcoinj.crypto.internal;
 
 import org.bitcoinj.base.Sha256Hash;
 import org.bouncycastle.crypto.digests.RIPEMD160Digest;
-import org.bouncycastle.crypto.digests.SHA512Digest;
-import org.bouncycastle.crypto.macs.HMac;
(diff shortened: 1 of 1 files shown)


schildbach:
Rebased.

core/src/main/java/org/bitcoinj/crypto/internal/CryptoUtils.java
```diff
@@ -59,13 +60,16 @@ public static byte[] digestRipeMd160(byte[] input) {
...
      */
     public static byte[] hmacSha512(byte[] key, byte[] data) {
-        SHA512Digest digest = new SHA512Digest();
-        HMac hMac = new HMac(digest);
-        hMac.init(new KeyParameter(key));
-        hMac.update(data, 0, data.length);
-        byte[] out = new byte[64];
-        hMac.doFinal(out, 0);
-        return out;
+        // HmacSHA512 is built-in on Java and Android for all versions we support
```
schildbach: This comment would be more useful if it would state the actual versions HmacSHA512 is built-in. Because if our supported versions change we would need to investigate from the beginning.

For Android I just found out [it supports HmacSHA152 since API level 1](https://developer.android.com/reference/kotlin/javax/crypto/Mac).

msgilligan: Fixed.

PR #3915, opened 2025-09-26
.gitlab-ci.yml, build.Containerfile: run `publishToMavenLocal` target on every subproject

schildbach:
`publishToMavenLocal` is a no-op on subprojects that don't use the `maven-publish` plugin, but runs on every subproject that does.


msgilligan (approved):

schildbach:
Pipeline on GitLab: https://gitlab.com/bitcoinj/bitcoinj/-/pipelines/2065709491

PR #3919, opened 2025-09-27
wallettemplate/build.gradle: add mergedModule requires

msgilligan:
The mergedModule is using and _should_ `require`:

* `java.logging`
* `jdk.unsupported`
* `org.slf4j`

but wallettemplate has worked fine without it until we run with Temurin JDK 24 or 25 (which requires the BA Jlink Plugin 3.1.14 to build) and then there are fatal exceptions at runtime when these modules are accessed.

This fix should work with all currently supported JDK versions and distributions.


schildbach (changes requested):
I'm fine with the change, but I'm struggling to understand the commit description. If you say "the mergedModule" what module is that? Is it "wallettemplate"?

I understand the commit is a fix to something. I think in the first line it would be good to mention what is being fixed – just a few words (one line). The "how" (add requires) goes into the longer description, if it's not obvious from the diff already.

schildbach:
Ah and isn't "This fix should work with all currently supported JDK versions and distributions." valid for all our commits? And in that sense it's redundant?

schildbach:
Rebased.

msgilligan:
The PR description was updated (on Sep 28) to address @schildbach 's questions (earlier on Sep 28)

Issue #2309, opened 2022-03-05
GitHub Actions: protobuf test failures due to `i` replaced by `ı`


schildbach:
Since a few days, our GitHub actions builds sometimes throw the following test failures. Here is a sample, but there are more of them, all very similar:

```
org.bitcoinj.wallet.DeterministicKeyChainTest > watchingSegwitChain FAILED
    org.junit.ComparisonFailure: expected:<...34\200H\366\312P"
      [issued_subkeys: 2
      lookahead_size: 10
      path: -2147483647
      path: 0
      sigs_required_to_spend: 1
    }
    public_key: "\002\237\202\357O\355r\020(]\235\b\3162\263N\331\320\216\246\262\376\260\352\224\313Bw\347~\310\336X"
    type: DETERMINISTIC_KEY

    deterministic_key {
      chain_code: "\336m6\270\332/\363\016\340=;\003\225\341S\024\262\332vM\201L\207\342\306=!\352A\f\272{"
      i]ssued_subkeys: 1
      l...> but was:<...34\200H\366\312P"
      [ıssued_subkeys: 2
      lookahead_size: 10
      path: -2147483647
      path: 0
      sigs_required_to_spend: 1
    }
    public_key: "\002\237\202\357O\355r\020(]\235\b\3162\263N\331\320\216\246\262\376\260\352\224\313Bw\347~\310\336X"
    type: DETERMINISTIC_KEY

    deterministic_key {
      chain_code: "\336m6\270\332/\363\016\340=;\003\225\341S\024\262\332vM\201L\207\342\306=!\352A\f\272{"
      ı]ssued_subkeys: 1
      l...>
        at org.junit.Assert.assertEquals(Assert.java:117)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at org.bitcoinj.wallet.DeterministicKeyChainTest.checkSerialization(DeterministicKeyChainTest.java:771)
        at org.bitcoinj.wallet.DeterministicKeyChainTest.watchingSegwitChain(DeterministicKeyChainTest.java:546)
```

If comparing the "expected" and the "but was" parts, the only difference is the `i` in "issued_subkeys" is replaced by `ı`, the dotless `i`. Note the `i` in "sigs_required_to_spend" isn't changed this way, though it should be derived the same way – both are names of fields in our protobuf.

Afaik the output is generated by the protobuf library, so I'd investigate into that first.

msgilligan:
[...]

PR #3920, opened 2025-09-27
wallettool: add module-info.java, ignore when Gradle < 8.5

msgilligan:
`wallettool` is built with JDK 17 and should have a `module-info.java`

### DIFF
--- a/wallettool/build.gradle
+++ b/wallettool/build.gradle
@@ -7,6 +7,7 @@ plugins {
     id 'org.graalvm.buildtools.native' version '0.11.0' apply false
 }
 
+def isModularBuild = GradleVersion.current() >= GradleVersion.version("8.5")
 def annotationProcessorMinVersion = GradleVersion.version("4.6")
 boolean hasAnnotationProcessor = GradleVersion.current() >= annotationProcessorMinVersion
 def junit5MinVersion = GradleVersion.version("4.6")
@@ -39,6 +40,10 @@ tasks.withType(JavaCompile) {
     options.compilerArgs.addAll(['--release', '17'])
     options.compilerArgs << '-Xlint:deprecation'
     options.encoding = 'UTF-8'
+    if (!isModularBuild) {
+        // Exclude the module-info.java file because Gradle 4.x doesn't handle module-info.java well.
+        source = source.filter { file -> !file.path.endsWith('module-info.java') }
+    }
 }
 
 javadoc.options.encoding = 'UTF-8'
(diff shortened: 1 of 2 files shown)


msgilligan:
The Debian Gradle builds are having a little trouble compiling `wallettool` with `module-info`. This PR will need a little more work to configure Debian Gradle 4.4  to compile a modular `wallettool`.

schildbach:
I'm fine with adding a module-info, but the special casing in build.gradle makes it appear unattractive to me. What do we gain from a modular wallet-tool? And is it worth the added complexity?

msgilligan:
> What do we gain from a modular wallet-tool?

A modular wallet-tool can be packaged with `jlink` which makes distribution and installation easier by bundling a known and minimized JRE. `jlink` is often used by packaging systems when packaging Java apps.

The module system is part of the [System Integrity by Default](https://openjdk.org/jeps/8305968) initiative. It enforces module boundaries at compile- and run-time which can prevent or warn about some egregious behaviors.

For example my testing in the last few days revealed that **ProtoBuf** is using `sun.misc.Unsafe` and we were forced to declare/permit it. We are now assured that other 3rd-party libraries we using are _not_ using `Unsafe`. And the module system will prevent any dependencies from using it in the future without our permission. One of the scariest bugs we've had [GitHub Actions: protobuf test failures due to i replaced by ı](https://github.com/bitcoinj/bitcoinj/issues/2309) may have been related to ProtoBuf's usage of `sun.misc.Unsafe`.

The module system will also prevent modules from using introspection to look into modules they should not be accessing, etc. For example, we added the following line in `wallet-tools`'s `module-info.java`:

```
    opens org.bitcoinj.wallettool to info.picocli;
```

This gives `picocli` permission to use reflection on the `wallettool` module, but does not give it permission to introspect `core` or `base` or any other module. So this alerts us and documents what `picocli` is doing and limits its scope to what is necessary.

And the quality improvements we get from using modules will help us improve and modularize `bitcoinj-core` and `bitcoinj-base` as well.

> And is it worth the added complexity?

~~I think it could be, but at this point I don't know exactly what changes will be required. Let's make that decision one I have something that works.~~ Yes, see below.

msgilligan:
> > And is it worth the added complexity?
> 
> I think it could be, but at this point I don't know exactly what changes will be required. Let's make that decision one I have something that works..

The added complexity is minimal: we just tell the `JavaCompile` task to ignore `module-info.java` when Gradle < 8.5.

In my opinion the benefits of a modular build of `wallet-tool` are worth this minimal added complexity.

PR #3922, opened 2025-09-28
wallettool/build.gradle: set mainModule in application block

msgilligan:
~~This is a child of #3920 but is independent of the use of the BA JLink plugin.~~ (rebased)

Sets the `mainModule` property in the `application` block signifying that the application should be modular. This means that the existing `installDist` task will build a launch script that sets up the module path.

When applications run on the module path they have stricter integrity guarantees. E.g. if a module is accessing sun.misc.unsafe (in module jdk.unsupported) this will result in a warning or an error. It was by running wallet-tool as a module that I discovered that ProtoBuf is doing this. So enforcing these checks will help us make our modules better as well as warn us about bad behavior in 3rd-party modules. See also the discussion/rationale in Issue #3928.

### DIFF
--- a/wallettool/build.gradle
+++ b/wallettool/build.gradle
@@ -55,6 +55,10 @@ if (GradleVersion.current() > GradleVersion.version("8.0")) {
     application {
         mainClass = walletToolMainClassName
         applicationName = walletToolName
+        if (isModularBuild) {
+            // setting mainModule tells installDist to create a launch script that uses module path
+            mainModule = "org.bitcoinj.wallettool"
+        }
     }
 } else {
     mainClassName = walletToolMainClassName


wallettool/build.gradle
```diff
@@ -55,6 +55,11 @@ if (GradleVersion.current() > GradleVersion.version("8.0")) {
     application {
         mainClass = walletToolMainClassName
         applicationName = walletToolName
+        // isModularBuild is not strictly necessary here because we're already Gradle 8+
+        if (isModularBuild) {
```
schildbach: If it's not necessary, then why use it? I'm not against it, but I'd like to know why.

msgilligan: When applications run on the module path they have stricter integrity guarantees.  E.g. if a module is accessing `sun.misc.unsafe` (in module `jdk.unsupported`) this will result in a warning or an error. It was by running wallet-tool as a module that I discovered that ProtoBuf is doing this. So enforcing these checks will help us make our modules better as well as warn us about bad behavior in 3rd-party modules.

msgilligan: @schildbach I also copied the above explanation into the PR description.

msgilligan: There is also some explanation in Issue #3928.

schildbach: Sorry, I meant the conditional `if (isModularBuild)`. Why use it if it's not necessary? For example, in wallettemplate we don't have a conditional around the "mainModule" declaration either.

msgilligan: Looking at it with fresh eyes, it is necessary. Because if `isModularBuild` is set to false (for some reason) then setting `mainModule` is wrong.

My comment saying it was "unnecessary" actually meant that in our _current_ build configurations `isModularBuild` will never be false at this point in the code.

msgilligan: I think the comment should just be removed.  (I think an alternative comment is unnecessary, it would just say something like "if this is a modular build, configure the application appropriately" and we already have a comment explaining how that setting affects the `installDist` task.

msgilligan: Good catch! Updated and force-pushed.

PR #3924, opened 2025-09-29
TestWithNetworkConnections, TestWithPeerGroup: handle nulls in tearDown()/stopPeerServer()

msgilligan:
These two null checks prevent `DefaultMultiCauseException: Multiple Failures` during the intermittent `BindException: Address already in use: bind` we are seeing. They make the tearDown a little more orderly.


schildbach:
Rebased.

PR #3796, opened 2025-04-27
ClientConnectionManager: Deprecate inheritence of Guava `Service`

msgilligan:
This contains a preliminary commit to simplify the JavaDoc of `ClientConnectionManager` before updating the comment as part of the main commit.

PR #3926, opened 2025-09-29
ClientConnectionManager: BUGFIX: stop() call stopAsync() not startAsync()

msgilligan:
This fixes a bug introduced by PR #3796.

Looks like I authored the bad commit in April of this year and it got merged on July 19, 2025.

I guess we didn't find it until now because it shows up intermittently and looks like existing intermittent failures.


schildbach (approved):

schildbach:
Nice catch!

msgilligan:
> Nice catch!

I felt like I was tilting at windmills or chasing "decoy" bugs on the "Address is in use" issue, but I did find this one "real" bug.

PR #3949, opened 2025-10-02
GitHub spam handling script

msgilligan:
With a little help from AI I quickly wrote this script to mitigate spam issues and comments.

Apparently, it is not possible
to automate reporting spam or deleting spam issues.

But this script closes, locks, and labels the issues and prs as well as blocking users from the current user account and the organization.

You need to run the script from the bitcoinj local repo.

./scripts/gh-handle-spam.sh <issue-or-pr-number>

### DIFF
--- a/scripts/gh-handle-spam.sh
+++ b/scripts/gh-handle-spam.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+# gh-handle-spam <issue-or-pr-number>
+# Requires: gh CLI + a PAT with user:write scope if blocking users
+
+set -euo pipefail
+
+if [ $# -ne 1 ]; then
+  echo "Usage: gh-handle-spam <issue-or-pr-number>"
+  exit 1
+fi
+
+NUM="$1"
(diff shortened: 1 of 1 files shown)


schildbach:
First of all, thanks for automating this. I'm not sure if I want to add the script to this repo since it's generic to all projects of GitHub and isn't related to bitcoinj. But we can keep it around on a branch for now until we've found a home.

Some tools react to GitHub users setting a label. For example, we could set the spam label manually and then the rest happens magically. How do projects do this? Can it be done as a GitHub action or something else that runs at GitHub?

msgilligan:
> Some tools react to GitHub users setting a label. For example, we could set the spam label manually and then the rest happens magically. How do projects do this? Can it be done as a GitHub action or something else that runs at GitHub?

We should be able to do that with a GitHub Action.

We'll see if this wave of spam goes away or if GitHub provides us with better built-in tools to handle it. If it continues to be a problem and we need to continue rolling our own solution we could try a GitHub Action approach.

PR #3865, opened 2025-08-08
base,core/build.gradle: make jspecify an api dependency

msgilligan:
Nullability annotations are part of the API for a library module. Make sure the Gradle metadata reflects this.

PR #3956, opened 2025-11-05
modular-architecture.md: two minor updates

msgilligan:
Two commits:

* core, not base depends on SLF4J (see PR #3957)
* JSpecify dependency is `api` not `implementation` (see PR #3865)

### DIFF
--- a/designdocs/modular-architecture.md
+++ b/designdocs/modular-architecture.md
@@ -88,7 +88,7 @@ flowchart TD
     SECPFFM --> SECPC[libsecp256k1]
     CORE --> P[ProtoBuf]
     CORE .-> S[SLF4J]
-    BASE .-> JS[JSpecify]
+    BASE --> JS[JSpecify]
 
 classDef external fill:#999;
 class G,S,JS,BC,SECPC,P external;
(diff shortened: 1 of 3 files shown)


designdocs/modular-architecture.md
```diff
@@ -87,7 +87,7 @@ flowchart TD
     SECPBC --> BC[Bouncy Castle]
     SECPFFM --> SECPC[libsecp256k1]
     CORE --> P[ProtoBuf]
-    BASE .-> S[SLF4J]
```
schildbach: At the moment, we've got these dependendies on slf4j:

```
$ grep -R slf4j base/
base/src/main/java/org/bitcoinj/base/internal/PlatformUtils.java:import org.slf4j.Logger;
base/src/main/java/org/bitcoinj/base/internal/PlatformUtils.java:import org.slf4j.LoggerFactory;
base/src/main/java/org/bitcoinj/base/package-info.java: * has no dependencies other than {@code slf4j-api} and {@code jspecify}.
base/build.gradle:    implementation 'org.slf4j:slf4j-api:2.0.16'
base/build.gradle:    testImplementation 'org.slf4j:slf4j-jdk14:2.0.16'
```

So I think it's too early to merge this commit.

msgilligan: You know I mixed up `secp-api` with `bitcoinj-base`. I originally thought `secp-api` would have an `slf4j-api` dependency, but then realized we didn't really need it. I somehow thought that as also the case with `bitcoinj-base`. But I think we can remove that dependency. Let me make a PR.

msgilligan: Actually, this document specifies what we are targeting for 0.18, not documenting what we have. So we don't have to fix it first.

msgilligan: Fixed in PR #3957

schildbach:
I have cherry-picked the "slf4j" commit and rebased.

PR #3957, opened 2025-11-06
base module: add UNKOWN runtime and os, remove use of slf4j

msgilligan:
Note that this also eliminates a nullable static field.

Update: and it allows us to make `os` and `runtime` into `final` fields, so I added a commit to do that.

### DIFF
--- a/base/src/main/java/org/bitcoinj/base/internal/PlatformUtils.java
+++ b/base/src/main/java/org/bitcoinj/base/internal/PlatformUtils.java
@@ -35,7 +35,7 @@ public class PlatformUtils {
 
     static {
         String runtimeProp = System.getProperty("java.runtime.name", "").toLowerCase(Locale.US);
-        if (runtimeProp.equals(""))
+        if (runtimeProp.isEmpty())
             runtime = PlatformUtils.Runtime.UNKNOWN;
         else if (runtimeProp.contains("android"))
             runtime = PlatformUtils.Runtime.ANDROID;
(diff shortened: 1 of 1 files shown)


schildbach:
I cherry-picked the removal of SLF4J changes (in two commits) and rebased.

base/src/main/java/org/bitcoinj/base/internal/PlatformUtils.java
```diff
@@ -30,35 +30,39 @@ public enum OS {
...
-    public static Runtime runtime = PlatformUtils.Runtime.UNKNOWN;
-    public static OS os = PlatformUtils.OS.UNKNOWN;
+    public static final Runtime runtime;
+    public static final OS os;
 
     static {
         String runtimeProp = System.getProperty("java.runtime.name", "").toLowerCase(Locale.US);
-        if (runtimeProp.equals(""))
-            PlatformUtils.runtime = PlatformUtils.Runtime.UNKNOWN;
+        Runtime runtimeEnum;
```
schildbach: Why the indirection via another local variable? (two occurences)

msgilligan: Fixed and force-pushed. (I also split the `.equals("")` to `.isEmpty()` conversion ins a separate conversion.

I think the reason I did it this way was an attempt to match a pattern that the IDE can detect for conversion to a switch expression, but I'm not sure that will work for this particular if/else sequence, we are many versions away form that feature, and as you imply the code is much cleaner without the intermediate variable.

schildbach (approved):

PR #3960, opened 2025-12-08
GHA graavm.yml, gradle.yml: Use macOS-15-intel

msgilligan:
macOS-13 has been dropped by GitHub.

PR #3961, opened 2025-12-08
GHA gradle.yml: Install JDK 8 via setup-java action

msgilligan:
We previously relied on it being already installed on the GitHub runner, but we need to upgrade the macOS runner to macOS-15-intel which does not have JDK 8 built-in.

Note that since we have configured setup-java to use the Temurin distro, JDK 8 will now be the Temurin build rather than whatever was included with the runner.

Note that this PR does not fix the macOS builds (which still rely on the removed `macOS-13` runner) but is a standalone commit that when applied installs Temurin JDK 8 on _all_ runners and can be demonstrated to work correctly. PR #3960 can be applied later/separately to migrate from `macOS-13` to `macOS-15-intel`

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -19,6 +19,9 @@ jobs:
           - java: '25'
             gradle: '8.14.3'
       fail-fast: false
+    env:
+      # Use env variable to specify the JDK version to use for the `testMinJdk` Gradle option
+      MIN_JDK: '8'
     name: ${{ matrix.os }} JDK ${{ matrix.java }}.${{ matrix.distribution }} Gradle ${{ matrix.gradle }}
     steps:
       - name: Git checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle.yml
```diff
@@ -27,7 +27,9 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: ${{ matrix.distribution }}
-          java-version: ${{ matrix.java }}
+          java-version: |
+            8
```
schildbach: I think it would be useful to add a comment about why this ominous "8" is need.

msgilligan: I had a comment, but I think it broke the YAML so I removed it. I actually think it would be better to create a named variable for "minJdk" or something similar. This will help us to add macOS aarch64 support (see PR #3456)

msgilligan: Fixed and rebased.

PR #3962, opened 2025-12-08
GHA gradle.yml: add macOS-15 (aarch64) using testMinJdk=11

msgilligan:
~~This is a child of PR #3960~~ Rebased.


schildbach (approved):

schildbach:
I kept the old comment, because it's describes the meaning of MIN_JDK rather than the conditional (which everybody can read in the line below).

PR #3968, opened 2025-12-11
wallettool module-info.java: add o.b.core transitive dependencies

msgilligan:
Add transitive dependencies of bitcoinj-core that are not strictly needed for the build, but prevent errors/warnings in IDEs (e.g. IntelliJ)

These lines can be removed when bitcoinj-core gets a module-info.

This now a child of PR #3973 which also helps IDEA.

### DIFF
--- a/wallettool/src/main/java/module-info.java
+++ b/wallettool/src/main/java/module-info.java
@@ -22,6 +22,11 @@ module org.bitcoinj.wallettool {
     requires info.picocli;
 
     requires org.bitcoinj.core;
+    // Since bitcoinj-core doesn't have a module-info and IDEs are generally unaware of the
+    // processing done by the jlink plugin, we include transitive dependencies here that would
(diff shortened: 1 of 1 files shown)


ItoroD:
Can we have these merged? As I saw the errors on my end after I did a rebase. Its true that its not strictly needed and things will still run. But I think its tidy that we add it.

msgilligan:
> Can we have these merged? As I saw the errors on my end after I did a rebase. Its true that its not strictly needed and things will still run. But I think its tidy that we add it.

I made this draft (and removed the "high-priority" label) because I was still seeing (new, different) errors in IDEA even with this PR. The solution, I believe, is to apply PR #3973 so I will rebase this PR on that one.

schildbach (approved):

schildbach:
Oops, sorry I have missed that this PR was demoted to draft and merged anyway.

Issue #3969, opened 2025-12-12
`wallet-tool` throws `DateTimeException` on `rotate` when `--date=YYYY-MM-DD` is provided


msgilligan:
If the `date` parameter is provided (in `YYYY-MM-DD` format) to the `rotate` command, a `DateTimeException` is thrown by the line:

```
            rotationTime = Instant.from(date);
```

This is because `date` is a `LocalDate` which can't be converted to an `Instant` by `Instant.from()`.

The example command and output is:

```
$ ./wallettool/build/install/wallet-tool/bin/wallet-tool rotate --chain=regtest --wallet=newregtest-wallet --date=2025-12-11
java.time.DateTimeException: Unable to obtain Instant from TemporalAccessor: 2025-12-11 of type java.time.LocalDate
	at java.base/java.time.Instant.from(Instant.java:381)
	at org.bitcoinj.wallettool.WalletTool.rotate(WalletTool.java:536)
	at org.bitcoinj.wallettool.WalletTool.call(WalletTool.java:449)
	at org.bitcoinj.wallettool.WalletTool.call(WalletTool.java:100)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2045)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2465)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2457)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2419)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2277)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2421)
	at picocli.CommandLine.execute(CommandLine.java:2174)
	at org.bitcoinj.wallettool.WalletTool.main(WalletTool.java:319)
Caused by: java.time.temporal.UnsupportedTemporalTypeException: Unsupported field: InstantSeconds
	at java.base/java.time.LocalDate.get0(LocalDate.java:700)
	at java.base/java.time.LocalDate.getLong(LocalDate.java:680)
	at java.base/java.time.Instant.from(Instant.java:376)
	... 12 more
```

To reproduce from scratch use:

```
./wallettool/build/install/wallet-tool/bin/wallet-tool create --net=regtest --wallet=regtest-wallet.wallet
[...]

PR #3970, opened 2025-12-12
WalletTool: Fix `date` to `rotationTime` conversion.

msgilligan:
`date` is a `LocalDate` which can't be converted to an `Instant` by `Instant.from()`

By using `atStartOfDay()` with a timezone, we can use `toInstant()`.

This fixes Issue #3969.

### DIFF
--- a/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
+++ b/wallettool/src/main/java/org/bitcoinj/wallettool/WalletTool.java
@@ -533,7 +533,7 @@ public class WalletTool implements Callable<Integer> {
         // Set a key rotation time and possibly broadcast the resulting maintenance transactions.
         Instant rotationTime = TimeUtils.currentTime();
         if (date != null) {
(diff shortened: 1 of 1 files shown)


schildbach:
Some context: First, the `date` option was converted from Date to LocalDate via bb8179082438f9ab3e3d22628987f8b6a7ce0a0d. Maybe that already introduced a slight incorrectness by using `ZoneOffset.UTC` while previously the system default time was used?

Then, via commit de9c39b35a4ce9e181e8ab2672f14362623d092e `Instant.from()` was used which indeed can't work.

schildbach (approved):

schildbach:
This PR changes the time zone back to system default. Which I think is fine because the `date` option didn't work at all from March 2023 to this day. The behaviour from pre-March 2023 should now be restored.

At some point, we should go over the methods to provide dates and times in WalletTool, I guess. I never gave it much priority because a couple of hours are usually not a big deal in the blockchain…

PR #3971, opened 2025-12-12
WalletTool: add @Nullable and check with NullAway/ErrorProne

msgilligan:
~~There are many issues remaining, this is just a partial attempt to investigate what is needed for the current `WalletTool` to be @NullMarked and checked with NullAway/ErrorProne.~~

The scope of this PR has been reduced. It no longer adds `@NullMarked` and instead just adds `@Nullable` fields and turns on  NullAway/ErrorProne checking going forward.

### DIFF
--- a/wallettool/build.gradle
+++ b/wallettool/build.gradle
@@ -1,4 +1,5 @@
 import org.gradle.util.GradleVersion
+import net.ltgt.gradle.errorprone.CheckSeverity
 
 plugins {
     id 'java'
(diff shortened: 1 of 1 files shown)


msgilligan:
Making WalletTool @NullMarked will require significant work. We should try this as part of or after @ItoroD's work.

Update: Additionally, we could start adding more nullability annotations without @Nullmarked or compile-time enforcement.

msgilligan:
This is superseded by PR #4132 which enables checking in the root build.gradle for all projects.

msgilligan:
Closing, now that PR #4132 (which turns on ErrorProne/NullAway for all subprojects) has been merged.

PR #3976, opened 2025-12-24
wallettemplate: target JDK 25

schildbach:
We would like to use JavaFX 25 which requires JDK 23 or later.

This is broken out from #3882 but includes necessary changes to the readme.


msgilligan (approved):
Approved! 💯🎉

PR #3979, opened 2025-12-26
build.yml: merge gradle.yml and graalvm.yml, add build_all_green aggregation job

msgilligan:


.github/workflows/build-gradle.yml
hasskell, line 1: Does it make sense to rename this to `build-gradle.yml` since we can consolidate same for maven and have two builds 
- `build-gradle.yml`
- `build-maven.yml`

msgilligan: Good idea. Updated and force-pushed.

.github/workflows/build-gradle.yml
```diff
@@ -66,3 +66,41 @@ jobs:
...
+        run: gradle test nativeCompile --init-script build-scan-agree.gradle --scan --info --stacktrace
+      - name: Upload wallet-tool as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: wallet-tool-${{ matrix.os }}-${{ matrix.java }}-gradle-${{ matrix.gradle }}
+          path: wallettool/build/native/nativeCompile/wallet-tool
+
+  build_all_green:
+    runs-on: ubuntu-24.04
+    needs: [build, build_graalvm]
```
hasskell, line 104: Super nit: you could also change `build` to `build-gradle` with dashes `-` instead of underscore and have same dash pattern for jobs similar what I did for maven build 

```
build-maven-all-green:
    runs-on:
      ubuntu-24.04 needs: [build-maven, build-graalvm]
      steps:
        - run: echo "All upstream jobs passed"
```

msgilligan: I agree that we should be consistent, but maybe we should consistently use `_`.

I used `_` mostly because that is what we are currently using in `secp256k1-jdk` and I patterned this PR after that repository. I am open to using `-`, but then we should change it in `secp256k1-jdk`.

The [example](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-jobs#example-creating-jobs) in the GitHub documentation for setting job ID uses `_`, but the community uses both `-` and `_`.  I found this _unresolved_ issue asking for a convention: https://github.com/orgs/community/discussions/39547

@schildbach do you have a preference?

hasskell: Sure thing since majority is already using `_` let me just quickly change this in maven so it is consistent with current version

schildbach: > [@schildbach](https://github.com/schildbach) do you have a preference?

Underscore is fine with me, but I also agree we should be consistent.

msgilligan: Looks, like we're going with **underscore**.

PR #3986, opened 2026-01-07
Replace Guava Hasher with MessageDigest in CheckpointManager

### LINKED ISSUE #3985: Replace Guava  `com.google.common.hash.Hasher` with `Sha256Hash` "builder"
`CheckpointManager` is using `com.google.common.hash.Hasher`.

This can be replaced with something like:

```
Sha256Hash.Digest digest = Sha256Hash.builder();
for (data : dataList) {
    digest.update(data);
}
Sha256Hash hash = digest.digest();
```

This is fairly simply to add and may be useful elsewhere.

nadavramon:
Fixes #3985

**Description**
Replaced the Guava `com.google.common.hash.Hasher` with the standard Java `MessageDigest` (via `Sha256Hash.newDigest()`) in `CheckpointManager`.

**Changes**
- Refactored `readTextual` to use `MessageDigest.update()` loop.
- Verified that the hash logic produces the correct output for existing checkpoints.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/CheckpointManager.java
+++ b/core/src/main/java/org/bitcoinj/core/CheckpointManager.java
@@ -16,9 +16,6 @@
 
 package org.bitcoinj.core;
 
-import com.google.common.hash.HashCode;
-import com.google.common.hash.Hasher;
-import com.google.common.hash.Hashing;
 import com.google.common.io.BaseEncoding;
 import org.bitcoinj.base.Sha256Hash;
 import org.bitcoinj.base.internal.TimeUtils;
@@ -116,7 +113,7 @@ public class CheckpointManager {
     }
 
     private Sha256Hash readTextual(InputStream inputStream) throws IOException {
-        Hasher hasher = Hashing.sha256().newHasher();
+        MessageDigest digest = Sha256Hash.newDigest();
         try (BufferedReader reader =
                      new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.US_ASCII))) {
             String magic = reader.readLine();
@@ -127,11 +124,12 @@ public class CheckpointManager {
                 reader.readLine(); // Skip sigs for now.
             int numCheckpoints = Integer.parseInt(reader.readLine());
             checkState(numCheckpoints > 0);
+
             // Hash numCheckpoints in a way compatible to the binary format.
-            hasher.putBytes(ByteBuffer.allocate(4).order(ByteOrder.BIG_ENDIAN).putInt(numCheckpoints).array());
+            digest.update(ByteBuffer.allocate(4).order(ByteOrder.BIG_ENDIAN).putInt(numCheckpoints).array());
             for (int i = 0; i < numCheckpoints; i++) {
                 byte[] bytes = BASE64.decode(reader.readLine());
-                hasher.putBytes(bytes);
+                digest.update(bytes);
                 ByteBuffer buffer = ByteBuffer.wrap(bytes);
                 StoredBlock block;
                 if (bytes.length == StoredBlock.COMPACT_SERIALIZED_SIZE)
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/core/CheckpointManager.java
```diff
@@ -127,11 +120,12 @@ private Sha256Hash readTextual(InputStream inputStream) throws IOException {
                 reader.readLine(); // Skip sigs for now.
             int numCheckpoints = Integer.parseInt(reader.readLine());
             checkState(numCheckpoints > 0);
-            // Hash numCheckpoints in a way compatible to the binary format.
-            hasher.putBytes(ByteBuffer.allocate(4).order(ByteOrder.BIG_ENDIAN).putInt(numCheckpoints).array());
+
```
msgilligan, line 127: I think the comment from line 130 still applies and should not be removed.

core/src/main/java/org/bitcoinj/core/CheckpointManager.java
```diff
@@ -32,19 +29,15 @@
 import org.jspecify.annotations.Nullable;
 import java.io.BufferedInputStream;
 import java.io.BufferedReader;
-import java.io.DataInputStream;
```
msgilligan, line 35: Please don't remove unused imports in the same _commit_ as the change. It is nice to be able see which imports were added and removed by the change without the noise of removing unused commits.

You can have 2 commits in the PR one that removes the unused imports and the other that implements the hashing change.

nadavramon:
@msgilligan Thanks for the feedback! 
I've restored the binary format comment and put the unused DataInputStream import back to keep the diff clean. 
Ready for re-review.

core/src/main/java/org/bitcoinj/core/CheckpointManager.java
```diff
@@ -36,15 +33,12 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.nio.Buffer;
```
msgilligan, line 39: This removal is unrelated to the change, as well.

core/src/main/java/org/bitcoinj/core/CheckpointManager.java
```diff
@@ -36,15 +33,12 @@
...
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.charset.StandardCharsets;
-import java.security.DigestInputStream;
```
msgilligan, line 43: This removal is unrelated to the change, as well.

core/src/main/java/org/bitcoinj/core/CheckpointManager.java
```diff
@@ -36,15 +33,12 @@
...
 import java.security.MessageDigest;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.util.Arrays;
```
msgilligan, line 47: This removal is unrelated to the change, as well.

msgilligan (changes requested):
Apparently, I was a little unclear in my previous review - apologies! I commented on the first line that removed an unused import unrelated to the change, but there were 4 in total. IDEs automatically remove them and it is a common gotcha, but we make an effort to have each commit apply a single, easily-reviewable change.

Also, you made fixes in an add-on commit, so this will need to be rebased before it can be merged. We can do it as part of the merge, but normally when contributors update PRs they squash and rebase (or in more complex cases use an interactive rebase and make changes to multiple commits in a single PR.)

nadavramon:
@msgilligan I've restored the remaining unused imports (Arrays, Buffer, DigestInputStream) so they no longer show as removed in the diff. 
I also squashed everything into a single commit as requested.
The only deleted imports now are the Guava ones (com.google.common.hash.*), 
which are directly related to this change.

PR #3988, opened 2026-01-07
WIP: add o.b.btc.BtcBlock and other interfaces

msgilligan:
This is WIP on an attempt to create some "universal" interfaces for key Bitcoin types. This is to allow special case implementations (e.g. FFM-based "flat" representations of Block, etc.) and interoperability with other implementations like Sparrow Wallet's **Drongo** (which is actually a fork of bitcoinj)

These interfaces would also help us with our immutability efforts, I think.

There are two major interoperability uses for these interfaces:

1. Methods and classes that need data from (for example) a "block" object could use the interface which would allow those methods to take any implementing class.
2. Classes that implement these interfaces could have copy constructors/factories that take the interface. For example: `public static Block of(BtcBlock block)`.

### DIFF
--- a/base/src/main/java/org/bitcoinj/btc/BtcSha256.java
+++ b/base/src/main/java/org/bitcoinj/btc/BtcSha256.java
@@ -1,0 +1,15 @@
+package org.bitcoinj.btc;
+
+import java.math.BigInteger;
+
+/**
+ *
+ */
+public interface BtcSha256 {
+    // Bitcoin serialization fornat (reversed)
+    byte[] serialize();
+    // Returns a hex string
+    String toString();
+    BigInteger toBigInteger();
+    byte[] toByteArray();
+}
--- a/base/src/main/java/org/bitcoinj/btc/StoredHeader.java
+++ b/base/src/main/java/org/bitcoinj/btc/StoredHeader.java
@@ -1,0 +1,11 @@
+package org.bitcoinj.btc;
+
+import java.math.BigInteger;
+
+// Extra fields for bitcoinj StoredBlock (or equivalent)
+// TODO: Do we want this?
+// TODO: Refactor bitcoin StoredBlock and BlockStore to use this.
+public interface StoredHeader {
+    BigInteger chainWork();
+    int height();
+}
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -25,6 +25,7 @@ import org.bitcoinj.base.Sha256Hash;
 import org.bitcoinj.base.VarInt;
 import org.bitcoinj.base.internal.Buffers;
 import org.bitcoinj.base.internal.TimeUtils;
+import org.bitcoinj.btc.BtcTransaction;
 import org.bitcoinj.core.LockTime.HeightLock;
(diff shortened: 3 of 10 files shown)


base/src/main/java/org/bitcoinj/btc/BtcSha256.java
schildbach, line 1: Why do we need an interface to this super-simple data class? Will there ever be other implementations?

base/src/main/java/org/bitcoinj/btc/StoredHeader.java
schildbach, line 1: I'd probably try to avoid this class, or at least not include it in a first stab.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -322,6 +323,10 @@ public Transaction() {
         vLockTime = LockTime.unset();
     }
 
+    public byte[] serialize() {
+        return write(ByteBuffer.allocate(messageSize())).array();
+    }
+
```
schildbach, line 357: This seems redundant, we do have a default `Message.serialize()` already.

schildbach:
We have this "immutable block/transaction" vs "block/transaction builder" idea. How do the interfaces from this PR relate? Or is this out of scope of this PR?

base/src/main/java/org/bitcoinj/btc/BtcSha256.java
msgilligan: I think there is a need for an implementation that uses a "slice" `MemorySegment` that points into a parent "MemorySegment". I'm also thinking the **Drongo** Sha256 type could implement this interface.

msgilligan:
> I have the vague feeling the BtcBlock, BtcBlockHeader, BtcTransaction could be useful. Perhaps a unit test would make this feeling a certainty? Or some migrated code within bitcoinj?

I think there are several places where migrated code in bitcoinj will prove the value of these interfaces, such as adding a `FinalizedTransaction` class that is immutable. Stay tuned, this is WIP...

> (I strongly dislike the "Btc" prefix and package name, but no worries during the concept phase.)

I'm open to suggestions. Since `Block`, etc. is already used in `core` we either have to rename those or use something different here. I was considering abbreviated names like `Blk`, `BlkHeader`, and `Tx`, but figure you'd like that even less.

msgilligan:
> We have this "immutable block/transaction" vs "block/transaction builder" idea. How do the interfaces from this PR relate? Or is this out of scope of this PR?

These are a prerequisite for that and I would say that it is out of scope for the code in this PR, but in scope for a group of PRs including this one.

core/src/main/java/org/bitcoinj/core/Transaction.java
msgilligan: Yeah, this was a workaround to a compiler issue (as I recall it was because `Message.serialize()` is a `default` method). There is probably a better solution.

base/src/main/java/org/bitcoinj/btc/StoredHeader.java
msgilligan: I'd like to leave it in for now because it is an important use case and something I would like to consider. It can be moved to a separate PR later.  (and/or implemented in a different module)

msgilligan:
I think the next major step for this PR is to add the component fields of transaction.

PR #3989, opened 2026-01-08
Script.writeBytes: Improve API and JavaDoc

msgilligan:
A sequence of commits to improve the API (parameter types and names, exception error message) and JavaDoc.

This is mainly in preparation for deprecating the method and replacing it with a ByteBuffer version. See DRAFT PR #4085

### DIFF
--- a/core/src/main/java/org/bitcoinj/script/Script.java
+++ b/core/src/main/java/org/bitcoinj/script/Script.java
@@ -41,7 +41,6 @@ import org.jspecify.annotations.Nullable;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -325,23 +324,26 @@ else if (ScriptPattern.isP2TR(this))
     ////////////////////// Interface for writing scripts from scratch ////////////////////////////////
 
     /**
-     * Writes out the given byte buffer to the output stream with the correct opcode prefix
+     * Writes a byte array to an output stream with the correct "push" opcode prefix
      * To write an integer call writeBytes(out, Utils.reverseBytes(Utils.encodeMPI(val, false)));
+     * @param os ByteArrayOutputStream to write to
+     * @param bytes byte array to prefix with push opcode and write
+     * @throws IOException shouldn't happen when using ByteArrayOutputStream
```
hasskell, line 331: It also throws RuntimeException when arrays is too long

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -325,23 +324,26 @@ else if (ScriptPattern.isP2TR(this))
...
+            os.write(bytes);
+        } else if (bytes.length < 65536) {
             os.write(OP_PUSHDATA2);
-            ByteUtils.writeInt16LE(buf.length, os);
-            os.write(buf);
+            ByteUtils.writeInt16LE(bytes.length, os);
+            os.write(bytes);
         } else {
-            throw new RuntimeException("Unimplemented");
+            throw new RuntimeException("byte array too long");
```
hasskell: Should you also add actual length if this ever happen think it is important datapoint to understand what is actual length

msgilligan: Good suggestion! Fixed.

core/src/main/java/org/bitcoinj/script/Script.java
msgilligan: Good point. But in general,  we don't document unchecked exceptions in JavaDoc. Partly because they are easy to get out of sync and partly because you generally are not catching them.

hasskell (approved):

schildbach (approved):

PR #4000, opened 2026-01-10
Refactor: Replace Guava MoreObjects with Java 8 StringJoiner globally

nadavramon:
This PR replaces all usages of Guava's MoreObjects.toStringHelper with standard Java 8 StringJoiner or string concatenation.

Contributes to #2000

Changes:

* Replaced MoreObjects.toStringHelper(this) with new StringJoiner(...) in complex classes (e.g., Peer, DeterministicKey).

* Replaced simple concatenations in classes like BloomFilter.

* Removed com.google.common.base.MoreObjects imports from all modified files.

Verification:

* gradlew :bitcoinj-core:test passed successfully.

* Verified toString() output logic remains consistent (null handling, formatting).

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/BloomFilter.java
+++ b/core/src/main/java/org/bitcoinj/core/BloomFilter.java
@@ -16,10 +16,10 @@
 
 package org.bitcoinj.core;
 
-import com.google.common.base.MoreObjects;
 import org.bitcoinj.base.Sha256Hash;
 import org.bitcoinj.base.internal.Buffers;
 import org.bitcoinj.base.internal.ByteUtils;
+import org.bitcoinj.core.internal.ToStringUtil;
 import org.bitcoinj.crypto.ECKey;
 import org.bitcoinj.script.Script;
 import org.bitcoinj.script.ScriptChunk;
@@ -157,11 +157,11 @@ public class BloomFilter implements Message {
 
     @Override
     public String toString() {
-        final MoreObjects.ToStringHelper helper = MoreObjects.toStringHelper(this).omitNullValues();
-        helper.add("data length", data.length);
-        helper.add("hashFuncs", hashFuncs);
-        helper.add("nFlags", getUpdateFlag());
-        return helper.toString();
+        return new ToStringUtil(this)
+                .add("data length", data.length)
+                .add("hashFuncs", hashFuncs)
+                .add("nFlags", getUpdateFlag())
+                .toString();
     }
 
     @Override
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -16,11 +16,13 @@
 
 package org.bitcoinj.core;
 
-import com.google.common.base.MoreObjects;
+import com.google.common.base.Strings;
+import com.google.common.base.Throwables;
 import org.bitcoinj.core.internal.GuardedBy;
 import org.bitcoinj.base.Coin;
 import org.bitcoinj.base.Sha256Hash;
 import org.bitcoinj.base.internal.TimeUtils;
+import org.bitcoinj.core.internal.ToStringUtil;
 import org.bitcoinj.core.listeners.AddressEventListener;
 import org.bitcoinj.core.listeners.BlocksDownloadedEventListener;
 import org.bitcoinj.core.listeners.ChainDownloadStartedEventListener;
@@ -404,15 +406,14 @@ public class Peer extends PeerSocketHandler {
 
     @Override
     public String toString() {
-        final MoreObjects.ToStringHelper helper = MoreObjects.toStringHelper(this).omitNullValues();
-        helper.addValue(getAddress());
-        helper.add("version", vPeerVersionMessage.clientVersion);
-        helper.add("subVer", vPeerVersionMessage.subVer);
-        if (vPeerVersionMessage.localServices.hasAny())
-            helper.add("services", vPeerVersionMessage.localServices.toString());
-        helper.add("time", TimeUtils.dateTimeFormat(vPeerVersionMessage.time));
-        helper.add("height", vPeerVersionMessage.bestHeight);
-        return helper.toString();
+        return new ToStringUtil(this)
+                .addValue(getAddress())
+                .add("version", vPeerVersionMessage.clientVersion)
+                .add("subVer", vPeerVersionMessage.subVer)
+                .addIf(vPeerVersionMessage.localServices.hasAny(), "services", vPeerVersionMessage.localServices)
+                .add("time", TimeUtils.dateTimeFormat(vPeerVersionMessage.time))
+                .add("height", vPeerVersionMessage.bestHeight)
+                .toString();
     }
 
     @Override
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -17,7 +17,8 @@
 
 package org.bitcoinj.core;
 
-import com.google.common.base.MoreObjects;
+import com.google.common.math.IntMath;
+import org.bitcoinj.core.internal.ToStringUtil;
 import org.bitcoinj.base.Address;
 import org.bitcoinj.base.Coin;
 import org.bitcoinj.base.Network;
@@ -27,6 +28,7 @@ import org.bitcoinj.base.internal.Buffers;
 import org.bitcoinj.base.internal.TimeUtils;
 import org.bitcoinj.core.LockTime.HeightLock;
 import org.bitcoinj.core.LockTime.TimeLock;
+import org.bitcoinj.core.internal.ToStringUtil;
 import org.bitcoinj.crypto.AesKey;
 import org.bitcoinj.base.internal.ByteUtils;
 import org.bitcoinj.core.TransactionConfidence.ConfidenceType;
@@ -724,9 +726,9 @@ public class Transaction implements Message {
 
     @Override
     public String toString() {
-        MoreObjects.ToStringHelper helper = MoreObjects.toStringHelper(this);
-        helper.addValue(toString(null, null));
-        return helper.toString();
+        return new ToStringUtil(this)
+                .addValue(toString(null, null))
+                .toString();
     }
 
     /**
--- a/core/src/main/java/org/bitcoinj/core/internal/ToStringUtil.java
+++ b/core/src/main/java/org/bitcoinj/core/internal/ToStringUtil.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright by the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bitcoinj.core.internal;
+
+import java.util.StringJoiner;
+
+/**
+ * A utility to help construct {@code toString} strings, replacing Guava's {@code MoreObjects.ToStringHelper}.
+ * <p>
+ * This class is internal and not part of the public API.
+ */
+public class ToStringUtil {
+    private final StringJoiner joiner;
+
+    /**
+     * @param instance The object instance to format
+     */
+    public ToStringUtil(Object instance) {
+        this.joiner = new StringJoiner(", ", instance.getClass().getSimpleName() + "{", "}");
+    }
+
+    /**
+     * Adds a name/value pair to the formatted output.
+     * If the value is null, it is ignored (mimicking Guava's omitNullValues).
+     *
+     * @param name  The name of the property
+     * @param value The value of the property
+     * @return this instance to support fluent chaining
+     */
+    public ToStringUtil add(String name, Object value) {
+        if (value != null) {
+            joiner.add(name + "=" + value);
+        }
+        return this;
+    }
+
+    /**
+     * Adds a name/value pair to the formatted output.
+     *
+     * @param name  The name of the property
+     * @param value The value of the property
+     * @return this instance to support fluent chaining
+     */
+    public ToStringUtil add(String name, long value) {
+        joiner.add(name + "=" + value);
+        return this;
+    }
+
+    /**
+     * Adds an unnamed value to the formatted output.
+     *
+     * @param value The value to add
+     * @return this instance to support fluent chaining
+     */
+    public ToStringUtil addValue(Object value) {
+        joiner.add(String.valueOf(value));
+        return this;
+    }
+
+    /**
+     * Adds an unnamed value to the formatted output only if the condition is true.
+     *
+     * @param condition The condition to check
+     * @param value     The value to add if condition is true
+     * @return this instance to support fluent chaining
+     */
+    public ToStringUtil addIf(boolean condition, String value) {
+        if (condition) {
+            joiner.add(value);
+        }
+        return this;
+    }
+
+    /**
+     * Adds a name/value pair to the formatted output only if the condition is true.
+     *
+     * @param condition The condition to check
+     * @param name      The name of the property
+     * @param value     The value to add if condition is true
+     * @return this instance to support fluent chaining
+     */
+    public ToStringUtil addIf(boolean condition, String name, Object value) {
+        if (condition) {
+            add(name, value);
+        }
+        return this;
+    }
+
+    @Override
+    public String toString() {
+        return joiner.toString();
+    }
+}
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -17,7 +17,6 @@
 
 package org.bitcoinj.crypto;
 
-import com.google.common.base.MoreObjects;
 import com.google.common.primitives.UnsignedBytes;
 import org.bitcoinj.base.Network;
 import org.bitcoinj.base.ScriptType;
@@ -25,6 +24,7 @@ import org.bitcoinj.base.internal.ByteUtils;
 import org.bitcoinj.base.Base58;
 import org.bitcoinj.core.NetworkParameters;
 import org.bitcoinj.base.Sha256Hash;
+import org.bitcoinj.core.internal.ToStringUtil;
 import org.bitcoinj.crypto.internal.CryptoUtils;
 import org.bouncycastle.math.ec.ECPoint;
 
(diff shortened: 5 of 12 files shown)


core/src/main/java/org/bitcoinj/core/BloomFilter.java
```diff
@@ -16,7 +16,7 @@
 
 package org.bitcoinj.core;
 
-import com.google.common.base.MoreObjects;
+
```
msgilligan: extra white space (here and in several other places)

core/src/main/java/org/bitcoinj/core/BloomFilter.java
```diff
@@ -157,11 +157,8 @@ public double getFalsePositiveRate(int elements) {
 
     @Override
     public String toString() {
-        final MoreObjects.ToStringHelper helper = MoreObjects.toStringHelper(this).omitNullValues();
-        helper.add("data length", data.length);
-        helper.add("hashFuncs", hashFuncs);
-        helper.add("nFlags", getUpdateFlag());
-        return helper.toString();
+        return "BloomFilter{data length=" + data.length + ", hashFuncs=" + hashFuncs + ", nFlags=" + getUpdateFlag()
+                + "}";
```
msgilligan: I would suggest the handling of each field should be on its own line, for ease of reading and ease of reading the diff.

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -1428,14 +1428,13 @@ public Script freshOutputScript(KeyPurpose purpose) {
...
-        if (isFollowing)
-            helper.addValue("following");
-        return helper.toString();
+        java.util.StringJoiner joiner = new java.util.StringJoiner(", ", "DeterministicKeyChain{", "}");
+        if (outputScriptType != null) joiner.add(outputScriptType.toString());
+        if (accountPath != null) joiner.add("accountPath=" + accountPath);
+        joiner.add("lookaheadSize=" + lookaheadSize);
+        joiner.add("lookaheadThreshold=" + lookaheadThreshold);
+        if (isFollowing) joiner.add("following");
+        return joiner.toString();
```
msgilligan: Maybe if we add a utility class this could be:

```
return org.bitcoinj.base.internal.ToStringUtil.for(this)
    .add(outputScriptType.toString())
    .add("accountPath", accountPath)
    .add("lookaheadSize", lookaheadSize)
    .add("lookaheadThreshold", lookaheadThreshold)
    .addIf(isFollowing, "following")
    .toString();
```

core/src/main/java/org/bitcoinj/core/BloomFilter.java
msgilligan: If we implement our own `ToStringUtil`, we could consider using it consistently. But just using `+` for the simple cases is OK, too. What do you think?

msgilligan (changes requested):
Thanks for this! I didn't realize that using `StringJoiner` is almost as functional as `ToStringHelper`!

In addition to a couple of minor change requests, I have one tentative suggestion: Can we try an (internal) `ToStringUtil` that can make things simpler. I think it can use method chaining (except perhaps in the case where try/catch needs to be used) and be even simpler than Guava's `ToStringHelper` for our use cases.

nadavramon:
Thanks for the great suggestion! I really like the fluent API approach.

I've pushed a new commit that:

1. Introduced ToStringUtil: Created a new internal utility class at org.bitcoinj.core.internal.ToStringHelper that wraps StringJoiner with a fluent .add() API.

2. Refactored Globally: Updated all files in this PR (including Peer, Transaction, DeterministicKeyChain, etc.) to use this new helper.

3. Formatting:

* Updated BloomFilter to place each field on its own line for better readability.

* Removed the extra whitespace issues you noted.

4. Verification: Validated everything passes with ./gradlew :bitcoinj-core:test.

core/src/main/java/org/bitcoinj/core/internal/ToStringUtil.java
```diff
@@ -0,0 +1,67 @@
...
+            joiner.add(value);
+        }
+        return this;
+    }
+
+    @Override
+    public String toString() {
+        return joiner.toString();
+    }
+}
```
msgilligan: Missing newline

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -69,6 +69,7 @@
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.StringJoiner;
```
msgilligan: New, unused import.

nadavramon:
Squashed this one as well.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -384,15 +385,22 @@ public boolean removeAddressEventListener(AddressEventListener listener) {
...
-        helper.add("time", TimeUtils.dateTimeFormat(vPeerVersionMessage.time));
-        helper.add("height", vPeerVersionMessage.bestHeight);
-        return helper.toString();
+        ToStringUtil helper = ToStringUtil.forObject(this)
+                .addValue(getAddress())
+                .add("version", vPeerVersionMessage.clientVersion);
+
+        if (vPeerVersionMessage.subVer != null) {
+            helper.add("subVer", vPeerVersionMessage.subVer);
+        }
```
msgilligan: Why can't we use the fluent `.add` method here? Can't it handle `null` properly?

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -384,15 +385,22 @@ public boolean removeAddressEventListener(AddressEventListener listener) {
...
+
+        if (vPeerVersionMessage.localServices.hasAny()) {
+            helper.add("services", vPeerVersionMessage.localServices);
+        }
```
msgilligan: I wonder if a fluent `.add(String name, Supplier<@Nullable String> valueSupplier)` might work here. Something like:

```
  .add("services", () -> vPeerVersionMessage.localServices.hasAny() ? vPeerVersionMessage.localServices : null)`
```

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -753,21 +753,24 @@ public int hashCode() {
...
+                .add("path", getPathAsString())
+                .add("depth", depth);
+
         Optional<Instant> creationTime = this.getCreationTime();
         if (!creationTime.isPresent())
             helper.add("creationTimeSeconds", "unknown");
         else if (parent != null)
             helper.add("creationTimeSeconds", creationTime.get().getEpochSecond() + " (inherited)");
         else
             helper.add("creationTimeSeconds", creationTime.get().getEpochSecond());
```
msgilligan: I think the fluent `.add()` that takes `Supplier<String>` might also work here. Or maybe (even simpler) the `String` value for `creationTimeSeconds` could be calculated before the `helper` is constructed?

core/src/main/java/org/bitcoinj/core/Peer.java
msgilligan: Or maybe the `String` value for "services" could just be computed before `helper` is constructed.

msgilligan (changes requested):
I think with a little more effort we can use the fluent style more consistently. See my suggestions on some of the cases. What do you think? Do you want to give them a try?

nadavramon:
I've refactored ToStringUtil to handle nulls internally and be fully fluent, as suggested.

* Refactor: ToStringUtil is now instantiable (new ToStringUtil(this)) and null-safe.

* Clean Call Sites: Removed if blocks in Peer.java and DeterministicKey.java in favor of cleaner pre-calculation and chaining.

* Cleanup: Fixed the newline and unused imports.

Validated with a clean build.

core/src/main/java/org/bitcoinj/core/internal/ToStringUtil.java
```diff
@@ -0,0 +1,63 @@
+package org.bitcoinj.core.internal;
```
msgilligan, line 17: This needs a standard bitcoinj copyright header. (The latest version say "Copyright by the original author or authors.")

core/src/main/java/org/bitcoinj/core/internal/ToStringUtil.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * A utility to help construct {@code toString} strings, replacing Guava's {@code MoreObjects.ToStringUtil}.
+ * <p>
+ * This class is internal and not part of the public API.
+ */
+public class ToStringUtil {
+    private final StringJoiner joiner;
+
+    /**
+     * Taking the object instance directly
```
msgilligan: The Javadoc should have `@param` and `@returns` tags.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -384,14 +384,13 @@ public boolean removeAddressEventListener(AddressEventListener listener) {
...
+                .addValue(getAddress())
+                .add("version", vPeerVersionMessage.clientVersion);
+
         helper.add("subVer", vPeerVersionMessage.subVer);
-        if (vPeerVersionMessage.localServices.hasAny())
-            helper.add("services", vPeerVersionMessage.localServices.toString());
-        helper.add("time", TimeUtils.dateTimeFormat(vPeerVersionMessage.time));
-        helper.add("height", vPeerVersionMessage.bestHeight);
+        helper.add("services", vPeerVersionMessage.localServices.hasAny() ? vPeerVersionMessage.localServices : null);
+
```
msgilligan: This could be fluent. It also looks like `time` and `height` were accidentally removed here.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -753,20 +753,14 @@ public int hashCode() {
...
-        if (!creationTime.isPresent())
-            helper.add("creationTimeSeconds", "unknown");
-        else if (parent != null)
-            helper.add("creationTimeSeconds", creationTime.get().getEpochSecond() + " (inherited)");
-        else
-            helper.add("creationTimeSeconds", creationTime.get().getEpochSecond());
-        helper.add("isEncrypted", isEncrypted());
-        helper.add("isPubKeyOnly", isPubKeyOnly());
+
+        long time = getCreationTime().map(java.time.Instant::getEpochSecond).orElse(0L);
```
msgilligan: I think the pre-calculated `time` should be a string. This changes behavior: `unknown` becomes `0`.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -753,20 +753,14 @@ public int hashCode() {
...
+
+        helper.add("creationTimeSeconds", parent != null ? time + " (inherited)" : time);
+
```
msgilligan: I think you lost part of your edits, some lines are missing here.

msgilligan (changes requested):
The latest commit looks incomplete and doesn't match your description of the changes. Make sure you force-pushed it correctly. I made few other requests before noticing this.

msgilligan (changes requested):
I think that with some additional effort _all_ use of `ToStringUtil` can e fluent. Do you agree?

nadavramon:
@msgilligan Agreed. 
ToStringUtil is now fully fluent. 
I also fixed the copyright header and restored the missing fields.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -753,20 +753,20 @@ public int hashCode() {
...
-            helper.add("creationTimeSeconds", creationTime.get().getEpochSecond());
+        final ToStringUtil helper = new ToStringUtil(this)
+                .add("pub", ByteUtils.formatHex(pub.getEncoded()))
+                .add("chainCode", ByteUtils.formatHex(chainCode))
+                .add("path", getPathAsString())
+                .add("depth", depth);
+
+        final String creationTime = getCreationTime()
+                .map(t -> String.valueOf(t.getEpochSecond()))
+                .orElse("unknown");
```
msgilligan, line 747: If you move this _above_ the creation of the `helper`, you can avoid the interruption in the "fluent chain". You should be able to do the same thing in the other places in this PR where the "fluent chain" is interrupted, so that a `helper` local is never defined and we have:

```
return new ToStringUtil(this)
    // multiple adds
.toString()
```

in every usage.

msgilligan (changes requested):

nadavramon:
@msgilligan I've updated the PR to follow the "Pure Fluent" style you requested.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -384,15 +384,14 @@ public boolean removeAddressEventListener(AddressEventListener listener) {
...
-        if (vPeerVersionMessage.localServices.hasAny())
-            helper.add("services", vPeerVersionMessage.localServices.toString());
-        helper.add("time", TimeUtils.dateTimeFormat(vPeerVersionMessage.time));
-        helper.add("height", vPeerVersionMessage.bestHeight);
-        return helper.toString();
+        return new ToStringUtil(this)
+                .addValue(getAddress())
+                .add("version", vPeerVersionMessage.clientVersion)
+                .add("subVer", vPeerVersionMessage.subVer)
+                .add("services", vPeerVersionMessage.localServices.hasAny() ? vPeerVersionMessage.localServices : null)
```
msgilligan: Wouldn't the  `.addIf()` method be a little simpler here?

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -753,21 +753,19 @@ public int hashCode() {
...
-        if (!creationTime.isPresent())
-            helper.add("creationTimeSeconds", "unknown");
-        else if (parent != null)
-            helper.add("creationTimeSeconds", creationTime.get().getEpochSecond() + " (inherited)");
-        else
-            helper.add("creationTimeSeconds", creationTime.get().getEpochSecond());
-        helper.add("isEncrypted", isEncrypted());
-        helper.add("isPubKeyOnly", isPubKeyOnly());
-        return helper.toString();
+        final String creationTime = getCreationTime()
```
msgilligan, line 745: Good catch that `getCreationTime()` already uses the paren when available!

core/src/main/java/org/bitcoinj/wallet/DeterministicSeed.java
```diff
@@ -201,14 +201,12 @@ public String toString() {
...
-            helper.addValue(toHexString()).add("mnemonicCode", getMnemonicString());
-        else
-            helper.addValue("unencrypted");
-        return helper.toString();
+        return new ToStringUtil(this)
+                .addIf(isEncrypted(), "encrypted")
+                .addIf(!isEncrypted() && includePrivate, toHexString())
+                .addIf(!isEncrypted() && !includePrivate, "unencrypted")
+                .addIf(!isEncrypted() && includePrivate && getMnemonicString() != null, "mnemonicCode", getMnemonicString())
+                .toString();
```
msgilligan: I think our attempt to use fluent-style here made this too hard to read and difficult to maintain. We should ether go back to the original if/else if/else or maybe try something like this:
  

```
        ToStringUtil helper = new ToStringUtil(this)
        if (isEncrypted())
            return helper.addValue("encrypted")
                              .toString();
        else if (includePrivate)
            return helper.addValue(toHexString())
                              .add("mnemonicCode", getMnemonicString())
                              .toString();
        else
            return helper.addValue("unencrypted")
                              .toString();
```

Because there are clearly three independent case. Too bad we can't use the new switch expression with guard statement stuff here, it would work well, I think.

msgilligan: It is very important that this code be correct because it can leak a private key if it is not. Ww should probably add tests.

core/src/main/java/org/bitcoinj/crypto/ECKey.java
```diff
@@ -1338,28 +1338,26 @@ public String getPrivateKeyAsWiF(Network network) {
...
+                ? String.valueOf(this.encryptedPrivateKey)
+                : null;
+
+        final boolean showUnencrypted = includePrivate && !isEncrypted() && hasPrivKey();
+        final String privHex = showUnencrypted ? getPrivateKeyAsHex() : null;
+        final String privWif = showUnencrypted ? getPrivateKeyAsWiF(network) : null;
+
+        return new ToStringUtil(this)
+                .add("pub HEX", getPublicKeyAsHex())
+                .addIf(privHex != null, "priv HEX", privHex)
```
msgilligan: We shouldn't have to use `addIf()` here and in similar usages. Since we were _always_ using `omitNullValues()` on the Guava helper we can have our default `add()` method implement that behavior.

core/src/main/java/org/bitcoinj/wallet/SendRequest.java
```diff
@@ -229,17 +229,16 @@ public static SendRequest childPaysForParent(Wallet wallet, Transaction parentTr
...
-        helper.add("ensureMinRequiredFee", ensureMinRequiredFee);
-        helper.add("signInputs", signInputs);
-        helper.add("aesKey", aesKey != null ? "set" : null); // careful to not leak the key
-        helper.add("coinSelector", coinSelector);
-        helper.add("shuffleOutputs", shuffleOutputs);
-        helper.add("recipientsPayFees", recipientsPayFees);
-        return helper.toString();
+        return new ToStringUtil(this)
+                .add("emptyWallet", emptyWallet)
+                .addIf(changeAddress != null, "changeAddress", changeAddress)
```
msgilligan: This doesn't need to be an `addIf`, nor some others in this method.

msgilligan (changes requested):

nadavramon:
@msgilligan updated the PR :

1. Mimicking Guava's omitNullValues().
2. Reverted to the `if/else` block for better readability and safety.
3. Removing explicit null checks (since ToStringUtil now handles them).

msgilligan (changes requested):
This is looking good!

The only remaining issue I see is _testing_. We should have unit tests for the `ToStringUtil` class as part of this PR. And we should also implement the following tests (in separate PRs) _before_ we merge this PR:

* Issue #4067 
* Issue #4068
* Issue #4069

nadavramon:
@msgilligan I have implemented all requested safety tests:

**1. ToStringUtil: Unit tests for null handling and fluency.
**2. DeterministicSeed (#4067): Verifies no private key leakage.
**3. DeterministicKeyChain (#4068): Verifies toString format and safety.
**4. DeterministicKey (#4069): Verifies output matches the new format.

msgilligan:
> **2. DeterministicSeed (#4067): Verifies no private key leakage.
> **3. DeterministicKeyChain (#4068): Verifies toString format and safety.
> **4. DeterministicKey (#4069): Verifies output matches the new format.

These should be three separate PRs and should work (i.e. pass) on the exiting toString() implementations. We'll merge them first, then we can verify that this PR did not change behavior.

Update: Sorry, I see I was a little unclear above  -- I edited that comment for clarity.

nadavramon:
@msgilligan Understood. I have removed the safety tests from this PR. 

I will open separate PRs for the other tests.

Update: I have split the safety tests into separate PRs as requested.
* #4071 
* #4072
* #4075

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -17,7 +17,8 @@
 
 package org.bitcoinj.core;
 
-import com.google.common.base.MoreObjects;
+import com.google.common.math.IntMath;
```
schildbach, line 20: Why is this importing `IntMath`?

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -16,11 +16,13 @@
 
 package org.bitcoinj.core;
 
-import com.google.common.base.MoreObjects;
+import com.google.common.base.Strings;
```
schildbach, line 19: Why these two?

core/src/test/java/org/bitcoinj/core/internal/ToStringUtilTest.java
```diff
@@ -0,0 +1,57 @@
...
+        assertEquals("TestObject{included=yes}", util.toString());
+    }
+
+    @Test
+    public void testAddValue() {
+        ToStringUtil util = new ToStringUtil(new TestObject());
+        util.addValue("plainValue");
+        assertEquals("TestObject{plainValue}", util.toString());
+    }
+}
```
schildbach, line 57: LF missing at EOF.

schildbach (changes requested):
I think this needs a rebase.

PR #4001, opened 2026-01-11
Refactor: Replace Guava VisibleForTesting with internal annotation

nadavramon:
This PR replaces the dependency on Guava's @VisibleForTesting with a lightweight internal annotation.

Contributes to #2000 (Remove Guava dependency)

Changes:

* Created org.bitcoinj.core.internal.VisibleForTesting.

* Replaced all com.google.common.annotations.VisibleForTesting imports with the new internal class.

* Verified via grep that no Guava usages for this annotation remain.

Verification:

* gradlew :bitcoinj-core:classes passed successfully.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/internal/VisibleForTesting.java
+++ b/core/src/main/java/org/bitcoinj/core/internal/VisibleForTesting.java
@@ -1,0 +1,15 @@
+package org.bitcoinj.core.internal;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * Annotates a program element that exists, or is more widely visible than otherwise necessary, only for use in test code.
+ * <p>
+ * Replaces Guava's {@code @VisibleForTesting}.
+ */
+@Documented
+@Retention(RetentionPolicy.CLASS)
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/bitcoinj/core/internal/VisibleForTesting.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Annotates a program element that exists, or is more widely visible than otherwise necessary, only for use in test code.
+ * <p>
+ * Replaces Guava's {@code @VisibleForTesting}.
+ */
+@Documented
+@Retention(RetentionPolicy.CLASS)
+public @interface VisibleForTesting {
+}
```
msgilligan: Missing newline.

nadavramon:
I understand the hesitation regarding the architectural approach. 
I'm happy to leave this PR open as a "ready-to-go" option while you decide. 
In the meantime, I'll continue chipping away at the other Guava dependencies

msgilligan:
@schildbach We're getting pretty close to having Guava removed and `@VisibleForTesting` is one of the remains usages. I think we previously discussed just replacing the annotation with a comment (rather than creating our own annotation.) Do you agree that is what we should do here?

PR #4002, opened 2026-01-11
Refactor: Replace Guava Strings and IntMath with standard Java 8

nadavramon:
This PR replaces usages of com.google.common.base.Strings and com.google.common.math.IntMath with standard Java 8 equivalents.

Contributes to #2000

Changes:

* BtcFormat.java: Replaced Strings.isNullOrEmpty() with null/empty checks and Strings.repeat() with Collections.nCopies.

* Transaction.java & Wallet.java: Replaced IntMath.checkedAdd() with Math.addExact().

* Transaction.java & Wallet.java: Replaced IntMath.divide(..., RoundingMode.CEILING) with standard integer arithmetic (x + y - 1) / y.

* Peer.java: Removed unused Strings import.

Verification:

* ./gradlew :bitcoinj-core:test passed successfully.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Peer.java
+++ b/core/src/main/java/org/bitcoinj/core/Peer.java
@@ -17,7 +17,6 @@
 package org.bitcoinj.core;
 
 import com.google.common.base.MoreObjects;
-import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
 import org.bitcoinj.core.internal.GuardedBy;
 import org.bitcoinj.base.Coin;
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -18,7 +18,6 @@
 package org.bitcoinj.core;
 
 import com.google.common.base.MoreObjects;
-import com.google.common.math.IntMath;
 import org.bitcoinj.base.Address;
 import org.bitcoinj.base.Coin;
 import org.bitcoinj.base.Network;
@@ -48,7 +47,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.jspecify.annotations.Nullable;
-import java.math.RoundingMode;
 import java.nio.BufferOverflowException;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -17,7 +17,6 @@
 package org.bitcoinj.core;
 
 import com.google.common.base.MoreObjects;
-import com.google.common.base.Strings;
```
msgilligan, line 20: Unrelated change.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -397,7 +395,7 @@ public int getWeight() {
     public int getVsize() {
         if (!hasWitnesses())
             return this.messageSize();
-        return IntMath.divide(getWeight(), 4, RoundingMode.CEILING); // round up
+        return (getWeight() + 3) / 4;
```
msgilligan, line 398: This makes me nervous. I'll need to check this more carefully.

core/src/main/java/org/bitcoinj/utils/BtcFormat.java
```diff
@@ -679,7 +678,7 @@ public Builder scale(int val) {
          *  <p>Note that by applying a pattern you override the configured formatting style of
          *  {@link BtcAutoFormat} instances.  */
         public Builder pattern(String val) {
-            if (!Strings.isNullOrEmpty(localizedPattern))
+            if (localizedPattern != null && !localizedPattern.isEmpty())
```
msgilligan, line 681: I think it would be better to create a `private` method `isNullOrEmpty()` inside the `BtcFormat` class. It would make the code much easier to read.

core/src/main/java/org/bitcoinj/core/Transaction.java
msgilligan: Perhaps we should define a simple function (method) for making this calculation, document it, and use it in both places. (And maybe even write a unit test.)

nadavramon:
Closing this PR as requested. I have split the changes into two separate PRs to decouple the String refactoring from the Math logic:

* #4003: Refactor: Replace Guava Strings in BtcFormat

* #4004: Refactor: Replace Guava IntMath with standard Java 8 Math

PR #4003, opened 2026-01-12
Refactor: Replace Guava Strings in BtcFormat

nadavramon:
Replaces Strings.isNullOrEmpty with a private helper isNullOrEmpty as requested. 
Also replaces Strings.repeat with Collections.nCopies. 
This separates the String logic from the Math logic.

### DIFF
--- a/core/src/main/java/org/bitcoinj/utils/BtcFormat.java
+++ b/core/src/main/java/org/bitcoinj/utils/BtcFormat.java
@@ -16,7 +16,6 @@
 
 package org.bitcoinj.utils;
 
-import com.google.common.base.Strings;
 import org.bitcoinj.base.Coin;
 import org.bitcoinj.utils.BtcAutoFormat.Style;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/utils/BtcFormat.java
```diff
@@ -1549,7 +1552,7 @@ public String coinCode() { synchronized(numberFormat) {
     public String pattern() { synchronized(numberFormat) {
         StringBuilder groups = new StringBuilder();
         for (int group : decimalGroups) {
-            groups.append("(").append(Strings.repeat("#", group)).append(")");
+            groups.append("(").append(String.join("", Collections.nCopies(group, "#"))).append(")");
```
msgilligan, line 1551: This should be in separate commit. You can either rebase this PR to have two commits or submit it in a separate PR.

core/src/main/java/org/bitcoinj/utils/BtcFormat.java
```diff
@@ -679,11 +678,15 @@ public Builder scale(int val) {
...
+            if (!isNullOrEmpty(localizedPattern))
                 throw new IllegalStateException("You cannot invoke both pattern() and localizedPattern()");
             pattern = val;
             return this;
-        } 
+        }
+
+        private static boolean isNullOrEmpty(String s) {
+            return s == null || s.isEmpty();
+        }
```
msgilligan: This is very minor and subjective, but as this is a trivial helper method, I would suggest putting it as near the bottom of the file as possible.

nadavramon:
I've split the changes into two separate commits as requested:

* Refactor: Replace Strings.isNullOrEmpty with private helper (moved helper to bottom of file)

* Refactor: Replace Strings.repeat with Collections.nCopies

PR #4004, opened 2026-01-12
Refactor: Replace Guava IntMath with standard Java 8 Math

nadavramon:
This PR removes dependencies on Guava IntMath by using standard Java 8 features and a safe internal helper, as discussed in the review for #4002.

Changes:

* New Helper: Added org.bitcoinj.core.internal.MathUtils.divideRoundUp(long, long) to handle ceiling division safely. It uses Math.addExact to prevent overflow and returns int (via Math.toIntExact) for type safety.

* Transaction & Wallet: Replaced IntMath.divide(..., RoundingMode.CEILING) with the new MathUtils.divideRoundUp.

* Safety: Replaced IntMath.checkedAdd with Math.addExact.

Verified with ./gradlew :bitcoinj-core:test.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -18,7 +18,6 @@
 package org.bitcoinj.core;
 
 import com.google.common.base.MoreObjects;
-import com.google.common.math.IntMath;
 import org.bitcoinj.base.Address;
 import org.bitcoinj.base.Coin;
 import org.bitcoinj.base.Network;
@@ -164,7 +163,7 @@ public class Transaction implements Message {
      * @see <a href="https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki">BIP 141</a>
      */
     public static int calculateVirtualTransactionSize(int weight) {
-        return IntMath.divide(weight, WITNESS_SCALE_FACTOR, RoundingMode.CEILING); // round up
+        return (weight + WITNESS_SCALE_FACTOR - 1) / WITNESS_SCALE_FACTOR; // round up
     }
 
     private final int protocolVersion;


msgilligan (changes requested):
The idea here is good, but I think we can do something better. 

On the previous PR I made a suggestion:

> Perhaps we should define a simple function (method) for making this calculation, document it, and use it in both places. 

When I said "this calculation", I wasn't clear enough. The calculation is (I believe) known as "calculating the Segwit discount".

We don't really need a general-purpose `divideRoundUp`, what we need is a function to calculate the "segwit discount" defined in BIP 141. This is an important implementation detail in Bitcoin and we should encapsulate it and document it clearly.

Let's try the following:

1. Add a static constant `WITNESS_SCALE_FACTOR` to `Transaction` with value `4`.
2. Move `divideRoundUp` to a `static` method in `Transaction` named `calculateVirtualTransactionSize`
3. Remove the `divisor` parameter and hard-code it to `WITNESS_SCALE_FACTOR`

This method can then have Javadoc comments explaining the calculation and referencing BIP 141, specifically:

> Virtual transaction size is defined as Transaction weight / 4 (rounded up to the next integer)

see: https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki

This improves the readability of our code and its mapping to the BIP _and_ has the side effect of dropping the Guava dependency.

(Note that `WITNESS_SCALE_FACTOR` is the name of the constant in the Bitcoin Core C++ code)

nadavramon:
I've refactored the code to follow BIP 141 logic as requested:

* Transaction.java: Added WITNESS_SCALE_FACTOR (4) and the static method calculateVirtualTransactionSize(long weight). Added the Javadoc link to BIP 141.

* Wallet.java: Updated estimateVirtualBytesForSigning to use the new static method from Transaction.

* Cleanup: Deleted the temporary MathUtils class entirely.

Validated with a clean build.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -151,6 +150,20 @@ private static int sortableBlockHeight(Transaction tx) {
...
+    /**
+     * The scale factor for Witness data in Segregated Witness transactions.
+     * @see <a href="https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki">BIP 141</a>
+     */
+    public static final int WITNESS_SCALE_FACTOR = 4;
+
+    /**
+     * Virtual transaction size is defined as Transaction weight / 4 (rounded up to the next integer).
+     * @see <a href="https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki">BIP 141</a>
+     */
```
msgilligan, line 164: This should have @param and @return tags.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -397,7 +410,7 @@ public int getWeight() {
     public int getVsize() {
         if (!hasWitnesses())
             return this.messageSize();
-        return IntMath.divide(getWeight(), 4, RoundingMode.CEILING); // round up
+        return calculateVirtualTransactionSize(getWeight()); // round up
```
msgilligan: The "round up" comment is unnecessary (and a little misleading)

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -5367,8 +5366,8 @@ private int estimateVirtualBytesForSigning(Script script) {
             } else if (ScriptPattern.isP2WPKH(script)) {
                 ECKey key = findKeyFromPubKeyHash(ScriptPattern.extractHashFromP2WH(script), ScriptType.P2WPKH);
                 Objects.requireNonNull(key, "Coin selection includes unspendable outputs");
-                return IntMath.divide(script.getNumberOfBytesRequiredToSpend(key, null), 4,
-                        RoundingMode.CEILING); // round up
+                return Transaction.calculateVirtualTransactionSize(script.getNumberOfBytesRequiredToSpend(key
+                        , null)); // round up
```
msgilligan: The "round up" comment is unnecessary (and a little misleading)

nadavramon:
Done.

* Transaction.java: Added @param and @return tags to calculateVirtualTransactionSize. Removed the unnecessary // round up comment.

* Wallet.java: Removed the // round up comment.

msgilligan (approved):
Approved! Thanks!

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -151,6 +150,22 @@ private static int sortableBlockHeight(Transaction tx) {
...
+    public static final int WITNESS_SCALE_FACTOR = 4;
+
+    /**
+     * Virtual transaction size is defined as Transaction weight / 4 (rounded up to the next integer).
+     * @param weight the transaction weight
+     * @return the virtual transaction size
+     * @see <a href="https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki">BIP 141</a>
+     */
+    public static int calculateVirtualTransactionSize(long weight) {
+        return (int) ((weight + WITNESS_SCALE_FACTOR - 1) / WITNESS_SCALE_FACTOR);
```
schildbach: I think we should make clearer at this point that we're rounding up here.

msgilligan: I had previously asked that a `// round up` comment be removed, so this is my fault.

I think summarizing the divide and roundup in the Javadoc comment would be better.

schildbach: Why would you include an implementation detail in the JavaDoc? I think JavaDocs are targetted at users (of the method) and those in general don't care about the implementation.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: I agree it's not necessary any more at this point, but instead it should be added to the impl of `Transaction.calculateVirtualTransactionSize()`.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
schildbach: I agree it's not necessary any more at this point, but instead it should be added to the impl of `Transaction.calculateVirtualTransactionSize()`.

core/src/main/java/org/bitcoinj/core/Transaction.java
nadavramon: I will wait for a consensus here before pushing the final update to avoid unnecessary noise in the commit history. Just let me know which direction you prefer!

msgilligan: I think we have consensus, please add a description to the method Javadoc that explains dividing and rounding. Maybe even copy the exact text in the BIP.

msgilligan: Oh my bad, Andreas is explicitly saying don't put it in the Javadoc...

msgilligan: > Why would you include an implementation detail in the JavaDoc? I think JavaDocs are targetted at users (of the method) and those in general don't care about the implementation.

For Bitcoin developers who are working at this level, I do not think it is an implementation detail that should be abstracted. I think it is an important part of the protocol that should be documented and understood.

I'll let @schildbach make the final call, though.

schildbach: I think what's important to me is when I read a formula that is not super-obvious, I'd expect a small hint either in form of an inline comment or by naming of a local variable. I would not read the JavaDoc for that. So I'd recommend adding something like `// round up` to the formula.

That said, we can keep the description of how to calculate virtual transaction size in the JavaDoc too – it certainly doesn't hurt.

nadavramon: I split the PR into two commits as requested:

* Implements the BIP 141 logic in Transaction (with Javadoc + inline comment) and updates Wallet to use it.

* Removes the unused Guava IntMath imports.

msgilligan: > I split the PR into two commits as requested:
> 
> * Implements the BIP 141 logic in Transaction (with Javadoc + inline comment) and updates Wallet to use it.
> * Removes the unused Guava IntMath imports.

I think what we want here is:

1.  to move the logic to `Transaction` but keep the Guava integer math.
2. Replace Guava integer math with built-in integer math.

Or in other words:

1. Consolidate existing logic without re-writing the math
2. Re-write the math to remove Guava

schildbach: I'm sorry, but that's not the separation I asked for.

First commit shall extract the method and use it, but not change its implementation.

Second commit changes its implementation.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -151,6 +150,23 @@ private static int sortableBlockHeight(Transaction tx) {
...
+    public static final int WITNESS_SCALE_FACTOR = 4;
+
+    /**
+     * Virtual transaction size is defined as Transaction weight / 4 (rounded up to the next integer).
+     *
+     * @param weight the transaction weight
+     * @return the virtual transaction size
+     * @see <a href="https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki">BIP 141</a>
+     */
+    public static int calculateVirtualTransactionSize(long weight) {
```
schildbach: Is there a reason you picked `long` for weight? In general, we use `int` for sizes as far as I can tell.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -5367,8 +5366,8 @@ private int estimateVirtualBytesForSigning(Script script) {
             } else if (ScriptPattern.isP2WPKH(script)) {
                 ECKey key = findKeyFromPubKeyHash(ScriptPattern.extractHashFromP2WH(script), ScriptType.P2WPKH);
                 Objects.requireNonNull(key, "Coin selection includes unspendable outputs");
-                return IntMath.divide(script.getNumberOfBytesRequiredToSpend(key, null), 4,
-                        RoundingMode.CEILING); // round up
+                return Transaction.calculateVirtualTransactionSize(script.getNumberOfBytesRequiredToSpend(key
+                        , null));
```
schildbach: I think these two lines can fit into one, or if you prefer to split it in 2 then can you put the comma at the EOL rather than the SOL?

core/src/main/java/org/bitcoinj/core/Transaction.java
nadavramon: I have restructured the commits and addressed the new feedback:

* Centralized the logic using int (instead of long) and fixed the formatting in Wallet.java.
* Swapped the Guava IntMath implementation for standard Java math and removed the imports.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -19,7 +19,6 @@
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ArrayListMultimap;
-import com.google.common.math.IntMath;
```
msgilligan, line 22: This change should actually be in the first commit. That commit removes the dependency on `IntMath` in `Wallet`.

nadavramon:
I moved the removal of the IntMath import in Wallet.java to the first commit, since that's where the dependency was actually dropped.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: This is fixed -- the change is now in the first commit.

PR #3999, opened 2026-01-10
Refactor: Replace Guava Lists.newArrayList with JDK equivalents globally

nadavramon:
This PR replaces all occurrences of Lists.newArrayList across the project with standard JDK 8 equivalents, as requested in the review of #3996.

Contributes to #2000

Supersedes: #3996 (This PR includes the fix for ECKeyTest.java and extends it to the rest of the codebase).

Changes:

Lists.newArrayList() → new ArrayList<>()

Lists.newArrayList(e1, e2) → Arrays.asList(e1, e2)

Lists.newArrayListWithExpectedSize(n) → new ArrayList<>(n)

Removed unused com.google.common.collect.Lists imports.

Verification:

Passed full core:test suite.

Verified compilation and tests for WalletTest, BasicKeyChainTest, TransactionInputTest, and others.

PR #4005, opened 2026-01-12
BasicKeyChainTest: Remove use of ArrayList type.

msgilligan:
2 commits to migrate away from use of `ArrayList`. This will make the transition away from Guava `Lists.newArrayList()` a little cleaner.  (See PR #3999)


msgilligan:
@nadavramon take a look at this PR.

PR #4010, opened 2026-01-13
TransactionConfidence: replace Guava `Iterators` with standard Java in `getBroadcastBy()`

msgilligan:
There is also a follow-on commit that updates Javadoc for `PeerAddress` now that it is thread-safe.


msgilligan:
A step towards #2000

core/src/main/java/org/bitcoinj/core/PeerAddress.java
```diff
@@ -45,8 +45,6 @@
 /**
  * A PeerAddress holds an IP address and port number representing the network location of
  * a peer in the Bitcoin P2P network. It exists primarily for serialization purposes.
- * <p>
- * Instances of this class are not safe for use by multiple threads.
```
schildbach, line 49: Shouldn't we replace this with

> Instances of this class are immutable.

then?

msgilligan: Sure. Updated, rebased, and force-pushed.

schildbach (approved):

PR #4014, opened 2026-01-15
Replace Guava UnsignedBytes.checkedCast with ByteUtils.longToByteExact

### LINKED ISSUE #4009: Replace Guava `UnsignedBytes.checkedCast`
I think adding `public static byte longToByteExact(long value)` to o.b.base.internal.ByteUtils would be the best solution.

The proposed name is consistent with the naming of similar methods in the JDK.

nadavramon:
This PR replaces the usage of Guava's `UnsignedBytes.checkedCast` with a method `ByteUtils.longToByteExact`, as requested in issue #4009.

## Changes
* Added `longToByteExact(long value)` to `org.bitcoinj.base.internal.ByteUtils`.
* Replaced the single usage of the Guava method in DeterministicKey.java.
* Updated `DeterministicKeyTest` to expect `ArithmeticException` instead of `IllegalArgumentException`.

Closes #4009

### DIFF
--- a/base/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
+++ b/base/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
@@ -159,6 +159,20 @@ public class ByteUtils {
     }
 
     /**
+     * Checks if the {@code long} value fits in an unsigned byte (0-255) and returns it as a {@code byte}.
+     * <p>
+     * The value is expected as an unsigned {@code long} as per the Java Unsigned Integer API.
+     *
+     * @param value value to be checked and cast
+     * @return the value as a {@code byte}
+     * @throws IllegalArgumentException if the value doesn't fit in an unsigned byte
+     */
+    public static byte longToByteExact(long value) {
+        checkArgument(value >= 0 && value <= 255, () -> "Value " + value + " out of range for unsigned byte");
+        return (byte) value;
+    }
+
+    /**
      * Write a 32-bit integer to a given byte array in little-endian format, starting at a given offset.
      * <p>
      * The value is expected as an unsigned {@code long} as per the Java Unsigned Integer API.
--- a/base/src/test/java/org/bitcoinj/base/internal/ByteUtilsTest.java
+++ b/base/src/test/java/org/bitcoinj/base/internal/ByteUtilsTest.java
@@ -330,4 +330,21 @@ public class ByteUtilsTest {
     public void testDecodeMPI() {
         assertEquals(BigInteger.ZERO, ByteUtils.decodeMPI(new byte[]{}, false));
     }
+
+    @Test
+    public void longToByteExact_valid() {
+        assertEquals(0, ByteUtils.longToByteExact(0));
+        assertEquals((byte) 255, ByteUtils.longToByteExact(255));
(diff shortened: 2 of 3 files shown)


msgilligan (changes requested):
This should be two commits:

* The refactor to use ByteUtils.longToByteExact() which throws `IllegalArgumentException`
* The change from `IllegalArgumentException` to `ArithmeticException`

The idea is "refactoring" commits should not change behavior. Note that we sometimes use `git bisect` to figure out when an issue might have been introduced.

I'm assuming you change the exception type to be more correct/consistent -- which we probably want to do if it doesn't break anything. But we will want to consider that change separately and if we do merge it we want it in a separate commit.

nadavramon:
@msgilligan I have split the changes into two separate commits as requested:

1. The first commit refactors the code to use ByteUtils.longToByteExact while maintaining the original IllegalArgumentException behavior. 
2. The second commit updates the method (and the test) to throw ArithmeticException for consistency with JDK standards.

msgilligan (approved):
I 100% approve the first commit. The second commit I approve if @schildbach thinks it's ok to change the exception that is thrown.

base/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -158,6 +158,22 @@ public static ByteBuffer writeInt32LE(long val, ByteBuffer buf) throws BufferOve
...
+     * The value is expected as an unsigned {@code long} as per the Java Unsigned Integer API.
+     *
+     * @param value value to be checked and cast
+     * @return the value as a {@code byte}
+     * @throws IllegalArgumentException if the value doesn't fit in an unsigned byte
+     */
+    public static byte longToByteExact(long value) {
+        if (value < 0 || value > 255) {
+            throw new IllegalArgumentException("Value " + value + " out of range for unsigned byte");
+        }
```
schildbach: Can you use `Preconditions.checkArgument()` for such checks. There are many usage examples in bitcoinj if you search for references.

schildbach (commented):
I also think the new method could maybe use some tests? Or do you think it's too trivial? I bet we draw inspirations from Guava as well…

base/src/main/java/org/bitcoinj/base/internal/ByteUtils.java
```diff
@@ -165,11 +165,11 @@ public static ByteBuffer writeInt32LE(long val, ByteBuffer buf) throws BufferOve
      *
      * @param value value to be checked and cast
      * @return the value as a {@code byte}
-     * @throws IllegalArgumentException if the value doesn't fit in an unsigned byte
+     * @throws ArithmeticException if the value doesn't fit in an unsigned byte
```
schildbach: Hmmm, is not being able to cast an "arithmetic condition"? From the ArithmeticException JavaDoc:

> Thrown when an exceptional arithmetic condition has occurred.

I'd lean slightly towards not merging this change, but I'm certainly not adamant about that.

In case we merge it: there is a generic `Preconditions.check()` for custom exceptions as well.

nadavramon:
@schildbach I've updated the PR based on your feedback:

The method now throws IllegalArgumentException again.
Switched to using Preconditions.checkArgument.
Added longToByteExact unit tests to ByteUtilsTest.

PR #4015, opened 2026-01-17
gradle.yml, graalvm.yml: update Gradle to 9.4.1

msgilligan:
PR is updated to Gradle 9.4.1, which works. This brings us JDK 26 build support.

It looks like our build doesn't work on Gradle 9.3.0, perhaps because of the following change:

> This results from Gradle adopting the incubating Test Event Reporting API internally.

See: https://docs.gradle.org/current/release-notes.html#test-reporting-improvements
However, as noted above it works on Gradle 9.4.1.

### DIFF
--- a/.github/workflows/build-gradle.yml
+++ b/.github/workflows/build-gradle.yml
@@ -11,7 +11,7 @@ jobs:
         os: [ubuntu-24.04, ubuntu-24.04-arm, macOS-15, macOS-15-intel, windows-2022]
         java: ['17', '21', '25']
         distribution: ['temurin']
-        gradle: ['8.14.4', '9.2.1']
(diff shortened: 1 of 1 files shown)


schildbach:
Hm, is this a known problem at Gradle? Does someone have more specific info about the incompatibility?

msgilligan:
> Hm, is this a known problem at Gradle? Does someone have more specific info about the incompatibility?

I didn't spend a lot of time looking at this, but our use of JUnit 4 seems to be one element of why our configuration is failing. Maybe should should open an issue in the Gradle project.

msgilligan:
Rebased and updated to Gradle 9.4.1 which is working.

PR #4017, opened 2026-01-17
Use `java.util.base64` instead of Guava and Bouncy Castle

msgilligan:


msgilligan:
cc: @nadavramon

nadavramon:
Thanks for handling the tools build issue @msgilligan! 
This looks like the right solution.

PR #4029, opened 2026-01-20
WalletAppKit: remove setCheckpoints()

msgilligan:
This is implemented incorrectly, and if used would fail. WalletAppKit will operate correctly by default using a pre-configured checkpoints file.

This commit has the side-effect of removing the `Closeables` Guava dependency.

### DIFF
--- a/core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
+++ b/core/src/main/java/org/bitcoinj/kits/WalletAppKit.java
@@ -17,7 +17,6 @@
 
 package org.bitcoinj.kits;
 
-import com.google.common.io.Closeables;
 import com.google.common.util.concurrent.AbstractIdleService;
 import org.bitcoinj.base.BitcoinNetwork;
 import org.bitcoinj.base.ScriptType;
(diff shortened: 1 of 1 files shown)


schildbach:
> This is implemented incorrectly, and if used would fail.

How do you mean? Do we have an issue for the failure/incorrect impl?

msgilligan:
> > This is implemented incorrectly, and if used would fail.
> 
> How do you mean? Do we have an issue for the failure/incorrect impl?

I misread the code and thought I saw a bug. The code is:

```
    public WalletAppKit setCheckpoints(InputStream checkpoints) {
        if (this.checkpoints != null)
            Closeables.closeQuietly(checkpoints);
        this.checkpoints = Objects.requireNonNull(checkpoints);
        return this;
    }
```

I somehow read it as `if (checkpoints != null) Closeables.closeQuietly(checkpoints);`

And since there are no tests or usage of this method anywhere in bitcoinj, I figured it was broken and we might s well remove it.

msgilligan:
While double-checking my mistake, I found b6ecc380 which we can revers to remove `Closeables.closeQuietly`.

msgilligan:
Closing because this was a mistake. PR #4029 removes the use of Guava `Closeables`.

PR #4032, opened 2026-01-20
DnsDiscovery: Don't create/dispose defaultThreadFactory for each thread 

msgilligan:
This is a child of (the first commit in) PR #4031

### DIFF
--- a/core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
+++ b/core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
@@ -77,8 +77,7 @@ public class DnsDiscovery extends MultiplexingDiscovery {
     @Override
     protected ExecutorService createExecutor() {
         return Executors.newFixedThreadPool(seeds.size(), r -> {
-            Thread thread = Executors.defaultThreadFactory().newThread(r);
-            thread.setName("DNS seed lookups");
+            Thread thread = new Thread(Thread.currentThread().getThreadGroup(), r, "DNS seed lookups", 0);
             thread.setDaemon(true);
             return thread;
         });


schildbach:
Rebased.

core/src/main/java/org/bitcoinj/net/discovery/DnsDiscovery.java
```diff
@@ -77,8 +77,7 @@ private static List<PeerDiscovery> buildDiscoveries(Network network, String[] se
     @Override
     protected ExecutorService createExecutor() {
         return Executors.newFixedThreadPool(seeds.size(), r -> {
-            Thread thread = Executors.defaultThreadFactory().newThread(r);
-            thread.setName("DNS seed lookups");
+            Thread thread = new Thread(Thread.currentThread().getThreadGroup(), r, "DNS seed lookups", 0);
```
schildbach, line 80: I wonder if this change affects the handling of the thread group. I assume previously it was set by the "default thread factory". Now, it is set to whatever group the thread calling `getPeers()` is in.

Not sure if this is a problem, but I'm not very confident with the handling of threads in general, and in bitcoinj specifically – so I'm a bit scared of such changes.

msgilligan: The newly-created `DefaultThreadFactory` returned by `defaultThreadFactory()` stores `Thread.currentThread().getThreadGroup()` from the thread calling its constructor and uses that in `newThread()`. 

So this change does not affect the assigned thread group.

schildbach: Thanks. Good to know.

PR #4039, opened 2026-01-20
Threading: Threading.UserThread improvements

msgilligan:
2 commits with minor enhancements (including making UserThread package-private -- nobody should be using it)

Update: maybe we want to create a more idiomatic method to set `WARNING_THRESHOLD`? It currently is a mutable static field, but with an uppercase name implying being `final`.

Also see WIP PR #3848


msgilligan:
This was rebased on master which took this PR a little further my making things `private`.

PR #4046, opened 2026-01-22
Transaction: introduce constant for defining dust

schildbach:
Dust is defined by a minimum feerate. Historically this has been based on the REFERENCE_DEFAULT_MIN_TX_FEE, however changing the dust limit changes which transactions are standard and should be done with care.

Hence, we decouple the two constants.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -145,6 +145,11 @@ public class Transaction implements Message {
     public static final Coin REFERENCE_DEFAULT_MIN_TX_FEE = Coin.valueOf(1_000); // 0.01 mBTC
 
     /**
+     * Minimum feerate for defining dust, in sats per kB.
+     */
+    public static final Coin DUST_RELAY_TX_FEE_RATE = Coin.valueOf(3_000); // per kB
+
+    /**
      * If using this feePerKb, transactions will get confirmed within the next couple of blocks.
      * This should be adjusted from time to time. Last adjustment: February 2017.
      */
--- a/core/src/main/java/org/bitcoinj/core/TransactionOutput.java
+++ b/core/src/main/java/org/bitcoinj/core/TransactionOutput.java
@@ -241,7 +241,7 @@ public class TransactionOutput {
      * and mined by default miners.
      */
     public Coin getMinNonDustValue() {
-        return getMinNonDustValue(Transaction.REFERENCE_DEFAULT_MIN_TX_FEE.multiply(3));
+        return getMinNonDustValue(Transaction.DUST_RELAY_TX_FEE_RATE);
     }
 
     /**


core/src/main/java/org/bitcoinj/core/Transaction.java
msgilligan, line 143: The reference to `feePerKb` in this comment is confusing to me. Is `REFERENCE_DEFAULT_MIN_TX_FEE ` a minimum fee or minimum fee-rate? If it is a minimum fee how can it be compared to `feePerKb`.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -144,6 +144,11 @@ private static int sortableBlockHeight(Transaction tx) {
      */
     public static final Coin REFERENCE_DEFAULT_MIN_TX_FEE = Coin.valueOf(1_000); // 0.01 mBTC
 
+    /**
+     * Minimum feerate for defining dust.
+     */
+    public static final Coin DUST_RELAY_TX_FEE = Coin.valueOf(3_000); // per kB
```
msgilligan: If this constant is a fee rate maybe it should have `RATE` in its name?

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -144,6 +144,11 @@ private static int sortableBlockHeight(Transaction tx) {
      */
     public static final Coin REFERENCE_DEFAULT_MIN_TX_FEE = Coin.valueOf(1_000); // 0.01 mBTC
 
+    /**
+     * Minimum feerate for defining dust.
```
msgilligan: We should probably specify the units here (i.e. `feePerKb`)

msgilligan:
It might make sense to consistently include `_RATE` in the name of _all_ constants that are fee rates. This would require `@Deprecated` tags on existing names pointing to the new names, but I think it would be worth it for clarity.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: It's a fee rate, which is historically then use to calculate how large (amount) a txout must be in order to cover the cost of itself and its (future) spending txin. If I understand it right. It's the calculation done in `public Coin getMinNonDustValue(Coin feePerKb)`.

The name of the new constant was copied from Bitcoin Core, which we do for a lot of these constants. But I agree, we should make it clear by name that it's a fee *rate*. So I'll rename it to `DUST_RELAY_TX_FEERATE`.

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: :+1:

core/src/main/java/org/bitcoinj/core/Transaction.java
schildbach: :+1:

core/src/main/java/org/bitcoinj/core/Transaction.java
msgilligan: See Issue #4053

PR #4055, opened 2026-01-22
SendRequest: make emptyWallet final

msgilligan:
This is now two commits on top of PR #4066 

1. Prepare to make `emptyWallet` final (with warning comment)
2. Make it final.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/SendRequest.java
+++ b/core/src/main/java/org/bitcoinj/wallet/SendRequest.java
@@ -58,10 +58,8 @@ public class SendRequest {
      * When emptyWallet is {@code true}, all coins selected by the coin selector are sent to the first output in {@link #tx}
      * (its {@code value} is ignored and set to {@link Wallet#getBalance()} <i>less the fee</i> required
      * for the transaction). Any additional outputs are removed.
-     * <p>
-     * This field should be treated as {@code final} and set via {@link SendRequest#emptyWallet(Transaction)} or {@link SendRequest#emptyWallet(Address)}.
      */
-    public boolean emptyWallet;
+    public final boolean emptyWallet;
 
     /**
      * "Change" means the difference between the value gathered by a transactions inputs (the size of which you


schildbach:
The JavaDoc changes could be merged in a separate commit.

msgilligan:
> I'm not sure about this change. Currently (in 0.17), SendRequest is more like a builder class. E.g. in BW it's used like this for the standard case of sending coins:
> 
> ```
>         final SendRequest sendRequest = finalPaymentIntent.toSendRequest();
>         sendRequest.emptyWallet =
>                 viewModel.paymentIntent.mayEditAmount() && amount.equals(Constants.NETWORK_PARAMETERS.getMaxMoney());
>         sendRequest.feePerKb = fees.get(viewModel.feeCategory.getValue());
>         sendRequest.memo = viewModel.paymentIntent.memo;
>         sendRequest.exchangeRate = amountCalculatorLink.getExchangeRate();
>         sendRequest.aesKey = encryptionKey;
> ```
> 
> So without a new builder concept I would refrain from making this field immutable.

My intention was to make all fields final, add new `private` constructors as necessary, and add new `static` factory methods as needed.

The first move in this direction was PR #2893

msgilligan:
> The JavaDoc changes could be merged in a separate commit.

See PR #4066 

This PR has been rebased on PR #4066 and split into two additional commits:

1. Add new constructor and static factory method, so bitcoinj never mutates `emptyWallet` and add comment saying that mutation is deprecated
2. Make `emptyWallet` `final`.

So the first commit can be merged and once BW and others are updated, we can make `emptyWallet` final.

core/src/test/java/org/bitcoinj/wallet/WalletTest.java
```diff
@@ -2658,7 +2657,6 @@ public void lowerThanDefaultFeeRate() throws InsufficientMoneyException {
         SendRequest emptyReq = SendRequest.emptyWallet(myAddress);
         emptyReq.feePerKb = fee;
         emptyReq.ensureMinRequiredFee = true;
-        emptyReq.emptyWallet = true;
```
msgilligan, line 2661: Note that this line (and similar one below) are unnecessary and could be removed without any other changes.

msgilligan:
@schildbach See Issue #4060 for a proposed direction.

PR #4083, opened 2026-01-27
drop BLOCKING_CLIENT_MANAGER from all TestWithNetworkConnection subclasses

msgilligan:
This is an experiment to see if it eliminates (or reduces) test failures.

(I think the issue(s) we are seeing is more likely with the NioServer, but which clients are used could be a factor)


msgilligan:
Closing as this did not seem to fix the problem.

PR #4089, opened 2026-01-29
test-support: remove dependency on bitcoinj-base

msgilligan:
The only class used was the `BitcoinNetwork` `enum`, which is replaced by the corresponding ID string.

This removes a circular dependency (in Maven's dependency model) needed for Issue #3840 -- we had previously attempted to remove the cycle with PR #3850.

### DIFF
--- a/test-support/build.gradle
+++ b/test-support/build.gradle
@@ -3,7 +3,7 @@ plugins {
 }
 
 dependencies {
-    api project(':bitcoinj-base')
+    api 'org.jspecify:jspecify:1.0.0'
 }
 
 tasks.withType(JavaCompile) {
--- a/test-support/src/main/java/org/bitcoinj/test/support/package-info.java
+++ b/test-support/src/main/java/org/bitcoinj/test/support/package-info.java
@@ -14,8 +14,7 @@
  * limitations under the License.
  */
 /**
- * Test support classes that depend upon {@code bitcoinj-base} and can be used in both
- * {@code bitcoinj-base} and {@code bitcoinj-core} tests.
+ * Test support classes that can be used in both {@code bitcoinj-base} and {@code bitcoinj-core} tests.
  */
 @NullMarked
 package org.bitcoinj.test.support;


msgilligan:
@schildbach @hasskell Rebased and split into two commits.

msgilligan:
@schildbach I know you feel like this is a step backward (loosing our nice `BitcoinNetwork` enum and using a `String` instead) but I think it is worth it in order to move forward with Maven.

Unfortunately, Maven multi-project builds handle dependencies at the JAR level and do not allow the test classpath of `base` to depend on the (non-test) classes in `test-support` without seeing a cycle. I did some searching and it seems this is a fundamental limitation of Maven and not easily worked around (I did some searching.)

One solution would be to move (all, some of) the tests from `base` to a new `base-test` module and this would break the circular dependency. I think this would be overkill at present (given that we only have one class in `test-support` and this PR resolves the issue)

We could also move the tests in `base` that need `AddressData` to `core` or to `integration-test` -- perhaps in a certain package, but that seems like an even greater separation between `base` code and tests.

There is some documentation of how Maven handles multiple-modules and scopes here:

* https://maven.apache.org/guides/mini/guide-multiple-modules.html
* https://maven.apache.org/repositories/dependencies.html

I'm adding this comment to document some of the research and decision making (and private feedback from @schildbach) that I have considered.

I think we should just merge this PR and move ahead with the Maven build. When we find another possible class for inclusion in `test-support` we can revisit the discussion. Note that if we use Maven we will get the benefit of being able to use JUnit 5 (or 6!) everywhere and will be able to run this with Debian's Maven.

hasskell:
> @schildbach I know you feel like this is a step backward (loosing our nice `BitcoinNetwork` enum and using a `String` instead) but I think it is worth it in order to move forward with Maven.
> 
> Unfortunately, Maven multi-project builds handle dependencies at the JAR level and do not allow the test classpath of `base` to depend on the (non-test) classes in `test-support` without seeing a cycle. I did some searching and it seems this is a fundamental limitation of Maven and not easily worked around (I did some searching.)
> 
> One solution would be to move (all, some of) the tests from `base` to a new `base-test` module and this would break the circular dependency. I think this would be overkill at present (given that we only have one class in `test-support` and this PR resolves the issue)
> 
> We could also move the tests in `base` that need `AddressData` to `core` or to `integration-test` -- perhaps in a certain package, but that seems like an even greater separation between `base` code and tests.
> 
> There is some documentation of how Maven handles multiple-modules and scopes here:
> 
> * https://maven.apache.org/guides/mini/guide-multiple-modules.html
> * https://maven.apache.org/repositories/dependencies.html
> 
> I'm adding this comment to document some of the research and decision making (and private feedback from @schildbach) that I have considered.
> 
> I think we should just merge this PR and move ahead with the Maven build. When we find another possible class for inclusion in `test-support` we can revisit the discussion. Note that if we use Maven we will get the benefit of being able to use JUnit 5 (or 6!) everywhere and will be able to run this with Debian's Maven.

Yea I agree with @msgilligan I have done some digging here https://github.com/bitcoinj/bitcoinj/issues/3840#issuecomment-3817851221 nothing really worked to keep Maven happy. So this seams like a legitimate tradeoff as of now

PR #4095, opened 2026-02-02
SendRequest: move use of Context to constructor

msgilligan:
This consolidates use of `Context` to the constructor and prepares for further refactoring.


schildbach:
If I remember right, these two defaults are only assigned to allow for easier testing. We could probably replace this with a static constructor `SendRequest.withNoFee()` (or similar) that would then be used in lots of the tests instead of setting a custom context.

PR #4096, opened 2026-02-02
PeerNetwork: new class to store TxConfidenceTable

msgilligan:
* Add PeerNetwork class
* (For now) construct it in the `Context` constructor
* Deprecate Context.getConfidenceTable()
* Update callers to use Context.peerNetwork().txConfidenceTable()

This should be rebased when PR #4094 is merged.

TODO: Make a default PeerNetwork singleton and update apps/samples to set it in the initial/propagated Context.


schildbach:
At the moment, this just seems to wrap TxConfidenceTable. I think it would be good to see the next steps (commits/PRs) after this before deciding to merge/continue.

Issue #840, opened 2015-06-11
Unit tests that test NIO are flaky


mikehearn:
Originally reported on Google Code with ID 474

```
I've tried to build the library on Debian Jessie/testing, amd64 and i686. Both fail.

Results :

Failed tests:   basicTimeoutTest(com.google.bitcoin.protocols.niowrapper.NioWrapperTest)

Tests in error: 
  testGeneratedChain(com.google.bitcoin.core.FullPrunedBlockChainTest)
  largeDataTest(com.google.bitcoin.protocols.niowrapper.NioWrapperTest): Address already
in use
  testConnectionEventHandlers(com.google.bitcoin.protocols.niowrapper.NioWrapperTest):
Address already in use
  testSimpleChannel(com.google.bitcoin.protocols.channels.ChannelConnectionTest): Address
already in use

Tests run: 246, Failures: 1, Errors: 4, Skipped: 0

The problem is a bit subtle, one test of NioWrapperTest fails and then it leaves an
open port so subsequent tests also fail because the address is already in use. The
surefire log is attached. The failing line is:
assertTrue(clientConnection1Closed.isDone() && serverConnection1Closed.isDone());

So it seems the connection isn't closed properly. I'm using OpenJDK 6b27-1.12.5-2,
maven 3.0.4-6.
```

Reported by `radioanonzoi` on 2013-10-30 07:04:56

<hr>
- _Attachment: [com.google.bitcoin.protocols.niowrapper.NioWrapperTest.txt](https://storage.googleapis.com/google-code-attachments/bitcoinj/issue-474/comment-0/com.google.bitcoin.protocols.niowrapper.NioWrapperTest.txt)_

PR #4098, opened 2026-02-03
fix: fixing integration test address collision

hasskell:
### Description 
So pipeline test is a bit flaky and keep getting socket address collisions that require restart test. I am replacing random addresses to 0 which basically let operating system to choose address

```
PeerGroupTest > [1] > transactionConfidence[1] FAILED
    java.net.BindException: Address already in use
        at java.base/sun.nio.ch.Net.bind0(Native Method)
        at java.base/sun.nio.ch.Net.bind(Net.java:577)
        at java.base/sun.nio.ch.ServerSocketChannelImpl.netBind(ServerSocketChannelImpl.java:344)
        at java.base/sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:301)
        at java.base/sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:89)
        at java.base/sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:81)
        at org.bitcoinj.net.NioServer.<init>(NioServer.java:78)
        at org.bitcoinj.testing.TestWithNetworkConnections.startPeerServer(TestWithNetworkConnections.java:153)
        at org.bitcoinj.testing.TestWithNetworkConnections.startPeerServers(TestWithNetworkConnections.java:133)
        at org.bitcoinj.testing.TestWithNetworkConnections.setUp(TestWithNetworkConnections.java:122)
        at org.bitcoinj.testing.TestWithPeerGroup.setUp(TestWithPeerGroup.java:78)
        at org.bitcoinj.testing.TestWithPeerGroup.setUp(TestWithPeerGroup.java:73)
        at org.bitcoinj.core.PeerGroupTest.setUp(PeerGroupTest.java:117)
```
@msgilligan I recall there was an item for this but just could not find which one 

### Testing 
Tested with remote. Test are still flaky but for other reasons, haven't see any address collisions after 3 runs

### DIFF
--- a/core/src/main/java/org/bitcoinj/net/NioServer.java
+++ b/core/src/main/java/org/bitcoinj/net/NioServer.java
@@ -134,4 +134,12 @@ public class NioServer extends AbstractExecutionThreadService {
         // Wake up the selector and let the selection thread break its loop as the ExecutionService !isRunning()
         selector.wakeup();
     }
+
+    /**
+     * Returns the port this server is bound to. Useful for testing when port 0 is used to let the OS allocate a port.
+     * @return the local port number
+     */
+    public int getPort() {
+        return sc.socket().getLocalPort();
+    }
 }
--- a/integration-test/src/test/java/org/bitcoinj/testing/TestWithNetworkConnections.java
+++ b/integration-test/src/test/java/org/bitcoinj/testing/TestWithNetworkConnections.java
@@ -56,7 +56,6 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.time.Duration;
-import java.util.Random;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -69,7 +68,6 @@ import static org.bitcoinj.base.internal.Preconditions.checkState;
  * Utility class that makes it easy to work with mock NetworkConnections.
  */
 public class TestWithNetworkConnections {
-    protected static final int TCP_PORT_BASE = 10000 + new Random().nextInt(40000);
     public static final int PEER_SERVERS = 5;
 
     protected static final NetworkParameters UNITTEST = UnitTestParams.get();
@@ -81,6 +79,7 @@ public class TestWithNetworkConnections {
     protected SocketAddress socketAddress;
 
     private final NioServer[] peerServers = new NioServer[PEER_SERVERS];
+    private final int[] peerServerPorts = new int[PEER_SERVERS];
     private final ClientConnectionManager channels;
     protected final BlockingQueue<InboundMessageQueuer> newPeerWriteTargetQueue = new LinkedBlockingQueue<>();
 
@@ -150,9 +149,10 @@ public class TestWithNetworkConnections {
                     }
                 };
             }
-        }, new InetSocketAddress(InetAddress.getLoopbackAddress(), TCP_PORT_BASE + i));
+        }, new InetSocketAddress(InetAddress.getLoopbackAddress(), 0));
         peerServers[i].startAsync();
         peerServers[i].awaitRunning();
+        peerServerPorts[i] = peerServers[i].getPort();
     }
 
     public void tearDown() throws Exception {
@@ -170,6 +170,11 @@ public class TestWithNetworkConnections {
             server.stopAsync();
             server.awaitTerminated();
         }
+    }
+
+    protected int getPeerServerPort(int i) {
+        checkArgument(i >= 0 && i < PEER_SERVERS);
+        return peerServerPorts[i];
     }
 
     protected InboundMessageQueuer connect(Peer peer, VersionMessage versionMessage) throws Exception {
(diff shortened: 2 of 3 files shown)


schildbach:
If this works, it'd be a great fix!

I have yet to understand how the port is picked and why the collisions have been happening in the first place.

msgilligan:
Thanks @hasskell for investigating this issue and making an attempt to fix it!

> @msgilligan I recall there was an item for this but just could not find which one

* Issue #840

Note that there are _other_ still unresolved intermittent failures. We have a parent (tracking) PR for known intermittent test failures:

* Issue #1984

I have spent some time troubleshooting and I believe there is a _root cause_ of Issue #840 and am hopeful that fixing the _cause_ will reduce our false failures by at least an order of magnitude.

msgilligan:
> If this works, it'd be a great fix!

From the investigations I've done I would expect that this change may _reduce_ the number of failures, but it may also hide the _root cause_ of the problem. I could be wrong, but when @hasskell says

> Test are still flaky but for other reasons
 
I fear that the root cause of _peer-networking-related_ problems has not been fixed.  (The most common non-networking related false-failure that I see lately is a random failure of a Bloom filter test. There is also another random-number-related occasional false-failure IIRC) 

It also may be incorrect to call these peer-networking failures "false-failures" because the root cause is possibly/probably an intermittent bug in NioServer, so there is a real bug in the code -- it's just showing up in random tests.

> I have yet to understand how the port is picked 

I just looked at the port-picking code a few days ago and the same range of ports is used in each test in test classes using `TestWithNetworkConnections`.  If you look at line 153 of `TestWithNetworkConnections`, you'll see`TCP_PORT_BASE + i` and that means the Peer `i` gets assigned that port (and the number of `PEER_SERVERS` is `5`.

> and why the collisions have been happening in the first place.

I believe the collisions are happening because some intermittent failure is occurring and the peers are not being closed. I could be wrong, though. But this is why I'm skeptical that this PR will resolve the issue.

@hasskell What are the errors your seeing when test are "flaky but for other reasons"?

hasskell:
@msgilligan your intuition could right here. Let me spend a bit more time here and add more logs to pinpoint actual issue. The other failures that I see is this:
```
PeerGroupTest > [1] STANDARD_ERROR
    11:19:59.093 463 PeerGroup$2.run: Exception when trying to build connections
    java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@460a78a0[Not completed, task = java.util.concurrent.Executors$RunnableAdapter@4cf213d0[Wrapped task = org.bitcoinj.core.PeerGroup$2@4235e43e]] rejected from org.bitcoinj.testing.TestWithPeerGroup$1$1@308ee05[Shutting down, pool size = 1, active threads = 1, queued tasks = 3, completed tasks = 4]
    	at java.base/java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2065)
    	at java.base/java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:833)
    	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor.delayedExecute(ScheduledThreadPoolExecutor.java:340)
    	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor.schedule(ScheduledThreadPoolExecutor.java:562)
    	at org.bitcoinj.testing.TestWithPeerGroup$1$1.schedule(TestWithPeerGroup.java:122)
    	at org.bitcoinj.core.PeerGroup$2.go(PeerGroup.java:606)
    	at org.bitcoinj.core.PeerGroup$2.run(PeerGroup.java:531)
    	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
    	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
    	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
    	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
    	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
    	at org.bitcoinj.utils.ContextPropagatingThreadFactory.lambda$newThread$0(ContextPropagatingThreadFactory.java:49)
    	at java.base/java.lang.Thread.run(Thread.java:840)  
```

Which I think is classical race condition when executor is shutting down before in-flight task is finished.

msgilligan:
> Which I think is classical race condition when executor is shutting down before in-flight task is finished.

I believe what is happening is that `executor` is in the process of being shut down and `triggerConnectionsJob` is trying to schedule a new task.  Note that `triggerConnections()` checks to see if `executor` is shut down before calling `triggerConnectionsJob`, but by the time it gets to line 606, something has shutdown the executor.

I would look earlier in the log and see if something is causing the `PeerGroup` to terminate before it is even fully started.

msgilligan:
@hasskell @schildbach See https://github.com/bitcoinj/bitcoinj/commit/006c00ac35790da0cf9446b3541e72e58b2fb17b for an earlier attempt to fix this -- since it only randomized at class initialization time, I don't think it actually did much to prevent collisions. But I thought we all should be aware of it.

Note that I've created DRAFT/Debugging PR #4114 that is adding instrumentation and other changes to help find the problem without worrying so much about whether each individual commit is something we want long term. There is currently a commit that _removes_ this randomization and another that displays the address/port the conflict occurs on, so we can see if there is a patter of where the issue is appearing.

PR #4102, opened 2026-02-04
PeerTest: clear uncaughtExceptionHandler after setting

msgilligan:
`badMessage()` sets Threading.uncaughtExceptionHandler to complete a local `CompletableFuture` at the start of the test, but does not set it back to `null` when finished, like the `exceptionListener()` test does.


schildbach (approved):

schildbach:
Nice catch!

PR #4103, opened 2026-02-04
PeerGroup: skip scheduling task in setupPinging() if executor is shutdown

msgilligan:
It is possible that the executor can be told to shutdown before setupPinging() is called. This prevents a `RejectedExecutionException` from being thrown. In some cases (if the PeerGroup is stopped before setupPinging() is called during PeerGroup startup) the `RejectedExecutionException` is ignored (but is logged) and in other cases it is not.

You can see the logged-and-ignored case in the bitcoinj-core `WalletAppKitTest` `launchAndClose()` and `launchAndCloseTwice()` tests. This commit should cause `RejectedExecutionException` to no longer show in the log for those tests.


msgilligan:
cc: @hasskell

PR #4104, opened 2026-02-04
PeerGroup: better handle shutdown in triggerConnectionsJob

msgilligan:
This is a child of PR #4103, but could potentially be applied separately.

This contains two commits:

1. Checks for shutdown before trying to schedule jobs with `executor`
2. Avoid doing unnecessary work and I/O when `executor` is shutdown


msgilligan:
I opened Issue #4105 after studying the current test failure here: https://github.com/bitcoinj/bitcoinj/actions/runs/21692274818/job/62556011781

msgilligan:
cc: @hasskell

PR #4108, opened 2026-02-05
integration-tests: narrow throws declarations in helper methods

msgilligan:
This ~~is a child of PR #4107 which narrows the throws clauses of classes in `core`~~ PR narrows throws declarations in integration test helper methods..


schildbach:
I've undone one change to a test method, rather than just a helper method.

Issue #4099, opened 2026-02-03
Refactor `NioServer` and `TestWithNetworkConnections`


msgilligan:
We should work to refactor these classes to improve readability and reliability (as well as remove the Guava dependency)

This should be done via a series of refactoring PRs and the following are some ideas for improvements.

NioServer:

- [ ] Refactor to not implement Guava `Service`
- [ ] Simplify class hierarchy if possible
- [ ] Verify and possibly improve error-handling that might result in ports not closed
- [x] `NullMarked` -- PR #4142
- [ ] Other improvements

TestWithNetworkConnections:

- [ ] Upgrade to use JUnit Jupiter: See (part 2) of Issue #2436
- [ ] Verify and possibly improve error-handling that might result in ports not closed
- [x] `NullMarked` -- PR #4141 
- [ ] Add code to detect when ports aren't closed?
- [ ] Simplify code with JDK 17 features
- [ ] Move some classes from `src/tests` to `src/main`?

This issue is based on [this comment](https://github.com/bitcoinj/bitcoinj/issues/4079) from Issue #4079.

msgilligan:
Here's a list of existing, related PRs:

* PR #3799 -- I just found this earlier PR that duplicates a commit in PR 4084
* PR #4081
* PR #4082
* PR #4083 -- this is an experiment, not a candidate for merging
* PR #4084

mkdev5:
@msgilligan Can I pick this issue up?

msgilligan:
> Can I pick this issue up?

Are you a human?

mkdev5:
What do you mean?

mkdev5:
@msgilligan I am interested in this issue. I will provide a good result.

msgilligan:
This is not a good issue for a first time contributor, but I will certainly review any PRs made in good faith.

mkdev5:
> > Can I pick this issue up?
> 
> Are you a human?

Haha, yes, I’m human. I totally understand you, especially with so many AI agents being used these days to fix bugs and issues.
But, can you let me know that it's not allowed to use AI to fix issues on this repo?

msgilligan:
We don't have an "official" AI policy for bitcoinj yet. We have seen some some drive-by Pull Requests that we believe were generated using chatbots that were fairly low quality.

[...]

PR #4114, opened 2026-02-09
TestWithNetworkConnections and subclasses: debugging branch

msgilligan:
This change does not seem to have eliminated the problem (though making TCP_PORT_BASE non-static might help) and makes troubleshooting a little harder.

At the very least we should remove this while troubleshooting if not permanently.

### DIFF
--- a/core/src/main/java/org/bitcoinj/net/NioServer.java
+++ b/core/src/main/java/org/bitcoinj/net/NioServer.java
@@ -110,6 +110,9 @@ public class NioServer extends AbstractExecutionThreadService {
             }
         } catch (IOException e) {
             log.error("Error trying to open/read from connection", e);
+        } catch (Throwable t) {
+            log.error("Throwable trying to open/read from connection", t);
+            throw new RuntimeException(t);
         } finally {
             // Go through and close everything, without letting IOExceptions get in our way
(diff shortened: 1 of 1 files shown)


mkdev5:
Hi @msgilligan, I've been looking into the intermittent `BindException` and I think the Guava `Service` lifecycle may be contributing to it.

When `NioServer.run()` encounters an unchecked exception (e.g. `CancelledKeyException`), it escapes the `catch (IOException)` block. Guava catches it and transitions the service to FAILED state. Then in `stopPeerServers()`, `awaitTerminated()` throws IllegalStateException on the failed server — which propagates through the loop without being caught, leaving the remaining servers' ports bound. The next test hits `BindException` on those ports.

I've opened a PR that refactors `NioServer` to not implement Guava `Service` (item 1 of #4099): https://github.com/bitcoinj/bitcoinj/pull/4161

The key behavioral change is that `awaitTerminated()` never throws — it simply waits for the thread to finish regardless of how it exited. It also widens the catch to `Exception`, adds `SO_REUSEADDR`, fixes a constructor resource leak, and guards against null from `accept()`.

msgilligan:
This will need to be rebased because of PR #4182 and others.

PR #4121, opened 2026-02-20
Transaction: require `amount` for `addSignedInput()`

msgilligan:
This is  a child of PR #4120 and is a breaking change that we should consider very carefully. I do not recommend merging this for 0.18. Although we might want to consider the `@NonNull` annotations without the `IllegalArgumentException`

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -957,9 +957,9 @@ public class Transaction implements Message {
         // Verify the API user didn't try to do operations out of order.
         checkState(!outputs.isEmpty(), () ->
                 "attempting to sign tx without outputs");
-        if (amount == null || amount.value <= 0) {
-            log.warn("Illegal amount value. Amount is required for SegWit transactions.");
-        }
+        Objects.requireNonNull(amount);
+        checkState(amount.value > 0, () ->
+                "Illegal amount value. Amount is required for SegWit transactions.");
         TransactionInput input = new TransactionInput(this, new byte[] {}, prevOut, amount);
         addInput(input);
         int inputIndex = inputs.size() - 1;


schildbach:
Well, this is something I have considered myself. Apart from some API breakage, what are you afraid of?

msgilligan:
> Apart from some API breakage, what are you afraid of?

Well, it's runtime breakage without compile-time breakage. This PR adds both `@NonNull` annotations (which for a typical project will probably be ignored) and `Objects.requireNonNull(amount);` which throws a `RuntimeException`.

We could add the `@NonNull` annotations now and wait for a future release to add the `RuntimeException`.

schildbach:
The thing is for SegWit the amount is mandatory for signing, and who is still using "legacy" signing? So I assume this will affect only a very small percentage of users.

And old apps likely still use the deprecated methods we just removed. Maybe our mistake was to not include a recommendation into the deprecation message to not simply pass null as an amount (which I assume many devs might be doing when they quickly try to clear as many deprecation warnings as possible).

msgilligan:
> The thing is for SegWit the amount is mandatory for signing, and who is still using "legacy" signing? So I assume this will affect only a very small percentage of users.

Agreed.

> And old apps likely still use the deprecated methods we just removed. Maybe our mistake was to not include a recommendation into the deprecation message to not simply pass null as an amount (which I assume many devs might be doing when they quickly try to clear as many deprecation warnings as possible).

I'll split this PR into 2 commits. The first with the `@NonNull` annotation (which would then be removed when the containing package becomes `@NullMarked`) and the second with the `Objects.requireNonNull()`. We can merge the first right away and consider the second separately. I'm OK with merging the second immediately if you are, but we can also wait till after the first alpha.

Update: We also should consider throwing an IllegalArgumentException when `amount` is zero or negative.

msgilligan:
@schildbach I split this into two commits:

1. Annotate `amount` parameter as `@NonNull` with no runtime enforcement
2. Throw `RuntimeException` if `amount` is null or <= 0

schildbach:
Merged the first commit and rebased.

PR #4124, opened 2026-02-22
core module: fix all existing blocking  issues and enable ErrorProne/NullAway

msgilligan:
This is ~~8~~ ~~7~~ ~~5~~ 3 commits. The first ~~7~~ ~~6~~ ~~4~~ 2 address various issues that would break the build if ErrorProne/NullAway were enabled. The final commit enables ErrorProne/NullAway. Individual commits can mostly be cherry-picked.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -1,12 +1,19 @@
 import org.gradle.util.GradleVersion
+import net.ltgt.gradle.errorprone.CheckSeverity
 
 plugins {
     id 'java-library'
     id 'com.google.protobuf'
     id 'maven-publish'
+    id "net.ltgt.errorprone" version "5.0.0" apply false
 }
 
 version = '0.18-SNAPSHOT'
+
+def supportsErrorProne = GradleVersion.current() >= GradleVersion.version("8.5") && JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_21)
+if (supportsErrorProne) {
+    apply plugin: 'net.ltgt.errorprone'
+}
 
 dependencies {
     api project(':bitcoinj-base')
@@ -25,6 +32,12 @@ dependencies {
     testImplementation 'nl.jqno.equalsverifier:equalsverifier:3.18.1'
     testImplementation 'org.hamcrest:hamcrest-library:3.0'
     testImplementation 'pl.pragmatists:JUnitParams:1.1.1'
+
+    if (supportsErrorProne) {
+        annotationProcessor "com.uber.nullaway:nullaway:0.13.1"
+        testAnnotationProcessor "com.uber.nullaway:nullaway:0.13.1"
+        errorprone "com.google.errorprone:error_prone_core:2.47.0"
+    }
 }
 
 // Prevent newer Gradle from switching to JRE version of Guava for tests.
@@ -48,6 +61,12 @@ tasks.withType(JavaCompile) {
     options.compilerArgs.addAll(['--release', '8'])
     options.compilerArgs << '-Xlint:deprecation'
     options.encoding = 'UTF-8'
+    if (supportsErrorProne) {
+        options.errorprone {
+            check("NullAway", CheckSeverity.ERROR)
+            option("NullAway:OnlyNullMarked")
+        }
+    }
 }
 
 javadoc.options.encoding = 'UTF-8'


core/src/main/java/org/bitcoinj/crypto/KeyCrypterScrypt.java
```diff
@@ -121,8 +121,7 @@ public KeyCrypterScrypt(ScryptParameters scryptParameters) {
         this.scryptParameters = Objects.requireNonNull(scryptParameters);
         // Check there is a non-empty salt.
         // (Some early MultiBit wallets has a missing salt so it is not a hard fail).
-        if (scryptParameters.getSalt() == null
-                || scryptParameters.getSalt().toByteArray() == null
+        if (scryptParameters.getSalt().toByteArray() == null
```
schildbach: How does the plugin know ScryptParameters.getSalt() can't be null? It's a protobuf class, and yes the `.proto` defines the field as required, but I don't see this reflected in the generated Java class.

Anyway, I think this commit's description could rely less on NullAway and explain more the real reason we could strip this check.

core/src/main/java/org/bitcoinj/signers/LocalTransactionSigner.java
```diff
@@ -137,7 +137,7 @@ public boolean signInputs(ProposedTransaction propTx, KeyBag keyBag) {
                     txIn = txIn.withScriptSig(ScriptBuilder.createEmpty());
                     txIn = txIn.withWitness(TransactionWitness.redeemP2WPKH(signature, key));
                 } else {
-                    throw new IllegalStateException(script.toString());
+                    throw new IllegalStateException("Unsupported script");
```
schildbach: Shouldn't we include some content of the script with the exception message?

core/src/test/java/org/bitcoinj/core/TransactionOutputTest.java
```diff
@@ -117,13 +117,13 @@ public void getMinNonDustValue() {
     @Test
     public void toString_() {
         TransactionOutput p2pk = new TransactionOutput(null, Coin.COIN, myKey);
-        p2pk.toString();
+        String s1 = p2pk.toString();
```
schildbach: I think this just clutters code. I'd rather ignore the specific check for this method, or maybe for all tests?

Alternatively, we could come of with a meaningful test about the contents of `toString()`, which assume would clear the error as a side-effect as well.

core/src/main/java/org/bitcoinj/crypto/KeyCrypterScrypt.java
msgilligan: > How does the plugin know ScryptParameters.getSalt() can't be null? It's a protobuf class, and yes the `.proto` defines the field as required, but I don't see this reflected in the generated Java class.

See https://errorprone.info/bugpattern/ImpossibleNullComparison

> 
> Anyway, I think this commit's description could rely less on NullAway and explain more the real reason we could strip this check.

We can update the commit description _or_ we can use `@SuppressWarnings("ImpossibleNullComparison")`. I'm going to change it to suppress the warning and we can decide later if we actually want to change the code.

msgilligan: I've rebased and force-pushed the change to use `@SuppressWarnings`.

I've also created PR #4127 which removes the null-check as was originally done here, but with an improved comment that explains the real reason for removing the check.

core/src/main/java/org/bitcoinj/signers/LocalTransactionSigner.java
msgilligan: Fixed and force-pushed.  (Note that it is actually the `scriptPubKey` that is unsupported)

core/src/test/java/org/bitcoinj/core/TransactionOutputTest.java
msgilligan: >meaningful test about the contents of `toString()`

Great suggestion. I added verification against a regular expression.

schildbach: Thanks! I cherry-pick this.

Still I think for similar cases we should consider disabling the error. While in production you'd try to avoid calling methods to "provoke" side effects, I think in unit tests that's a common (and to some extent welcome) practise.

core/src/main/java/org/bitcoinj/signers/LocalTransactionSigner.java
schildbach: Picked.

core/src/test/java/org/bitcoinj/core/TransactionOutputTest.java
msgilligan: > that's a common (and to some extent welcome) practice.

Although typically all it really tests is that the method doesn't throw rather than verifying correct operation.

schildbach:
Rebased.

schildbach:
I think I've now cherry-picked everything from this PR that is useful on its own.

If we merge the rest (enabling the plugin), then I think we should merge it squashed with the `@SuppressWarning` annotations, as those are an artifact of the plugin.

msgilligan:
This is superseded by PR #4132 which enables checking in the root build.gradle for all projects without using/needing the `checkedProjects` list and has a single commit which also contains the annotations to suppress warnings.

msgilligan:
Closing, now that PR #4132 (which turns on ErrorProne/NullAway for all subprojects) has been merged.

PR #4127, opened 2026-02-22
KeyScripterScrypt: remove checks for `.getSalt()`

msgilligan:
These checks are unnecessary because `salt` is a `required` ProtoBuf field.

Note that there is also a commit to suppress ErrorProne warnings for this operation in PR #4124 that I suggest we merge first and consider this PR later/separately.


msgilligan:
Closing in favor of PR #4134 which throws `IllegalArgumentException` if salt is missing or too short.

PR #4164, opened 2026-03-17
NioServer: add reproducer for awaitTerminated() throwing on FAILED state

mkdev5:
When connectionOpened() throws a RuntimeException, the unchecked exception escapes run()'s catch(IOException) block. Guava transitions the service to FAILED state, and awaitTerminated() throws IllegalStateException instead of returning normally.

In TestWithNetworkConnections.stopPeerServers(), this exception propagates through the loop, preventing remaining servers from being stopped. Their ports remain bound, causing BindException in subsequent tests.

This test currently FAILS, demonstrating the bug.

### DIFF
--- a/core/src/test/java/org/bitcoinj/net/NioServerTest.java
+++ b/core/src/test/java/org/bitcoinj/net/NioServerTest.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2013 Google Inc.
+ *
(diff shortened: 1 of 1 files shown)


mkdev5:
@msgilligan please let me know the process you want.

schildbach:
Rebased.

PR #4175, opened 2026-03-22
TestWithPeerGroup, TestWithNetworkConnections: fix test teardown to prevent port leaks

### LINKED ISSUE #1984: Resolve (MS-Windows) unit test issues.

mkdev5:
Fixes #1984 (partially — addresses the teardown-related root causes of flaky integration tests).

2 commits:

`TestWithPeerGroup`: the `tearDown()` method was calling `super.tearDown()` (which stops the NioServers) before stopping the `peerGroup`. This left the PeerGroup with active connections to dead servers, causing `CancelledKeyException` and `RejectedExecutionException`. The fix reverses the order — stop the PeerGroup first, then the servers. `super.tearDown()` is also moved to a `finally` block so servers are always cleaned up even if stopping the PeerGroup throws.

`TestWithNetworkConnections`: if `stopPeerServer(i)` throws (e.g. `IllegalStateException` from `awaitTerminated()` on a FAILED NioServer), the `stopPeerServers()` loop would abort and the remaining servers would never be stopped, leaking bound ports and causing `BindException` in subsequent tests. Each iteration is now wrapped in try-catch so all servers get a chance to shut down.

### DIFF
--- a/integration-test/src/main/java/org/bitcoinj/test/integration/peer/TestWithNetworkConnections.java
+++ b/integration-test/src/main/java/org/bitcoinj/test/integration/peer/TestWithNetworkConnections.java
@@ -166,8 +166,13 @@ public class TestWithNetworkConnections {
     }
 
     protected void stopPeerServers() {
-        for (int i = 0 ; i < PEER_SERVERS ; i++)
-            stopPeerServer(i);
+        for (int i = 0 ; i < PEER_SERVERS ; i++) {
+            try {
+                stopPeerServer(i);
+            } catch (Exception e) {
+                // continue stopping remaining servers
+            }
+        }
     }
 
     protected void stopPeerServer(int i) {
--- a/integration-test/src/test/java/org/bitcoinj/test/integration/peer/TestWithNetworkConnectionsTest.java
+++ b/integration-test/src/test/java/org/bitcoinj/test/integration/peer/TestWithNetworkConnectionsTest.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2026 the bitcoinj contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


msgilligan:
I would really like to see a reproducible, failing test for the "Address in use" error(s) and then see that these commits fix it. To move in that  direction I have created:

* PR #4182

More work is needed to actually create a reproducer, but this approach should allow us to write tests of the `TestWithNetworkConnections` class itself, which should be valuable.

mkdev5:
@msgilligan Can I write a reproducer test in this PR?

msgilligan:
I would recommend working on it in a separate branch/PR, then adding it at the end of this one to see if this one fixes it.

But you may also want to take a look at what I'm doing in PR #4182 -- by restructuring things I think we'll be able to write more tightly targeted reproducers and clean up the test infrastructure in general.

mkdev5:
Just added a reproducer test.

msgilligan:
At first glance this looks good! I think we will want to rebase it on 4182, assuming that is merged.

mkdev5:
> At first glance this looks good! I think we will want to rebase it on 4182, assuming that is merged.

Thank you!

msgilligan:
You already need a rebase. It also might be helpful to submit a _DRAFT_ PR of just the test, so we can see it fail.

mkdev5:
> You already need a rebase. It also might be helpful to submit a _DRAFT_ PR of just the test, so we can see it fail.

Done!

msgilligan:
The last test run demonstrates that this fix does not (at least completely) fix the issue we are seeing in CI.

mkdev5:
> The last test run demonstrates that this fix does not (at least completely) fix the issue we are seeing in CI.

Yeah, this PR fixes 1 cause. If you prefer, I can look into SO_REUSEADDR / port allocation changes here.

mkdev5:
@msgilligan can you please let me know your feedback?

PR #4181, opened 2026-03-24
TestWithNetworkConnections: Don't randomize TCP_PORT_BASE

msgilligan:
This randomization does not prevent the 'address already in use' errors we have been seeing. It only randomizes the value upon class loading, so it is not randomized between tests. Since it is not necessary or helpful we should remove it.

### DIFF
--- a/integration-test/src/main/java/org/bitcoinj/test/integration/peer/TestWithNetworkConnections.java
+++ b/integration-test/src/main/java/org/bitcoinj/test/integration/peer/TestWithNetworkConnections.java
@@ -72,7 +72,7 @@ import static org.bitcoinj.base.internal.Preconditions.checkState;
  * Utility class that makes it easy to work with mock NetworkConnections.
  */
 public class TestWithNetworkConnections {
-    protected static final int TCP_PORT_BASE = 10000 + new Random().nextInt(40000);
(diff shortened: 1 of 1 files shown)


integration-test/src/main/java/org/bitcoinj/test/integration/peer/TestWithNetworkConnections.java
```diff
@@ -72,7 +72,7 @@
  * Utility class that makes it easy to work with mock NetworkConnections.
  */
 public class TestWithNetworkConnections {
-    protected static final int TCP_PORT_BASE = 10000 + new Random().nextInt(40000);
+    protected static final int TCP_PORT_BASE = 10000;
```
schildbach, line 75: Shouldn't we rename the constant to `TCP_PORT` in the same commit, since it's not a base any more?

schildbach (commented):
I agree. However one comment.

integration-test/src/main/java/org/bitcoinj/test/integration/peer/TestWithNetworkConnections.java
msgilligan: It's still a "base" because there are 5 ports starting at that port number.

schildbach (approved):

PR #4197, opened 2026-04-26
BuildCheckpoints: fix textual checkpoints to not include padding

schildbach:
This fixes a regression introduced with migrating from Guava base64 to JDK:

Two commits:
- First one extracts the base64 encoder to a variable.
- Second fixes textual checkpoints to not include padding. `CheckpointManager` chokes on padding.

Afaict we didn't backport the bug to 0.17.


msgilligan (approved):
Approved.

PR #4206, opened 2026-05-12
modular-architecture.md: Split 0.18 release into two, add 0.20

msgilligan:
Split the secp-api refactoring out of 0.18 and into 0.19 and create an 0.20 release with what was previously release 0.19.

Also make some other minor updates and improvements.


schildbach (approved):
0.20 is a little too far out for my taste, but yeah better have a plan than not having one…

PR #4187, opened 2026-03-29
CryptoUtils: Use JCA to access RIPEMD160 and SHA3

msgilligan:
Three commits: one for RIPEMD160 and one for SHA3 and a third to `clone()` instances for improved performance.

Issue #3964, opened 2025-12-09
Add RIPEMD160 implementation to `bitcoinj-base` module


msgilligan:
We are trying to reduce our (direct) dependencies on Bouncy Castle. In particular, see PR #3912.

A RIPEMD160 implementation is small, relatively simple and is only used to hash already hashed pubkeys, so I think it is a reasonable candidate for inclusion directly in `bitcoinj-base`.

For reference we could look at the Bouncy Castle implementation or the implementation in [nayuki/Bitcoin-Cryptography-Library](https://github.com/nayuki/Bitcoin-Cryptography-Library):

* [Ripemd160.java](https://github.com/nayuki/Bitcoin-Cryptography-Library/blob/master/java/io/nayuki/bitcoin/crypto/Ripemd160.java)
* [Ripemd160Test.java](https://github.com/nayuki/Bitcoin-Cryptography-Library/blob/master/java/io/nayuki/bitcoin/crypto/Ripemd160Test.java)

George43566:
> We are trying to reduce our (direct) dependencies on Bouncy Castle. In particular, see Issue [#3912](https://github.com/bitcoinj/bitcoinj/issues/3912).
> 
> A RIPEMD160 implementation is small, relatively simple and is only used to hash already hashed pubkeys, so I think it is a reasonable candidate for inclusion directly in `bitcoinj-base`.
> 
> For reference we could look at the Bouncy Castle implementation or the implementation in [nayuki/Bitcoin-Cryptography-Library](https://github.com/nayuki/Bitcoin-Cryptography-Library):
> 
> * [Ripemd160.java](https://github.com/nayuki/Bitcoin-Cryptography-Library/blob/master/java/io/nayuki/bitcoin/crypto/Ripemd160.java)
> * [Ripemd160Test.java](https://github.com/nayuki/Bitcoin-Cryptography-Library/blob/master/java/io/nayuki/bitcoin/crypto/Ripemd160Test.java)

All you need to do is to deposit 200 to this address. bitcoin address below and it will be pulled.All you need to do is to deposit  50$ bitcoin to this address. bitcoin address below and it will be pulled immediately and rectify for you ..bc1q9947hg38qca6zxmjw5ca6p9w922tz4vprxx3pq.  To make it recognize and clear errors To make it recognize and clear errors

Crystora:
[...]

PR #4208, opened 2026-05-13
Address: add static constructor that takes ECPublicKey

msgilligan:
Allows the `base` module -- without `bitcoinj-core` and eventually without Bouncy Castle on the classpath to create addresses (using Sha256 and RIPEMD160).

See also:

* Issue #3964  
* PR #4187

### DIFF
--- a/base/src/main/java/org/bitcoinj/base/Address.java
+++ b/base/src/main/java/org/bitcoinj/base/Address.java
@@ -16,6 +16,7 @@
 
 package org.bitcoinj.base;
 
+import java.security.interfaces.ECPublicKey;
 import java.util.Comparator;
 
 /**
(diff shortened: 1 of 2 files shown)


schildbach:
What's the usecase for this? Would you like to add a unit test?

msgilligan:
> What's the usecase for this? Would you like to add a unit test?

I will add a unit test when the prerequisites are merged and I can implement the TODOs in the code. (The code is currently not finished or working.)

The use case would be any JDK-based Bitcoin application that wants to use `o.b.base` but not `o.b.core` (because maybe they have their own implementations of transactions or wallets, etc.) and wants to create an `Address`.

msgilligan:
@schildbach asked:

> What's the usecase for this?

Now that `ECKey` implements `ECPublicKey`, once this PR is finished/merged users will be able to:

```
ECKey key = ...
Address.fromKey(key, ScriptType.P2WPKH, MAINNET)
```

msgilligan:
>  Would you like to add a unit test?

I've added two unit tests: one that creates an address from an `ECKey` and another that creates one from an arbitrary (Secp256k1) subclass of `ECPublicKey`

liamgilligan (approved):

msgilligan:
I've decided to make this feature lower-priority, for a few reasons:

* We need to resolve this issue first: https://github.com/bitcoinj/secp256k1-jdk/issues/520 
* I think we need to get an 0.18 release as soon as possible and this

PR #4210, opened 2026-05-17
HDKeyDerivation: update seed bit-length checks for master key gen

### LINKED ISSUE #4209: BIP32: Incorrect seed bit-length validation for master key generation
I found that [DeterministicKey createMasterPrivateKey](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/crypto/HDKeyDerivation.java#L64) provides input check for `byte[] seed` . 
```
        checkArgument(seed.length > 8, () ->
                "seed is too short and could be brute forced");
```
However, the constraints [defined in BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki?plain=1#L148)

> Generate a seed byte sequence S of a chosen length (between 128 and 512 bits; 256 bits is advised) from a (P)RNG.

are different. I wonder whether there is a specific reason for this, or whether this effectively 9-byte lower bound was chosen arbitrarily. In any case, I would recommend adhering to the specification as closely as possible.

Found using the [bitcoinfuzz](https://github.com/bitcoinfuzz/bitcoinfuzz) tool.

kuliq23:
Resolves #4209

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/HDKeyDerivation.java
+++ b/core/src/main/java/org/bitcoinj/crypto/HDKeyDerivation.java
@@ -56,14 +56,16 @@ public final class HDKeyDerivation {
      * Generates a new deterministic key from the given seed, which can be any arbitrary byte array. However resist
      * the temptation to use a string as the seed - any key derived from a password is likely to be weak and easily
      * broken by attackers (this is not theoretical, people have had money stolen that way). This method checks
-     * that the given seed is at least 64 bits long.
+     * that the given seed is at least 128 bits long and at most 512 bits long (as per BIP32).
      *
      * @throws HDDerivationException if generated master key is invalid (private key not between 0 and n inclusive)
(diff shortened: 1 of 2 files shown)


msgilligan (approved):
Looks good. I think this should be included in the 0.18 release.

schildbach:
I pushed some trivial unit test fixes on top. I will squash the fixes when this PR is merged.

schildbach:
I pushed fixes for the remaining failing tests on top, but these change the actual test vectors.

@msgilligan  can you review these last commits?

schildbach (approved):
Everything squashed into one commit looks good to me.

liamgilligan (approved):

msgilligan (approved):
Since we're changing tests, we might want to add tests that verify the `IllegalArgumentException` when the seed is too short or too long.

PR #4212, opened 2026-06-26
ChildNumber/HDPath: accept H, h, apostrophe for hardened derivation; add toString(char)

dukeh3:
## Summary

- `ChildNumber.parse()` now accepts `H`, `h`, and `'` (apostrophe) as hardened derivation suffixes
- Add `ChildNumber.toString(char)` to serialize with a chosen hardening symbol (e.g. `'H'`, `'h'`, or `'\''`)
- Add `HDPath.toString(char)` to serialize an entire path with a chosen hardening symbol
- Default `toString()` behavior unchanged (uses `H`)

## Test plan

- [x] Existing `HDPathTest` vectors pass unchanged
- [x] New parse test vectors verify `m/44'/0'/0'` and `m/44h/0h/0h` parse correctly
- [x] `ChildNumber.parse()` tests for `H`, `h`, and `'` — all equivalent
- [x] `ChildNumber.toString(char)` tests for apostrophe, lowercase h, uppercase H
- [x] `HDPath.toString(char)` tests for full paths, partial paths, mixed hardened/non-hardened
- [x] Roundtrip test: parse apostrophe path, serialize with all three symbols

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/ChildNumber.java
+++ b/core/src/main/java/org/bitcoinj/crypto/ChildNumber.java
@@ -68,13 +68,14 @@ public class ChildNumber implements Comparable<ChildNumber> {
 
     /**
      * Parse a single child number.
+     * Accepts {@code H}, {@code h}, or {@code '} as the hardened suffix.
      *
-     * @param str string of the form "1" or "1H"
+     * @param str string of the form "1", "1H", "1h", or "1'"
      * @return child number instance
      * @throws NumberFormatException when parsing fails
      */
     public static ChildNumber parse(String str) {
-        boolean isHard = str.endsWith("H") || str.endsWith("'");
+        boolean isHard = str.endsWith("H") || str.endsWith("h") || str.endsWith("'");
         String nodeNumber = isHard ?
                 str.substring(0, str.length() - 1) :
                 str;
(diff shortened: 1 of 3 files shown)


msgilligan:
A correction to the  above comment:

> * BIP-32 specifies apostrophe as the standard notation (e.g. `m/44'/0'/0'`), but the parser currently only recognizes `H`

BIP-32 _does not_ specify apostrophe notation. It is a community-notation that is mentioned in later BIPs, but BIP-32 itself uses `H` (subscript) to indicate hardened children.

There is also a comment in the code that makes this incorrect statement.

core/src/main/java/org/bitcoinj/crypto/ChildNumber.java
```diff
@@ -74,7 +74,7 @@ public ChildNumber(int i) {
      * @throws NumberFormatException when parsing fails
      */
     public static ChildNumber parse(String str) {
-        boolean isHard = str.endsWith("H");
+        boolean isHard = str.endsWith("H") || str.endsWith("'");
```
msgilligan: We should allow lower-case `h` here as well.

core/src/test/java/org/bitcoinj/crypto/HDPathTest.java
```diff
@@ -235,6 +235,18 @@ private PathVector[] parseTestVectors() {
             new PathVector (
                 "1 / 2 / 3 /",
                 HDPath.partial(new ChildNumber(1, false), new ChildNumber(2, false), new ChildNumber(3, false))
+            ),
+
+            // BIP-32 standard apostrophe notation
```
msgilligan: This comment is incorrect.  Just remove "BIP-32 standard" and it will be fine.

msgilligan (changes requested):
Looks good, but I have two minor requests.

PR #4214, opened 2026-07-09
PeerGroup: fix startAsync() to complete exceptionally, rather than swallow exceptions

msgilligan:


core/src/main/java/org/bitcoinj/core/PeerGroup.java
```diff
@@ -1163,10 +1163,10 @@ public CompletableFuture<Void> startAsync() {
         vRunning = true;
         vUsedUp = true;
         executorStartupLatch.countDown();
+        log.info("Starting ...");
```
schildbach: I'm not sure if that is what was intended when adding this message. Before, it was logging when it's starting (as the message sais). Now, it is logging that `startAsync()` was called.

Also, if it stays like this, it could be put into an `else` branch of the above `log.warn` conditional, because the two logs can be redundant.

schildbach (changes requested):
Just a comment about the first commit.

core/src/main/java/org/bitcoinj/core/PeerGroup.java
msgilligan: The `log.info("Starting...")` is the only thing that happens on the `runAsync` thread _before_ calling `channels.start()`.  I want to be able to just call channel.start() directly as it is simpler/cleaner (it is already asynchronous and returns a future.) The purpose of the log is to indicate that the PeerGroup is starting and I don't think it really matters which thread it is made from.

I will move it to an else branch sibling of the `log.warn`...

msgilligan: Rebased!

schildbach (approved):

PR #4220, opened 2026-07-19
PeerNetwork: new class to store TxConfidenceTable

msgilligan:
* Add PeerNetwork class
* (For now) construct it in the `Context` constructor
* Deprecate Context.getConfidenceTable()
* Update callers to use Context.peerNetwork().txConfidenceTable()

TODO: Make a default PeerNetwork singleton and update apps/samples to set it in the initial/propagated Context.


msgilligan:
This supersedes PR #4096.

PR #4223, opened 2026-07-19
ContextPropagatingThreadFactory: move to o.b.core package

msgilligan:
This is a step towards deprecating `Context` entirely. The move will allow ContextPropagatingThreadFactory to use (soon-to-be) package private methods in `Context` and will also allow ContextPropagatingThreadFactory itself to be made package-private.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Context.java
+++ b/core/src/main/java/org/bitcoinj/core/Context.java
@@ -17,7 +17,6 @@
 package org.bitcoinj.core;
 
 import org.bitcoinj.base.Coin;
-import org.bitcoinj.utils.ContextPropagatingThreadFactory;
 import org.bitcoinj.wallet.SendRequest;
 import org.slf4j.Logger;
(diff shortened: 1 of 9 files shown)


schildbach (commented):
At the moment, Bitcoin Wallet is stil using `ContextPropagatingThreadFactory`. The pattern is always like this:

```
ExecutorService threadPool = Executors.newFixedThreadPool/newSingleThreadExecutor(…,
                        new ContextPropagatingThreadFactory("request"));
```

I had hoped the class can still be used until the point where `Context` is not necessary any more.

msgilligan:
> At the moment, Bitcoin Wallet is stil using `ContextPropagatingThreadFactory`. 

After this PR, you should be able to just change the package you are importing from.

> I had hoped the class can still be used until the point where `Context` is not necessary any more.

It can still be used if you change the import package. I guess we should hold off on making it package private until it is no longer needed by non-internal classes.

PR #4228, opened 2026-07-20
build-gradle.yml: Use setup-java in place of setup-graalvm

msgilligan:


schildbach (changes requested):
This PR will make different versions of setup-gradle (v4/v5) be used in the same build. Would it be an option to update the other usage(s) first?

msgilligan:
> This PR will make different versions of setup-gradle (v4/v5) be used in the same build. Would it be an option to update the other usage(s) first?

Yes, that was just an oversight. I'll fix it.

msgilligan:
> update the other usage(s) first?

@schildbach fixed, rebased, force-pushed.

schildbach (approved):

PR #4234, opened 2026-07-23
dependabot.yml: add junit group definition

msgilligan:


schildbach:
Seems like we don't really need this:

- dependabot grouped JUnit 6 correctly, without special config
- if JUnit 7 shows up, we would probably want to be notified – at that point we can decide to ignore

Wdyt?

msgilligan:
I agree.

msgilligan:
Actually, we still have some JUnit 5 usage. So I was *hoping* the "no major updates" would upgrade modules using 5 to the latest 5.x, and modules using 6 to the latest 6.x. I'm not sure if that would work correctly.

schildbach:
It's JUnit 4 usage, and that one uses an entirely different group+artifact ID.

msgilligan:
OK, we just got separate updates for JUnit 6.1.3:

* PR #4279
* PR #4282

So I rebased this PR, removed the ignore change, and added a `:` to the `org.junit:*` pattern.

schildbach (approved):

PR #4257, opened 2026-07-29
SPVBlockStore: fix and improve cursor wrap-around

### LINKED ISSUE #1084: Thread stuck in SPVBlockStore.get
I run a tomcat server using with a web app using bitcoinj 0.13.2.

Every once in a while (after a few days, or weeks) the java process starts consuming 100% of a core. It turns out that this thread is constantly running:

```
  "NioClientManager" daemon prio=10 tid=0x00007fa438054000 nid=0x37c3 runnable [0x00007fa40ca2f000]
   java.lang.Thread.State: RUNNABLE
        at org.bitcoinj.store.SPVBlockStore.get(SPVBlockStore.java:215)
        at org.bitcoinj.core.BlockChain.getStoredBlockInCurrentScope(BlockChain.java:160)
        at org.bitcoinj.core.AbstractBlockChain.tryConnectingOrphans(AbstractBlockChain.java:810)
        at org.bitcoinj.core.AbstractBlockChain.add(AbstractBlockChain.java:418)
        at org.bitcoinj.core.AbstractBlockChain.add(AbstractBlockChain.java:308)
        at org.bitcoinj.core.BlockChain.add(BlockChain.java:165)
        at org.bitcoinj.core.Peer.endFilteredBlock(Peer.java:945)
        at org.bitcoinj.core.Peer.processMessage(Peer.java:342)
        at org.bitcoinj.core.PeerSocketHandler.receiveBytes(PeerSocketHandler.java:180)
        at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:223)
        at org.bitcoinj.net.NioClientManager.handleKey(NioClientManager.java:86)
        at org.bitcoinj.net.NioClientManager.run(NioClientManager.java:122)
        at com.google.common.util.concurrent.AbstractExecutionThreadService$1$2.run(AbstractExecutionThreadService.java:60)
        at com.google.common.util.concurrent.Callables$3.run(Callables.java:93)
        at org.bitcoinj.utils.ContextPropagatingThreadFactory$1.run(ContextPropagatingThreadFactory.java:35)
        at java.lang.Thread.run(Thread.java:745)
```

I checked (in the above example of the problem) that nid=0x37c3 is the same as "top -H" displays as the thread (14275 decimal) consuming the cpu.

I've generated many (10-20) thread dumps and it was the exact line SPVBlockStore.java:215 in all of them.

The only way to "fix" this is to kill -9 the java process and restart it.

I've tried to analyze the code, but couldn't find anything obvious causing this. Maybe someone here could shed some light on this? Do you have any suggestions on how I could debug this further?

schildbach:
Two commits:

1. SPVBlockStore: fix cursor not wrapping around in get()
~~2. SPVBlockStore: apply the wrap-around logic to `setRingCursor()` as well~~

### DIFF
--- a/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
+++ b/core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
@@ -269,10 +269,6 @@ public class SPVBlockStore implements BlockStore {
         lock.lock();
         try {
             int cursor = getRingCursor();
-            if (cursor == fileLength) {
-                // Wrapped around.
-                cursor = FILE_PROLOGUE_BYTES;
-            }
             ((Buffer) buffer).position(cursor);
             Sha256Hash hash = block.getHeader().getHash();
             notFoundCache.remove(hash);
@@ -405,13 +401,19 @@ public class SPVBlockStore implements BlockStore {
 
     protected static final int FILE_PROLOGUE_BYTES = 1024;
 
-    /** Returns the offset from the file start where the latest block should be written (end of prev block). */
+    /**
+     * Returns the offset from the start of file where the latest block should be written. Normally this is after
+     * the previous block. But if that's also the end of file, the cursor gets wrapped around to start, after the
+     * prologue.
+     *
+     * @return cursor as offset in bytes
+     */
     int getRingCursor() {
         Objects.requireNonNull(buffer);
-        int c = buffer.getInt(4);
-        checkState(c >= FILE_PROLOGUE_BYTES, () ->
-                "integer overflow");
-        return c;
+        int cursor = buffer.getInt(4);
+        checkState(cursor >= FILE_PROLOGUE_BYTES && cursor <= fileLength, () ->
+                "cursor out of range: " + cursor);
+        return cursor < fileLength ? cursor : FILE_PROLOGUE_BYTES;
     }
 
     private void setRingCursor(int newCursor) {


core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -407,13 +403,16 @@ public void close() throws BlockStoreException {
...
+     * the previous block. But if that's also the end of file, the cursor gets wrapped around to after the prologue.
+     */
     int getRingCursor() {
         Objects.requireNonNull(buffer);
-        int c = buffer.getInt(4);
-        checkState(c >= FILE_PROLOGUE_BYTES, () ->
-                "integer overflow");
-        return c;
+        int cursor = buffer.getInt(4);
+        checkState(cursor >= FILE_PROLOGUE_BYTES || cursor > fileLength, () ->
```
msgilligan: Shouldn't this be `cursor >= FILE_PROLOGUE_BYTES && cursor < fileLength`? And adjusting _after_ the check will never happen, but that's ok since the adjustment is done in `setRingCursor()`, right?

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -415,10 +415,15 @@ int getRingCursor() {
...
     }
 
-    private void setRingCursor(int newCursor) {
-        checkArgument(newCursor >= 0);
+    /**
+     * Sets the new cursor as an offset from the start of file. If the cursor is set to the end of file, it gets
+     * wrapped around to after the prologue.
+     */
+    private void setRingCursor(final int newCursor) {
+        checkArgument(newCursor >= FILE_PROLOGUE_BYTES || newCursor > fileLength, () ->
```
msgilligan: This should be `&& newCursor <= fileLength`

msgilligan (changes requested):

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
schildbach: OMG yes my mistake. However it's `cursor >= FILE_PROLOGUE_BYTES && cursor <= fileLength`, as the eof position is explicitly allowed (but then gets wrapped). Fix pushed.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
schildbach: Same as above. Fix pushed.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
msgilligan: Well, I was suggesting we remove the wrap in `getRingCursor()` since all writes go through `setRingCursor()` which performs the wrap. In that case `<` would be the correct check for valid state.

schildbach: But we'd need a migration then. Note that existing stores have the cursor at the eof.

We can do that of course. But maybe to fix the actual problem, we should perhaps merge just the first commit and not change the storage semantics.

msgilligan: Right, existing stores. Let's keep this PR simple and go with just the first commit.

msgilligan (approved):

schildbach:
Ok, I rebased this PR on the recent SPVBlockStore changes and squashed the fix. I also removed the second commit and its fix.

Requesting re-review under the assumption that the second commit may never come up.

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
```diff
@@ -405,13 +401,17 @@ public void close() throws BlockStoreException {
 
     protected static final int FILE_PROLOGUE_BYTES = 1024;
 
-    /** Returns the offset from the file start where the latest block should be written (end of prev block). */
+    /**
+     * Returns the offset from the start of file where the latest block should be written. Normally this is after
+     * the previous block. But if that's also the end of file, the cursor gets wrapped around to start, after the
+     * prologue.
+     */
```
msgilligan, line 410: Nit: since you rewrote the comment, why not add a `@return` tag and have it document that the return value is an "offset in bytes".

msgilligan (changes requested):
Minor request for Javadoc change...

core/src/main/java/org/bitcoinj/store/SPVBlockStore.java
schildbach: Done.

msgilligan (approved):

PR #4261, opened 2026-07-30
Block: add getGenesis(), SPVBlockStore: use Network in constructors

msgilligan:
Existing SPVBlockStore constructors that take NetworkParameters are deprecated.

Note that the RestoreFromSeed example is now able to completely remove imports of NetworkParameters and subclasses.

BlockImporter can drop NetworkParameters once the MemoryFullPrunedBlockStore constructors are updated to take Network.

This is a step towards our goal of (typical) applications using only (Bitcoin)Network in API calls.

This also opens the possibility of migrating genesis block parameters out of network parameters.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Block.java
+++ b/core/src/main/java/org/bitcoinj/core/Block.java
@@ -19,6 +19,7 @@ package org.bitcoinj.core;
 
 import com.google.common.annotations.VisibleForTesting;
 import org.bitcoinj.base.Difficulty;
+import org.bitcoinj.base.Network;
 import org.bitcoinj.base.Sha256Hash;
 import org.bitcoinj.base.VarInt;
 import org.bitcoinj.base.internal.Buffers;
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/bitcoinj/core/Block.java
```diff
@@ -269,6 +270,15 @@ public Block(long version, Sha256Hash prevHash, Sha256Hash merkleRoot, Instant t
...
 
+    /**
+     * Get the standard full Genesis Block for a network
+     * @param network the network
+     * @return a complete block (should be treated as immutable)
+     */
+    public static Block getGenesis(Network network) {
+        return NetworkParameters.of(network).getGenesisBlock();
+    }
+
```
schildbach, line 281: Hm, I think the special genesis block is more an artifact of a network or a block chain, but not really of a block.

So I'd argue perhaps we should have a helper in `Network` instead. (I'm surprised we don't have than already.)

However, even if we go ahead with this PR, I think the refactoring to introduce this method should be on a separate commit.

msgilligan: > So I'd argue perhaps we should have a helper in Network instead.

I think genesis block should be constructed by `Block` for three reasons:

1. They are instances of `Block` and the standard static factory method pattern seems to apply here.
2. We already have several `createGenesis()` static factories in `Block` (immediately below this new method)
3. We can't put it in `Network` because `Block` and other dependencies are in `o.b.core` not `o.b.base` (notably the current `NetworkParameters` constants, thought those could be moved)

> the refactoring to introduce this method should be on a separate commit.

Agree.

msgilligan:
@schildbach Oops, re-requested review and I have not created a separate commit. I will do that soon. In the meantime, do my arguments for why `getGenesis()` should be in `Block` makes sense?

msgilligan:
As requested, I factored the creation of `Block.getGenesis(Network network)` to a new PR:

* PR #4352 

Once that is merged this can be rebased.

PR #4265, opened 2026-08-01
Wallet, MemoryFullPrunedBlockStore: return unmodifiable lists when using UTXOProvider

msgilligan:


core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4690,7 +4690,7 @@ private List<UTXO> getStoredOutputsFromUTXOProviderInternal() throws UTXOProvide
                         getActiveKeyChain().getLeafKeys().stream())
                 .map(ECKey::fromPublicOnly) // don't pass private keys to UTXO provider
                 .collect(Collectors.toList());
-        return new ArrayList<>(utxoProvider.getOpenTransactionOutputs(keys));
+        return Collections.unmodifiableList(new ArrayList<>(utxoProvider.getOpenTransactionOutputs(keys)));
```
schildbach, line 4693: Why not take the ArrayList away at the same time?

schildbach (changes requested):

core/src/main/java/org/bitcoinj/wallet/Wallet.java
msgilligan: Because that would lose the defensive copy. Since we don't know (for sure) what the provider is doing and `unmodifiable` is a wrapper view and the wrapped list may be modifiable.

schildbach (approved):

PR #4283, opened 2026-08-11
ECKey, DeterministicKey: use BC ECPoint internally, prepare to deprecate `LazyECPoint`

msgilligan:
This is a draft/proof-of-concept, but I think it is pretty close to being ready. So preliminary reviews are welcome.

All use of `LazyECPoint` can now be deprecated. We should also remove the `ECPublicKey` superclass we added to `LazyECPoint` since that was never released and add it to ECKey.

This (when finalized) will supersede:

* PR #4278 
* PR #3756

PR #4284, opened 2026-08-11
ECKey: remove deprecated message signing/verification methods

msgilligan:
The changes to the unit tests make them a little more flexible (e.g. they could be modified to us a parameterized output ScriptType.)

cc: @johnzweng (who deprecated the methods a few years back)

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/ECKey.java
@@ -30,7 +30,6 @@ import org.bitcoinj.base.internal.TimeUtils;
 import org.bitcoinj.base.internal.ByteUtils;
 import org.bitcoinj.base.VarInt;
 import org.bitcoinj.crypto.internal.CryptoUtils;
-import org.bitcoinj.crypto.utils.MessageVerifyUtils;
 import org.bitcoinj.wallet.Wallet;
 import org.bouncycastle.asn1.ASN1InputStream;
 import org.bouncycastle.asn1.ASN1Integer;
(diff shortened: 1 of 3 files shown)


msgilligan:
PR #4283 should be rebased after this is merged.

schildbach (commented):
I wonder if we should first migrate the test cases that sign/verify arbitrary messages to `MessageVerifyUtilsTest`, and at the same time make them use `MessageVerifyUtils`.

Then this PR should apply cleanly, possibly without any more changes to unit tests. Wdyt?

msgilligan:
> migrate the test cases that sign/verify arbitrary messages to `MessageVerifyUtilsTest`, and at the same time make them use `MessageVerifyUtils`.

I think we should do this at a later time. I would like to (eventually) do some refactoring on `MessageVerifyUtils` and write `MessageVerifyUtilsTest`  at that time.

> Then this PR should apply cleanly, possibly without any more changes to unit tests. Wdyt?

I think we should rebase and merge as is and write `MessageVerifyUtilsTest` later.

msgilligan:
> this PR should apply cleanly

@schildbach It was a very simple rebase and now applies cleanly.

schildbach:
Ok I agree. The tests remain testing ECKey and DeterministicKey in nature. Moving them to MessageVerifyUtilsTest is no win at the moment.

schildbach (approved):

johnzweng:
Thanks for taking me in CC. Looks good to me, too, but as I’m only a very occasional contributor, my word probably shouldn’t carry too much weight here anyway.

Thanks for your work. 🙂

PR #4285, opened 2026-08-11
ECKeyTest: use assertions in checkSomeBytesAreNonZero(), other code cleanup

msgilligan:
9 commits that fall into roughly 3 categories:

1. Fix `checkSomeBytesAreNonZero` so it fails tests when the check fails (its boolean result was being ignored)
2. Use Jackson instead of "Poor Man's JSON Parser" for reading some .json files
3. Miscellaneous test cleanup (mostly driven by IDE suggestions)

See the individual commits for details.


schildbach:
> The correct assertion is that getSecretBytes() is non-null.

Sure about that? The assert says it must be null.

schildbach (approved):
One open question about a possible mistake in a commit description, see comment above.

msgilligan:
> > The correct assertion is that getSecretBytes() is non-null.
> 
> Sure about that? The assert says it must be null.

Oops! Fixed and rebased!

PR #4293, opened 2026-08-12
ECKey: make `pub` LazyECPoint member private

msgilligan:
This is a child of PR #4291.

The first two (additional) commits remove the last two non-private uses of `pub`. The third commit makes `pub` `private`.


schildbach:
Rebased.

schildbach (approved):

PR #4299, opened 2026-08-13
ECKey: make `priv` private, add getNullableS() accessor

msgilligan:
Make the BigInteger private key `priv` a private member. Add the
`protected` getter `getNullableS()` to return the key or null. This method is similarly
named to Java's ECPrivateKey.getS(), but that method does not return `null`.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
@@ -201,7 +201,7 @@ public class DeterministicKey extends ECKey {
     /** @deprecated use {@link #withParent(DeterministicKey)} */
     @Deprecated
     public DeterministicKey(DeterministicKey keyToClone, DeterministicKey newParent) {
-        this(keyToClone.priv, keyToClone.getPubKeyPoint(), keyToClone.childNumberPath.size(), newParent,
+        this(keyToClone.getNullableS(), keyToClone.getPubKeyPoint(), keyToClone.childNumberPath.size(), newParent,
                 newParent.getFingerprint(), keyToClone.chainCode, keyToClone.childNumberPath, null, null);
     }
 
@@ -343,7 +343,7 @@ public class DeterministicKey extends ECKey {
      * @return this key without private key
      */
     public DeterministicKey withoutPrivateKey() {
-        return priv == null && encryptedPrivateKey == null && keyCrypter == null ?
+        return !super.hasPrivKey() && encryptedPrivateKey == null && keyCrypter == null ?
                 this :
                 new DeterministicKey(null, getPubKeyPoint(), depth, parent, parentFingerprint, chainCode, childNumberPath,
                         null, null);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -343,7 +343,7 @@ public byte[] getPrivKeyBytes33() {
      * @return this key without private key
      */
     public DeterministicKey withoutPrivateKey() {
-        return priv == null && encryptedPrivateKey == null && keyCrypter == null ?
+        return getNullableS() == null && encryptedPrivateKey == null && keyCrypter == null ?
```
schildbach: Rather than `getNullableS() == null` I think we should use something like `!hasPrivateKey()`.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -438,7 +438,7 @@ public boolean hasPrivKey() {
 
     @Override
     public byte @Nullable [] getSecretBytes() {
-        return priv != null ? getPrivKeyBytes() : null;
+        return getNullableS() != null ? getPrivKeyBytes() : null;
```
schildbach: Same here.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -537,7 +537,7 @@ private BigInteger findOrDeriveEncryptedPrivateKey(KeyCrypter keyCrypter, AesKey
     private DeterministicKey findParentWithPrivKey() {
         DeterministicKey cursor = this;
         while (cursor != null) {
-            if (cursor.priv != null) break;
+            if (cursor.getNullableS() != null) break;
```
schildbach, line 540: Here.

core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java
```diff
@@ -546,9 +546,9 @@ private DeterministicKey findParentWithPrivKey() {
     @Nullable
     private BigInteger findOrDerivePrivateKey() {
         DeterministicKey cursor = findParentWithPrivKey();
-        if (cursor == null || cursor.priv == null)
+        if (cursor == null || cursor.getNullableS() == null)
```
schildbach, line 549: Here.

schildbach (commented):
See comments.

msgilligan:
`!hasPrivKey()` on `DeterministicKey` is different from `priv == null`, because it also checks whether parents have private keys...

Maybe we can use `!super.hasPrivKey()`?

schildbach:
…or introduce a method that doesn't take parents into account.

schildbach:
I'm also not a fan of names like `getNullableX()`. I'd rather avoid that an API method can return `null`.

msgilligan:
> I'm also not a fan of names like `getNullableX()`. I'd rather avoid that an API method can return `null`.

I agree. But in this case (as a temporary solution) I think it is ok/necessary.

One use case: we are chaining constructors and the called constructor takes a Nullable private key.

schildbach:
Do we actually *need* to make `priv` fully private? Maybe it's not worth it, at least not at the moment?

schildbach:
> Maybe we can use `!super.hasPrivKey()`

It could also have unexpected semantics, just as with `this.hasPrivKey()`.

msgilligan:
> Do we actually _need_ to make `priv` fully private? Maybe it's not worth it, at least not at the moment?

I think we should. I want to replace it with `SecpPrivKey` in the next release and have done that already in my test branch.  This PR is a backport from that test branch.

liamgilligan (approved):

schildbach (approved):

PR #4300, opened 2026-08-13
ECKey: Use getW() rather than getPubKeyPoint() for equality assertions

msgilligan:
This will convert from BC to Java for each comparison, but prepares us for deprecating (or making protected) getPubKeyPoint() in the future.


schildbach:
Rebased.

schildbach:
I wonder if we should migrate to comparing keys in these cases, not the points or "W".

msgilligan:
> I wonder if we should migrate to comparing keys in these cases, not the points or "W".

Do you mean `ECKey` or `ECPublicKey`?

hasnnbhan1-beep (commented):
bc1q50nfearh80ckcgqy5cwmtsrud4lvat8fyja4hy

liamgilligan (approved):

schildbach (approved):

PR #4304, opened 2026-08-16
ECKey: simplify encodeToDER(), toASN1()

msgilligan:
See the 5 individual commits for details.

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/ECKey.java
+++ b/core/src/main/java/org/bitcoinj/crypto/ECKey.java
@@ -62,6 +62,7 @@ import org.bouncycastle.math.ec.ECPoint;
 import org.bouncycastle.math.ec.FixedPointCombMultiplier;
 import org.bouncycastle.math.ec.FixedPointUtil;
 import org.bouncycastle.math.ec.custom.sec.SecP256K1Curve;
+import org.bouncycastle.util.BigIntegers;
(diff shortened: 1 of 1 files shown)


schildbach:
Rebased.

schildbach:
I've merged all but the last commit for now.

The last one is not a simplification, but inlines code from another method. Not sure if the reduction of private key copied actually buys us something, as we can never get rid of at least one copy.

Maybe in future FFM will help to keep control?

schildbach:
Perhaps we should simply deprecate and remove ASN.1? It (and nowadays mostly also DER) isn't welcome in the Bitcoin world, so why interface? Are you aware of any usecases?

msgilligan:
> I've merged all but the last commit for now.

Yes, let's drop that one. We've deprecated `toASN1()` now anyway.
 
> Not sure if the reduction of private key copied actually buys us something, as we can never get rid of at least one copy.

Well, if we do it systematically and cut down on the number of copies of private keys floating around in the heap, we will reduce the chance of key leaks.

> 
> Maybe in future FFM will help to keep control?

Yes, `secp-ffm` will definitely help keep the number of copies of private key data under better control. `secp-bouncy` with supporting changes in **bitcoinj** may help, too.

msgilligan:
Since we've deprecated `toASN1()` we can just close this.

Closing.

PR #4306, opened 2026-08-16
ECKey: Read ASN1 with fromByteArray() rather than ASN1InputStream

msgilligan:
Child of PR #4305 

Note that making this same change to replace the other `ASN1InputStream` breaks tests due to `fromByteArray()` being strict about extra bytes in the stream.


schildbach:
Rebased. Is this PR ready for review/merging?

msgilligan:
I think we should close this PR since this is deprecated anyway. 

This PR may change behavior, so let's leave things as they are until removal.

schildbach:
> I think we should close this PR since this is deprecated anyway.

Agree.

PR #4307, opened 2026-08-16
WIP: ECKey.ECDSASignature: decodeFromDER() -- two possible changes

msgilligan:
These two changes need further review:

1. Verify `seq.size() == 2` and throw otherwise. If an invalid signature
   is parsed, the existing code will throw `ArrayIndexOutOfBoundsException`
   not `SignatureDecodeException` with possible unintended behavior.
2. Cast `seq` to `ASN1Sequence` which is more broad. We might want
   the more specific check, but we probably want to throw `SignatureDecodeException`
   if the format is not what we expected.


schildbach (commented):
Looks good to me.

At the moment, the sequence could theoretically be longer than 2 and the remainder will be ignored. This change will throw in this case. I expect we want the latter anyway, but should take the change into account.

PR #4313, opened 2026-08-17
DeterministicKeyChain: verify seed when decoding a key chain from protobuf

schildbach:
Addresses https://github.com/project-loupe/audit-bitcoinj/issues/331.

When decoding a key chain from protobuf format, always derive the seed from the mnemonic words and verify the stored seed against it.

Also remove helper method `DeterministicSeed.optionalSeedFromMnemonic()` and add a test.

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DeterministicSeed.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DeterministicSeed.java
@@ -141,8 +141,8 @@ public class DeterministicSeed implements EncryptableItem {
     }
 
     // For use in DeteministicKeyChain.fromProtobuf() only
-    static DeterministicSeed fromProtobuf(String mnemonicString, byte @Nullable [] seed, String passphrase, @Nullable Instant creationTime) {
-        return new DeterministicSeed(optionalSeedFromMnemonic(splitMnemonicCode(mnemonicString), passphrase, seed), splitMnemonicCode(mnemonicString), creationTime);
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/bitcoinj/wallet/DeterministicSeed.java
```diff
@@ -278,8 +273,7 @@ public DeterministicSeed decrypt(KeyCrypter crypter, String passphrase, AesKey a
         checkState(isEncrypted());
         Objects.requireNonNull(encryptedMnemonicCode);
         List<String> mnemonic = decodeMnemonicCode(crypter.decrypt(encryptedMnemonicCode, aesKey));
-        byte[] seed = encryptedSeed == null ? null : crypter.decrypt(encryptedSeed, aesKey);
-        return new DeterministicSeed(optionalSeedFromMnemonic(mnemonic, passphrase, seed), mnemonic, creationTime);
+        return new DeterministicSeed(seedFromMnemonic(mnemonic, passphrase), mnemonic, creationTime);
```
msgilligan, line 276: I think this change should be in a separate commit or possibly even a separate PR. This commit/PR should come before the other change.

Note that with this change the cached `encryptedSeed` functionality is entirely removed.

msgilligan (changes requested):
We should split the disabling of the cached `encryptedSeed` into a separate PR (or at least separate commit)

There is further cleanup/simplification that can be done if use of `encryptedSeed` is removed.

msgilligan:
See:

*  PR #4314

PR #4314, opened 2026-08-17
DeterministicSeed/DeterministicKeyChain: remove cached encryptedSeed data

msgilligan:
The seed will always be recalculated from the mnemonic. A `null` value will be serialized via ProtoBuf for backward-compatibility with old wallets.


schildbach:
The protobuf has cached copies for both cases: encrypted and unencrypted.

```
  // The seed for a deterministic key hierarchy.  Derived from the mnemonic,
  // but cached here for quick startup.  Only applicable to a DETERMINISTIC_MNEMONIC key entry.
  optional bytes deterministic_seed = 8;

  // Encrypted version of the seed
  optional EncryptedData encrypted_deterministic_seed = 9;
```

I wonder for the sake of uniformity we should remove both?

schildbach:
Also:

- Do we take a significant performance hit if wallets restored from protobuf need to derive from their mnemonic *always*? I have to check where the key stretching happens exactly.
- Are there any wallets out there that have a mismatching seed vs mnemonic. Those would be at risk for loosing funds. In theory, this case should not happen…

PR #4332, opened 2026-08-23
Move FullBlockTestGenerator and related tests to integration-test

msgilligan:


schildbach:
What's the reason for this move, and why is `first-100k-blocks.dat` moving in the first place?

msgilligan:
> What's the reason for this move 

We have wanted to move integration tests to the integration test module for a long time:

* See: https://github.com/bitcoinj/bitcoinj/issues/1984#issuecomment-619239187

`FullBlockTestGenerator` is not a test, but a test support class. `BitcoindComparisonTool` is a command-line tool. Moving these two classes to integration tests means they can be better utilized.

I have a modified version of `BlockFileLoaderTest` that I optimized and used to speed up block processing and would like to commit to the repository (it uses `java.util.concurrent.Flow` for "reactive" multithreading, that's what gives it a performance boost.)

I think these tests can be considerable improved and their coverage expanded and being able to use Java 17 and JUnit 6, will make that much easier.

> why is `first-100k-blocks.dat` moving in the first place?

The five files being moved are all integration tests (and test integration test data) and need to be moved as a group because they are linked together as follows:

+`first-100k-blocks.dat`
|--`FullPrunedBlockChainTest`
|--`BlockFileLoaderTest`


+`FullBlockTestGenerator`
|--`BitcoindComparisonTool`
|--`FullPrunedBlockChainTest`

schildbach (approved):

PR #4343, opened 2026-08-26
Context: make isRelaxProofOfWork() package-private

msgilligan:
~~Two commits~~:

* Context: make isRelaxProofOfWork() package-private
* ~~Context: fix Javadoc link to private method in Block~~

Rebased.


schildbach:
I fixed the JavaDoc link by removing it entirely. I see no point in linking to the Block class if we can't point to the specific method.

Issue #4093, opened 2026-02-02
Deprecate and remove `Context` object.


msgilligan:
We have been working towards this goal for years now, but I think we can finish this as part of the 0.18 release.

There are 5 fields in `Context`:

1. `boolean relaxProofOfWork`  -- only used in `Block.checkProofOfWork()` and only used for unit testing. 
2. `int eventHorizon` -- only used in `Wallet.notifyNewBestBlock()`
3. `boolean ensureMinRequiredFee` --  only used in building `SendRequest` instances
4. `Coin feePerKb`-- only used in building `SendRequest` instances
5. `TxConfidenceTable txConfidenceTable` -- used in `Wallet`, `Peer`, `PeerGroup` (indirectly), and `Transaction`

This issue is a tracking issue for a stepwise deprecation and removal of `Context` and its use as a `ThreadLocal`. The steps should look roughly like.

1. Eliminating all use of `Context` within **bitcoinj** focusing on the 5 fields above. For backwards-compatibility we can probably leave the getters for these five fields and have them provide the instances/data that are now stored elsewhere.
2. Remove the requirement for and use of `Context` as a  `ThreadLocal` for normal usage (i.e. everywhere but in the tests)
3. Remove the the usage of `ThreadLocal` `Context` in the tests
4. Deprecate all `public` references to `Context`  --- hopefully this can be included in Milestone 0.18.
5. Remove the `Context` class entirely.

msgilligan:
See PR #4096 as a step towards moving the `TxConfidenceTable` to another object/instance/singleton.

Crystora:
@msgilligan Can I pick this up? I am interested! I'd love to contribute.

msgilligan:
See: 

* PR #4221 
* PR #4223
* PR #4224

and this WIP PR shows a strategy for replacing `Context` in the public API with `PeerNetwork`:

* PR #4225

PR #4344, opened 2026-08-26
Context: auto-set default Context, remove strict mode enforcement.

msgilligan:
Two commits:

1. `enableStrictMode()` is deprecated, but this commit removes enforcement even if `enableStrictMode()` is called.
2. Rather than fail if no `lastConstructed` context is available, creating a default context.

This implements step 2 in:

* Issue #4093

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Context.java
+++ b/core/src/main/java/org/bitcoinj/core/Context.java
@@ -80,7 +80,6 @@ public class Context {
     }
 
     private static volatile Context lastConstructed;
-    private static boolean isStrictMode;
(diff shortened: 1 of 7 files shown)


schildbach:
I see no chance of merging this before we find a new home for `confidenceTable`. Or alternatively completely rethink and rewrite how transaction confidence works.

schildbach:
If we want to reduce the "impact" of strict mode (or the side effects of not using it), we need to reduce calls to `Context.get()` and `Context.getOrCreate()`. Once that is all gone from non-test code, we can remove strict mode.

PR #4362, opened 2026-08-29
ListCryptoProviders: tool to list relevant Java Cryptography Providers

msgilligan:
This tool will install Bouncy Castle if it is on the classpath. It will then look for ECDSA providers and test whether they can sign and verify on curve secp256k1. It then lists all providers implementing each algorithm in internal list  `BITCOIN_ALGOS`.


tools/src/main/java/org/bitcoinj/tools/ListCryptoProviders.java
```diff
@@ -0,0 +1,141 @@
...
+            System.out.println("No ECDSA at all.");
+            return;
+        }
+
+        System.out.println("Signature." + ECDSA_SIG_ALGO + ":");
+        for (Provider p : candidates) {
+            boolean byName = ecdsaWorks(p, ECDSA_CURVE);
+            boolean byOid  = byName || ecdsaWorks(p, ECDSA_OID);
+            System.out.printf(" • %-12s %-8s %-40s  %s=%b\n",
+                    p.getName(), p.getVersionStr(), shortenInfo(p.getInfo()), ECDSA_CURVE, byName || byOid);
```
liamgilligan, line 92: `byName || byOid` simplifies to `byOid`, since `byOid = byName || ecdsaWorks(...)`.

I think we should just set `byOid = ecdsaWorks(...)`, and keep the `printf` as is.

tools/src/main/java/org/bitcoinj/tools/ListCryptoProviders.java
```diff
@@ -0,0 +1,141 @@
...
+            Signature verifier = Signature.getInstance(ECDSA_SIG_ALGO, p);
+            verifier.initVerify(kp.getPublic());
+            verifier.update(msg);
+            return verifier.verify(sig);
+        } catch (GeneralSecurityException e) {
+            return false;
+        }
+    }
+
+    static String shortenInfo(String info) {
```
liamgilligan, line 149: Could you add a comment for this method, it's the only whose function is not clear (to me).

liamgilligan (changes requested):

tools/src/main/java/org/bitcoinj/tools/ListCryptoProviders.java
msgilligan: Good catch. Fixed and force-pushed!

tools/src/main/java/org/bitcoinj/tools/ListCryptoProviders.java
msgilligan: Added.

liamgilligan (approved):

PR #4374, opened 2026-09-12
Base58: protect alphabet from mutation

schildbach:
Make the constant `ALPHABET` private and provide an accessor `alphabet()` to use instead.

Add tests to verify the alphabet cannot be mutated any more.

### DIFF
--- a/base/src/main/java/org/bitcoinj/base/Base58.java
+++ b/base/src/main/java/org/bitcoinj/base/Base58.java
@@ -1,6 +1,5 @@
 /*
- * Copyright 2011 Google Inc.
- * Copyright 2018 Andreas Schildbach
+ * Copyright by the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,7 +20,10 @@ import org.bitcoinj.base.exceptions.AddressFormatException;
 import org.bitcoinj.base.internal.ByteUtils;
 
 import java.math.BigInteger;
+import java.nio.CharBuffer;
 import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Base58 is a way to encode Bitcoin addresses (or arbitrary data) as alphanumeric strings.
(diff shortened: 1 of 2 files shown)


base/src/main/java/org/bitcoinj/base/Base58.java
```diff
@@ -58,6 +60,18 @@ public class Base58 {
...
+    /**
+     * Get the Base58 alphabet.
+     *
+     * @return alphabet as a list of characters
+     */
+    public static List<Character> alphabet() {
+        return CharBuffer.wrap(ALPHABET)
+                .chars()
+                .mapToObj(c -> (char) c)
+                .collect(Collectors.toList());
```
msgilligan, line 72: If we replace `Collectors.toList()` with our utility `StreamUtils.toUnmodifiableList()` the returned list will be unmodifiable and will throw `UnsupportedOperationException` if modification is attempted.

base/src/test/java/Base58Test.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    @Test
+    public void alphabetIsImmutable() {
+        List<Character> alphabet1 = Base58.alphabet();
+        char previousChar = alphabet1.remove(0);
+        alphabet1.add(0, '₿');
+        List<Character> alphabet2 = Base58.alphabet();
+        assertEquals(previousChar, alphabet2.get(0).charValue());
+    }
```
msgilligan, line 37: If `StreamUtils.toUnmodifiableList()` is used as suggested above, this test can use `assertThrows` to test for unmodifiability.

msgilligan (changes requested):
If we want to return an (immutable/unmodifiable) `List` I have some requested changes.

Alternatively, we could return a `String` which is immutable. We should also be able to use a `String` internally. I'll make a separate PR to explore this option.

msgilligan:
> I'll make a separate PR to explore [using `String`].

See PR #4378

schildbach:
> Alternatively, we could return a `String` which is immutable.

For the API method `alphabet()`, I'd rather not return a String. An alphabet is a collection of characters, not a String.

For the internal storage, I'm more relaxed about what type to use.

msgilligan:
> > Alternatively, we could return a `String` which is immutable.
> 
> For the API method `alphabet()`, I'd rather not return a String. An alphabet is a collection of characters, not a String.

An alphabet is an _ordered_ collection/sequence of characters. So is `String` which implements [CharSequence](https://docs.oracle.com/javase/8/docs/api/java/lang/CharSequence.html) which is a read-only sequence of `char` values.

Yes, Strings are often serialized/output in their (ordered) entirety, but they are still a collection/sequence of characters. 

`String` is an immutable, built-in type that does everything we need. We should use it. Using a generic `List<Character>` is very inefficient (although in a far-distant Valhalla future it may eventually be flattened to the equivalent of a `char[]`.)


> For the internal storage, I'm more relaxed about what type to use.

We could make `ALPHABET`  `private` and provide an `alphabet()` accessor (see https://github.com/bitcoinj/bitcoinj/pull/4378/commits/15b2d30fea124f8e071661cfa90359b302cf1b93.) And the accessor could return the narrower `CharSequence` type (which has a `toString()` method) but I see no harm in it returning `String` directly. Note that `String` has the `toCharArray()` method which returns a (mutable) character array, so we could use that to return `char[]`, but if the user really wants that they can get it from `String` themselves.

